PR #7, opened 2025-11-11
Port create-react-app to vite

sats-23:
create-react-app (cra) was [sunset](https://react.dev/blog/2025/02/14/sunsetting-create-react-app) in Feb 2025

Porting to vite as per [official doc](https://www.robinwieruch.de/vite-create-react-app/) as a long term solution
This also resolves our CVEs (0 vulnerabilities post migration)


spyre-rag/ui/vite.config.js
```diff
@@ -0,0 +1,18 @@
+import { defineConfig } from 'vite';
+import react from '@vitejs/plugin-react';
+
+
+export default defineConfig(() => {
+  return {
+    build: {
+      outDir: 'build',
+    },
```
mayuka-c: Default dist would be better but this would require change in Dockerfile while copying.
If its too much, we can go ahead with build itself

PR #25, opened 2025-11-12
Populate AppTemplate and Version into pod templates

mayuka-c:
Signed-off-by: mayuka-c <Mayuka.C@ibm.com>

Populate labels in Rag template

image

### DIFF
--- a/ai-services/internal/pkg/cli/helpers/helper.go
+++ b/ai-services/internal/pkg/cli/helpers/helper.go
@@ -134,7 +134,8 @@ func FetchContainerStartPeriod(runtime runtime.Runtime, containerNameOrId string
 }
 
 type AppMetadata struct {
-	// TODO: Include other variables too
+	Name                  string     `yaml:"name,omitempty"`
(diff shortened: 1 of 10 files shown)


ai-services/internal/pkg/cli/helpers/helper.go
```diff
@@ -134,7 +134,8 @@ func FetchContainerStartPeriod(runtime runtime.Runtime, containerNameOrId string
 }
 
 type AppMetadata struct {
-	// TODO: Include other variables too
+	Name                  string     `yaml:"name,omitempty"`
+	Version               string     `yaml:"version,omitempty"`
 	SMTLevel              *int       `yaml:"smtLevel,omitempty"`
 	PodTemplateExecutions [][]string `yaml:"podTemplateExecutions"`
 }
```
mayuka-c: Currently we are using the method defined in this file to load Metadata in create. So I had to populate it here. Later most of the methods in this file will be removed and we can use the methods defined in template.go.

Will take it up as part of refactoring as most of the business logic would be implemented with this

PR #41, opened 2025-11-14
Use upsert instead of insert to overwrite the chunkid to avoid duplicates in milvus

dharaneeshvrd:
Add checksum logic to avoid re-processing the pdf files(convert and generate text and table files)

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/application.go
+++ b/ai-services/cmd/ai-services/cmd/application/application.go
@@ -24,6 +24,7 @@ func init() {
 	ApplicationCmd.AddCommand(stopCmd)
 	ApplicationCmd.AddCommand(startCmd)
 	ApplicationCmd.AddCommand(infoCmd)
+	ApplicationCmd.AddCommand(logsCmd)
 	ApplicationCmd.AddCommand(model.ModelCmd)
 	ApplicationCmd.PersistentFlags().StringVar(&vars.ToolImage, "tool-image", vars.ToolImage, "Tool image to use for downloading the model(only for the development purpose)")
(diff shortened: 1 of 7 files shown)


spyre-rag/ingest/doc_utils.py
```diff
@@ -403,13 +403,24 @@ def extract_document_data(input_paths, out_path, llm_model, llm_endpoint, force=
...
-            (Path(out_path) / f"{Path(path).stem}_clean_text.json").exists() and
-            (Path(out_path) / f"{Path(path).stem}_tables.json").exists()
-        )
-    ]
+    filtered_input_paths = []
+    for path in input_paths:
+        f = (Path(out_path) / f"{Path(path).stem}.checksum")
+        if f.exists():
+            checksum = f.read_text()
+            if checksum != generate_file_checksum(path) \
```
manju956: Maybe we can add a comment to explain the logic here for better readability

manju956 (approved):
lgtm

spyre-rag/ingest/doc_utils.py
```diff
@@ -403,13 +403,26 @@ def extract_document_data(input_paths, out_path, llm_model, llm_endpoint, force=
     pipeline_options.table_structure_options.do_cell_matching = True
     pipeline_options.do_ocr = False
 
-    # Skip files that already exist
-    filtered_input_paths = [
```
Niharika0306: @dharaneeshvrd wasn't this logic already handling processing duplicates of files if already present? 
But however, to update a document if something changed, we are adding this checksum logic. Is my understanding right?

PR #48, opened 2025-11-17
[CLEANUP] Port fmt.print usages to klog

sats-23:


ai-services/internal/pkg/cli/helpers/models.go
```diff
@@ -77,6 +78,6 @@ func DownloadModel(model, targetDir string) error {
 	if err != nil {
 		return fmt.Errorf("failed to execute command: %w", err)
 	}
-	fmt.Printf("Model downloaded successfully\n")
+	logger.Infoln("Model downloaded successfully\n")
```
mayuka-c: There's an addition \n. Since we are using Infoln not required

PR #61, opened 2025-11-18
Adding github action to build go binary for PR

adarshagrawal38:


.github/workflows/go.yaml
```diff
@@ -0,0 +1,64 @@
...
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: 'ai-services/go.mod'
+
+      - run: go version
+      
+      - name: Install dependencies
+        run: |
+          cd ai-services
+          go mod tidy
```
mkumatag: we should do tidy, rather - I would prefer running this and report error if anything changed after this tidy run, this will gives us the confidence that user has updated the mod files properly when any changes made.

.github/workflows/go.yaml
```diff
@@ -0,0 +1,64 @@
...
+  build:
+    name: Build go binary
+    runs-on: ubuntu-24.04-ppc64le
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+      
+      - name: Setup Go
```
mkumatag: let us follow above way of setup go reading via go.mod file

mkumatag (changes requested):

.github/workflows/go.yaml
adarshagrawal38: then shall I use `ubuntu-latest` instead of `ubuntu-24.04-ppc64le`, then we can use `actions/setup-go@v6`

.github/workflows/go.yaml
adarshagrawal38: okay sure, added `git diff` check

.github/workflows/go.yaml
adarshagrawal38: PR is updated, PTAL

ai-services/Makefile
```diff
@@ -9,7 +9,7 @@ BUILDDATE ?= $(shell date -u +%Y-%m-%dT%H:%M:%SZ)
 
 .PHONY: build
 build:
-	CGO_ENABLED=0 $(GO) build $(BUILDFLAGS) \
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) $(GO) build $(BUILDFLAGS) \
```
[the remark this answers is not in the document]
adarshagrawal38: yes correct

PR #62, opened 2025-11-19
UI: Setup Dev proxy and remove WebVitals

mayuka-c:
- Setup backend proxy during Dev (This would avoid manually configuring localhost while doing fetch in local dev)
- Remove WebVitals as it part of CRA and not being used

Validation
---
image

image

### DIFF
--- a/ai-services/go.mod
+++ b/ai-services/go.mod
@@ -4,6 +4,7 @@ go 1.24.1
 
 require (
 	github.com/containers/podman/v5 v5.6.2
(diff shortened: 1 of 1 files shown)


mayuka-c:
> lgtm @mayuka-c is just doing a final testing along with nginx server, to make sure proxying doesn't get affected.

Have tested out by building and deploying. PTAL

image

image

Niharika0306 (approved):

PR #63, opened 2025-11-19
Run go mod tidy to fix a dependency

mayuka-c:
Rerunning go mod tidy as one of the dependency that was used here: https://github.com/IBM/project-ai-services/pull/47 was marked as indirect but rather should be a direct dependency.

Reran `go mod tidy` to resolve


mkumatag (approved):
/lgtm

PR #78, opened 2025-11-20
Configure reranker spyre

mayuka-c:


ai-services/assets/applications/RAG/templates/vllm-server.yaml.tmpl
```diff
@@ -114,14 +125,36 @@ spec:
         periodSeconds: 30
         timeoutSeconds: 5
         failureThreshold: 3
+      env:
+        - name: VLLM_MODEL_PATH
+          value: "/models/BAAI/bge-reranker-large"
```
mkumatag: we don't have an approval for this model I believe, we need to use the `BAAI/bge-reranker-v2-m3` model.

mayuka-c: Just testedout, looks to be working fine with BAAI/bge-reranker-v2-m3 model too. Have updated, PTAL

ai-services/assets/applications/RAG/templates/vllm-server.yaml.tmpl
```diff
@@ -53,6 +58,8 @@ spec:
           value: "32768"
         - name: MAX_BATCH_SIZE
           value: "32"
+        - name: MASTER_PORT
```
mkumatag: what is this port?

mayuka-c: This is the port which is required by torch VLLM CoreEngine
By default it uses port 12355

Below is the error which I was facing since we run multiple vllm services within the same pod.
```
ERROR 11-19 16:20:12 [multiproc_executor.py:597] torch.distributed.DistNetworkError: The server socket has failed to listen on any local network address. port: 12355, useIpv6: false, code: -98, name: EADDRINUSE, message: address already in use
```

 So MASTER_PORT env allows us to change this port which it uses.

mkumatag (approved):
LGTM

PR #81, opened 2025-11-20
Adding workflow to spyre rag ui image

adarshagrawal38:


spyre-rag/ui/Makefile
```diff
@@ -1,12 +1,13 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=rag-ui
 TAG?=v0.0.3
```
mkumatag: let us bump to the next version as we have code update in this repo

PR #98, opened 2025-11-22
Fix usage prints

mayuka-c:
- Move all the user errors to PreRunE
- SilenceUsage at the beginning of RunE -> This has to be done at the beginning of RunE and not at the end of PreRunE because cobra returns the flag errors after PreRunE hook (like MarkFlagRequired)
- Avoid printing usages for internal errors

Usage Errors
---
image

Internal Errors
---
image


Shubhamag12:
looks good, but can you check why error is getting printed like that in attached screenshot?

mayuka-c:
> looks good, but can you check why error is getting printed like that in attached screenshot?

Looks like somewhere format specifier used is wrong, let me check

mayuka-c:
@Shubhamag12 Fixed and updated the image. PTAL

PR #99, opened 2025-11-22
Implement Retry utility and add Retries in create

mayuka-c:
- Implement a new retry utility method
- Add retries for bootstrap configure and model download as part of `application create` flow

Sample Output
---
image

### DIFF
--- a/ai-services/cmd/ai-services/cmd/bootstrap/validate.go
+++ b/ai-services/cmd/ai-services/cmd/bootstrap/validate.go
@@ -125,7 +125,7 @@ func RunValidateCmd(skip map[string]bool) error {
 				s.Fail(err.Error())
 				validationErrors = append(validationErrors, fmt.Errorf("%s: %w", ruleName, err))
 			case constants.ValidationLevelWarning:
-				logger.Warningf(err.Error())
+				s.Stop("Warning: " + err.Error())
 			}
 		} else {
 			s.Stop(rule.Message())


ai-services/internal/pkg/utils/retry.go
```diff
@@ -0,0 +1,55 @@
...
+	var err error
+
+	// Run the function initially and if no error do not proceed with retry attempts
+	err = fn()
+	if err == nil {
+		return nil
+	}
+
+	for i := range attempts {
+		logger.Infof("\n[Retry] Attempt %d/%d\n", i+1, attempts, 0)
```
manju956: Does it print what is being retried like Configure/model download/application create ?

ai-services/internal/pkg/utils/retry.go
```diff
@@ -0,0 +1,55 @@
...
+
+		if err = fn(); err == nil {
+			return nil
+		}
+
+		// At Last attempt — stop
+		if i == attempts-1 {
+			break
+		}
```
Shubhamag12: this actually means that we are retrying `attempts + 1` times, right?

mayuka-c: So essentially attempts here means -> on failure how many times you want to retry.
So first do a normal function call, On failure keep retrying until 3 times more

manju956: Yeah first defecto try + attempts * retry.

manju956 (approved):

PR #105, opened 2025-11-23
Bump up the base image version

adarshagrawal38:


images/etcd/Makefile
```diff
@@ -1,11 +1,12 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=etcd
-TAG?=3.6.5
+ETCD_VERSION?=3.6.5
+TAG?=v0.0.1
```
mkumatag:
```suggestion
TAG?=${ETCD_VERSION}-1
```

mkumatag (changes requested):
let us retain the original versions of the components so that it becomes easy to identify what version we are running post deployment.

mkumatag (approved):
LGTM

PR #112, opened 2025-11-24
Make changes to consume ui port from values.yml and override as part of params

mayuka-c:


ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -607,20 +607,31 @@ func checkForPodStartAnnotation(podAnnotations map[string]string) string {
...
 	hostPortMapping := map[string]string{}
 
-	isContainerPortExposeAnnotation := func(annotation string) (string, bool) {
-		matches := vars.ContainerPortExposeAnnotationRegex.FindStringSubmatch(annotation)
-		if matches == nil {
-			return "", false
-		}
-		return matches[2], true
-	}
+	if portMappings, ok := podAnnotations[constants.PodPortsAnnotationKey]; ok {
```
mkumatag: lets do something like this for the better reading:

```golang
portMappings, ok := podAnnotations[constants.PodPortsAnnotationKey]
if !ok{
  hostPortMapping
}

portMapping := strings.SplitSeq(portMappings, ",")
		for p := range portMapping {
			p = strings.TrimSpace(p)
			if p == "" {
				continue
			}
......
```

PR #124, opened 2025-11-25
Handle queries when db not populated

Niharika0306:
This PR handles the following:
1. When db is not ingested, it raises a 503 error when completions request is made. 
2. Handles the scenario on the UI to communicate the same message to end-user

Screenshot 2025-11-25 at 12 25 57 AM
Screenshot 2025-11-25 at 8 53 50 AM

### DIFF
--- a/spyre-rag/src/retrieve/backend_server.py
+++ b/spyre-rag/src/retrieve/backend_server.py
@@ -4,7 +4,7 @@ import logging
 import os
 import time
 
-from common.db_utils import MilvusVectorStore
+from common.db_utils import MilvusVectorStore, MilvusNotReadyError
 from common.llm_utils import query_vllm_stream, query_vllm_models
 from common.misc_utils import get_model_endpoints, set_log_level
 from retrieve.backend_utils import search_and_answer_backend, search_only
@@ -97,6 +97,8 @@ def get_reference_docs():
             use_reranker,
             vectorstore=vectorstore
         )
+    except MilvusNotReadyError as e:
+        return jsonify({"error": str(e)}), 503   # Service unavailable
     except Exception as e:
         return jsonify({"error": repr(e)})
     return Response(
@@ -145,6 +147,8 @@ def chat_completion():
             use_reranker,
             vectorstore=vectorstore
         )
+    except MilvusNotReadyError as e:
+        return jsonify({"error": str(e)}), 503   # Service unavailable
     except Exception as e:
         return jsonify({"error": repr(e)})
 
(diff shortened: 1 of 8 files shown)


Niharika0306:
p.s. I had also implemented a banner on top to highlight this error before user-input , with a recheck action to check the state without user-input, and thus blocking it. 
Screenshot 2025-11-25 at 12 25 08 AM

But due to some alignment issues, user might miss this and hence to keep a clean action for now - reverted to the above version.

spyre-rag/ui/src/components/customSendMessage.jsx
```diff
@@ -170,7 +196,12 @@ async function customSendMessage(request, _options, instance) {
     });
 
   } catch (err) {
-    console.error("Error during streaming:", err);
+
+    instance.updateIsLoadingCounter('decrease');
+
+    let errorText = "Sorry, something went wrong during streaming.";
```
mkumatag: If we overwrite this variable with `err.error` immediately after this statement, is the initial assignment necessary? Also, `Something went wrong` is a bit vague, but at the same time, are we sure we want to expose raw internal error messages directly to the user via err.error?

spyre-rag/src/retrieve/backend_server.py
```diff
@@ -156,6 +160,21 @@ def chat_completion():
...
+        emb_endpoint = emb_model_dict['emb_endpoint']
+        emb_max_tokens = emb_model_dict['max_tokens']
+        status = vectorstore.check_db_populated(emb_model, emb_endpoint, emb_max_tokens)
+        if status==True:
+            return jsonify({"ready": True}), 200
+        else:
+            return jsonify({"ready": False, "message": "No data ingested"}), 200
+        
+    except Exception as e:
+        return jsonify({"ready": False, "message": str(e)}), 500
```
mayuka-c: Since we are exposing all these APIs, so now we expect them also to use this to cover this scenario (Data not ingested)?

Is it not possible to determine this as part of chat completions endpoint as this would reduce one more call to be made?

spyre-rag/ui/src/components/customSendMessage.jsx
```diff
@@ -11,6 +11,32 @@ const client = new OpenAI({
...
+            nickname: "DocuAgent",
+            user_type: UserType.BOT,
+          },
+        },
+      });
+      return;
+      }
+  } catch (e) {
+    console.log(`Error: ${e.message}`);
+  }
```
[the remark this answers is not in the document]
Niharika0306: Addressed.

manju956 (approved):
lgtm

mayuka-c (approved):
LGTM

spyre-rag/ui/src/components/customSendMessage.jsx
mkumatag: no need to tell sorry, we can keep the error message straight forward text, may be something like: `Error occurred during active stream`

mkumatag (changes requested):

mayuka-c (approved):
LGTM

PR #129, opened 2025-11-25
Use torch protobuf from devpi

yussufsh:


yussufsh:
Checking on this... I do not think we need numpy<2.0 anymore.

mkumatag:
Hope we have done some level of testing with document ingestion?!

yussufsh:
Yes testing was done.

PR #141, opened 2025-11-26
AISERVICES-244: stopping ingest if no files found to ingest

iv1111:
* stopping ingest if no files found to ingest


iv1111:
> make sure to bump the version in makefile and template file to next version

bumped up version in Makefile, what other template do you mean?

PR #145, opened 2025-11-26
Add TOC processing

dharaneeshvrd:
Enable metadata by default
Improvement on finding header text font size by loading pdf once instead of everytime

Improvement stats:
- Converting IBM R1150 intro redbook earlier took around 9 mins, with the TOC processing it reduced to 6 min.
- It also solved the font size extraction issue seen while converting AIX commands pdf which had over 2000 pages by loading the pdf page one time instead of every time a header is processed, which reduced from 17000 secs to 4 secs.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/start.go
+++ b/ai-services/cmd/ai-services/cmd/application/start.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/containers/podman/v5/pkg/domain/entities/types"
+	"github.com/project-ai-services/ai-services/internal/pkg/constants"
 	"github.com/project-ai-services/ai-services/internal/pkg/logger"
 	"github.com/project-ai-services/ai-services/internal/pkg/runtime/podman"
 	"github.com/project-ai-services/ai-services/internal/pkg/utils"
@@ -47,7 +48,7 @@ Note: Logs are streamed only when a single pod is specified, and only after the 
 			return fmt.Errorf("failed to connect to podman: %w", err)
 		}
 
-		return startApplication(cmd, runtimeClient, applicationName, startPodNames)
+		return startApplication(runtimeClient, applicationName, startPodNames)
 	},
 }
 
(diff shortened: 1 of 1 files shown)


mkumatag (changes requested):
Please make sure we bump the version number in the makefile and the template.

spyre-rag/src/ingest/toc.py
```diff
@@ -0,0 +1,157 @@
...
+    tocHeader = titleAndLevel["title"]
+    score = fuzz.partial_ratio(title.lower(), tocHeader.lower())
+    if score >= threshold:
+        return "#" * int(titleAndLevel["level"])
+    return ""
+
+def get_toc(file):
+    toc = {}
+    with open(file, "rb") as fp:
+        try:
```
iv1111: Here exception handling doesn't cover other possible exceptions. Is that intended?

spyre-rag/src/common/llm_utils.py
```diff
@@ -1,10 +1,22 @@
...
 from concurrent.futures import ThreadPoolExecutor, as_completed
 
 from common.misc_utils import get_logger
 from common.settings import get_settings
 
+POOL_SIZE = 10 
+
+adapter = HTTPAdapter(
+    pool_connections=POOL_SIZE, 
+    pool_maxsize=POOL_SIZE,
```
mkumatag: these values will be interesting to deal with! we need to tune them according to our endpoint we will be talking to, e.g: tokenise/embedding apis can take more load and chat completion can limit to batch size allowed for any specific model(e.g: 32)

spyre-rag/src/ingest/pdf_page_resolver.py
```diff
@@ -0,0 +1,108 @@
+from enum import Enum, auto
+from typing import Any, Optional
+from typing import Any
+from pdfminer.pdfdocument import PDFDocument
+from pdfminer.pdfpage import PDFPage, LITERAL_PAGE
+from pdfminer.pdftypes import PDFObjRef
+
+"""
+Code is referred from https://pdfminersix.readthedocs.io/en/latest/howto/toc_target_page.html
```
mkumatag:
```python
# This logic is adapted from the pdfminer.six project. Ref: https://pdfminersix.readthedocs.io/en/latest/howto/toc_target_page.html
# Original Copyright (c) 2019 Yusuke Shinyama, Philippe Guglielmetti & Pieter Marsman.
# Used under the MIT License.
# <add description/text if you have modified/added anything!>
```

mkumatag:
I will defer the code review to @Niharika0306 as I'm not familiar with this codebase

mkumatag:
@dharaneeshvrd please don't forget to bump the version in the makefile

PR #151, opened 2025-11-28
Expose RAG Backend port

mayuka-c:
image

### DIFF
--- a/ai-services/assets/applications/RAG/values.yaml
+++ b/ai-services/assets/applications/RAG/values.yaml
@@ -2,7 +2,7 @@ ui:
   # @description Host port for the RAG UI. If unspecified, a random available port is assigned. Specify a port number to use a custom value.
   port: ""
   # @hidden
-  image: icr.io/ai-services-cicd/rag-ui:v0.0.7
+  image: icr.io/ai-services-cicd/rag-ui:v0.0.8
 
 backend:
   # @description Host port for the OpenAI-compatible RAG service. Defaults to unexposed; assign a port to enable external access.
--- a/ai-services/cmd/ai-services/cmd/bootstrap/bootstrap.go
+++ b/ai-services/cmd/ai-services/cmd/bootstrap/bootstrap.go
@@ -3,6 +3,7 @@ package bootstrap
 import (
 	"fmt"
 
+	"github.com/charmbracelet/lipgloss"
 	"github.com/project-ai-services/ai-services/internal/pkg/logger"
 	"github.com/project-ai-services/ai-services/internal/pkg/validators/root"
 	"github.com/spf13/cobra"
@@ -15,13 +16,23 @@ func BootstrapCmd() *cobra.Command {
 		Short: "Initializes AI services infrastructure",
 		Long: `Bootstrap and configure the infrastructure required for AI services.
 
-The bootstrap command prepares and validates the environment needed
+The bootstrap command configures and validates the environment needed
 to run AI services on Power11 systems, ensuring prerequisites are met
 and initial configuration is completed.
 
 Available subcommands:
-  validate   - Check system prerequisites and configuration
-  configure  - Configure and initialize the AI services infrastructure`,
+  configure - Configure performs below actions
+  • Installs podman on host if not installed
+  • Runs servicereport tool to configure required spyre cards
+  • Initializes the AI services infrastructure
+
+  validate - Checks below system prerequisites
+  • Root user
+  • Power11 server
+  • RHEL OS
+  • LPAR affinity
+  • Spyre cards availability
+  • ServiceReport validation`,
 		Example: `  # Validate the environment
   ai-services bootstrap validate
 
@@ -46,6 +57,10 @@ Available subcommands:
 			}
 
 			logger.Infoln("LPAR boostrapped successfully")
+			logger.Infoln("----------------------------------------------------------------------------")
+			style := lipgloss.NewStyle().Foreground(lipgloss.Color("#32BD27"))
+			message := style.Render("Re-login to the shell to reflect necessary permissions assigned to vfio cards")
+			logger.Infoln(message)
 			return nil
 		},
 	}
--- a/ai-services/cmd/ai-services/cmd/bootstrap/configure.go
+++ b/ai-services/cmd/ai-services/cmd/bootstrap/configure.go
@@ -90,8 +90,6 @@ func RunConfigureCmd() error {
 	s.Stop("Spyre cards configuration validated successfully.")
 
 	logger.Infoln("LPAR configured successfully")
-	logger.Infoln("-------------------------------------------------------------------")
-	logger.Infoln("Re-login to the shell to relect necessary permissions assigned to vfio cards")
 
 	return nil
 }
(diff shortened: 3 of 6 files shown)


ai-services/internal/pkg/cli/helpers/steps.go
```diff
@@ -199,6 +199,11 @@ func fetchDataSpecificInfo(data any, format string) (string, error) {
 
 	err = tmpl.Execute(&result, data)
 	if err != nil {
+		// if there is an error executing the template (Can occur for example if there is no port set)
```
mkumatag: can you elaborate in what situation this will happen?

ai-services/assets/applications/RAG/values.yaml
mkumatag: this needs to be updated with default behaviour saying how this reacts when user doesn't specify/use this param!

ai-services/assets/applications/RAG/templates/chat-bot.yaml.tmpl
```diff
@@ -54,6 +54,9 @@ spec:
           value: "RAG_DB"
         - name: MILVUS_COLLECTION_NAME
           value: "{{ .AppName  }}"
+      ports:
```
mkumatag: here also we can add if statement, block only when port value is nonzero

ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -821,6 +821,10 @@ func constructPodDeployOptions(podAnnotations map[string]string) map[string]stri
 
 	// loop over each of the hostPortMappings to construct the 'publish' option
 	for containerPort, hostPort := range hostPortMappings {
+		if hostPort == "0" {
```
mkumatag: wondering with above if block in the template, we can skip this here! not sure what is the best solution.

mkumatag (changes requested):

ai-services/assets/applications/RAG/templates/chat-bot.yaml.tmpl
mayuka-c: Yes this would be required and better to define it explicity which is the containerport

ai-services/internal/pkg/cli/helpers/steps.go
mayuka-c: This would occur as in the output_vars.yaml we are asking for the port value with "    format: "(index (index .InfraConfig.PortBindings \"5000/tcp\") 0).HostPort"

So if the port itself is not exposed, the index at 0 would fail with error while template execute. So in such scenarios, we will go ahead use the defaultValue if port is not exposed

ai-services/assets/applications/RAG/templates/chat-bot.yaml.tmpl
```diff
@@ -7,7 +7,7 @@ metadata:
     ai-services.io/template: "{{ .AppTemplateName }}"
     ai-services.io/version: "{{ .Version }}"
   annotations:
-    ai-services.io/ports: "{{ .Values.ui.port }}:3000"
+    ai-services.io/ports: "{{ .Values.ui.port }}:3000,{{ .Values.backend.port }}:5000"
```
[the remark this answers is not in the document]
mayuka-c: Sure we can do it

ai-services/internal/pkg/cli/helpers/steps.go
mayuka-c: I have a couple of ways that we can tackle this. I had documented one of the approach in box note which I had written for handling `application upgrade` and 'application rollback'

ai-services/assets/applications/RAG/values.yaml
mayuka-c: Done

ai-services/assets/applications/RAG/steps/info.md
```diff
@@ -1,14 +1,23 @@
...
+{{- end }}
+{{- else }}1. Chatbot UI is not exposed.
+{{- end }}
 
-3. If you want to do the ingestion again, execute below command and wait for the ingestion to be completed before accessing the chatbot to query the new data.
+{{ if ne .BACKEND_PORT "" }}
+{{- if eq .BACKEND_STATUS "running" }}2. Chatbot Backend is available to use at http://{{ .HOST_IP }}:{{ .BACKEND_PORT }}.
+{{- else }}2. Chatbot Backend is unavailable to use. Please make sure '{{ .AppName }}--chat-bot' pod is running.
+{{- end }}
+{{- else }}2. Chatbot Backend is not exposed.
```
mkumatag: why do we want to print this explicit message? then some newbie user may think that I need to expose this backend API or otherwise something is wrong. We will to avoid the message if not really required.

ai-services/assets/applications/RAG/values.yaml
```diff
@@ -1,10 +1,12 @@
 ui:
-  # @description The host port on which the RAG UI will run
+  # @description The host port on which the RAG UI will run. If not set uses random available host port. If set to '0' does not expose the port.
   port: ""
   # @hidden
   image: icr.io/ai-services-cicd/rag-ui:v0.0.7
 
 backend:
+  # @description The host port on which the RAG Backend will run. If not set uses random available host port. If set to '0' does not expose the port.
+  port: ""
```
mkumatag:
```suggestion
  port: "0"
```

ai-services/assets/applications/RAG/values.yaml
```diff
@@ -1,10 +1,12 @@
 ui:
-  # @description The host port on which the RAG UI will run
+  # @description The host port on which the RAG UI will run. If not set uses random available host port. If set to '0' does not expose the port.
```
mkumatag:
```suggestion
  # @description The host port for the RAG UI. If unspecified, a random available port is assigned. Specify a port number to use a custom value.
```

mkumatag (changes requested):

ai-services/assets/applications/RAG/steps/info.md
mayuka-c: The problem was with the templating as if we are using numbering so if no exist cant change the numbering for subsequent ones and also it adds a new line if we hit the else block

image

mayuka-c: Maybe instead of numbering we can go ahead with '-' but removing extra line spaces is becoming a problem

ai-services/assets/applications/RAG/steps/next.md
```diff
@@ -3,4 +3,10 @@
 2. Start the ingestion with below command to feed the documents placed in previous step into the DB
 `ai-services application start {{ .AppName }} --pod={{ .AppName }}--ingest-docs`
 
-3. Chatbot is available to use at http://{{ .HOST_IP }}:{{ .UI_PORT }}.
+{{ if ne .UI_PORT "" }}3. Chatbot UI is available to use at http://{{ .HOST_IP }}:{{ .UI_PORT }}.
+{{- else }}3. Chatbot UI is not exposed.
+{{- end }}
+
+{{ if ne .BACKEND_PORT "" }}4. Chatbot Backend is available to use at http://{{ .HOST_IP }}:{{ .BACKEND_PORT }}.
+{{- else }}4. Chatbot Backend is not exposed.
```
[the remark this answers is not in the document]
mayuka-c: Done

PR #154, opened 2025-11-28
Adding job to scan cves in images

adarshagrawal38:


.github/workflows/minio-image.yml
```diff
@@ -16,6 +16,41 @@ env:
...
+        run: uname -m
+
+      - name: Podman login to ICR
+        uses: redhat-actions/podman-login@v1
+        with:
+          registry: ${{ env.ICR_REGISTRY }}
+          username: ${{ secrets.ICR_USERNAME }}
+          password: ${{ secrets.ICR_APIKEY }}
+      
+      - name: Pull Image
```
mkumatag: this may not work in the PR workflow.

mkumatag (changes requested):

.github/workflows/minio-image.yml
```diff
@@ -16,6 +16,41 @@ env:
...
+        run: |
+          podman pull ${{ env.ICR_REGISTRY }}/${{ env.ICR_NAMESPACE }}/minio:RELEASE.2025-10-15T17-29-55Z-2
+      
+      - name: Grype cve scan
```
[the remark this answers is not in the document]
adarshagrawal38: yes, will do that now I am doing experiments.

.github/workflows/minio-image.yml
adarshagrawal38: yes true, 

trivy scanner fails when we run cves scan on the build image, 

so I was trying to see output when image is in `remote`

.github/workflows/minio-image.yml
```diff
@@ -14,21 +14,37 @@ on:
...
+      
+      - name: Trivy vulnerability scan
+        id: trivy
+        uses: aquasecurity/trivy-action@0.33.1
+        with:
+          input: images/minio/${{ env.IMAGE_FILE_NAME }}
+          format: table
+          severity: CRITICAL,HIGH,MEDIUM
+          exit-code: 1
+        continue-on-error: true
```
mayuka-c: Just a suggestion,
Instead of continue on error, why dont we have a `trivyignore file` maintained, so if we want to ignore a CVE we can add this in a file.
In this way you dont have skip actual errors here and all the ignored CVEs are maintained in a file which can be relooked later when actual fix for CVE is available.

yussufsh: @adarshagrawal38 We need to fail the check when any CVE is found. This will always pass.
What we want is try to complete scan to list all CVEs instead of just exiting out at first encounter. If this is not possible it is fine to just fail.

.github/workflows/minio-image.yml
```diff
@@ -14,21 +14,37 @@ on:
...
+          IMAGE_NAME=$(make image-name REGISTRY=${{ env.ICR_REGISTRY }}/${{ env.ICR_NAMESPACE }} )
+          podman save > ${{ env.IMAGE_FILE_NAME }} $IMAGE_NAME
+      
+      - name: Trivy vulnerability scan
+        id: trivy
+        uses: aquasecurity/trivy-action@0.33.1
+        with:
+          input: images/minio/${{ env.IMAGE_FILE_NAME }}
+          format: table
+          severity: CRITICAL,HIGH,MEDIUM
```
yussufsh: This will eventually be `HIGH,CRITICAL`.
ie. Fail on any HIGH severity CVEs.

.github/workflows/minio-image.yml
```diff
@@ -14,21 +14,37 @@ on:
...
+        working-directory: images/minio
+        run: |
+          IMAGE_NAME=$(make image-name REGISTRY=${{ env.ICR_REGISTRY }}/${{ env.ICR_NAMESPACE }} )
+          podman save > ${{ env.IMAGE_FILE_NAME }} $IMAGE_NAME
+      
+      - name: Trivy vulnerability scan
+        id: trivy
+        uses: aquasecurity/trivy-action@0.33.1
+        with:
+          input: images/minio/${{ env.IMAGE_FILE_NAME }}
```
yussufsh: We should not block the PR where a fix is not available.
`ignore-unfixed` can be used for that purpose.
cc @pravin-dsilva This is just a precheck and I hope we are anyway going to scan the images for CVEs periodically.

yussufsh (changes requested):

mkumatag:
I feel its worth adding this CVE table in the PR post run(adds lot of value)

images/minio/Makefile
```diff
@@ -4,6 +4,7 @@ MINIO_RELEASE=RELEASE.2025-10-15T17-29-55Z
 TAG?=${MINIO_RELEASE}-2
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
+
```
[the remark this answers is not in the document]
adarshagrawal38: done

.github/workflows/minio-image.yml
adarshagrawal38: added `ignore-unfixed` param to true

mkumatag:
> > I feel its worth adding this CVE table in the PR post run(adds lot of value)
> 
> 
> 
> As PR is coming from forked repo there are `permission` issues for posting comments in PR

I see, let us stick to reporting in the PR job then..

mkumatag (dismissed):

mayuka-c (dismissed):
LGTM,

One suggestion, better to maintain trivyignore for each ContainerFile we have like for minio, etcd, rag etc as each image will have their own CVE's and help us to distinguish the CVE which we have ignored.

.github/workflows/minio-image.yml
```diff
@@ -36,30 +36,9 @@ jobs:
...
-          severity: CRITICAL,HIGH,MEDIUM
-          exit-code: 1
-          ignore-unfixed: true
\ No newline at end of file
+  scan:
+    uses: ./.github/workflows/scanner.yml
+    with:
+      path: 'images/minio'
+      image-name: 'minio'
+      registry: 'icr.io/ai-services-cicd'
```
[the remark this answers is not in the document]
adarshagrawal38: Using `repository var` instead of `env`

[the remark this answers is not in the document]
yussufsh: It shows `pre-scan` check in the pr :(
I think we can foget about reusable workflow for now.

PR #161, opened 2025-11-28
AISERVICES-280

iv1111:


mayuka-c (commented):
LGTM,

@iv1111 I guess you need to bump up the tag for RAG backend image in makefile

iv1111:
> LGTM,
> 
> @iv1111 I guess you need to bump up the tag for RAG backend image in makefile

Done that + using http methods specifically in Flask now instead of `route`.

PR #162, opened 2025-12-01
Show Pod Health in `application ps`  and container names with status

mayuka-c:
Technical Details
---
- Show Pod healthy/Unhealthy as part of `application ps`. Pod is set to healthy/unhealthy only for running pods. If pod is healthy then its considered to be ready to serve or else not. With this we will know whether the pod eventhough its running, are all the containers healthy or not.

- Also show container health along with names.

image


ai-services/cmd/ai-services/cmd/application/ps.go
```diff
@@ -175,3 +180,51 @@ func getContainerNames(runtimeClient *podman.PodmanClient, pod *types.ListPodsRe
...
 }
+
+func getPodStatus(runtimeClient *podman.PodmanClient, pInfo *types.PodInspectReport) string {
+	// if the pod Status is running, make sure to check if its healthy or not
+	if pInfo.State == "Running" {
+		healthyContainers := 0
+		for _, container := range pInfo.Containers {
+			cInfo, err := runtimeClient.InspectContainer(container.ID)
+			if err != nil {
+				// skip container if inspect failed
```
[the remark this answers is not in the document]
mayuka-c: Done

ai-services/cmd/ai-services/cmd/application/ps.go
```diff
@@ -95,12 +97,20 @@ func runPsCmd(runtimeClient *podman.PodmanClient, appName string) error {
...
-			//Skip pods which are not linked to ai-services
+			// skip pods which are not linked to ai-services
+			continue
+		}
+
+		// do pod inspect
+		pInfo, err := runtimeClient.InspectPod(pod.Id)
+		if err != nil {
+			// log and skip pod if inspect failed
+			logger.Errorf("Failed to do pod inspect: '%s' with error: %v", pod.Id, err)
```
manju956: Can we have pod name instead of id, will be useful to understand better when the error message is seen on console.

ai-services/cmd/ai-services/cmd/application/ps.go
```diff
@@ -158,13 +164,13 @@ func getContainerNames(runtimeClient *podman.PodmanClient, pod *types.ListPodsRe
 		cInfo, err := runtimeClient.InspectContainer(container.Id)
 		if err != nil {
 			// skip container if inspect failed
+			logger.Infof("failed to do container inspect for podID: '%s', containerID: '%s' with error: %v", pod.Id, container.Id, err, 2)
```
manju956: here too, can we have podname and container name instead of IDs?

ai-services/cmd/ai-services/cmd/application/ps.go
mayuka-c: Done

PR #164, opened 2025-12-01
Fix sos config issue in bootstrap configure

manju956:
Servicereport tool when run in repair mode, adds below 2 lines to `/etc/sos/sos.conf`. Hence its required to volume mount `/etc/sos` host directory during `ai-services bootstrap configure` as well.
```
podman.logs = true
podman.all = true
```


manju956:
> Have we never ran the bootstrap configure on a fresh machine? If not we have to do it to uncover the issues like this.

Yeah right. this issue was caught by @manalilatkar on one such fresh setup.

PR #167, opened 2025-12-02
Use podman cli way of pod start

manju956:
- With SDK way of starting pod as part of `ai-services application start` flow, we are seeing both vLLM and reranker containers getting crash with error `The Linux VFIO kernel module could not create the TCE window for the sPAPR IOMMU.  This is an unexpected condition`. However, when switch to podman cli way of starting pod, it seemless works, both vllm and reranker containers successfully booted. 
We are sticking to this cli way of starting pod until the root cause of why SDK way of starting pod resulting in crash is identified.

### DIFF
--- a/.github/workflows/minio-image.yml
+++ b/.github/workflows/minio-image.yml
@@ -8,6 +8,7 @@ on:
   pull_request:
     paths:
       - 'images/minio/**'
+      - .github/workflows/minio-image.yml
     branches:
       - main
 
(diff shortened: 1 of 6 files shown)


manju956:
> * With SDK way of starting pod as part of `ai-services application start` flow, we are seeing both vLLM and reranker containers getting crash with error `The Linux VFIO kernel module could not create the TCE window for the sPAPR IOMMU.  This is an unexpected condition`. However, when switch to podman cli way of starting pod, it seemless works, both vllm and reranker containers successfully booted.
>   We are sticking to this cli way of starting pod until the root cause of why SDK way of starting pod resulting in crash is identified.

bug ticket - https://jsw.ibm.com/browse/AISERVICES-313

manju956:
> @manju956 @Shubhamag12 can one of you create the future jira issue for this To-Do before merging this?

ticket created: https://jsw.ibm.com/browse/AISERVICES-316

PR #168, opened 2025-12-02
Limit concurrent requests at backend-server

Niharika0306:
I have added the semaphore locks to handle concurrent requests, and set it to 32 ( this was set keeping in mind the MAX_BATCH_SIZE - to allow parallel requests). 
This PR handles the display on the UI side as well. 
Screenshot 2025-12-02 at 10 22 32 AM
Screenshot 2025-12-02 at 10 57 30 AM

However, the actual intent is still not fully served. 
since the /chat/completions api returns a stream generator as response, the lock only waits till /reference api , and frees as soon as the stream generator is returned, whereas ideally it should wait till the streaming is completed. 

This should be handled by wrapping the generator explicitly in the limit_concurrency function I guess.

p.s. I achieved the above logs and testing by putting a wait-time of 10 secs in /chat/completion api

- The PR also removes /generate api from the server, and its utility functions that are unused.

### DIFF
--- a/spyre-rag/src/Makefile
+++ b/spyre-rag/src/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=rag
-TAG?=v0.0.18
+TAG?=v0.0.19
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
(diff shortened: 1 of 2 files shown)


iv1111 (commented):
```
def my_stream():
        try:
           for chunk in query_vllm_stream(...):
               yield chunk
        finally:
            chat_stream_semaphore.release() # when our stream finishes for any reason only then we release semaphore

.......

resp_text = stream_with_context(my_stream())
```

I think something like that would solve stream locking issue.

PR #171, opened 2025-12-02
Integrating cves check scan workflow

adarshagrawal38:


mkumatag:
> @mkumatag Merging this will lead to a rebuild for each image. Do we want that? What do you suggest?

which is bad :) rebuilding is fine but we shouldn't be pushing it!

PR #172, opened 2025-12-02
Add new 'Pod Created Column' in `application ps`

mayuka-c:
**Technical Details**
---
Adding a new POD creation column as part of application ps. Also making the CREATED column more human readable.

**Output**
---
image


mkumatag (approved):
All these column ordering and the text we are printing needs to be revisited later post GA

PR #174, opened 2025-12-02
Address another round of ingestion logs improvements

dharaneeshvrd:
Add support for passing LOG_LEVEL to the ingestion
Add progress bar for various activities when DEBUG log level is set


mayuka-c (commented):
@dharaneeshvrd What would happen if the Log_level set by user is wrong for Ex: "xyz". Will the backend server crash?

mayuka-c:
> > @dharaneeshvrd What would happen if the Log_level set by user is wrong for Ex: "xyz". Will the backend server crash?
> 
> No, it will ignore.

So it uses Info by default?

mkumatag (approved):

dharaneeshvrd:
hold

PR #175, opened 2025-12-02
Update and include a few dependencies for docling

yussufsh:


images/rag-base/Makefile
```diff
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=rag-base
-TAG?=v0.0.8
+TAG?=v0.0.9
```
dharaneeshvrd: We are changing the rag image's FROM version once it successfully builds is it?

dharaneeshvrd (approved):
lgtm

PR #177, opened 2025-12-03
Do early return if all the pods for given app are already deployed

mayuka-c:
Technical Details
---
- If all the pods for given app are already deployed, then do an early exit instead of proceeding further.

image

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/create.go
+++ b/ai-services/cmd/ai-services/cmd/application/create.go
@@ -63,7 +63,19 @@ var createCmd = &cobra.Command{
 		if len(rawArgParams) > 0 {
 			argParams, err = utils.ParseKeyValues(rawArgParams)
 			if err != nil {
-				return fmt.Errorf("error validating params flag: %v", err)
+				return fmt.Errorf("error validating params flag: %w", err)
+			}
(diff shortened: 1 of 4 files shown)


mayuka-c:
> @mayuka-c Any specific reason for adding this check? what happen with and without this fix?

Since most of our operation is idempotent, doesnt cause any issue but simply it does image pull and other checks (next steps). So wanted to avoid this and just do early exit right at the beginning if all are deployed

mayuka-c:
Also one of the suggestion from the testing team was to terminate early

mkumatag:
> > @mayuka-c Any specific reason for adding this check? what happen with and without this fix?
> 
> Since most of our operation is idempotent, doesnt cause any issue but simply it does image pull and other checks (next steps). So wanted to avoid this and just do early exit right at the beginning if all are deployed

please add all these details in the description of the PR, sometime background is helpful :)

mayuka-c:
> > > @mayuka-c Any specific reason for adding this check? what happen with and without this fix?
> > 
> > 
> > Since most of our operation is idempotent, doesnt cause any issue but simply it does image pull and other checks (next steps). So wanted to avoid this and just do early exit right at the beginning if all are deployed
> 
> please add all these details in the description of the PR, sometime background is helpful :)

Sure will be adding it in description. Thanks @mkumatag :)

PR #178, opened 2025-12-03
Update RAG ChatBot Favicon

mayuka-c:
image


mkumatag (changes requested):
please bump the version in both makefile and template

adarshagrawal38:
> @mkumatag @adarshagrawal38 I was thinking why not have a workflow_dispatch job in github action. So that we need not have to bump version for every single change. We can trigger manually the build and push whenever required

hey @mayuka-c , or shall we bump version up, When there is a release cut going to happen

Then on release cut we can publish the image to icr, instead of publish image after every code change.

Shall we discuss Image version idea [here](https://ibm-systems.slack.com/archives/C09M6728MA4/p1764758675627129)

PR #180, opened 2025-12-03
Reduce memory limits for vllm and ingest

dharaneeshvrd:
Set restartPolicy to Never for ingest-docs & clean-docs, since these are not meant to be restarted on failure


ai-services/assets/applications/rag/templates/ingest-docs.yaml.tmpl
```diff
@@ -19,9 +19,9 @@ spec:
         - "ingest"
       resources:
         requests:
-          memory: "128Gi"
+          memory: "32Gi"
```
mkumatag: can you make both of these memory values to 40Gi?

mkumatag (approved):

PR #185, opened 2025-12-04
change platform rule name to rhel

Shubhamag12:
https://jsw.ibm.com/browse/AISERVICES-325


mkumatag (approved):
we will add the rhaiis when we really find a way to add it back.

PR #193, opened 2025-12-04
Use proper request pool size

dharaneeshvrd:
Reduced the batch size used in classify table method from 128 to 32 to match the vLLM's max batch size


yussufsh (commented):
lgtm
Thanks!

mkumatag (approved):

PR #207, opened 2025-12-08
Refactoring of backend server and reranker utils.

iv1111:
- lower log level for document excerpts
- lower log level for score threshold
- moved logging start to earlier time in backend server
- parametrised backend server logging level
- removed unused `main` function
- removed default function values

### DIFF
--- a/ai-services/assets/applications/rag/values.yaml
+++ b/ai-services/assets/applications/rag/values.yaml
@@ -8,7 +8,7 @@ backend:
   # @description Host port for the OpenAI-compatible RAG service. Defaults to unexposed; assign a port to enable external access.
   port: "0"
   # @hidden
-  image: icr.io/ai-services-cicd/rag:v0.0.22
+  image: icr.io/ai-services-cicd/rag:v0.0.23
 
 ingest:
   # @hidden
--- a/spyre-rag/src/Makefile
+++ b/spyre-rag/src/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=rag
-TAG?=v0.0.22
+TAG?=v0.0.23
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
 CONTAINER_BUILDER?=podman


dharaneeshvrd (commented):
@iv1111 once we add the config in values.yaml, we have to refer it in pod deployment like [this](https://github.com/IBM/project-ai-services/blob/221f6fc5c5f1d5be36f0ee0fec1d324684d79d7b/ai-services/assets/applications/rag/templates/ingest-docs.yaml.tmpl#L44)
Please make necessary changes

mkumatag:
> @iv1111 can we please remove the default values here
> 
> https://github.com/IBM/project-ai-services/blob/221f6fc5c5f1d5be36f0ee0fec1d324684d79d7b/spyre-rag/src/retrieve/reranker_utils.py#L30
> 
> ?
> @mkumatag hope we can accommodate this minor cleanup!

we are always overriding it right? and there is one more line post that also need to be altered. If functionality wise it is not making any impact - please avoid making any change for now.

mkumatag (changes requested):

Niharika0306:
> > @iv1111 We have to move this block https://github.com/IBM/project-ai-services/blob/main/spyre-rag/src/retrieve/backend_server.py#L198toL205 before importing other modules, to propagate the log level to those modules, can you please take care of this as well?
> 
> @dharaneeshvrd Done, PTAL.
> 
> > Yes, we are always overriding it. These are some public instance links that was pointed by research team.
> > but yes, as you said - its not super important now.
> 
> @Niharika0306 but it looks like here we call it without passing last three params relying on default values: https://github.com/iv1111/project-ai-services/blob/2a865c06399b6be6d80548c7929a583b59736128/spyre-rag/src/retrieve/reranker_utils.py#L69
> 
> @mkumatag Since this PR won't go into GA1, thus I could handle the change here right away.

yes, which is called only when the script is invoked separately. Nonetheless, if this PR is not being merged for GA1, we can go ahead and remove the call and the defaults too.

spyre-rag/src/retrieve/backend_server.py
```diff
@@ -1,10 +1,21 @@
...
+level = os.getenv("LOG_LEVEL", "").removeprefix("--").lower()
+if level != "":
+    if "debug" in level:
+        log_level == logging.DEBUG
+    elif not "info" in level:
+        raise Exception(f"Unknown LOG_LEVEL passed: '{level}'")
+set_log_level(log_level)
+
+
+from flask import Flask, request, jsonify, Response, stream_with_contex
```
dharaneeshvrd: typo
```suggestion
from flask import Flask, request, jsonify, Response, stream_with_context
```

iv1111: Good catch, thanks.

dharaneeshvrd (approved):

PR #197, opened 2025-12-05
validate params from values.yaml

Shubhamag12:
# Description
We were only validation params which are present as part of description (Exposed to user). As a result, params that are hidden but can be overridden, was not working

# Fix
Validate all params instead of what is exposed

# Testing

Case 1: valid param
```
[root@sagarwal ai-services]# ./bin/ai-services application create temp --template rag --params "ingest.log_level=DEBUG"
Validating the LPAR environment before creating application 'temp'...
Creating application 'temp' using template 'rag'
✔ SMT level configured successfully
No existing pods found for application: temp
Downloading container images required for application template rag:
Downloading image: icr.io/ai-services/tools:0.5...
Pulling image icr.io/ai-services/tools:0.5...
Trying to pull icr.io/ai-services/tools:0.5...
^C

[root@sagarwal ai-services]# ./bin/ai-services application create temp --template rag --params "ui.port=12312"
Validating the LPAR environment before creating application 'temp'...
Creating application 'temp' using template 'rag'
✔ SMT level configured successfully
No existing pods found for application: temp
Downloading container images required for application template rag:
Downloading image: icr.io/ai-services/tools:0.5...
Pulling image icr.io/ai-services/tools:0.5...
^C
```

Case 2: Invalid param
```
[root@sagarwal ai-services]# ./bin/ai-services application create temp --template rag --params "u.port=12312"
Error: failed to load params for application: unsupported parameter: u.port

[root@sagarwal ai-services]# ./bin/ai-services application create temp --template rag --params "ui.ort=12312"
Error: failed to load params for application: unsupported parameter: ui.ort
```

PR #209, opened 2025-12-09
Fix Fetch Values PreRun during `application create`

mayuka-c:
Fix for this: https://github.com/IBM/project-ai-services/pull/197

- Earlier the values was read from values.yml only when the `params` flag was passed as part of `application create`
- Do Values Map fetch even though the `params` arg is not passed


mkumatag:
make sure we test with both param set vs default

PR #213, opened 2025-12-09
Bug fix - Include tokenize/detokenize APIs to call via request SESSION to avoid ephemeral port exhaustion issue

dharaneeshvrd:
- Fix error log for vllm calls
- Add clear comment on the reasoning behind SESSION usage
```
# SESSION object will be used by instruct and embedding endpoints. Hence keeping pool_connections = 2
    # Need to use SESSION object for following reasons:
    # - To limit the number of concurrent requests getting created to instruct vLLM's API to 32
    # - To fix the ephemeral port exhaustion issue during chunking, since numerous tokenize calls are made to embedding server
```


dharaneeshvrd:
Apologise for missing this scenario on yesterday's [PR](https://github.com/IBM/project-ai-services/pull/202)

PR #215, opened 2025-12-09
This is proposed addition for performance data logging

carll99:
Here is the first pass at what we would like to see enabled for performance tracking.

Posting this as requested from earlier meeting.  Will discuss the PR in our next meeting.

### DIFF
--- a/spyre-rag/src/retrieve/backend_server.py
+++ b/spyre-rag/src/retrieve/backend_server.py
@@ -2,6 +2,7 @@ from flask import Flask, request, jsonify, Response, stream_with_context
 import json
 import logging
 import os
+import time
 from threading import BoundedSemaphore
 from functools import wraps
 
@@ -18,7 +19,8 @@ TRUNCATION  = True
 emb_model_dict = {}
 llm_model_dict = {}
 reranker_model_dict = {}
-
+from common.misc_utils import get_logger
+logger = get_logger("backend_server")
 settings = get_settings()
 concurrency_limiter = BoundedSemaphore(settings.max_concurrent_requests)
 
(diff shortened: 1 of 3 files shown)


spyre-rag/src/retrieve/backend_server.py
```diff
@@ -145,7 +147,10 @@ def chat_completion():
             return jsonify({"error": "Server busy. Try again shortly."}), 429
 
         try:
+            start_time = time.time()
             vllm_stream = query_vllm_stream(prompt, docs, llm_endpoint, llm_model, stop_words, max_tokens, temperature, stream, dynamic_chunk_truncation=TRUNCATION)
+            request_time = time.time() - start_time
+            logger.info(f"Perf data: rag answer time = {request_time}")
```
dharaneeshvrd: What is the difference between `rag answer time` and `llm inferencing time` in llm_utils? It looks it would be produce same timings.

spyre-rag/src/retrieve/backend_server.py
```diff
@@ -145,7 +147,10 @@ def chat_completion():
             return jsonify({"error": "Server busy. Try again shortly."}), 429
 
         try:
+            start_time = time.time()
             vllm_stream = query_vllm_stream(prompt, docs, llm_endpoint, llm_model, stop_words, max_tokens, temperature, stream, dynamic_chunk_truncation=TRUNCATION)
```
iv1111: This call to `query_vllm_stream` in streaming mode returns a generator and not a final answer. This generator (like a C++ iterator) is then read chunk by chunk with each piece being sent to Web UI even after this Python backend function returns. So if you want to measure properly you should probably test non-streaming case only.

iv1111: @Niharika0306 chime in please to confirm.

iv1111:
@carll99 You could try out this for request id retrieval:

```
from flask import request
.....
req_id = id(request)
```

PR #217, opened 2025-12-10
Fix release version part of the workflow

mkumatag:
This fixes the Release name not populating the name properly(version)


adarshagrawal38 (approved):
LGTM

PR #218, opened 2025-12-10
Fix patch links and use commits instead

yussufsh:
The torchvision patch file name was changed yesterday in the build-scripts repo. This led to a build failure for rag-base docling Containerfile. 

Now we are locked on a commit.

Also, using upstream build repo link for docling patch instead of the forked repo link used before.

### DIFF
--- a/images/rag-base/Makefile
+++ b/images/rag-base/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=rag-base
-TAG?=v0.0.12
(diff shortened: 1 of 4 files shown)


mkumatag (changes requested):
better carry the patches here in the repo instead of pulling them runtime!

adarshagrawal38:
> bingo! we have some build failure.!
> 
> @adarshagrawal38 I see image is built twice for PR, wondering if we can optimise this workflow to use runners optimal.

yes we can optimize, to avoid rebuilding image.

PR #219, opened 2025-12-10
This is proposed addition for performance data logging

carll99:
This is an updated PR with just the sign-off added.  Goal was to have the very minimal code changes to give very basic logging info for the initial release.  More detailed, code invasive, timings will we worked on for a future release/PR.


iv1111:
What is the state of this?

PR #267, opened 2026-01-16
enable ai-services to run as non-root-user

Shubhamag12:
Context: https://ibm.ent.box.com/file/2102027246213

# Description
## Problem Statement
Currently, ai-services only supports execution as the root user. 

## Solution
This PR enables ai-services to run as a non-root user. The tool now supports a hybrid approach:
- **bootstrap (validate/configure)**: Runs with `sudo` for one-time system setup
- **runtime operations (application commands)**: Run as a regular user without elevated privileges

## User Setup Requirements

To use ai-services as a non-root user, the following one-time setup is required:

### 1. Add user to sudo group (wheel)
```bash
sudo usermod -aG wheel <username>
```
This grants the user permission to execute specific commands with sudo when needed.

### 2. Enable systemd user session persistence
```bash
sudo loginctl enable-linger <username>
```
This ensures the user's systemd instance persists even when not logged in, which is required for:
- Rootless podman socket to remain active
- User services to run in the background

### Rootless Podman
- Containers run as the regular user on the host (not root)
- Uses user namespaces: processes appear as "root" inside containers but are actually the user's UID on the host
- Podman socket created at `/run/user/<UID>/podman/podman.sock` (user-specific, not system-wide)

### Directory Permissions
- Bootstrap creates directories with ownership set to `$SUDO_USER`
- Regular user has full read/write access to application directories
- No root privileges required for runtime operations

### Environment Variables
- `XDG_RUNTIME_DIR`: Automatically set if missing (required for user systemd operations)
- `SUDO_USER`: Used during bootstrap to determine the actual user for ownership settings

## Q&A

Q: Why do we need to add the non-root user to the sudo group?
A: We still need certain commands to be executed with sudo, like `ai-services bootstrap`, as it configures and modifies the env which requires root permissions. 

Q: Which commands would be run without sudo?
A: Except `bootstrap`, every other command would be run without sudo.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/create.go
+++ b/ai-services/cmd/ai-services/cmd/application/create.go
@@ -375,15 +375,28 @@ func getSMTLevel(output string) (int, error) {
 
 func setSMTLevel() error {
 	/*
-		1. Fetch current SMT level
-		2. Fetch the target SMT level
+		1. Fetch the target SMT level
+		2. Fetch current SMT level
 		3. Check if SMT level is already set to target value
 		4. If not, set it to target value
 		5. Verify again
 	*/
 
-	// 1. Fetch Current SMT level
-	cmd := exec.Command("ppc64_cpu", "--smt")
+	// 1. Fetch the target SMT level
+	targetSMTLevel, err := getTargetSMTLevel()
+	if err != nil {
+		return fmt.Errorf("failed to get target SMT level: %w", err)
+	}
+
+	if targetSMTLevel == nil {
+		// No SMT level specified in metadata.yaml
+		logger.Infof("No SMT level specified in metadata.yaml. Keeping it to current level")
+
+		return nil
+	}
+
+	// 2. Fetch Current SMT level
+	cmd := exec.Command("sudo", "ppc64_cpu", "--smt")
 	out, err := cmd.CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("failed to check current SMT level: %v, output: %s", err, string(out))
@@ -392,19 +405,6 @@ func setSMTLevel() error {
 	currentSMTlevel, err := getSMTLevel(string(out))
 	if err != nil {
 		return fmt.Errorf("failed to get current SMT level: %w", err)
-	}
-
-	// 2. Fetch the target SMT level
-	targetSMTLevel, err := getTargetSMTLevel()
-	if err != nil {
-		return fmt.Errorf("failed to get target SMT level: %w", err)
-	}
-
-	if targetSMTLevel == nil {
-		// No SMT level specified in metadata.yaml
-		logger.Infof("No SMT level specified in metadata.yaml. Keeping it to current level: %d\n", currentSMTlevel)
-
-		return nil
 	}
 
 	// 3. Check if SMT level is already set to target value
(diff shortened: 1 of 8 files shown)


ai-services/internal/pkg/runtime/podman/podman.go
```diff
@@ -47,6 +49,42 @@ func NewPodmanClient() (*PodmanClient, error) {
...
+		u.Uid,
+	), nil
+}
+
+func userPodmanURI(uid int) string {
+	return fmt.Sprintf(
+		"unix:///run/user/%d/podman/podman.sock",
+		uid,
+	)
+}
```
mayuka-c: Just a check. Did all the pods deployment worked fine with non root?
Did you test by deploying the RAG application?

ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -375,38 +375,38 @@ func getSMTLevel(output string) (int, error) {
...
 
 		return nil
 	}
 
+	// 2. Fetch Current SMT level
+	cmd := exec.Command("sudo", "ppc64_cpu", "--smt")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to check current SMT level: %v, output: %s", err, string(out))
+	}
```
mayuka-c: If we are using sudo, will this prompt for the password to be entered by user here?

ai-services/internal/pkg/runtime/podman/podman.go
Shubhamag12: I tried this on a 3 spyre card setup, where all pods were in healthy state except vllm as it required 4 cards for instruct and 1 for reranker. I configured instruct to use 2 cards and hence it was unhealthy

ai-services/cmd/ai-services/cmd/application/create.go
Shubhamag12: > but I think here it might fail, as exec.Command doesn't attach to terminal stdin by default afaik.

confirmed this, I have disabled the spinner for smt level, since we might need user input for this

Shubhamag12: I dont think we should do that
Its upto admin on how the user is setup, passwordless or with password

manju956: Ok, in that case we can have documentation section to capture this?

Shubhamag12: yes, we should have that

Shubhamag12: User will anyway have permissions to run as sudo, and since moving it to configure.go cannot be possible as we want to keep this smt value configurable as per application, thats why thought of adding sudo.
CC: @mkumatag

mayuka-c: Not sure. Maybe we can have a default constant value set I guess during bootstrap. 
For now, I guess we need to document that only root user and non root users with sudo privileges can run.

ai-services/internal/pkg/validators/root/root.go
```diff
@@ -20,11 +20,18 @@ func (r *RootRule) Name() string {
...
 
-	logger.Infoln("Checking root privileges", logger.VerbosityLevelDebug)
+		return nil
+	}
 
-	if euid != 0 {
-		return fmt.Errorf("current user is not root (EUID: %d)", euid)
+	if euid != 0 && os.Getenv("XDG_RUNTIME_DIR") == "" {
+		uid := os.Getuid()
+		logger.Infoln("running command as %s", uid, logger.VerbosityLevelDebug)
```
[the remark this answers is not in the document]
Shubhamag12: wait, my bad, while creating the application we get `failed to pull image registry.redhat.io/rhaiis/vllm-spyre-rhel9:3.2.5: reading JSON file "/run/containers/1001/auth.json": open /run/containers/1001/auth.json: permission denied` because it is not able to lookup the directory.
So yes, this is required.

Apologies for creating the confusion.

mkumatag (changes requested):
initial comments, overall flow looks really complicated, might need some more time to go through this code and the description.

mkumatag (approved):

Shubhamag12:
Since there are a lot more changes we had in, will keep this as a reference PR and will raise new PR as per requirements

PR #321, opened 2026-02-13
provide language support

Niharika0306:
This PR provides functionality for language_detection and  prompt selection based on language.


spyre-rag/src/retrieve/backend_utils.py
```diff
@@ -1,11 +1,35 @@
...
 settings = get_settings()
 
+def cosine(a, b):
+    return np.dot(a, b) / (np.linalg.norm(a) * np.linalg.norm(b))
+
+def detect_language(query, emb_model, emb_endpoint, max_tokens):
+    embedder = get_embedder(emb_model, emb_endpoint, max_tokens)
+    user_emb = embedder.embed_query(query)
+    ref_texts = {
+        "en": "This is an English sentence.",
```
Niharika0306: Will move these centroids( eng and german ref statements) to settings.json, and will see if I can store the embedding computation in memory somewhere as part of the vectorstore, instead of computing for every query.

PR #341, opened 2026-02-19
Application Create for Openshift - 2

mayuka-c:
- Use Helm Wait strategy: Watcher
- Add default timeout in app metadata.yml
- Support `timeout` flag as part of application creation to override the default timeout.
- Add spinner

### DIFF
--- a/ai-services/assets/applications/rag-dev/metadata.yaml
+++ b/ai-services/assets/applications/rag-dev/metadata.yaml
@@ -3,6 +3,5 @@ description: "Retrieval Augmented Generation (RAG) application that combines a v
               and a retrieval mechanism to provide accurate and context-aware responses based on ingested documents."
 hidden: true
 smtLevel: 2
-runtimes:
-  - name: openshift
-    timeout: 30m
+openshift:
+  timeout: 20m
--- a/ai-services/assets/applications/rag-dev/metadata.yaml
+++ b/ai-services/assets/applications/rag-dev/metadata.yaml
@@ -3,3 +3,6 @@ description: "Retrieval Augmented Generation (RAG) application that combines a v
               and a retrieval mechanism to provide accurate and context-aware responses based on ingested documents."
 hidden: true
 smtLevel: 2
+runtimes:
+  - name: openshift
+    timeout: 30m
--- a/ai-services/cmd/ai-services/cmd/application/create.go
+++ b/ai-services/cmd/ai-services/cmd/application/create.go
@@ -223,7 +223,7 @@ func initOpenShiftFlags() {
 		&timeout,
 		"timeout",
 		0, // default
-		"Timeout for the operation (e.g. 10s, 2m, 1h)",
+		"Timeout for the operation (e.g. 10s, 2m, 1h). Supported for runtime set to openshift only.",
 	)
 }
 
--- a/ai-services/cmd/ai-services/cmd/application/create.go
+++ b/ai-services/cmd/ai-services/cmd/application/create.go
@@ -23,12 +23,12 @@ import (
 
 // Variables for flags placeholder.
 var (
-	// common flags
+	// common flags.
 	templateName string
 	rawArgParams []string
 	argParams    map[string]string
 
-	// podman flags
+	// podman flags.
 	skipModelDownload     bool
 	skipImageDownload     bool
 	skipChecks            []string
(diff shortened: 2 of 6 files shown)


ai-services/assets/applications/rag-dev/metadata.yaml
```diff
@@ -3,3 +3,6 @@ description: "Retrieval Augmented Generation (RAG) application that combines a v
               and a retrieval mechanism to provide accurate and context-aware responses based on ingested documents."
 hidden: true
 smtLevel: 2
+runtimes:
+  - name: openshift
+    timeout: 30m
```
yussufsh:
```suggestion
openshift:
    timeout: 30m
```
Suggest this structure for easier reference instead of looping and checking by name.

ai-services/internal/pkg/application/openshift/create.go
```diff
@@ -8,22 +8,51 @@ import (
...
 	tp := templates.NewEmbedTemplateProvider(templates.EmbedOptions{Runtime: vars.RuntimeFactory.GetRuntimeType()})
+
+	s := spinner.New("Setting the operation timeout...")
+	s.Start(ctx)
+	// populate the operation timeout
+	if timeout == 0 {
+		// load metadata.yml to read the app metadata
+		appMetadata, err := tp.LoadMetadata(opts.TemplateName, false)
+		if err != nil {
+			return fmt.Errorf("failed to read the app metadata: %w", err)
```
Shubhamag12:
```suggestion
            s.Fail("failed to create application")
			return fmt.Errorf("failed to read the app metadata: %w", err)
```

Shubhamag12 (commented):
minor comments, rest lgtm

ai-services/internal/pkg/application/openshift/create.go
```diff
@@ -3,52 +3,123 @@ package openshift
...
+	}
+
+	//nolint:godox
+	// TODO: Step4: print Next steps
+
+	return nil
+}
+
+func getOperationTimeout(ctx context.Context, tp templates.Template, opts types.CreateOptions) (time.Duration, error) {
+	s := spinner.New("Setting the operation timeout...")
```
yussufsh: I do not think a spinner is needed for fetching timeout value.

ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -22,15 +23,21 @@ import (
...
 	skipImageDownload     bool
 	skipChecks            []string
-	rawArgParams          []string
-	argParams             map[string]string
 	valuesFiles           []string
 	rawArgImagePullPolicy string
 	imagePullPolicy       image.ImagePullPolicy
+
+	// openshift flags.
+	timeout time.Duration
```
yussufsh: Do we need validation for this new param?

ai-services/internal/pkg/application/openshift/create.go
mayuka-c: For now I have added a spinner.
I had a closer look at our create PreRunE here: https://github.com/IBM/project-ai-services/blob/0d43cbe8eb69072d2dc9228d1e8d3c83155b72ee/ai-services/cmd/ai-services/cmd/application/create.go#L45.

Looks like we are already reading appMetadata there. So I guess I can set the timeout there itself before even coming to the internal pkg layer. So probably will be moving this there.

I have created a new task as quite a few things that we validate in PreRunE needs to be modified based on Podman or Openshift runtimes. So will be taking up seperately.

Shubhamag12 (dismissed):
lgtm

ai-services/internal/pkg/helm/helm.go
```diff
@@ -34,36 +48,54 @@ func NewHelm(namespace string) (*Helm, error) {
...
-	//nolint:godox
-	// TODO: Replace the WaitStrategy to watcher and also add timeout
-	upgradeClient.WaitStrategy = "hookOnly"
+	upgradeClient.WaitStrategy = kube.StatusWatcherStrategy
+
+	upgradeClient.Timeout = opts.Timeout
+	if upgradeClient.Timeout <= 0 {
+		// use default timeout if set to 0 or negative
+		upgradeClient.Timeout = defaultWaitingTimeout
+	}
```
yussufsh: Why are we maintaining 2 default timeouts? Can we directly use what is provided considering `getOperationTimeout` does most of the stuff.

yussufsh (changes requested):

yussufsh (approved):
lgtm

PR #342, opened 2026-02-19
German language support

Niharika0306:
After couple of iterations and experimentations, we have locked on this approach for now to support german questions and german context. 
The language detection and response translation will be handled by the prompt, and the cross-lingual information retrieval is handled by the granite-278m-multilingual embedding model which is already being used. 
Will upload some test results here soon.


dharaneeshvrd:
We can close this since we are going to use dual prompt based on detected language

PR #343, opened 2026-02-19
remove invalid aria-label from Carbon Header

pranithraoibm:
Removed the empty aria-label from the Carbon `<Header>` component to resolve the aria_attribute_valid accessibility violation.
The header already contains visible text (DigitalAssistant), so an explicit ARIA label was unnecessary.
No functional or UI changes.

Report before code change:
image
 
 Report after code change:
 image
 Issue ID `2624070542` is no longer present


iv1111 (approved):
LGTM.

PR #407, opened 2026-03-04
Backend integration into Catalog-ui

pranithraoibm:
Add NGINX reverse proxy to route /api requests from the UI to the backend service, avoiding CORS, and update the Containerfile to serve the UI via NGINX.

### DIFF
--- a/ai-services/assets/applications/rag-dev/openshift/steps/info.md
+++ b/ai-services/assets/applications/rag-dev/openshift/steps/info.md
@@ -2,7 +2,7 @@ Day N:
 
 {{- if eq .UI_STATUS "running" }}
 
-- Chatbot UI is available to use at http://{{ .UI_ROUTE }}.
+- Chatbot UI is available to use at https://{{ .UI_ROUTE }}.
 {{- else }}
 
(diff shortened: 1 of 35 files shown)


mayuka-c (commented):
This wouldnt be required. I feel since our frontend is served with nginx we can make use of it as reverse proxy. Also both UI and backend runs within the same cluster and within same namespace (so the domain remains the same and we are not using routes, but rather UI talks to backend with service endpoint).

Just add a proxy pass within the nginx conf file here: https://github.com/IBM/project-ai-services/blob/main/catalog-ui/nginx.conf.tmpl#L21 and we dont have to deal with CORS.

cc. @mkumatag

mayuka-c:
Also any reason for adding this?. Were you facing the issue running locally?
If yes, do let me know, I can help you with. There is proxy that you can set up in Vite too to avoid all these.

mkumatag (changes requested):

mayuka-c (commented):
Rest LGTM

mayuka-c (commented):
LGTM

mayuka-c (dismissed):

mkumatag (changes requested):

mkumatag (approved):

PR #412, opened 2026-03-04
override rollup to 4.59.0 to address path traversal vulnerability

pranithraoibm:
Fixes the Rollup path traversal vulnerability by overriding the transitive dependency to `rollup@^4.59.0`


mkumatag (approved):
LGTM

PR #441, opened 2026-03-10
Restrict query_length in backend - chat

Niharika0306:
What does this PR do?
- Keeps a check on the query_length in the chat app, and returns appropriate errors on  respective cases.

( inherently, there are already exceptions by 
 1. reranker -> to restrict query + doc to 1024 tokens
 2. opensearch in case of sparse mode of search to 1024 tokens ) 
  
  But in order to catch it gracefully, and use a more realistic limit - we have restricted the query_length to 512 tokens.

### DIFF
--- a/spyre-rag/src/retrieve/app.py
+++ b/spyre-rag/src/retrieve/app.py
@@ -15,7 +15,7 @@ from common.llm_utils import create_llm_session, query_vllm_stream, query_vllm_n
 from common.misc_utils import get_model_endpoints, set_log_level, set_request_id
 from common.settings import get_settings
 from common.perf_utils import perf_registry
-from retrieve.backend_utils import search_only
+from retrieve.backend_utils import search_only, validate_query_length
 from retrieve.response_utils import (
     ReferenceRequest,
     ReferenceResponse,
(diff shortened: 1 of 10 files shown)


spyre-rag/src/settings.json
```diff
@@ -20,6 +20,7 @@
   "temperature": 0.0,
   "max_input_length": 6000,
   "prompt_template_token_count": 250,
+  "max_query_token_length": 512,
```
dharaneeshvrd: @manalilatkar 
Do you think we need to consider German language here while deciding this?

spyre-rag/src/retrieve/app.py
```diff
@@ -129,6 +132,9 @@ async def get_reference_docs(req: ReferenceRequest) -> ReferenceResponse:
         # Store metrics in registry for reference endpoint
         perf_registry.add_metric(perf_stat_dict)
         
+    except HTTPException:
+        # Re-raise HTTPException without modification
```
manalilatkar: please log the exception detail before raising.

spyre-rag/src/retrieve/app.py
```diff
@@ -211,6 +201,16 @@ async def chat_completion(req: ChatCompletionRequest) -> ChatCompletionResponse
         reranker_model = reranker_model_dict['reranker_model']
         reranker_endpoint = reranker_model_dict['reranker_endpoint']
         
+        # Validate query length
+        is_valid, token_count, error_msg = await asyncio.to_thread(
```
manalilatkar: What is the point of returning token_count here? It is not being used anywhere except the message which already contains the token count.

dharaneeshvrd:
Yes lgtm

dharaneeshvrd (approved):

PR #470, opened 2026-03-13
[Openshift] Helm Chart for Rag prod

mayuka-c:
- Remains exactly same as rag-dev helm chart.
- Except that in reranker-isvc it currently uses spyre serving runtime and uses 1 spyre card. Also changes in arg and env respectively.
- Also took care of running OpenSearch in restricted-v2.


mkumatag:
> * Except that in reranker-isvc it currently uses spyre serving runtime and uses 1 spyre card. Also changes in arg and env respectively.

how much is the difference we have seen running inference with reranker on cpu vs spyre? if not much - let us run the reranker on cpu itself

mkumatag (approved):
Let us merge this for now to unblock the test and other folks who want to use.

yussufsh (approved):
lgtm

PR #494, opened 2026-03-17
[Digitize-UI] Proper Error Handling in Jobs and Docs Page

mayuka-c:
- Error Handling for 5xx errors in Jobs and Docs Page.
- Handle individual job/ doc errors by using a tip next to status which shows the detailed information about the error. Also adds overflow with scrollable y-axis if error message is too big.
- Also added status icon checks based on all the enums defined for digitize.

Made With Bob


pranithraoibm (approved):
LGTM

PR #524, opened 2026-03-23
Upgrade flatted to 3.4.2 (security fix)

pranithraoibm:
- Updated transitive dependency flatted to 3.4.2
- Eliminates risk of prototype pollution via parse()


pranithraoibm:
@mayuka-c I checked the UI locally without the backend, all looks good with this change. Wanted to check using the PR build pipeline thing but that machine is currently used for testing so build failed.

mayuka-c (approved):
LGTM

PR #575, opened 2026-04-01
[OpenShift] Configure and Enable OpenSearch HA

mayuka-c:
- Configuring opensearch cluster by creating 3 replicas of OpenSearch statefulset.
- Creating a opensearch headless service for the nodes within the cluster to communicate with each other.
- Making necessary changes in backend service to have the `number_of_replicas` set to `0-all` and `number_of_shards` set to 1.


mkumatag (changes requested):

mayuka-c:
After doing thorough validations looks like PDB is required without which when scaling down, the openshift may bring down nodes simultaneously due to which faced issues multiple times causing the cluster to be not in good condition. So have added it back.

Other than I have set the replicas as 1 by default and also made changes so that the pvc gets removed when we scale down the pods.

mkumatag:
> After doing thorough validations looks like PDB is required without which when scaling down, the openshift may bring down nodes simultaneously due to which faced issues multiple times causing the cluster to be not in good condition. So have added it back.

I didn't understand this! scaling down the statefulset is bringing down the worker node?

ai-services/assets/applications/rag-cpu/podman/templates/chat-bot.yaml.tmpl
```diff
@@ -75,6 +75,8 @@ spec:
           value: "{{ .Values.opensearch.auth.username }}"
         - name: OPENSEARCH_PASSWORD
           value: "{{ .Values.opensearch.auth.password }}"
+        - name: OPENSEARCH_NUM_SHARDS
```
mkumatag, line 78: let us parameter this as well, somehow we need to classify the basic vs advanced so that user won't get lost with so many params :)

mayuka-c:
> > After doing thorough validations looks like PDB is required without which when scaling down, the openshift may bring down nodes simultaneously due to which faced issues multiple times causing the cluster to be not in good condition. So have added it back.
> 
> I didn't understand this! scaling down the statefulset is bringing down the worker node?

No not the worker node (Not sure if I caused some confusion here :-)). Here node means the OpenSearch pod (as for OpenSearch cluster each pod considers to be a node).

So when we scale down, it is not just that one of the pod (OpenSearch node) gets taken down. But the OpenSearch Cluster tries to restart the other nodes too. When it does the restart of other pods, if all the pods are taken down at the same time (depends), then the cluster gets broken (at a time atleast 1 pod within the cluster should be active). So with PDB it ensures that atleast 1 pod is running (active at a given time) within the cluster so that it doesnt cause any issues.

ai-services/assets/applications/rag-cpu/podman/templates/chat-bot.yaml.tmpl
mayuka-c: Done. I have made paramterized for OpenShift.
I have not done it for Podman since it will always be a single node OpenSearch running. So for a single node, ideal would be to have a single shard. PLMK

mkumatag:
> So when we scale down, it is not just that one of the pod (OpenSearch node) gets taken down. But the OpenSearch Cluster tries to restart the other nodes too. When it does the restart of other pods, if all the pods are taken down at the same time (depends), then the cluster gets broken (at a time atleast 1 pod within the cluster should be active). So with PDB it ensures that atleast 1 pod is running (active at a given time) within the cluster so that it doesnt cause any issues.

this make sense, thanks.

ai-services/assets/applications/rag-cpu/podman/templates/chat-bot.yaml.tmpl
mkumatag: this make sense but let us make sure we document somewhere..

mkumatag (approved):
LGTM

yussufsh (approved):

PR #584, opened 2026-04-02
Code reorg proposal

ryanmarc:


dharaneeshvrd (approved):
Looks great!
Thank you!

iv1111 (commented):
- aren't we supposed to call it `rag` instead of `chatbot`?
- not sure `services` would be the right term for it because it couples both standalone services as well as deployable architectures etc, it might need deeper folder structure or a different term if we want to pack all into one folder


Otherwise looks good to me.

mkumatag (approved):
overall structure LGTM

iv1111 (approved):
> > * not sure `services` would be the right term for it because it couples both standalone services as well as deployable architectures etc, it might need deeper folder structure or a different term if we want to pack all into one folder
> 
> this can go as a incremental change when we really figure out how we want to handle those things.

Agree.

PR #591, opened 2026-04-07
[Podman] Application create and start commands to use podman sdk instead of cli

mayuka-c:
- Earlier podman cli was used for podman kube play and podman start command due to issues with bringing up the vllm pod.
- This PR addresses the issues and migrates to podman sdk way.
- This PR would be required for the Catalog service as it will be the Catalog server backend bringing up the pods and it uses the sdk to talk to host podman.
- Also avoids the step to relogin after running bootstrap.


mkumatag (approved):
LGTM

yussufsh (approved):

PR #592, opened 2026-04-07
Diagnostics dump in case of exception/crash

sats-23:
[AISERVICES-871](https://jsw.ibm.com/browse/AISERVICES-871)

-Helps dump important info in case of an exception/crash
-Currently handles via 3 modules -> stderr monitor, Exception handler and Signal handler
-Has potential to catch OOMs, C-level exceptions, PID exhaustion, Memory exhaustion, FD limit, CPU limit

image

### DIFF
--- a/spyre-rag/src/common/diagnostic_logger.py
+++ b/spyre-rag/src/common/diagnostic_logger.py
@@ -255,6 +255,54 @@ class DiagnosticLogger:
             self.logger.error(f"Error collecting CPU info: {e}")
             return {"error": str(e)}
     
+    def get_numa_info(self) -> Dict[str, Any]:
+        """
+        Collect NUMA node information.
+        
+        Returns:
+            Dictionary containing NUMA node count.
+        """
+        try:
+            import subprocess
+            
+            # Try to get NUMA node count using lscpu
+            try:
+                result = subprocess.run(
+                    ['lscpu'],
+                    capture_output=True,
+                    text=True,
+                    timeout=2
+                )
+                
+                if result.returncode == 0:
+                    # Parse lscpu output for NUMA node(s) line
+                    for line in result.stdout.splitlines():
+                        if 'NUMA node(s):' in line:
+                            fields = line.split()
+                            if fields:
+                                numa_count = int(fields[-1])
+                                return {"numa_node_count": numa_count}
+            except (subprocess.TimeoutExpired, subprocess.SubprocessError, ValueError) as e:
+                self.logger.debug(f"Could not get NUMA info via lscpu: {e}")
+            
+            # Fallback: try to read from /sys/devices/system/node/
+            try:
+                node_path = Path("/sys/devices/system/node")
+                if node_path.exists():
+                    # Count node* directories
+                    numa_nodes = [d for d in node_path.iterdir() if d.is_dir() and d.name.startswith('node')]
+                    numa_count = len(numa_nodes)
+                    if numa_count > 0:
+                        return {"numa_node_count": numa_count}
+            except Exception as e:
+                self.logger.debug(f"Could not read NUMA info from /sys: {e}")
+            
+            return {"numa_node_count": "N/A"}
+        
+        except Exception as e:
+            self.logger.error(f"Error collecting NUMA info: {e}")
+            return {"error": str(e)}
+    
     def get_file_descriptor_info(self) -> Dict[str, Any]:
         """
         Collect file descriptor usage information.
@@ -363,7 +411,7 @@ class DiagnosticLogger:
             Dictionary containing network connection statistics.
         """
         try:
-            connections = self.process.connections(kind='inet')
+            connections = self.process.net_connections(kind='inet')
             
             connection_summary = {
                 "total_connections": len(connections),
@@ -430,6 +478,11 @@ class DiagnosticLogger:
         self.logger.info("--- CPU INFORMATION ---")
         cpu_info = self.get_cpu_info()
         for key, value in cpu_info.items():
+            self.logger.info(f"{key}: {value}")
+        
+        self.logger.info("--- NUMA INFORMATION ---")
+        numa_info = self.get_numa_info()
+        for key, value in numa_info.items():
             self.logger.info(f"{key}: {value}")
         
         self.logger.info("--- FILE DESCRIPTOR INFORMATION ---")
@@ -516,8 +569,6 @@ class StderrMonitor:
         self.original_stderr = sys.stderr
         self._stop_event = threading.Event()
         self._monitor_thread: Optional[threading.Thread] = None
-        self.cooldown_seconds = 3
-        self._last_diagnostic_dump_time = 0.0
         
         # Error patterns to watch for
         self.error_patterns = [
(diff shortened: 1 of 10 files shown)


mkumatag (changes requested):
can we explore this option at framework level, may be via some command in the podman or openshift instead of pushing these into the application level?

sats-23:
> > I suggest we keep this diagnostics dump as simple as possible for our debugging and initial triage reference and go ahead with full Prometheus solution to plug into our exposed perf_metrics endpoint
> 
> Much of the data we're collecting is OS-specific. Can we ensure that there's an option to disable this so that I can run on Mac as well? Additionally, please measure and report how long it takes to collect this data. I'm concerned about using `du/df` commands because they may hang in environments where network mount paths are involved. If possible, let's avoid collecting disk usage information altogether.

-The metrics collection is instantaneous (no wait time at all). The exception is dumped and immediately followed by the diagnostics data.
-I have added a timeout of 1s to the disk usage calls, so adding those, we should have the metrics in ~1s-1.5s.
-In terms of MacOS, everything is expected too work except the `cgroup-based PID count`, which will gracefully fall back to `resource.RLIMIT_NPROC`. But in case it's redundant, we can remove it as well.

spyre-rag/src/summarize/app.py
```diff
@@ -23,8 +23,10 @@
 
 set_log_level(log_level)
 
-from common.llm_utils import query_vllm_summarize, query_vllm_summarize_stream
-from common.misc_utils import get_model_endpoints, set_request_id, create_llm_session, configure_uvicorn_logging
+from common.llm_utils import create_llm_session, query_vllm_summarize, query_vllm_summarize_stream
+from common.misc_utils import get_model_endpoints, set_request_id, configure_uvicorn_logging
```
dharaneeshvrd:
```suggestion
from common.llm_utils import query_vllm_summarize, query_vllm_summarize_stream
from common.misc_utils import get_model_endpoints, set_request_id, create_llm_session, configure_uvicorn_logging
```

spyre-rag/src/digitize/app.py
```diff
@@ -41,6 +42,8 @@
 
 logger = get_logger("digitize_server")
 
+diagnostic_logger, stderr_monitor, signal_handler = setup_comprehensive_crash_handler(logger)
```
dharaneeshvrd, line 45: Any reason for returning this unused handlers here?

spyre-rag/src/common/diagnostic_logger.py
```diff
@@ -0,0 +1,740 @@
...
+    
+    def get_disk_usage(self, paths: Optional[list] = None) -> Dict[str, Any]:
+        """
+        Collect disk usage information with timeout protection.
+        
+        Uses a short timeout (1000ms) to prevent hanging on unresponsive
+        network mounts (NFS, SMB, etc.) especially on macOS.
+        
+        Args:
+            paths: List of paths to check. Defaults to ["/", "/tmp", current working directory].
```
dharaneeshvrd, line 358: Can we add /var?
var contains docling models and venv, lets collect that as well

spyre-rag/src/digitize/app.py
sats-23: Only stderr_monitor is used later to stop at shutdown. 
Other 2 were just returned for uniformity and future usage.

spyre-rag/src/summarize/app.py
sats-23: Done

spyre-rag/src/common/diagnostic_logger.py
sats-23: Done

spyre-rag/src/common/diagnostic_logger.py
```diff
@@ -0,0 +1,740 @@
...
+        Args:
+            logger: Logger instance for warnings
+            diagnostic_logger: DiagnosticLogger instance for detailed dumps
+        """
+        self.logger = logger
+        self.diagnostic_logger = diagnostic_logger
+        self.original_stderr = sys.stderr
+        self._stop_event = threading.Event()
+        self._monitor_thread: Optional[threading.Thread] = None
+        self.cooldown_seconds = 3
```
dharaneeshvrd: its unused, if not required please remove, since `diagnostic_logger` have its own `cooldown_seconds`

spyre-rag/src/common/diagnostic_logger.py
```diff
@@ -0,0 +1,740 @@
...
+        self._last_diagnostic_dump_time = 0.0
```
dharaneeshvrd: this one too

spyre-rag/src/common/diagnostic_logger.py
dharaneeshvrd, line 1: I think we should give it a try in openshift env once before merging this.

spyre-rag/src/common/diagnostic_logger.py
```diff
@@ -0,0 +1,740 @@
...
+            return {
+                "current_usage": f"{round(mem_info.rss / (1024 ** 3), 2)} GB",
+                "limit": limit,
+                "system_total": f"{round(virtual_mem.total / (1024 ** 3), 2)} GB"
+            }
+        except Exception as e:
+            self.logger.error(f"Error collecting memory info: {e}")
+            return {"error": str(e)}
+    
+    def get_cpu_info(self) -> Dict[str, Any]:
```
dharaneeshvrd, line 225: can you please collect the numa count as well?

spyre-rag/src/common/diagnostic_logger.py
```diff
@@ -0,0 +1,740 @@
...
+    
+    def get_network_connections(self) -> Dict[str, Any]:
+        """
+        Collect network connection information.
+        
+        Returns:
+            Dictionary containing network connection statistics.
+        """
+        try:
+            connections = self.process.connections(kind='inet')
```
dharaneeshvrd: seems `connections` is deprecated. 
Please use `net_connections`

dharaneeshvrd:
@manju956 do you think we can collect any info related to the spyre cards like pci or vfio info?

manju956:
> @manju956 do you think we can collect any info related to the spyre cards like pci or vfio info?

there is pci cards health and sos report we could collect w.r.t spyre cards i believe. I will try to find if we have steps to collect these readily.

spyre-rag/src/common/diagnostic_logger.py
sats-23: Removed

spyre-rag/src/common/diagnostic_logger.py
sats-23: Removed

spyre-rag/src/common/diagnostic_logger.py
sats-23: Done, added

spyre-rag/src/common/diagnostic_logger.py
sats-23: Done

dharaneeshvrd (approved):
@mkumatag can you please approve if there are no comments?

mkumatag (approved):
I'm leaning toward's @dharaneeshvrd and other reviews and approving this PR for now. Thanks.

PR #597, opened 2026-04-08
refactor(spyre-rag): migrate settings.json into shared config 

dharaneeshvrd:


manalilatkar (commented):
mostly LGTM. 
@dharaneeshvrd please convey to the testing team to test all the testcases after this change so nothing is missed.

Niharika0306 (dismissed):
Overall LGTM. 
One small nit -  we may have to update the README.md  with latest info and mention the env vars that are available to change.

ryanmarc (changes requested):
I left 2 comments in the main conversation on this PR that I think should be addressed, or at least discussed based on this current approach.

- A base config with service specifics defined within the specific service folders
- Using `pydantic-settings`

spyre-rag/src/common/misc_utils.py
```diff
@@ -258,5 +252,9 @@ def validate_pdf_file(filename: str, content) -> None:
     if len(content) == 0:
         raise ValueError(f"File is empty: {filename}")
 
+    pdf_signature = b'%PDF'
```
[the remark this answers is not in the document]
dharaneeshvrd: Reverted this.

dharaneeshvrd:
@ryanmarc Agree with your comments. Tried to address it, ptal!

dharaneeshvrd:
@shwetha-s-poojary I feel that is too internal and no need to configure that as a param I believe. Will see in future if there is a need for those config and add accordingly.

dharaneeshvrd:
> we may have to update the README.md with latest info and mention the env vars that are available to change.

@Niharika0306 There is no change in the env names, only the way we consume is modified. But I agree, we need to document the available envs in a README, I would like to take care of that in a following PR.

spyre-rag/src/common/config.py
```diff
@@ -0,0 +1,366 @@
...
+
+
+class LLMConfig(BaseSettings):
+    """LLM model and generation settings."""
+
+    model_config = SettingsConfigDict(
+        env_prefix="",
+        case_sensitive=True,
+        extra="ignore",
+    )
```
ryanmarc: `env_prefix=""` and `extra="ignore"` are the default values, there's no reason to set them here.

Why set case_sensitive to `True`? 

Using this as an example:
```
granite_3_3_8b_instruct_context_length: int = Field(
    alias="GRANITE_3_3_8B_INSTRUCT_CONTEXT_LENGTH",
    default=32768,
    ge=1,
    description="Context length for Granite 3.3 8B Instruct model",
)
```
If `case_sensitive` is just left as the default of false, the `alias` here and everywhere else isn't needed. Pydantic settings would automatically find the uppercase env variable when using `LLMConfig.granite_3_3_8b_instruct_context_length`

spyre-rag/src/common/config.py
```diff
@@ -0,0 +1,366 @@
...
+    llm: LLMConfig = Field(default_factory=LLMConfig)
+    language: LanguageConfig = Field(default_factory=LanguageConfig)
+    model_endpoints: ModelEndpointsConfig = Field(default_factory=ModelEndpointsConfig)
+    vector_store: VectorStoreConfig = Field(default_factory=VectorStoreConfig)
+
+
+# Global settings instance
+settings = Settings()
+
+# Backward compatibility: expose settings as module-level constants
```
ryanmarc: This whole section just adds and extra layer, do we really need this extra backwards compatibility when everywhere using this settings change is already being updated anyway?

spyre-rag/src/digitize/settings.py
```diff
@@ -1,41 +1,196 @@
...
 from pathlib import Path
 
-# Directory paths
-CACHE_DIR = Path(os.getenv("CACHE_DIR", "/var/cache"))
-DOCS_DIR = CACHE_DIR / "docs"
-JOBS_DIR = CACHE_DIR / "jobs"
-STAGING_DIR = CACHE_DIR / "staging"
-DIGITIZED_DOCS_DIR = CACHE_DIR / "digitized"
+from pydantic import Field
+from pydantic_settings import BaseSettings, SettingsConfigDict
```
ryanmarc, line 8: Based on the other comments, if you followed those suggestions, here you could also import `from common.config import Settings as CommonSettings`.

With that imported you could define the digitize settings like you already are below, but then create a new `Settings` composition layer here that could be imported into the digitize app.

Something like 
```
class Settings(BaseSettings):
    common: CommonSettings = Field(default_factory=CommonSettings)
    digitize: DigitizeConfig = Field(default_factory=DigitizeConfig)
```

With that, the digitize app.py would only need to import this one config file.

spyre-rag/src/digitize/app.py
```diff
@@ -1,31 +1,20 @@
...
-if level != "":
-    if "debug" in level:
-        log_level = logging.DEBUG
-    elif not "info" in level:
-        logging.warning(f"Unknown LOG_LEVEL passed: '{level}', defaulting to INFO.")
-
-set_log_level(log_level)
+from common.misc_utils import set_log_level, get_logger
+import common.config as common_config
+import digitize.config as config
```
ryanmarc: Based on the comment from the config.py file in the digitize folder, this could be just `from digitize.config import Settings`

ryanmarc:
I was reviewing this again and adding some comments about how pydantic settings is being used. The module level exports of everything seem like an extra layer to keep up with that duplicate the values from the settings class to module level values.

I was adding some additional comments about that, but as i followed that through, the only way to accomplish using the settings class in a more standard fashion would require a pretty big refactor of the way the rest of the services are currently built. They all tend to expect module level values rather than class imports.

I'll go ahead an unblock this PR, but we should continue the discussion on a bigger refactor to address this.

ryanmarc (dismissed):

dharaneeshvrd:
@ryanmarc 
Thanks for the detailed comments. Since I m pretty new to pydantic, some of these items are new to me. 
Tried to address all of your comments. ptal!

dharaneeshvrd:
Can we please review this PR again? Addressed all the comments and rebased with main.

manalilatkar (approved):
LGTM.
As long as we've done basic testing for digitize, summarize and chatbot, this should be good to go.

dharaneeshvrd:
Yes basic functionality testing is done.

PR #601, opened 2026-04-09
build: Clean up hf cache from rag-base

yussufsh:
1. Cleans up 500MB worth of cache from the image
```
[root@4e079a8c423e /]# du -sh /root/.cache/huggingface/
506M    /root/.cache/huggingface/
```
2. Also, rag-base update is needed to fix the High sev python version issues.

### DIFF
--- a/images/rag-base/Containerfile
+++ b/images/rag-base/Containerfile
@@ -23,4 +23,5 @@ RUN --mount=type=bind,from=prebuilder,source=/wheels,target=/wheels,ro \
         --extra-index-url https://wheels.developerfirst.ibm.com/ppc64le/linux \
         --prefer-binary && \
     pip cache purge && \
-    python download_docling_models.py
+    python download_docling_models.py && \
+    rm -rf ~/.cache/huggingface
--- a/images/rag-base/Makefile
+++ b/images/rag-base/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=rag-base
-TAG?=v0.0.20
+TAG?=v0.0.21
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
 CONTAINER_BUILDER?=podman


images/rag-base/Containerfile
```diff
@@ -23,4 +23,5 @@ RUN --mount=type=bind,from=prebuilder,source=/wheels,target=/wheels,ro \
         --extra-index-url https://wheels.developerfirst.ibm.com/ppc64le/linux \
         --prefer-binary && \
     pip cache purge && \
-    python download_docling_models.py
+    python download_docling_models.py && \
+    rm -rf /root/.cache/huggingface
```
mkumatag: let us ensure we test properly and it is using the downloaded models, but I'm curious what data it was caching there in that dir?

mkumatag: can we avoid /root and keep this relative path, e.g: `rm -rf ~/.cache`

mkumatag (changes requested):

images/rag-base/Containerfile
yussufsh: > can we avoid /root and keep this relative path, e.g: rm -rf ~/.cache

I had the same instinct, but this is what Bob said... 😄 
```
Use /root/.cache/huggingface (Current Implementation is Correct)
The current implementation using /root/.cache/huggingface is correct for this Containerfile.

Reasoning
Use /root/.cache/huggingface because:

Container context: During the Docker/Podman build, commands run as root user by default
Explicit path: /root/.cache/huggingface is the absolute path and works reliably in all contexts
No shell expansion needed: Doesn't rely on ~ being expanded correctly
Build-time certainty: Guarantees the correct path regardless of shell or environment
Why NOT ~/.cache:

~ requires shell expansion which may not work consistently in all RUN contexts
Less explicit and could be ambiguous
/root/.cache would also work but /root/.cache/huggingface is more specific
```

yussufsh: Verified conversion and chatbot are working fine.
```
[root@onprem133726179 ai-services]# podman exec -ti yus-test--digitize-backend-server bash
/var/cache/jobs/bb9fd20e-6675-442e-99d9-9c8dce8c0ffc_status.json
[root@yus-test--digitize rag]# ls ~/.cache/huggingface
ls: cannot access '/root/.cache/huggingface': No such file or directory
[root@yus-test--digitize rag]# cat /var/cache/jobs/bb9fd20e-6675-442e-99d9-9c8dce8c0ffc_status.json
{
    "job_id": "bb9fd20e-6675-442e-99d9-9c8dce8c0ffc",
    "job_name": "job1",
    "operation": "ingestion",
    "status": "completed",
    "submitted_at": "2026-04-09T16:39:28.046182Z",
    "completed_at": "2026-04-09T16:40:15.186127Z",
    "documents": [
        {
            "id": "6bde5aa5-1e5b-4975-bde8-030e0b920a5a",
            "name": "IBM PowerHA SystemMirror for AIX Best Practices - redp5117.pdf",
            "status": "completed"
        }
    ],
    "stats": {
        "total_documents": 1,
        "completed": 1,
        "failed": 0,
        "in_progress": 0
    },
    "error": null
}
```

yussufsh: Changed to use `~/` as well.

yussufsh: btw docling model files are present here:
```
[root@yus-test--digitize rag]# du -sh /var/docling-models/
506M    /var/docling-models/
[root@yus-test--digitize rag]# ls /var/docling-models/
docling-project--docling-layout-heron  docling-project--docling-models
```

mkumatag (approved):
/lgtm

PR #607, opened 2026-04-10
docs(proposals): add proposal for index management strategy

ryanmarc:


docs/proposals/opensearch_index_management.md
mkumatag, line 1: There is concept called **Index State Management (ISM)** which takes care some of the stuff what you mentioned in this proposal, wondering if we can explore that so that we can offload the management to the opensearch itself?

### Optimizing OpenSearch Storage with Index State Management (ISM)

When managing a growing Retrieval-Augmented Generation (RAG) vector database, simply deleting or updating old embeddings doesn't immediately free up disk space. Because OpenSearch is built on Apache Lucene, deleted documents are merely marked with a "tombstone" bit. They continue to take up disk space and degrade search performance until the underlying files (segments) are merged.

By combining an **Index Alias** with an automated **ISM Policy**, we can create a self-cleaning, auto-scaling architecture that rolls over growing indices and forcibly purges deleted entries to reclaim disk space.

#### 1. The Architecture: Index Aliases
Your application should never write directly to a physical index (e.g., `vector-index-000001`). Instead, it connects to an **Alias** (e.g., `vector-write-alias`). 
* The alias routes all new incoming vectors to the currently active index.
* When that index gets too large, ISM automatically creates `vector-index-000002`, points the write alias to the new index, and turns the old one into a read-only state with zero downtime.

#### 2. The Clean-Up: Force Merging Deleted Entries
Once an index has rolled over and is no longer receiving new writes, it is the perfect time to clean it up. We use ISM to trigger a **Force Merge**. 
* **The Mechanism:** Force merging tells OpenSearch to take all the fragmented segments on disk and rewrite them into a single, optimized segment. 
* **Reclaiming Space:** During this rewrite process, OpenSearch finally discards all the "tombstoned" (deleted/updated) documents. This physically reclaims the disk space and shrinks the index size footprint drastically.
* **Performance Boost:** For vector search (k-NN), a single force-merged segment means the system only has to query one HNSW graph instead of dozens, significantly reducing latency and RAM usage.

#### 3. Example ISM Policy
Here is a blueprint of how this automated lifecycle looks in an ISM policy JSON. It dictates that an index should roll over when it hits 30GB, and immediately afterward, it should force merge to clean up the deleted entries.

```json
{
  "policy": {
    "description": "Auto-rollover and purge deleted vectors",
    "default_state": "active_write",
    "states": [
      {
        "name": "active_write",
        "actions": [
          {
            "rollover": {
              "min_size": "30gb",
              "min_index_age": "30d"
            }
          }
        ],
        "transitions": [
          {
            "state_name": "optimize_and_purge"
          }
        ]
      },
      {
        "name": "optimize_and_purge",
        "actions": [
          {
            "force_merge": {
              "max_num_segments": 1
            }
          }
        ],
        "transitions": []
      }
    ]
  }
}
```

ryanmarc: This largely describes the pattern I wrote up in the proposal. But I described using soft deleted to avoid ever having any tombstone items. 

In general, you should never have to force merge indexes. The underlying system ideally handles this automatically.

If you have specific areas of the proposal you’d like to think about changing please comment on that area.

mkumatag: I totally get where you're coming from. Do you think we could look into ISM to handle this? I'm worried that if we do it at the application level, the code will get complicated-especially as we add support for Oracle and any other vector DBs in the future. It might also be safer to let the system automate the cleanup so we don't have to worry about disk space or slower queries if a manual reindex is missed.

ryanmarc: This would have a similar issue on performance over time - rollover with ISM would create a new empty index each time and mark the old one as read only. Over time you'd end up with an alias that referenced back to lots of indexes and search peformance over that would degrade. In what i've outlined here, there would be a reindex to from previous index to a new index with the items flagged as deleted left out. Then the previous index would be completely dropped. Going this route would ensure that, aside from the time that this process is running, there's only ever 1 index.

Regarding Oracle or different vector DB's, the complexity would still exist if they don't all support ISM in the exact same way. We could make sure they process we have is properly extracted so that the steps or exact process is just an implementation detail of the process.

ISM for rollover is a valid pattern, but, in my opinion, it's not the right one for a RAG vector index.

One take away from this is that in the proposal I have here I will add a step that does a force_merge after doing the _reindex to make sure the segments are down to 1 and the new index is optimized. Additionally, I'll make a note about making sure that `refresh_interval` is set to `-1` and `number_of_replicas` is set to `0` when the reindex occurs to optimze that action.

I get what you are saying about having something like this be automatic so it doesn't get missed. Perhaps we could discuss automatically creating a cronjob to run this on some schedule. Either every X days, or periodically check the index for the size / number of deleted docs.

mkumatag: I want to ensure our design is generic enough to support future vector databases. My goal is to avoid rewriting this logic or creating 'OpenSearch-only' features that don't translate to other providers. As a test case, could we look at how this logic would work with Oracle Database 26AI? I want to make sure we aren't carrying too much backend-specific complexity at the application layer.

ryanmarc: I don’t know a lot about that database, but from some initial research, it does not have the same problem that opensearch does with deletes. Things are cleaned up and the space is immediately available, unlike opensearch.

mkumatag: hmm, then what will happen when `/v1/admin/reindex` is called in the non-opensearch implementations?

ryanmarc: Yeah - it wouldn't really apply and i agree it would feel strange to have that seemingly generic endpoint that may not apply. I know you said that it shouldn't be so bound to opensearch, but in this case, the issue is specific to how opensearch indexes work. Perhaps this is something we should think about building in as part of the opensearch deployment so if that's the vector db that is used then this cleanup just happens automatically with a cron and isn't an admin endpoint that has to actually be called.

PR #609, opened 2026-04-13
[Podman] Models download use podman sdk instead of CLI

mayuka-c:
- This avoids to maintain the podman cli client in the Catalog backend server and rather use the podman sdk for downloading the models.

### DIFF
--- a/ai-services/internal/pkg/cli/helpers/models.go
+++ b/ai-services/internal/pkg/cli/helpers/models.go
@@ -3,13 +3,13 @@ package helpers
 import (
 	"fmt"
 	"os"
-	"os/exec"
 	"strings"
 
 	"github.com/project-ai-services/ai-services/internal/pkg/cli/templates"
 	"github.com/project-ai-services/ai-services/internal/pkg/constants"
 	"github.com/project-ai-services/ai-services/internal/pkg/logger"
 	"github.com/project-ai-services/ai-services/internal/pkg/models"
+	"github.com/project-ai-services/ai-services/internal/pkg/runtime/podman"
 	"github.com/project-ai-services/ai-services/internal/pkg/vars"
 )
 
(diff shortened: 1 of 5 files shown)


ai-services/internal/pkg/runtime/podman/podman.go
```diff
@@ -293,6 +295,69 @@ func (pc *PodmanClient) ContainerExists(nameOrID string) (bool, error) {
...
+	Options     []string
+}
+
+// RunContainerWithSpec creates, starts, waits for, and removes a container with the given configuration.
+// Returns the exit code of the container.
+func (pc *PodmanClient) RunContainerWithSpec(
+	image string,
+	command []string,
+	mounts []ContainerMount,
+	terminal, stdin bool,
```
mkumatag: I'm not sure if adding so many param is really a nice idea considering we may add few more flags/options in the future, let us find a better solution here

ai-services/internal/pkg/cli/helpers/models.go
```diff
@@ -52,28 +52,42 @@ func DownloadModel(model, targetDir string) error {
...
 		fmt.Sprintf("/models/%s", model),
 	}
-	cmd := exec.Command(command, args...)
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
-	cmd.Stdin = os.Stdin
-	err := cmd.Run()
+
+	// Run container with spec
+	exitCode, err := runtimeClient.RunContainerWithSpec(vars.ToolImage, command, mounts, true, true)
```
mkumatag: I have seen plenty of dead containers post running these container runs, podman run does contain the flag called `-rm` for removing the containers after they stop, please find a suitable flag to add that here so that after the finish it will just be removed.

mkumatag (changes requested):

ai-services/internal/pkg/cli/helpers/models.go
mkumatag: e.g:
```
[root@onprem133720074 ~]# podman ps -a | grep -i exited
dd6406d3d1ca  icr.io/ai-services/tools:0.7                 hf download ibm-g...  34 minutes ago  Exited (0) 34 minutes ago                                                                        cool_swartz
e5f1dc0ecb50  icr.io/ai-services/tools:0.7                 hf download BAAI/...  34 minutes ago  Exited (0) 34 minutes ago                                                                        gracious_payne
5420f1fdce2a  icr.io/ai-services/tools:0.7                 hf download ibm-g...  34 minutes ago  Exited (0) 34 minutes ago                                                                        jolly_diffie
[root@onprem133720074 ~]#
```

yussufsh: He has added defer function to clean up the container. Before there was no `-rm` flag.

mkumatag: > He has added defer function to clean up the container. Before there was no `-rm` flag.

ofcourse, that was my concern, we didn't handle this before..

ai-services/internal/pkg/runtime/podman/podman.go
```diff
@@ -293,6 +295,69 @@ func (pc *PodmanClient) ContainerExists(nameOrID string) (bool, error) {
...
+) (int32, error) {
+	// Create container spec
+	s := specgen.NewSpecGenerator(image, false)
+	s.Terminal = &terminal
+	s.Stdin = &stdin
+	s.Command = command
```
mkumatag: There is option available in the `ContainerBasicConfig` which can be used here itself.

```
rm := true
s.Remove = &rm
```

mkumatag: @mayuka-c @yussufsh this instead of defer..

ai-services/internal/pkg/cli/helpers/models.go
mayuka-c: Will be removing the defer and using the `Remove` as specified below

ai-services/internal/pkg/runtime/podman/podman.go
mayuka-c: Thanks, will make the change and test it out

ai-services/internal/pkg/runtime/podman/podman.go
mayuka-c: Yes true, maybe I can move the specgen out of here from podman.go and make the specgen as input param so that its more generic and the caller can specify whatever he wants within specgen.

ai-services/internal/pkg/runtime/podman/podman.go
mayuka-c: Verified the container is getting deleted once it exits.

ai-services/internal/pkg/runtime/podman/podman.go
mayuka-c: I have made the changes and I feel now it looks good. PTAL and LMK.

mkumatag (approved):

mayuka-c:
Merging this, with this all the CLI client will be converted to make use of SDK client for Podman.

PR #610, opened 2026-04-13
[Podman] Asset structure for Initial Catalog assets

mayuka-c:
- Asset structure for catalog services.
- Create pod template assets.
- Make Catalog command unhidden.
- Make Catalog api-server command has hidden.


mayuka-c:
Note: Will be taking up creating a new `ai-services catalog bootstrap` command in the next PR which will be responsible for deploying the catalog assets.

ai-services/cmd/ai-services/cmd/catalog/apiserver.go
```diff
@@ -26,9 +26,10 @@ func NewAPIServerCmd() *cobra.Command {
 		adminPasswordHash      string
 	)
 	apiserverCmd := &cobra.Command{
-		Use:   "apiserver",
-		Short: "Manage AI Services API server",
-		Long:  `The apiserver command allows you to manage the AI Services API server, including starting, stopping, and checking the status of the server.`,
+		Use:    "apiserver",
+		Short:  "Manage AI Services API server",
+		Long:   `The apiserver command allows you to manage the AI Services API server, including starting, stopping, and checking the status of the server.`,
+		Hidden: true,
```
mkumatag: I think we can remove this now and make this command available and also let us build the catalog by default without any explicit tag.

ai-services/assets/catalog/podman/values.yaml
```diff
@@ -0,0 +1,12 @@
...
+  image: icr.io/ai-services-cicd/catalog-ui:0.0.1
+
+backend:
+  port: ""
+  image: icr.io/ai-services-cicd/ai-services:0.0.2
+  jwtSecret: ""
+  runtime: ""
+  adminPasswordHash: ""
+  podman:
+    uri: "/run/podman/podman.sock"
```
mkumatag, line 12: I'm not sure how you we will handle this when we run as non-root.

mkumatag: mark it as TODO somewhere and track this closely,

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
```diff
@@ -0,0 +1,76 @@
...
+      volumeMounts:
+        - name: podman-socket
+          mountPath: {{ .Values.backend.podman.uri }}
+        - name: iommu-groups
+          mountPath: /sys/kernel/iommu_groups
+          readOnly: true
+      resources:
+        requests:
+          podman.io/device=/dev/vfio: 4
+          memory: "1Gi"
```
mkumatag, line 63: do we really need 1Gi memory? let us add some task for tracking the real usage and size it appropriately.

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
```diff
@@ -0,0 +1,76 @@
...
+  volumes:
+    - name: podman-socket
+      hostPath:
+        # For rootless podman: /run/user/$(id -u)/podman/podman.sock
+        # For root podman: /run/podman/podman.sock
+        path: {{ .Values.backend.podman.uri }}
+        type: Socket
+    - name: iommu-groups
+      hostPath:
+        path: /sys/kernel/iommu_groups
```
mkumatag, line 75: why do we need this?

mkumatag (changes requested):

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
mayuka-c: It is used for calculating the PCI Addresses of free spyre cards: https://github.com/IBM/project-ai-services/blob/609eff11f5e72641fc0bc39c9ead762ad9dc7cf9/ai-services/internal/pkg/cli/helpers/helper.go#L141

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
mayuka-c: Yeah we would need to test it out later. Just added this for now. Will be revisiting later.

ai-services/assets/catalog/podman/values.yaml
mayuka-c: The values specified in values.yaml are just the default values. We will be overrding these as part of `ai-services bootstrap` command (a new catalog subcommand). More details can be found here: https://ibm.ent.box.com/file/2195323746717.

ai-services/cmd/ai-services/cmd/catalog/apiserver.go
mayuka-c: Sure will be doing it

ai-services/assets/catalog/podman/values.yaml
mayuka-c: This subcommand would override the runtime, adminPassword, podman socket uri, jwt secret while deploying catalog assets

ai-services/cmd/ai-services/cmd/catalog/apiserver.go
mayuka-c: Made the changes. Please refer this commit: https://github.com/IBM/project-ai-services/pull/610/changes/29fb6c9c123ca4405f75daccf24fdd000dddc753

Also I have hidden subcommands like `api-server` and `hashpw` as I feel we might not need these subcommands to be exposed to end users.

mkumatag: we need to expose to the end users, in this case end users are admin who will be deploying the catalog service itself and I don't mind if they know about these commands. Its okay to expose..

mayuka-c: Have removed the hidden.

But just a query, admin need not have to run these commands right as we will be giving 1 single command `ai-services catalog bootstrap` which takes care of everything?
- apiserver command would bring up the rest server (which is used in the catalog backend container to bring up rest server)
- hashpw command would be to get the hash of admin password which would be done internally within `catalog bootstrap`

mkumatag: It is okay.. I don't want to hide everything., unnecessary complexities, if not hidden - we can also ask user to use hashpw if they want to change the password in the future

ai-services/Makefile
```diff
@@ -5,13 +5,8 @@ CONTAINER_BUILDER?=podman
...
 GO := go
-# Feature flags - set ENABLE_CATALOG_API=true to include catalog API server
-# Default is false - catalog API is opt-in for development/CI/CD only
-ENABLE_CATALOG_API ?= false
-BUILDTAGS := exclude_graphdriver_btrfs containers_image_openpgp remote
-ifeq ($(ENABLE_CATALOG_API),true)
-    BUILDTAGS := $(BUILDTAGS) catalog_api
-endif
+# Build tags
+BUILDTAGS := exclude_graphdriver_btrfs containers_image_openpgp remote catalog_api
```
mkumatag: we don't need to add `catalog_api` if we are removing this tag..

mkumatag (commented):
one minor comment, otherwise LGTM

ai-services/Makefile
mayuka-c: Done

yussufsh (approved):

mkumatag (approved):

PR #611, opened 2026-04-13
AI -SERVICES 879: proposal for async summarization jobs

manalilatkar:


docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -357,3 +357,418 @@ curl -X POST http://localhost:5000/v1/summarize \
...
+---
+## 3. Non-Goals
+
+- **Horizontal scaling:** Like the digitize service, the summarization job system is designed for single-replica deployment.
+- **Replacing the sync endpoint:** `POST /v1/summarize` continues to work as-is for quick, interactive use cases.
+- **Multi-file jobs:** Each job processes exactly one document. Clients that need to summarize multiple files should submit one job per file.
+
+---
+
+## 4. New Endpoints
```
dharaneeshvrd, line 869: | Method | Endpoint | Description |
|--------|----------|-------------|
| POST | `/v1/summarize/jobs` | Submit a single file for async summarization. Returns a job_id. |
| GET | `/v1/summarize/jobs` | List all summarization jobs with pagination and filters. |
| GET | `/v1/summarize/jobs/{job_id}` | Get detailed status and metadata of a specific job. |
| GET | `/v1/summarize/jobs/{job_id}/result` | [Revised] Retrieve the summarization result as a sub-resource of the job. |
| DELETE | `/v1/summarize/jobs/{job_id}` | Delete a specific job record and its associated result. |
| DELETE | `/v1/summarize/jobs` | Bulk delete all jobs and results. Requires confirm=true. |

dharaneeshvrd: I feel the results should be part of jobs, in case of digitization, the documents were considered a separate entity. But that is not the case here.

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -357,3 +357,418 @@ curl -X POST http://localhost:5000/v1/summarize \
...
+|:---|:---|:---|:---|
+| `file` | file | Yes | A single `.txt` or `.pdf` file to summarize. |
+| `length` | integer | No | Desired summary length in words. |
+| `job_name` | string | No | Optional human-readable label for the job. |
+
+**Validation rules:**
+
+- Exactly one file must be provided.
+- The file must have a `.txt` or `.pdf` extension.
+- The file's word count must be within the `MAX_INPUT_WORDS` limit (same validation as the sync endpoint).
```
dharaneeshvrd: I want to take a diversion here, it makes sense for limiting the input token size for sync summarization.
If it is asynchronous, there should not be a limit in my opinion.
I have come across this blog where they do summarization for chunks of documents and combine the summaries to provide a document level summary.
https://www.ibm.com/think/tutorials/llm-text-summarization-alice-in-wonderland-granite-docling

manalilatkar: will work on this.

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -357,3 +357,596 @@ curl -X POST http://localhost:5000/v1/summarize \
...
+The approach follows a two-phase pipeline:
+
+**Phase 1 — Chunk and Summarize:**
+1. Split the input text into chunks that each fit within the model's context window (accounting for the system prompt and estimated output tokens).
+2. Summarize each chunk independently with a single LLM call, producing a chunk-level summary.
+
+**Phase 2 — Merge:**
+3. Concatenate all chunk-level summaries.
+4. If the combined summaries fit within the context window, make a single final LLM call to produce a unified summary.
+5. If the combined summaries still exceed the context window, fail the job with error `FILE_SIZE_OVER_LIMIT`.
```
dharaneeshvrd: But why?
At this point, summarization is completed more than 50%.
In case if we want to limit, it should happen before doing the summaries for chunks.

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -357,3 +357,596 @@ curl -X POST http://localhost:5000/v1/summarize \
...
+## 7. Storage Layout
+
+All job-related state is persisted under `/var/cache/summarize/`, which is a persistent volume mounted by the container runtime.
+
+```
+/var/cache/summarize/
+├── staging/
+│   └── {job_id}/
+│       └── report.pdf
+├── jobs/
```
dharaneeshvrd: Can we consider to store these metadata directly to db?
Let's implement it like that.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: implemented.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: incorporated the change.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: fixed approach.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: incorporated MAC's postgresql approach. not handling any setup concerns. Just assuming DB is already available.

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -831,3 +831,726 @@ The abstraction levels are now built into the API. UI should present these optio
...
+**Phase 2 — Merge:**
+4. Concatenate all chunk-level summaries.
+5. If the combined summaries fit within the context window, make a single final LLM call to produce a unified summary.
+6. In the future, if there is a demand for extra large files for this service, we can implement to apply Phase 1 recursively.
+
+### 6.3 Decision Logic
+
+```
+input_word_count = word_count(extracted_text)
+input_tokens = input_word_count / TOKEN_TO_WORD_RATIO
```
dharaneeshvrd: Let's use the tokenize api to decide the input_tokens

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -831,3 +831,726 @@ The abstraction levels are now built into the API. UI should present these optio
...
+
+For the async job path, this limit is unnecessarily restrictive. Long documents — reports, research papers, books — are a primary use case for background summarization. Instead of rejecting them, the async worker uses a **chunk-and-merge** strategy inspired by the approach described in the [IBM Granite + Docling summarization tutorial](https://www.ibm.com/think/tutorials/llm-text-summarization-alice-in-wonderland-granite-docling).
+
+### 6.2 Strategy Overview
+
+The approach follows a two-phase pipeline:
+
+**Phase 1 — Chunk and Summarize:**
+1. Split the input text into chunks that each fit within the model's context window (accounting for the system prompt and estimated output tokens).
+2. If the combined estimated summaries of all chunks exceed the context window, fail the job with error `FILE_SIZE_OVER_LIMIT`.
```
dharaneeshvrd: Can you please explain this on a high level how it's going to be done?

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -831,3 +831,726 @@ The abstraction levels are now built into the API. UI should present these optio
...
+
+if input_tokens + prompt_tokens + estimated_output_tokens <= MAX_MODEL_LEN:
+    → Direct summarization (single LLM call)
+else:
+    → Chunked summarization (multi-pass)
+```
+
+The threshold is the same `MAX_INPUT_WORDS` calculation already used by the sync endpoint — the async path simply uses it as a routing decision rather than a hard rejection.
+
+### 6.4 Chunking Strategy
```
dharaneeshvrd, line 1155: Hope chunk level summaries also follow summary level given by user!

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -831,3 +831,726 @@ The abstraction levels are now built into the API. UI should present these optio
...
+
+**Step 3 — Sentence-level fallback:**
+
+If a single paragraph exceeds `MAX_INPUT_WORDS` on its own (e.g., a very long unbroken block of text), fall back to sentence-level splitting within that paragraph. This uses the `SentenceSplitter` library (already a dependency in the project via `sentence_splitter`) to split the oversized paragraph into sentences, then greedily packs sentences into sub-chunks the same way.
+This fallback is expected to be rare in practice — a single paragraph would need to exceed ~20,000 words. It primarily guards against edge cases like concatenated text extracted from PDFs where paragraph boundaries are lost.
+
+**Step 4 — Overlap:**
+
+To preserve context at chunk boundaries, the last sentence of the previous chunk is prepended to the next chunk. This overlap can be configurable (default: 1 sentence, typically ~20–50 words). The overlap is small enough to not meaningfully reduce the available space for new content, but sufficient to give the model continuity for anaphora resolution and topic transitions.
+
```
dharaneeshvrd, line 1183: Individual chunk summaries would follow the current prompts? or do we need specialised prompts?

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -831,3 +831,726 @@ The abstraction levels are now built into the API. UI should present these optio
...
+            Worker->>vLLM: POST /v1/chat/completions (chunk i)
+            vLLM-->>Worker: Chunk summary i
+        end
+
+        Worker->>Worker: Concatenate chunk summaries
+
+        alt Merged summaries fit in context window
+            Worker->>vLLM: POST /v1/chat/completions (merge prompt)
+            vLLM-->>Worker: Final unified summary
+        else Still exceeds (recursive)
```
dharaneeshvrd: this is not true right?

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -831,3 +831,726 @@ The abstraction levels are now built into the API. UI should present these optio
...
+    "chunking": {
+        "strategy": "chunked",
+        "total_chunks": 12,
+        "completed_chunks": 7,
+        "phase": "summarizing"
+    }
+}
+```
+
+The `chunking` field is only present when the chunked strategy is in use. The `phase` field can be `"summarizing"` (chunk-level summaries in progress) or `"merging"` (producing the final unified summary). This allows clients polling `GET /v1/summarize/jobs/{job_id}` to see meaningful progress for long-running jobs.
```
dharaneeshvrd: > The `chunking` field is only present when the chunked strategy is in use.

I feel the columns should be static, maybe value for a column with json type can be different, please redesign this to have common columns in both type of summarization.

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -831,3 +831,726 @@ The abstraction levels are now built into the API. UI should present these optio
...
+
+---
+
+## 7. Storage Layout
+
+### 7.1 Overview
+
+Job and document metadata is persisted in **PostgreSQL**, consistent with the recent migration of the digitize service from JSON files to a relational database. The summarization service shares the same PostgreSQL instance and uses its own tables within the same database.
```
dharaneeshvrd, line 1305: what do we want to store in document here?
Hope we dont have a separate documents table

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -831,3 +831,726 @@ The abstraction levels are now built into the API. UI should present these optio
...
+| `job_id` | `String(255)` | No (PK) | UUID generated at job creation. |
+| `job_name` | `String(500)` | Yes | Optional human-readable label provided by the user. |
+| `status` | `String(50)` | No | One of: `accepted`, `in_progress`, `completed`, `failed`. |
+| `submitted_at` | `DateTime(tz)` | No | Timestamp when the job was created. |
+| `completed_at` | `DateTime(tz)` | Yes | Timestamp when the job finished (completed or failed). `NULL` while active. |
+| `error` | `Text` | Yes | Error message if the job failed. `NULL` on success. |
+| `document_name` | `String(500)` | No | Original filename of the uploaded file (e.g., `report.pdf`). |
+| `document_original_word_count` | `Integer` | Yes | Word count of the extracted text. Set after text extraction, `NULL` before. |
+|`level`     | `String(20)`   | No | Summarization detail level. One of: brief, standard, detailed. Defaults to standard.|
+| `chunking` | `JSONB` | Yes | Chunk progress metadata. `NULL` for direct summarization. See below. |
```
dharaneeshvrd: > Chunk progress metadata. `NULL` for direct summarization. See below.

sentence is not completed.

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -831,3 +831,726 @@ The abstraction levels are now built into the API. UI should present these optio
...
+### 5.1 POST /v1/summarize/jobs — Create Summarization Job
+
+**Content-Type:** `multipart/form-data`
+
+**Form parameters:**
+
+| Parameter | Type | Required | Description |
+|:---|:---|:---|:---|
+| `file` | file | Yes | A single `.txt` or `.pdf` file to summarize. |
+| `level` | string | Optional | Abstraction level for summary: `brief`, `standard`, or `detailed`. Length is automatically calculated based on input size and level. If not specified, the model determines the summary length automatically. |
```
dharaneeshvrd: Since we have only one config param(not length & level), can we keep the standard level as default?

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -831,3 +831,726 @@ The abstraction levels are now built into the API. UI should present these optio
...
+       - Implications and significance\n
+    3. Preserve ALL numerical data EXACTLY\n
+    4. A summary under {min_words} words is considered incomplete\n\n
+Section summaries:
+{merged_chunk_summaries}
+
+Unified summary:
+```
+
+**User prompt — without level:**
```
dharaneeshvrd: IMO we can get rid of this by keeping default level and maintain only one prompt.

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -831,3 +831,726 @@ The abstraction levels are now built into the API. UI should present these optio
...
+
+### 6.4 Chunking Strategy
+
+The input text is split into chunks of 'less than `MAX_INPUT_WORDS`' words each. The chunking approach:
+
+- Splits on paragraph boundaries (double newlines) to preserve coherent units of text.
+- Falls back to sentence boundaries if a single paragraph exceeds the chunk size.
+- Maintains a small overlap (configurable, default ~50 words) between consecutive chunks to preserve context at boundaries.
+
+#### 6.5 Splitting Strategy
```
dharaneeshvrd, line 1163: Can we have some parallelism here(summarising chunks) to speed up the summarization?
Maybe thread pool size of 4 can be considered

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: yes, will mention that separately.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: yes, current prompt should work for chunk summaries. Will mention that in a new section before 6.6 Merge prompt.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: forgot to correct this. will do.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: will think about this.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: no separate document table. Only job table is mentioned. document metadata means info like document name and word count.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: I meant to say `See below in 7.2.3`

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: ok

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: fixed.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: fixed as per discussion.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: mentioned in 6.6

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: mentioned in 6.6

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: fixed.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: fixed as per discussion.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: fixed.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: fixed.

docs/proposals/summarization_endpoint_design_doc.md
manalilatkar: modified approach in Section 8.

dharaneeshvrd (approved):

PR #618, opened 2026-04-14
chore: add digitize ui image version checks

yussufsh:
1. Add checks for digitize ui in Makefile.
2. Fix `check-image-names.yml` to include digitize ui tag updates.

### DIFF
--- a/.github/workflows/bump-makefile-version-component.yml
+++ b/.github/workflows/bump-makefile-version-component.yml
@@ -1,0 +1,76 @@
+name: Bump Makefile Version Component Check
+
+on:
+  workflow_call:
+    inputs:
+      component_path:
+        required: true
+        type: string
+        description: 'Component path to watch for changes (e.g., "spyre-rag/src")'
+      values_yaml_key:
+        required: false
+        type: string
+        description: 'Key in values.yaml files (e.g., "backend", "ui", "digitizeUi")'
+      additional_notes:
+        required: false
+        type: string
+        description: 'Additional notes to display in error message'
+
+jobs:
+  check-version-bump:
+    name: Verify TAG is bumped in ${{ inputs.component_path }}/Makefile
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      - name: Detect changed paths
+        id: changes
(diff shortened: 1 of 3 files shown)


.github/workflows/bump-makefile-version.yml
```diff
@@ -155,4 +156,52 @@ jobs:
...
           echo "✅ TAG bumped: $BASE_TAG → $HEAD_TAG"
 
+  check-digitize-ui-version-bump:
+    name: Verify TAG is bumped in digitize-ui/Makefile
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@v5
+        with:
```
adarshagrawal38: I think having a reusable workflow to check version bump, will be a good idea.
So that we don't have to repeat same steps when we enable version check in image

yussufsh: I will create a new pr for refactoring. Can we merge this as I want to use this in npm packages upgrades.

yussufsh: Ignore above comment. Used Bob to make a new reusable workflow.
Added a new commit with the changes, please help review again.

.github/workflows/bump-makefile-version-component.yml
```diff
@@ -0,0 +1,76 @@
...
+              echo ""
+              echo "${{ inputs.additional_notes }}"
+            fi
+            echo ""
+            exit 1
+          fi
+
+          echo "✅ TAG bumped: $BASE_TAG → $HEAD_TAG"
+
+# Made with Bob
```
adarshagrawal38: please add a new line here

.github/workflows/bump-makefile-version-component.yml
```diff
@@ -0,0 +1,76 @@
...
+        description: 'Additional notes to display in error message'
+
+jobs:
+  check-version-bump:
+    name: Verify TAG is bumped in ${{ inputs.component_path }}/Makefile
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@v5
```
adarshagrawal38:
```suggestion
        uses: actions/checkout@v6
```

adarshagrawal38 (approved):
LGTM

PR #620, opened 2026-04-15
Conversational RAG Proposal

sats-23:


docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1092 @@
...
+* Existing tools, SDKs, and clients work without modification
+* Future-proof design aligned with industry standards
+
+**4. Acceptable Trade-offs**
+* Request payload increase (~2-3KB for 10 messages) is negligible for modern networks
+* Conversation history of 10 messages fits comfortably within typical HTTP request limits
+* LocalStorage persistence (5-10MB limit) is more than sufficient for conversation data
+
+**5. User Privacy**
+* Conversations never leave user's browser unless explicitly sent to backend
```
Niharika0306, line 165: Didn't understand the 1st point - since we send conversation as part of the requests to the backend. The 2nd point makes sense though.

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1092 @@
...
+The system enables natural conversational flows:
+
+* **Follow-up Questions**: "Tell me more about that" references previous context
+* **Pronoun Resolution**: "What about it?" understands the subject from history
+* **Topic Continuity**: Maintains coherent discussion across multiple turns
+* **Context Switching**: Gracefully handles topic changes within conversation
+
+### 4.2 Conversation Persistence
+
+* **LocalStorage Integration**: Full conversations persist across page refreshes, including retrieved documents
```
Niharika0306: Are we planning to store retrieved docs as well , as part of the history component?

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1092 @@
...
+* **Privacy-Focused**: All data stays in browser, no server-side tracking
+* **Automatic Cleanup**: Old conversations can be manually cleared
+* **Retrieved Documents Storage**: Each assistant response stores the documents retrieved from the vector database, enabling transparency and debugging
+
+### 4.3 Intelligent Context Management
+
+* **Sliding Window**: Maintains last 10 messages (5 turns) for optimal context
+* **Token Budget**: Allocates context window efficiently:
+  - Conversation history: ~5,000 tokens
+  - Retrieved documents: ~3,000 tokens
```
Niharika0306: For Retrieved documents, the calculation is more likely higher : ( 3 * 512 * 5)

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1092 @@
...
+```python
+from chatbot.conversation_utils import get_conversation_context
+
+@app.post("/v1/chat/completions")
+async def chat_completion(req: ChatCompletionRequest):
+    # Extract current query and conversation history
+    current_query, conversation_history = get_conversation_context(req.messages)
+    
+    # Retrieve documents for current query
+    docs, perf_stats = await retrieve_and_rerank(current_query)
```
Niharika0306: Here we can't use current_query directly for retrieval ( It needs to rephrased by the LLM first, in order to get the right meaning based on context)

docs/proposals/conversational_rag.md
sats-23: So far, yes, the calculation is inclusive of the retrieved chunks.
But the idea behind it is mostly for tracking/debugging, in case we want to run it slim, we can remove it as well and keep just the responses.

docs/proposals/conversational_rag.md
sats-23: It is a generic statement mentioning that data stays in user's browser session until the user explicitly submits a chat request

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1092 @@
+# Design Proposal: Conversational RAG with Client-Side Session Management
+
+---
+
+## 1. Executive Summary
+
+The **Conversational RAG Enhancement** transforms the existing single-turn RAG chatbot into a multi-turn conversational system capable of maintaining context across multiple exchanges. By implementing a **client-side stateless architecture**, the system enables natural follow-up questions, pronoun resolution, and contextual understanding while maintaining 100% OpenAI API compatibility and backend statelessness. This design eliminates server-side session management complexity, ensures horizontal scalability, and provides a seamless upgrade path for existing deployments.
```
mkumatag, line 7: I like the idea of handling this at a client side :)

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1092 @@
...
+```
+
+### 10.3 Integration Tests
+
+| Test Case | Expected Behavior |
+|-----------|-------------------|
+| Single-turn query | Works as before (backward compatible) |
+| Follow-up question | References previous context |
+| Topic switch | Handles gracefully |
+| Long conversation (>10 turns) | Truncates oldest messages |
```
mkumatag, line 1161: may be we can start with truncate but later explore if we can compress or summarise with summarisation like api and use that or any other methodologies used in the RAG world!

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1092 @@
...
+* **True Statelessness**: Backend remains completely stateless, enabling unlimited horizontal scaling
+* **Zero Infrastructure**: No session storage, Redis, or database required
+* **Simplified Deployment**: No additional components to deploy or manage
+* **Cost Effective**: No infrastructure costs for session management
+* **Privacy-Focused**: Conversation data stays in user's browser, never stored server-side
+* **Resilient**: No session store dependency, no single point of failure
+* **OpenAI Compatible**: Pure standard API, works with existing tools and clients
+* **Fast Implementation**: Minimal backend changes, faster time to production
+
+**Cons:**
```
mkumatag, line 133: do we have any impact of the resource usage point of view? like browser memory and the ui react code running?

docs/proposals/conversational_rag.md
mkumatag: This is not that simple as we think, we can't blindly search vector DB for any latest query.

e.g:
Q. what is spyre
A. spyre is an AI accelerator for power hardware 
Q. is this supported on power 11

In this case - I don't if we really get a relevant answer from RAG if we just send last Q?! we need to spend more time how to deal with such situations!

mkumatag (changes requested):

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1092 @@
...
+
+User: "Can you give examples?"
+Bot: "Based on the machine learning concepts I just explained, here are some examples..."
+```
+
+## 3. Proposed Architecture
+
+### 3.1 Design Philosophy
+
+The solution adopts a **client-side stateless** approach where:
```
manju956: are we right in calling client side as stateless?

sats-23: It meant client-side logic, stateless on server 
Will correct it

docs/proposals/conversational_rag.md
sats-23: Yep!

docs/proposals/conversational_rag.md
sats-23: 512 in this context is the llm_max_tokens?
We will probably have to revisit the question of whether to consider keeping retrieved docs or not?

docs/proposals/conversational_rag.md
dharaneeshvrd: Like @Niharika0306 mentioned, we need to rephrase the question for every follow up question using LLM based on history of questions before generating the embeddings to search against the vdb. Please take care!

docs/proposals/conversational_rag.md
sats-23: Storing chat history in the browser uses some memory (~5-10KB for 10-20 messages). 
UI stays responsive because the code is optimized to only re-render when needed. 
For very long conversations, we probably need to limit message count or add cleanup to keep performance smooth.

docs/proposals/conversational_rag.md
sats-23: Revised plan is to not include retrieved docs in the history.
We may include metadata storage just for debugging purpose, but it will not be part of the final changes

docs/proposals/conversational_rag.md
Niharika0306: here, 512 was in context with the chunk-size of each doc.

docs/proposals/conversational_rag.md
sats-23: Updated

docs/proposals/conversational_rag.md
sats-23: The approach has been updated 
PTAL

docs/proposals/conversational_rag.md
sats-23: I have updated the numbers with latest approach not carrying any previously retrieved docs (only current query)

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1132 @@
...
+
+### 3.1 Design Philosophy
+
+The solution adopts a **client-side session management** approach with a **stateless backend**, where:
+
+* **Backend**: Remains completely stateless, no session storage
+* **Frontend**: Manages conversation message history in browser memory, excluding retrieved document payloads
+* **API**: Pure OpenAI standard, no modifications to request/response format
+* **Scalability**: Truly stateless backend enables unlimited horizontal scaling
+
```
dharaneeshvrd, line 50: workflow does not contain the rephrase step, please add it.

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1132 @@
...
+    
+    # Build rephrasing prompt
+    rephrase_prompt = f"""Given the conversation history and the current question, rephrase the current question to be a standalone, self-contained query that can be used for semantic search. The rephrased query should:
+1. Replace pronouns (this, that, it, they) with specific nouns from context
+2. Include all necessary context to understand what is being asked
+3. Be concise and focused on the search intent
+4. Maintain the original question's meaning
+
+Conversation History:
+{conversation_history}
```
dharaneeshvrd, line 350: We need to validate this does nt exceeds the LLM's context window, if it exceeds, we need to truncate the earlier messages before sending.

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1132 @@
...
+* Reduced operational costs (no session store monitoring/maintenance)
+
+### 3.3 Key Architectural Decisions
+
+| Decision | Choice | Rationale |
+|----------|--------|-----------|
+| **Session Management** | Client-side | No backend state, unlimited scalability |
+| **API Compatibility** | Pure OpenAI standard | No breaking changes, ecosystem compatibility |
+| **History Storage** | Frontend (Browser Memory) | Ephemeral by design, resets on refresh, user-controlled |
+| **History Truncation** | Frontend (Last 10 messages) | Simple, predictable, no token counting needed |
```
dharaneeshvrd: > no token counting needed

How? We are currently allowing upto 6000 input tokens, I feel there should be a validation wherever this is involved.

docs/proposals/conversational_rag.md
dharaneeshvrd: Agree, we can get started with the truncation, later we need to look at adding support for rolling summary.

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1132 @@
...
+* **Follow-up Questions**: "Tell me more about that" references previous context
+* **Pronoun Resolution**: "What about it?" understands the subject from history
+* **Topic Continuity**: Maintains coherent discussion across multiple turns
+* **Context Switching**: Gracefully handles topic changes within conversation
+
+### 4.2 Ephemeral Conversation State
+
+* **In-Memory Only**: Conversation messages exist only in browser memory for the current page session
+* **Refresh Resets State**: A page refresh, restart, or tab close clears the conversation by design
+* **Simple Client-Side Handling**: No session IDs, no session lifecycle management, and no browser persistence layer
```
dharaneeshvrd, line 205: Hope we can extend the client side to handle these things later with an external storage!
Please keep that in mind while you are working on this!

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1132 @@
...
+* **Privacy-Focused**: Conversation messages are not stored server-side; retrieved documents are not persisted in history
+* **Fresh Retrieval Per Turn**: Each turn performs a new vector database retrieval; retrieved documents are used only for that turn's prompt construction
+
+### 4.3 Intelligent Context Management
+
+* **Sliding Window**: Maintains last 10 messages (5 turns) for optimal context
+* **Token Budget**: Allocates context window efficiently:
+  - Conversation history (messages only, no docs): ~2,000 tokens
```
dharaneeshvrd, line 214: I feel this would be little higher than 2000, since 512 * 5 + input query should be greater than 2k tokens

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1132 @@
...
+  - Retrieved documents (current turn only): ~3,000 tokens
```
dharaneeshvrd, line 215: can you please explain this calculation?

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1132 @@
...
+| Original Query | Conversation Context | Rephrased Query |
+|----------------|---------------------|-----------------|
+| "Is this supported on Power 11?" | Previous: "What is Spyre?" → "Spyre is an AI accelerator..." | "Is Spyre supported on Power 11?" |
+| "How do I install it?" | Previous: "What is OpenShift?" | "How do I install OpenShift?" |
+
+**Performance Considerations:**
+
+1. **Latency Impact**: Adds ~200-500ms per query (LLM call)
+   - Acceptable trade-off for accuracy improvement
+   - Can be optimized with smaller/faster models
```
dharaneeshvrd: Currently this is not possible

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1132 @@
...
+
+**Configuration:**
+
+Add to `settings.json`:
+```json
+{
+  "query_rephrasing": {
+    "enabled": true,
+    "skip_if_no_history": true,
+    "max_history_tokens": 500,
```
dharaneeshvrd: purpose of this setting?

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1132 @@
...
+
+```json
+{
+  "conversation": {
+    "enabled": true,
+    "recommended_max_history_messages": 10,
+    "client_side_management": true
+  },
+  "prompts": {
+    "query_vllm_conversational": "You are a helpful AI assistant with access to a knowledge base.\n\nPrevious Conversation:\n{conversation_history}\n\nRetrieved Context:\n{context}\n\nCurrent Question:\n{question}\n\nInstructions:\n- Answer based on retrieved context AND conversation history\n- Reference previous exchanges when relevant\n- Be conversational and natural\n\nAnswer:",
```
dharaneeshvrd: Why not keep the conversation history as part of the payload itself like generic payload.
Not sure is this a proper way to provide the conversation history to the vllm

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1132 @@
...
+
+### 8.1 Backend Configuration (`settings.json`)
+
+```json
+{
+  "conversation": {
+    "enabled": true,
+    "recommended_max_history_messages": 10,
+    "client_side_management": true
+  },
```
dharaneeshvrd: UI is taking care of this right? do we need here as well?

docs/proposals/conversational_rag.md
sats-23: Yes, we can have a sliding window logic and include last n based on context window

docs/proposals/conversational_rag.md
sats-23: Updated

docs/proposals/conversational_rag.md
sats-23: Sure

docs/proposals/conversational_rag.md
sats-23: - System prompt/instructions: ~300 tokens
- Conversation history (messages only, no docs): ~2,000 tokens
- Retrieved documents (current turn only): ~3,000 tokens
- Current query: ~512 tokens
- Total input budget: ~5,812 tokens

This is the updated calculation, we may need to tweak things based on real factors when debugging edge cases and implement the sliding window logic to help maximize context within the max input length

docs/proposals/conversational_rag.md
sats-23: It was created in view of the storing retrieved docs, removing it

docs/proposals/conversational_rag.md
sats-23: Yep, Bob slop, removing it

docs/proposals/conversational_rag.md
sats-23: Added

docs/proposals/conversational_rag.md
sats-23: New calculation has been updated, PTAL

docs/proposals/conversational_rag.md
```diff
@@ -447,93 +460,224 @@ Add to `settings.json`:
...
+      "role": "user",
+      "content": "What is Spyre?"
+    },
+    {
+      "role": "assistant",
+      "content": "Spyre is an AI accelerator for Power hardware..."
+    },
+    {
+      "role": "user",
+      "content": "You are a helpful AI assistant.\n\nRetrieved Context:\n[Fresh RAG docs about Power 11 support]\n\nInstructions: Answer based on context AND history...\nIs it supported on Power 11?"
```
dharaneeshvrd: It looks odd to me, maybe we need to do some validation around this and decide whether this prompt should be passed as system prompt and last message should contain only the context and the question.

sats-23: That's right, I have updated the approach, PTAL

docs/proposals/conversational_rag.md
```diff
@@ -906,18 +984,52 @@ The frontend sends only the conversation message history; it does not include pr
...
-5. Use conversational prompt template with message history + newly retrieved docs for the current turn + current query
-6. Generate contextual response without reusing previously retrieved documents from earlier turns
+5. Build message array:
+   - Initial system prompt (conversational behavior)
+   - Conversation history (user/assistant pairs)
+   - Final system prompt (RAG context + rephrased query)
+   - Current user question
+6. Generate contextual response
+
+**Message Array Sent to LLM:**
```
dharaneeshvrd, line 957:
```
{
  "messages": [
    {
      "role": "system",
      "content": "You are a helpful, conversational AI assistant. Always use the provided context to inform your answers while maintaining the flow of the conversation."
    },
    {
      "role": "user",
      "content": "What is machine learning?"
    },
    {
      "role": "assistant",
      "content": "Machine learning is a subset..."
    },
    {
      "role": "user",
      "content": "Retrieved Context:\n[NEW ML examples documents]\n\nInstruction: Use the context above to answer the following.\n\nQuestion: Can you give examples?"
    }
  ]
}
```


I dont think passing multiple system prompt is a general recommended way. Can we do it like the sample I attached?

Niharika0306: I agree. We can do the other way as well , but this makes it more tighter to the openai API.

sats-23: As this is a minor change, during implementation, I suggest we try both approaches and proceed with the one that behaves well in multiple test cases

sats-23:
@mkumatag Here is the reference to the Rephrased Query concept in our adaptation - https://developers.llamaindex.ai/python/examples/chat_engine/chat_engine_condense_question/

docs/proposals/conversational_rag.md
dharaneeshvrd: Sure

docs/proposals/conversational_rag.md
dharaneeshvrd: I think this is the possible value.
- Retrieved documents (current turn only): ~1500 tokens (3 chunks * 512)
- Conversation history (messages only, no docs): It is difficult to calculate, since token length of user's query can't be determined. It could be (5 * 512) + (5 * avg query token length).

But anyway like we discussed the input token length restriction's for user's query of current turn, So no need to keep the token budget under 6k like you have calculated.

But now I m wondering do we really need to allow 6k tokens for the input query?
Maybe we need to revisit this part and reconfigure it.
@Niharika0306 @sats-23

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1448 @@
...
+### 3.1 Design Philosophy
+
+The solution adopts a **client-side session management** approach with a **stateless backend**, where:
+
+* **Backend**: Remains completely stateless, no session storage
+* **Frontend**: Manages conversation message history in browser memory, excluding retrieved document payloads
+* **API**: Pure OpenAI standard, no modifications to request/response format
+* **Scalability**: Truly stateless backend enables unlimited horizontal scaling
+
+```
```
mkumatag: just see if we can convert all these diagrams to mermaid

mkumatag:
> @mkumatag Here is the reference to the Rephrased Query concept in our adaptation - https://developers.llamaindex.ai/python/examples/chat_engine/chat_engine_condense_question/

can you please include this reference somewhere in the doc and more references if you have..

docs/proposals/conversational_rag.md
```diff
@@ -0,0 +1,1448 @@
...
+    
+    "initial_system_message_de": "Sie sind ein hilfreicher, gesprächiger KI-Assistent. Interagieren Sie natürlich mit Benutzern, behalten Sie den Kontext über das Gespräch hinweg bei und geben Sie klare, genaue Antworten. Seien Sie freundlich und professionell.",
+    
+    "rag_system_message_de": "Abgerufener Kontext:\n{context}\n\nUmformulierte Anfrage: {rephrased_query}\n\nAnweisungen:\n- Beantworten Sie die Frage des Benutzers basierend auf dem oben abgerufenen Kontext\n- Verweisen Sie auf die Konversationshistorie, wenn relevant\n- Seien Sie gesprächig und natürlich in Ihren Antworten\n- Wenn der Kontext die Antwort nicht enthält, sagen Sie dies klar\n- Zitieren Sie Quellen aus dem Kontext, wenn möglich\n- Die umformulierte Anfrage zeigt, wie die Frage mit dem Gesprächskontext interpretiert wurde"
+  }
+}
+```
+
+### 5.2 Frontend Changes (Moderate)
```
mkumatag, line 690: also wondering if you want to add some button like reset which will reset the context?

mkumatag (changes requested):

docs/proposals/conversational_rag.md
sats-23: We could add one or just mention somewhere in the app info to refresh page for a new session. 
What do you suggest @dharaneeshvrd @Niharika0306

dharaneeshvrd: Agree, I think this is suffice.

mkumatag: adding anymore doc will not be helpful, leave it as is for now :)

docs/proposals/conversational_rag.md
sats-23: Done, PTAL

docs/proposals/conversational_rag.md
dharaneeshvrd: This is still not removed, can remove the whole block.

mkumatag (approved):

docs/proposals/conversational_rag.md
sats-23: Done

dharaneeshvrd (approved):

PR #621, opened 2026-04-15
addendum to the doc about summarization improvements

Niharika0306:
This PR plans on certain summarization improvements:
1.  Use abstract levels = summary_level [ brief, standard, detailed ] for length customisation from user , instead of direct word count
2. Enhance prompt for better summarisation , giving importance to the facts and details.
3. Some technical improvements on parameter levels - reasoning given in the doc
4. Enhance the flow for accurate length implementations.

Here's a sample to show how the new flow is handling the classes accurately to the described length.
 
 
Screenshot 2026-04-13 at 11 01 10 AM
Screenshot 2026-04-13 at 11 04 05 AM
Screenshot 2026-04-13 at 11 27 30 AM
Screenshot 2026-04-13 at 11 27 35 AM
Screenshot 2026-04-13 at 11 28 01 AM
Screenshot 2026-04-13 at 11 28 12 AM

### DIFF
--- a/docs/proposals/summarization_endpoint_design_doc.md
+++ b/docs/proposals/summarization_endpoint_design_doc.md
@@ -369,7 +369,7 @@ curl -X POST http://localhost:5000/v1/summarize \
 
 ## 1. Overview of Changes
 
-This addendum documents the improvements made to the summarization endpoint, introducing an abstraction-level based approach (`summary_level` parameter) to replace the direct word count specification (`length` parameter). These changes significantly improve length compliance, token utilization, and summary quality.
+This addendum documents the improvements made to the summarization endpoint, introducing an abstraction-level based approach (`level` parameter) to replace the direct word count specification (`length` parameter). These changes significantly improve length compliance, token utilization, and summary quality.
 
 ## 2. Why the Summary Length Approach Changed
 
@@ -386,7 +386,7 @@ The previous approach using direct word count (`length` parameter) had several l
 
 ### 3.1 Summary Levels
 
-The new implementation uses **abstraction levels** (`summary_level` parameter) instead of direct word counts:
+The new implementation uses **abstraction levels** (`level` parameter) instead of direct word counts:
 
 | Level | Multiplier | Description | Use Case |
 |-------|------------|-------------|----------|
@@ -414,9 +414,9 @@ The new implementation uses **abstraction levels** (`summary_level` parameter) i
 
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
-| summary_level | string | Optional | Abstraction level for summary: `brief`, `standard`, or `detailed`. Length is automatically calculated based on input size and level. If not specified, the model determines the summary length automatically. |
+| level | string | Optional | Abstraction level for summary: `brief`, `standard`, or `detailed`. Length is automatically calculated based on input size and level. If not specified, the model determines the summary length automatically. |
 
-**Note**: The `length` parameter is still supported for backward compatibility but is **deprecated and will be removed in the next release**. Please migrate to using `summary_level` instead.
+**Note**: The `length` parameter is still supported for backward compatibility but is **deprecated and will be removed in the next release**. Please migrate to using `level` instead.
 
 ### 4.2 Updated Architecture Diagram
 
@@ -432,7 +432,7 @@ graph LR
     VLLM["External vLLM<br/>endpoint"]
 
     %% Edges
-    U -->|"Input text/file +<br/>summary_level"| API
+    U -->|"Input text/file +<br/>level"| API
     API -->|"Summary"| U
     API <--> VLLM
 ```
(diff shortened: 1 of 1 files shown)


manalilatkar (dismissed):
LGTM

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -20,8 +20,8 @@ This document describes the design and implementation of a summarization endpoin
 |--------|------|-------------|----------------------------------------------------------------------------------------------------|
 | text | string | Conditional | Plain text content to summarize. Required if file is not provided.                                 |
 | file | file | Conditional | File upload (.txt or .pdf). Required if text is not provided.                                      |
-| length | integer | Conditional | Desired summary length in no. of words                                                             |
-| stream | bool | Conditional | if true, stream the content value directly. Default value will be false if not explicitly provided |
+| summary_level | string | Optional | Abstraction level for summary: `brief`, `standard` (default), or `detailed`. Length is automatically calculated based on input size and level. |
```
dharaneeshvrd: I think instead of removing, we need to provide this as a new option. Since we already did a release with this param, we should not remove immediately in next release rather we should document that this new option would work well and could be used to configurable summaries.

@mkumatag could you please provide your opinion here?

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -50,7 +50,110 @@ Error response:
...
+```
+
+**Key features:**
+- Improvement over the prior prompt to give importance to facts and details present in the document.
+- Explicit target, min, and max word counts
+- Strong directive: "MUST approach X words - do NOT stop early"
+- Detailed checklist of what to include
+- Clear boundaries for acceptable length range
+
+#### User Prompt without Length Specification (Updated)
```
dharaneeshvrd: ~~Where are we using this?~~
nm, I got this.

dharaneeshvrd (commented):
I thought this would be added as an addendum like Manali did with Async Proposal.
It's better to do that instead of editing the existing proposal IMO.

dharaneeshvrd (commented):
I would like to see a table showcasing what would the summary length in words wrt the summary levels when a file submitted with maximum allowed input.
Please add this too.

docs/proposals/summarization_endpoint_design_doc.md
mkumatag: well, ideally we should deprecate it in this version and remove in the next version like `v2/summaries`.

mkumatag: and also I didn't like summary_length term there, may be its worth revisiting the name we choose these apis

Niharika0306: You mean the 'summary_level' term @mkumatag ?  Sure, we'll look into that.

docs/proposals/summarization_endpoint_design_doc.md
```diff
@@ -357,3 +357,462 @@ curl -X POST http://localhost:5000/v1/summarize \
...
+```
+
+**Key Features:**
+- Improvement over prior prompt to emphasize facts and details
+- Explicit target, min, and max word counts
+- Strong directive: "MUST approach X words - do NOT stop early"
+- Detailed checklist of what to include
+- Clear boundaries for acceptable length range
+
+### 5.3 User Prompt without Length Specification (Updated)
```
dharaneeshvrd, line 488: If we are going to apply default summary_level for all the requests. when this prompt would get used?

docs/proposals/summarization_endpoint_design_doc.md
mkumatag: yes `summary_level`

docs/proposals/summarization_endpoint_design_doc.md
dharaneeshvrd: I feel we can keep only one prompt with default summary level for all the requests, when user passes different summary level calculate the values accordingly and format the prompt.

docs/proposals/summarization_endpoint_design_doc.md
dharaneeshvrd: how about `mode`?

manalilatkar: @mkumatag how about summary_depth, verbosity_level or granularity?

docs/proposals/summarization_endpoint_design_doc.md
Niharika0306: Okay. Currently - when no length input is given, we are not mentioning any word count in the prompt, except for the max_tokens that were calculated from the co-efficient which is passed to the llm call. And were letting the model decide. 
The standard level as default is misrepresented in the proposal.
But if we want to keep it as default - then I can do the said changes in implementation.

dharaneeshvrd: Ok in that case can you remove the reference of default?
And in next release when length param is removed, this change can be done. wdyt?

Niharika0306: I think that makes sense, since we have to support both for now. 
I will remove the default reference.

docs/proposals/summarization_endpoint_design_doc.md
mkumatag: let us discuss in the slack thread created and finalise this..

docs/proposals/summarization_endpoint_design_doc.md
Niharika0306: Done.

Niharika0306:
> I would like to see a table showcasing what would the summary length in words wrt the summary levels when a file submitted with maximum allowed input. Please add this too.

Done.

docs/proposals/summarization_endpoint_design_doc.md
Niharika0306: Changed it to 'level' now.

dharaneeshvrd (approved):

PR #622, opened 2026-04-15
Jenkins pipeline to build and sign ai-services images.

adarshagrawal38:
The pipeline takes a release tag as an input parameter, then builds and signs images corresponding to that tag.
Pipeline archive public key which will be used to verify signature of the image.

image


iv1111 (approved):

iv1111:
@adarshagrawal38 merge when ready.

PR #629, opened 2026-04-16
Fix: Change log level for container logs

Shubhamag12:
JIRA: https://jsw.ibm.com/browse/AISERVICES-932

Changed logging in container logs from logger.Errorln(line) to logger.Infoln(line) so that stderr output is displayed as-is without the ERROR prefix. This may be because many containers write all their logs to stderr regardless of the actual log level.

### DIFF
--- a/ai-services/internal/pkg/runtime/podman/podman.go
+++ b/ai-services/internal/pkg/runtime/podman/podman.go
@@ -220,7 +220,7 @@ func (pc *PodmanClient) streamContainerLogs(ctx context.Context, containerNameOr
 				if !ok {
 					return
 				}
-				logger.Errorln(line)
+				logger.Infoln(line)
 			}
 		}
 	}()


Shubhamag12:
Ref:
```
--- Following logs for pod: rag-app-1776228260--clean-docs ---
Streaming logs for container: rag-app-1776228260--clean-docs-clean-docs
ERROR:  2026-04-15 06:07:40 - digitize_utils     - INFO     - Found 2 document IDs in /var/cache/docs

ERROR:  2026-04-15 06:07:42 - OpenSearch         - INFO     - Successfully deleted 1094 chunks for 2 documents from rag_db_913c2e34e90d89c2245dd0bf674fdf40
```

Info logs are written to stderr
Since logs levels are determined already as seen above `INFO`, its better to not have a prefix as ERROR, INFO etc

ai-services/internal/pkg/runtime/podman/podman.go
```diff
@@ -220,7 +220,7 @@ func (pc *PodmanClient) streamContainerLogs(ctx context.Context, containerNameOr
 				if !ok {
 					return
 				}
-				logger.Errorln(line)
+				logger.Infoln(line)
```
yussufsh, line 223: Do we need similar change for OpenShift runtime log stream?

yussufsh (changes requested):
Is this a change for a bug? please add JIRA reference or add a fix prefix to the commit message.

ai-services/internal/pkg/runtime/podman/podman.go
Shubhamag12: We are using Infoln in openshift

yussufsh (approved):

PR #631, opened 2026-04-17
vLLM Authentication Proposal

sats-23:
[AISERVICES-913](https://jsw.ibm.com/browse/AISERVICES-913)

### DIFF
--- a/docs/proposals/vllm_authentication.md
+++ b/docs/proposals/vllm_authentication.md
@@ -277,14 +277,17 @@ response = requests.post(instruct_url, headers=headers, json=payload)
 
 #### QnA Service - Chat Completion Endpoint
 
-The QnA service's `/v1/chat/completions` endpoint uses `llm_utils.py` functions to communicate with vLLM. Authentication is handled in the `llm_utils` module by adding the Authorization header to all vLLM requests.
+The QnA service's `/v1/chat/completions` endpoint should validate vLLM authentication before doing any retrieval or generation work. The check is performed by calling `GET /v1/models` using the same Authorization header that will be used for the later chat completion request. If the auth check succeeds, the normal QnA flow continues. If it fails, the endpoint returns early with an authentication error for vLLM.
 
 **Implementation in `common/llm_utils.py`:**
 
 ```python
 import os
+import time
 import requests
+import common.misc_utils as misc_utils
 from common.misc_utils import get_logger
+from common.retry_utils import retry_on_transient_error
 
 logger = get_logger("LLM")
 
@@ -292,33 +295,47 @@ logger = get_logger("LLM")
 VLLM_INSTRUCT_API_KEY = os.getenv("VLLM_INSTRUCT_API_KEY", "")
 
 def get_vllm_headers():
-    """
-    Get headers for vLLM API calls, including authentication if configured.
-    
-    Returns:
-        dict: Headers dictionary with Authorization header if API key is set
-    """
+    """Get headers for vLLM API calls, including auth if configured."""
     headers = {
         "accept": "application/json",
-        "Content-type": "application/json"
+        "Content-type": "application/json",
     }
-    
-    # Add Authorization header if API key is configured
+
     if VLLM_INSTRUCT_API_KEY:
         headers["Authorization"] = f"Bearer {VLLM_INSTRUCT_API_KEY}"
         logger.debug("Using vLLM API key for authentication")
-    
+
     return headers
 
-# Update existing functions to use get_vllm_headers()
 
-def query_vllm_payload(question, documents, llm_endpoint, llm_model, stop_words, max_new_tokens, temperature,
-                stream, lang):
+@retry_on_transient_error(max_retries=3, initial_delay=1.0, backoff_multiplier=2.0)
+def query_vllm_models(llm_endpoint):
+    """Used both for listing models and as an auth/availability preflight check."""
+    if misc_utils.SESSION is None:
+        raise RuntimeError("LLM session not initialized. Call create_llm_session() first.")
+
+    logger.debug("Querying VLLM models for auth validation")
+    response = misc_utils.SESSION.get(
+        f"{llm_endpoint}/v1/models",
+        headers=get_vllm_headers(),
+    )
+    response.raise_for_status()
+    return response.json()
+
+
+def validate_vllm_auth(llm_endpoint):
+    """
+    Validate that the configured credentials can access vLLM.
+    Returns True on success, raises requests.HTTPError on failure.
+    """
+    query_vllm_models(llm_endpoint)
+    return True
+
+
+def query_vllm_payload(question, documents, llm_endpoint, llm_model, stop_words, max_new_tokens, temperature, stream, lang):
     # ... existing context and prompt logic ...
-    
-    # Use the new header function
+
     headers = get_vllm_headers()
-    
     payload = {
         "messages": [{"role": "user", "content": prompt}],
         "model": llm_model,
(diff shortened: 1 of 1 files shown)


docs/proposals/vllm_authentication.md
```diff
@@ -0,0 +1,372 @@
...
+**Key Storage - Hybrid Approach** (Similar to OpenSearch auth password):
+
+| Environment | Storage Method | How Keys Are Used |
+|-------------|---------------|-------------------|
+| **Podman** | Values file directly | Keys read from values, passed as env vars to containers |
+| **OpenShift** | Kubernetes Secret (created from values) | Keys stored in Secret, mounted/referenced in deployments |
+
+**Key Generation & Deployment Flow**:
+```
+1. Generate Keys (if authEnabled=true and keys empty)
```
mkumatag: can we expect the user to supply the key manually if not just just don't set? and store the key in the secret so that we can use that wherever required!

sats-23: If my understanding is right, are you suggesting, 
We keep auth disabled by default and if the user provides the key, we use that key, else keep auth disabled?

mkumatag: yes, that is what I'm thinking.. or can we even have a default password! I'm just trying to avoid any extra logic required for generating anything(this will become application/service specific logic which I want to avoid), let me know if you any different approach.

manju956: I think its not a good practice to keep authN disabled by default. We can have some default vllm auth keys for each vllm service.

docs/proposals/vllm_authentication.md
```diff
@@ -0,0 +1,372 @@
...
+## 5. Configuration Structure
+
+### 5.1 values.yaml Schema
+
+```yaml
+vllm:
+  authEnabled: true  # Default: true (authentication enabled)
+
+instruct:
+  apiKey: ""  # Auto-generated if authEnabled=true and empty
```
manju956: Can we keep these keys in base64 encoded format just to avoid storing the keys in plain text?

docs/proposals/vllm_authentication.md
sats-23: Yes, we can have default auth keys for each service which are overwritten if passed by user.

docs/proposals/vllm_authentication.md
sats-23: Sure

sats-23: Updated

docs/proposals/vllm_authentication.md
dharaneeshvrd: +1 to a default password/keys

> this will become application/service specific logic which I want to avoid

Can we have an init container or similar way to generate these keys via FIPS complaint encryption algorithm and use it in vllm? And same init container can be used to generate and use it in the services(digitize, chatbot, summarize).

sats-23: I have updated the approach to use an FIPS compliant algo to generate an api key using password (default/user provided) within the startup command of each service and programmatically in python for each client (passed to auth header). PTAL

docs/proposals/vllm_authentication.md
sats-23: I have removed the base64 encoding. 
Current approach is to store plain text password similar to open search password.
This is later converted into a key using PBKDF2 (runtime) before actual usage

docs/proposals/vllm_authentication.md
```diff
@@ -0,0 +1,640 @@
...
+      - -c
+      - |
+        # Derive API key using PBKDF2 (FIPS-compliant)
+        # Password is used directly as cipher text input
+        export VLLM_API_KEY=$(echo -n "$VLLM_INSTRUCT_PASSWORD" | openssl enc -pbkdf2 \
+          -pass stdin \
+          -S "$(echo -n "$VLLM_SERVICE_SALT" | xxd -p)" \
+          -iter 600000 \
+          -md sha256 \
+          -P 2>/dev/null | grep "key=" | cut -d'=' -f2)
```
mkumatag: can we remove this and depend on user to supplied the credentials? we will learn more on this and later enhance this logic if needed.

mkumatag (changes requested):

docs/proposals/vllm_authentication.md
mkumatag: because in a micro service architecture it is almost close to impossible to know the credentials generated automatically runtime by the other service.

dharaneeshvrd: Applying same encryption algorithm on same user password on vllm & all the services would generate same key, that's the logic behind this.
Where do you think this won't be the case Manju?

sats-23: Yes, goal is to use same input cipher with same salt, and same number of iterations against same algorithm.

mkumatag: This will add an additional overhead on both the consumer's side as well, for example: if someone wants to use this VLLM via a `curl`, they will have to derive this key beforehand, which adds more complexity.

And also with this you are going to change spec for lot of service which may be problematic is what my opinion, wondering how folks are handling this in the real deployment of vLLM containers?

dharaneeshvrd: @mkumatag 
Do you have plans to expose vLLM to the end user?
Currently vLLMs are not exposed to the end user in both podman & openshift, I think we deliberately did this with the intention of exposing only the AIS services.

> how folks are handling this in the real deployment of vLLM containers?

Agree, we need to follow what is the generic standard, Let's wait for @sats-23 to look at this and come back

sats-23: In real world scenario, `--api-key` is used as a second layer of defense, as it only protects against few select endpoints like `/v1/chat/completions.` 
Typically, it is deployed with an API gateway that protects ALL endpoints, rate limits and provides audit trail logs.

But `--api-key` is the way to go to ensure only service to service talk at the app level. 

In extended scenarios like OpenShift, the Route/Load Balancer handles SSL/TLS termination.

dharaneeshvrd: > only protects against few select endpoints like /v1/chat/completions.

What are the endpoints are not covered as part of this?

External gateway might not be possible for this release.
If we go with `--api-key` for now, is there any recommended way on encrypting the key or plain text can be used?

mkumatag: Hmm, we need to take a step back and see what is required:
1. Secure the chat completion endpoint we are exposing (likely via our backend).
2. Implement security for the individual vLLM service (this will come in handy when we deploy independent instances for use with other services). - Influenced from UI designs from design team.

sats-23: [This](https://docs.vllm.ai/en/stable/usage/security/#protected-endpoints-require-api-key) lists all endpoints that are covered vs not covered. 

There is no recommended way of encrypting the key. 
All usages I could spot online are using plain text. 
I believe our crypto algorithm just adds a thin layer of abstraction beyond that. 
The official vLLM doc also [lists](https://docs.vllm.ai/en/stable/usage/security/#:~:text=Important:%20Do%20not%20rely%20exclusively%20on%20%2D%2Dapi%2Dkey%20for%20securing%20access%20to%20vLLM.%20Additional%20security%20measures%20are%20required%20for%20production%20deployments.) security implications with this var

dharaneeshvrd: ok I think for now we can go ahead with this approach(--api-key with plain text) and when we implement system wide authorization and authentication, an external gateway can be integrated.

> Secure the chat completion endpoint we are exposing (likely via our backend).

For this one, we need to add an extra auth layer for Q&A service for this endpoint alone.

sats-23: So, If I get it right
-We simplify the logic to use plain text instead of cryptography 
-We add a reverse proxy or similar solution for QnA service against /v1/chat/completion alone

Please confirm?

mkumatag: > We add a reverse proxy or similar solution for QnA service against /v1/chat/completion alone

wondering if we can just use the vllm's help for this?

sats-23:
@mkumatag @dharaneeshvrd 
The approach has been updated as per discussion, PTAL

docs/proposals/vllm_authentication.md
```diff
@@ -0,0 +1,421 @@
...
+Application 'my-app' created successfully
+=====================================
+```
+
+**What Happens:**
+1. `vllm.authEnabled` is `true` by default
+2. Hardcoded password is used across all services (instruct, embedding, reranker)
+3. Password is passed directly to vLLM servers via env var
+4. Client services use the same hardcoded password in Authorization headers
+5. No password storage in values file - password is hardcoded in code
```
dharaneeshvrd: Users should know the default password, we can follow the opensearch flow where default password is defined in values.yaml and user can override it using params. 
Using application template command user know about this param and default value.

docs/proposals/vllm_authentication.md
```diff
@@ -0,0 +1,421 @@
...
+```
+
+#### Behavior Matrix
+
+| authEnabled | vLLM Behavior |
+|-------------|---------------|
+| true | Authentication enabled with hardcoded password |
+| false | Authentication disabled |
+
+### 6.2 Client-Side (FastAPI Python Services)
```
dharaneeshvrd, line 232: I dont see how we are going to handle the auth for /v1/chat/completion in QnA service, please take care of that

docs/proposals/vllm_authentication.md
mkumatag: can we add this auth only to the vLLM instruct which is exposed outside others we can have still no-auth?

docs/proposals/vllm_authentication.md
```diff
@@ -0,0 +1,421 @@
...
+### Current State
+- vLLM services (instruct, embedding, reranker) are deployed **without authentication**
+- Any client with network access can consume vLLM APIs
+- No access control or audit trail for API usage
+- Security risk in production environments
+
+### Requirements
+1. **Secure by Default**: Authentication must be enabled automatically during application creation
+2. **Simple Configuration**: Single hardcoded password used across all services
+3. **Opt-Out Mechanism**: Users can disable authentication via explicit flag for development/testing
```
mkumatag: we can consider an empty API_KEY as an no-auth?

docs/proposals/vllm_authentication.md
```diff
@@ -0,0 +1,421 @@
...
+
+### Current State
+- vLLM services (instruct, embedding, reranker) are deployed **without authentication**
+- Any client with network access can consume vLLM APIs
+- No access control or audit trail for API usage
+- Security risk in production environments
+
+### Requirements
+1. **Secure by Default**: Authentication must be enabled automatically during application creation
+2. **Simple Configuration**: Single hardcoded password used across all services
```
mkumatag: let us use API_KEY everywhere instead of password term, how about this?


```suggestion
2. **Simple Configuration**: User-Provided API Key - The system authenticates using an API key supplied directly by the user.
```

docs/proposals/vllm_authentication.md
```diff
@@ -0,0 +1,421 @@
...
+            +--[Invalid/Missing]-----> 401 Unauthorized
+```
+
+### 3.2 System Components
+
+| Component | Role | Implementation |
+|-----------|------|----------------|
+| **vLLM Server** | Validates password using env var | Native vLLM support (v0.4.1+) |
+| **Client Services** | Include `Authorization: Bearer <password>` header | Python utilities (misc_utils, emb_utils, llm_utils) |
+| **Configuration** | Controls auth behavior with hardcoded password | values.yaml with vllm.authEnabled flag |
```
mkumatag: do we really need `vllm.authEnabled`? can we just have a helm way - don't supply this flag to vLLM if it is an empty.

mkumatag (changes requested):

docs/proposals/vllm_authentication.md
dharaneeshvrd: I think this is better.

sats-23: Sure, will do

docs/proposals/vllm_authentication.md
sats-23: The no-flag approach has been updated

docs/proposals/vllm_authentication.md
sats-23: @dharaneeshvrd If we go with a no-flag approach (no authEnabled), then by default we will have auth disabled. In case the user passes a key, it will be enabled 

@mkumatag sure, we shall remove it on the other services

docs/proposals/vllm_authentication.md
sats-23: Done, added

docs/proposals/vllm_authentication.md
```diff
@@ -0,0 +1,460 @@
...
+    return headers
+
+# Example usage in API calls
+headers = get_vllm_instruct_headers()
+response = requests.post(instruct_url, headers=headers, json=payload)
+```
+
+#### QnA Service - Chat Completion Endpoint
+
+The QnA service's `/v1/chat/completions` endpoint uses `llm_utils.py` functions to communicate with vLLM. Authentication is handled in the `llm_utils` module by adding the Authorization header to all vLLM requests.
```
dharaneeshvrd: Don't we need to add this auth for chat completion in QnA service?
:(

sats-23: Sorry for misunderstanding, updated the approach
PTAL

dharaneeshvrd (dismissed):

manju956 (dismissed):
lgtm

docs/proposals/vllm_authentication.md
```diff
@@ -0,0 +1,510 @@
...
+```
+
+#### Integration Points
+
+1. Load template values with `vllm.instruct.apiKey` field (default: empty)
+2. User can override via `--params vllm.instruct.apiKey=<value>`
+3. Render templates with user-provided API key if set
+4. Deploy application
+
+**Implementation Flow**:
```
mkumatag: can we remove this portion, we shouldn't be doing anything in the application create workflow

mkumatag (changes requested):

docs/proposals/vllm_authentication.md
sats-23: Done, removed

dharaneeshvrd (approved):

mkumatag (approved):

PR #635, opened 2026-04-17
Catalog Backend Design proposal

mayuka-c:
- Overall Architecture
- Catalog Installations and CLI commands
- DB Designs
- API Designs

Copied mostly from the docs maintained in IBM box.

There will be changes on DB design like
- PK for application table
- Removal of architecture table
- Removal of users table in accordance with storing in podman secrets.
- Preferring to have seperate infra table. Reasons mentioned in the doc. I do have documented clubbing services and infra table into one single table with its Pros and Cons. Open for any suggestions :) -> Edit Based on todays meeting we go with 1:1 and stay with having only services table and later in next release we will go with separate ones.

- API calls with appropriate endpoints, schema, implementation steps (For applications endpoints).


docs/proposals/catalog/architecture.md
```diff
@@ -126,66 +155,65 @@ To ensure strict feature and security parity, both the **Catalog UI** and the **
...
 ### 6.1 Container Images
 
-The solution is packaged into two primary container images. These are hosted in a enterprise registry (e.g., ICR) and pulled during the bootstrap phase.
+The solution is packaged into three primary container images. These are hosted in an enterprise registry (e.g., ICR) and pulled during the bootstrap phase.
 
 | Image Alias | Image Name | Base OS / Tech Stack | Role |
 | --- | --- | --- | --- |
 | **API Server** | `catalog-api:v1` | Red Hat UBI 9 (Minimal) / Go | Orchestration, Auth, & Infrastructure Interfacing |
 | **Catalog UI** | `catalog-ui:v1` | Red Hat UBI 9 (Nginx or Equivalent) / React | Carbon-based Web Portal & Asset Hosting |
+| **Database** | `postgresql:18` | Official Image | Persistent data storage for users, deployments, and audit logs |
```
mkumatag: there is always some limitation on a images we directly or use from the docker.io, please refer usage guidelines.

mayuka-c: Sorry my bad we will be using the postgress-ppc64le available in icr.io. Have updated it.
Here is the image: icr.io/ppc64le-oss/postgresql-ppc64le:16.3.

Note: Currently it is 16.3, we may need to have 18+ for any CVE fixes and also there are performance optimizes like UUID_V7 support and others.

docs/proposals/catalog/catalog-services-installation-proposal.md
```diff
@@ -0,0 +1,549 @@
...
+#### Flags
+
+| Flag | Type | Required | Default | Description |
+|------|------|----------|---------|-------------|
+| `--runtime` / `-r` | string | Yes | - | Target runtime environment (podman, openshift) |
+| `--params` | string[] | No | [] | Inline parameters to configure catalog service (comma-separated key=value pairs) |
+
+#### Interactive Password Input
+
+For enhanced security, the configure command prompts for the administrator password interactively rather than accepting it as a command-line flag. This approach:
```
yussufsh, line 111: Is there any way to change the password?
What happens when a user forgets the password? Any way to recover?

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1399 @@
...
+- `GET /api/v1/applications` - List all deployments
+- `GET /api/v1/applications/{appName}` - Get deployment details
+- `POST /api/v1/applications` - Create new deployment
+- `PUT /api/v1/applications/{appName}` - Update deployment
+- `DELETE /api/v1/applications/{appName}` - Delete deployment
+- `GET /api/v1/applications/{appName}/ps` - Get pod/container health status
+
+#### Catalog Endpoints
+- `GET /api/v1/architectures` - List available architectures
+- `GET /api/v1/architectures?name="Digital Assistant"` - Get architecture details
```
yussufsh: Let's use id here, which will be defined in the metadata.
For `Digital Assistant`, id would be `rag`.

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1399 @@
...
+
+- `GET /api/v1/services` - List available services
+- `GET /api/v1/services?name="Summarization"` - Get service details
+- `GET /api/v1/services/params?name="Summarization"` - Get service custom params
```
yussufsh: Same as above:
Let's use id here, which will be defined in the metadata.
For `Summarization` service, id would be `summarization`.

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1399 @@
...
+- If skip-cleanup=false (default), clean up all application data
+- Clean up database records after successful deletion
+- For OpenShift: delete namespace and all resources
+- For Podman: stop and remove all containers
+
+---
+
+#### 5.2.6 Get Pod/Container Health Status
+
+**Endpoint:** `GET /api/v1/applications/{appName}/ps`
```
yussufsh: Is this shown on the catalog ui in any way today?
IMO we should use the service table for service status and endpoints.

Just wondering where this will be used in future, except the cli. Maybe mark it as a stretch for the future.

yussufsh: We will need API to show h/w resources allocated and used by an application.

docs/proposals/catalog/db-design-proposal.md
```diff
@@ -0,0 +1,600 @@
...
+
+**Table Name:** `services_infra`
+
+| Column Name | Data Type | Constraints | Description |
+|-------------|-----------|-------------|-------------|
+| service_id  | UUID      | PRIMARY KEY, FOREIGN KEY | References services(id) |
+| infra_id    | UUID      | PRIMARY KEY, FOREIGN KEY | References infra(id) |
+
+**Note:** This is a many-to-many relationship table with a composite primary key.
+
```
yussufsh, line 168: These are mostly needed for reusable service instances. Let's mark them for the future.
When we allow reusing an existing instances new tables will be easy to plug-in as right now they will be empty. Also `Infrastructure` looks like the same columns as `Service` table, we can reuse it later.

docs/proposals/catalog/catalog-services-installation-proposal.md
mayuka-c: As discussed, if user forgets the password he need to reset the password.
For resetting, I will no longer keep the password in DB, but rather as a secret in podman. They probably can use `echo "new_secret_data" | podman secret create --replace <secret_name> -` to replace the password and restart the pod manually.

If time permitts maybe we can expose one more catalog subcommand through CLI to reset the admin password.

docs/proposals/catalog/db-design-proposal.md
mayuka-c: Have moved it at the end. For now there will be only 2 tables (Applications and services) table.
PTAL

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: The `ps` command would be to monitor the Health stats of individual pods in a application. I guess we may need this along with showing logs of pods streamed maybe in future?

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Done

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Done

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: `We will need API to show h/w resources allocated and used by an application.`
Where exactly are we showing this?. I thought we didnt want to show any resources right?

mayuka-c:
Have updated the description with recent updates done after today's internal squad discussion. PTAL

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+
+All endpoints (except `/auth/*`) require JWT Bearer token authentication:
+```
+Authorization: Bearer <access_token>
+```
+
+**Token Lifecycle:**
+- Access tokens expire after 15 minutes
+- Refresh tokens valid for 7 days
+- Token blacklisting on logout
```
mkumatag, line 128: This is in the memory rigthnow, we need to explore pushing this to DB once we have PostgreSQL in place

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+   - Parse token and extract expiry time
+   - If token is already invalid, treat as success (idempotent operation)
+3. **Blacklist Addition**: Call `TokenBlacklist.Add(token, expiryTime)` to:
+   - Add token to in-memory blacklist map
+   - Store with expiry time for automatic cleanup
+4. **Response**: Return success message "logged out"
+5. **Error Handling**: Return 400 for missing token, 500 for blacklist failures
+
+**Blacklist Implementation:**
+- Uses `InMemoryTokenBlacklist` with map[string]time.Time storage
```
mkumatag: As I mentioned above we need to start exploring storing the list in the DB for persistent.

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+- Background garbage collection runs every 1 minute to remove expired tokens
+- `Contains()` method checks blacklist and removes expired entries on-the-fly
+- **Note**: In-memory implementation only suitable for single-instance deployments
+- **Production**: Use Redis with TTL or distributed cache (Memcached) for multi-instance setups
```
mkumatag: One more DB in the prod will add more complexity, may be a goroutine to take care of this once in a while.

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+|-------|------|-------------|
+| data | array | Array of application objects |
+| pagination | object | Pagination metadata |
+
+**Application Object (data[]):**
+| Field | Type | Description |
+|-------|------|-------------|
+| id | string | Application ID (Primary Key, immutable, used for resource naming) |
+| deployment_name | string | User-friendly display name of the deployment |
+| deployment_type | string | Type of deployment: "Architecture" or "Service" |
```
mkumatag: lets keep them in lowercase

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+| type | string | Application type: "Digital Assistant" for architectures, "Summary" for summarization services |
```
mkumatag, line 557: will this be unique?

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+**Example Requests:**
+```
+# Get first page with default settings
+GET /api/v1/applications
+
+# Get second page with 50 items per page
+GET /api/v1/applications?page=2&page_size=50
+
+# Get running applications sorted by id
+GET /api/v1/applications?status=Running&sort_by=id&sort_order=asc
```
mkumatag: these params like `status`, `type`, `sort_by` or `sort_order` is not present in the above param, if we don't need then let us remove this from here

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+| type | string | Service type (e.g., "QA-Chatbot", "Summary", "Digitization") |
+| endpoints | array | Array of endpoint objects |
+| version | string | Service version |
+| created_at | string | Creation timestamp |
+| updated_at | string | Last update timestamp |
+
+**Endpoint Object:**
+| Field | Type | Description |
+|-------|------|-------------|
+| name | string | Endpoint name: "ui", "backend", or "api" |
```
mkumatag: need to add more detail here like what kind of endpoint will be used by the end customer vs internal other services, and adding examples here will be better for understanding

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+  }
+}
+```
+
+**Request Schema:**
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| deployment_name | string | Yes | User-friendly display name (3-100 chars) |
+| deployment_type | string | Yes | Deployment type: "Architecture" or "Service" |
+| type | string | Yes | Application type (e.g., "Digital Assistant", "Summary") |
```
mkumatag: there are multiple type fields here may create confusion

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+| template | string | Yes | Template ID (e.g., "rag" is the ID for "Digital Assistant" type) |
```
mkumatag: why do we need this if we are calling `Digital Assistant` as RAG

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+| created_by | string | Yes | Username of the user creating the application |
+| services | array | No | Array of service objects to deploy. If not provided, all services from the architecture template will be deployed. For Service deployment_type, this should contain a single service. |
+| params | object | No | Key-value pairs of custom parameters for the deployment |
```
mkumatag: going forward if we consider params are more close to services then can we shift this to services field?

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+
+**Implementation Notes:**
+1. **Token Validation**: Validate JWT token from Authorization header
+
+2. **User Validation**:
+   - Validate that `created_by` matches the authenticated user from the JWT token
+   - Extract user ID from JWT token and verify it corresponds to the provided `created_by` username
+   - Return 403 Forbidden if `created_by` doesn't match the authenticated user
+
+3. **ID Generation**:
```
mkumatag: can this be UUID? I see in the update operation you are allowing user to update the deployment_name, this will soon become outdated and will create more confusion.

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+|-----------|------|----------|-------------|
+| id | string | Yes | Application ID |
+
+**Request Body:** None
+
+**Response (200 OK):**
+```json
+{
+  "id": "rag-production",
+  "pods": [
```
mkumatag, line 1169: is it possible to get the pods/service mapping somehow so that it becomes easy to map and play with them when needed.

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+    {
+      "pod_id": "a1b2c3d4e5f6",
+      "pod_name": "rag-production-chat-ui",
+      "status": "Running (Ready)",
+      "created": "2d5h",
+      "exposed": "8080, 8081",
```
mkumatag: why not an array of ports?

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+```
+
+**Response Schema:**
+| Field | Type | Description |
+|-------|------|-------------|
+| id | string | Architecture template ID |
+| name | string | Architecture template name |
+| description | string | Description of the architecture |
+| version | string | Architecture version |
+| type | string | Type (architecture) |
```
mkumatag, line 1286: if they are always architecture then why do you need this type filed here?

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+| certified_by | string | Certification authority |
+| services | array | Array of service objects |
+| links | object | Related links (demo, code, documentation) |
+
+**Service Object Schema:**
+| Field | Type | Description |
+|-------|------|-------------|
+| id | string | Service ID |
+| version | string | Version constraint |
+| optional | boolean | Whether service is optional (only present if true) |
```
mkumatag, line 1379: if they are optional, then will also have to think about giving an optional checkmark during the deployment as well.

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+```json
+[
+  {
+    "id": "chat",
+    "name": "Question and Answer",
+    "description": "Answer questions in natural language by sourcing general & domain-specific knowledge",
+    "version": "1.0.0",
+    "type": "service",
+    "certified_by": "IBM",
+    "architectures": ["rag", "rag-cpu", "rag-dev"]
```
mkumatag: do we really need this and where this will be used?

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+- Read architecture metadata from `ai-services/assets/architectures/{id}/metadata.yaml`
+- Parse YAML and convert to JSON response format
+
+---
+
+#### 5.3.3 List Available Services
+
+**Endpoint:** `GET /api/v1/services`
+
+**Description:** Retrieves a list of all deployable service templates. Dependency-only services are excluded from this list.
```
mkumatag, line 1393: do we have any thoughts like deploy vllm instruct and use them in multiple places?!

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+**Error Responses:**
+- `401 Unauthorized` - Invalid or missing access token
+- `500 Internal Server Error` - Server error
+
+**Implementation Notes:**
+- Check out the proposal https://github.com/IBM/project-ai-services/pull/636
+- Read services from `ai-services/assets/services/` directory
+- Filter OUT services that have `dependency_only: true` in their metadata
+- Only return deployable services (chat, summarization, digitization)
+- Dependency-only services (opensearch, embedding, instruct, reranker) should NOT be included in the response
```
mkumatag, line 1456: what is thought process behind not exposing these dependent only services?

docs/proposals/catalog/application-deployment-api-proposal.md
```diff
@@ -0,0 +1,1599 @@
...
+| id | string | Yes | Service template ID (e.g., "rag", "summarize") |
+
+**Request Body:** None
+
+**Example Request:**
+```
+GET /api/v1/services/rag/params
+```
+
+**Response (200 OK):**
```
mkumatag, line 1688: is there a way to classify the params like mandatory, advanced etc so that UI can hide the advanced ones by default?

mkumatag: and your create workflow should respect this spec and return early error..

mkumatag (changes requested):

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Thanks for pointing out

Looks like I completely missed out on the blacklisting part and thought of only user storage part (Which we now store it in podman secret as it is only a single admin user)

I will try to model this in Postgres DB

mayuka-c: But upon thinking more, do we really want to store the blacklisted tokens as it is currently used for a single user (admin)?
- So currently if an admin logs in, he will get access to the token for 15 mins.
- Upon logging out, we store the token in in-memory as it is still not expired and move it to Blacklist.
- So if a same token is used, then we error back with 409 (Forbidden).
- And we do have a garbage collector which runs every minute to remove all the black listed tokens which have expired.

Here's my analysis
- We can still keep it in memory as there will be no im-memory issue since we keep removing the tokens.
- If lets say a Pod restarts then the memory is lost, so during the time if admin has logged in, then he need to relogin back.
- If lets say there was a blacklist token and during that time if a pod restarts, then that token is lost. This is the only time when there will be problem till the token expires (which is in worst case scenario 15 mins, which should be fine IMO or if required we can even make it 10 mins). Looks like people prefer having shortlived tokens with expiry set to around 5mins.
- This avoids the overhead of maintaining a new table just to store the tokens which involves basically the IAM management which I feel will not be required in next release as we anyway have to move to a dedicated IAM tool which simplifies all of these.

Based on my analysis, I feel we can still be rely on in-memory as it is simple, easy and works fine for a single admin user but open for any suggestions

PLMK your thoughts

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Pls check my above comment

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Pls check my above comment

mayuka-c: Probably I will remove Production one based on my above analysis

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Yeah I removed from the description, will remove the examples

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: The ID here would be the initial deployment name that the user will provide. So it will be the normalized form of it.

The reason for doing this is because currently we are using the app name for prefixing the pod names (in Podman) and namespace name (in OpenShift).

So if a user again uses the same deployment name then we do not want to add one more entry but rather say application already exist.

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: The endpoints object here is for each service like Digitize, summarization.
So the name here would be the container name (in Podman), I think that is how we can differentiate and make it more generic to know which endpoint is for what.

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: The `type` here is something which I have taken from Figma where they call Digital Assistant, Summary as type.
The `deployment_type` is whether its an architecture or service deployment.

PLMK if you have any better suggestions for these fields. Note: this would translate to the same column names in DB.

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: We wanted to keep the rag as it is as how we are using it as part of `application create` flow where we take rag as template. So the rag would act as id for Digital Assistant (We are thinking Digitial Assistant can be editable).

This is the thought process. @yussufsh Please correct if I'm wrong

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Here is the explanation to it: https://github.com/IBM/project-ai-services/pull/635#discussion_r3146319135

So the problem is we were using appName today to prefixing, for filtering out the pod with labels and also namespace name (OpenShift).

Open for any discussions/ suggestions.

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: With the current design, the params is tied to application.
So as part of application we read params like "opensearch.password", so here opensearch automatically means the service is `opensearch`.

So do you mean we decouple the above behavior. if yes this might require a significant change to read params in a diff way.

mayuka-c: I feel we can still it read it same way, where we have the logic that here `"opensearch.password"` -> the first part always corresponds to a service (That is before dot) and next part represents the field.

PLMK your thoughts

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Didnt quite get it, can you pls elaborate.

It already does listing of all the pods for a given app.
Pods service mapping meaning you want to know which pod is part of which service?. But isnt this something you already know by seeing the pod name.

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Done

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Yes true
Removed

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Done have added it in deployment.
But I dont think we will be doing optional for current release. But will help us in next release.

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: I dont think UI will utilize it anywhere IMO.

@yussufsh Can you please let me know if its required or I will remove or was there anything in your mind by exposing rag, rag-cpu and rag-dev here?

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: I dont think we are doing this for current release if I'm not wrong.
cc. @yussufsh

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: I dont think we will be even returning endpoints as part of list available deployable services as you can see from the response.

But the Catalog endpoints were added by @yussufsh while he was working on asset structure. So would let him say if we can remove it.

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: This uses the json schema (very similar to values.schema.json in helm chart) here we can specify `required` if the field is required.

Advanced I dont think so its there, we may need to implement a new annotation I guess

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Also there are multiple things we need to consider for reusability like lets say during application deployment (lets say app-1), user passed opensearch.password with "test123", this is something we might need to persist in DB or as a secret file (But DB we cannot store password), secret file means we can do (We can create a new secret file, store the reference of it in DB) and then while deploying tell the service to read from the secret.

But the problem with secret file approach is it works in podman, but in OpenShift it fails, as I cannot reference secret deployed in other namespace or else need to maintain the secret object across all the namesapces which causes maintainability issues if we want to do secrets rotations.

I'm thinking we might need a new centralized secret management tool like Vault, Openbao which stores the config files and we reuse it across.

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: The scenario I'm considering is:
1. User logs in and get the token
2. Log out - this is when it gets added into the blacklist
3. restart the pod - blacklist is gone
4. User can use the refresh token which is logout out and get the new access token.

There is chance that user can create the access token from the blacklisted stale refresh token is a security risk!

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: I'm talking about type field

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: Do you think we also have a type of endpoint that would enable UI to create intelligent messaging based on these endpoint types? Maybe they could also perform additional operations, such as if an endpoint is an API/REST endpoint, can they read the Swagger spec (a futuristic thing) and generate some intelligent documentation or info? just a thought!

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: IMO: these things needs to be revisited for sure, I'm not convinced with these fields

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: As per today's discussion as all these fields are pushed towards to service layer and definitely need some logic for sure, whether <service>.<param> here and send the params to the each service or services will have unique keys, worth spending some time and architect based on the leaning from @yussufsh 's POC

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: my point is - if you are allowing user to update the deployment name post deployment then this whole logic becomes invalid.

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: I feel it will be better if we somehow map the pods to services! then user will have clue what service got failed or also it will be easy for us to say - because of what pod and service the application is problemetic, it will be helpful during the debug time.

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: let us remove this mapping for now and will explore if really needed later.

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: yes, this adds more complexity for now but keep the architecture little more flexible so that we can enhance later if needed! there are ways of accessing the secrets from a different namespaces(but do we really wanna do that? - not sure!) or try something else!

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: we will revisit this later

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: let us explore.. basic vs advanced if there is a way to do so.

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Yeah maybe we can do that, also we need to see how we can store this type in endpoint object during application deployment flow. Will keep a note of it.

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Ah ok, any reason we want type to be unique?

It will be "Digital Assistant", "Summarize" etc, basically the application type

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Yes as I mentioned in my 3rd point, definitely there is a security risk.
```
If lets say there was a blacklist token and during that time if a pod restarts, then that token is lost. This is the only time when there will be problem till the token expires (which is in worst case scenario 15 mins, which should be fine IMO or if required we can even make it 10 mins). Looks like people prefer having shortlived tokens with expiry set to around 5mins.
```

I will create a new table for this and document.

Also just a query,  I'm still thinking is it even worth doing it as it is a single user and we will anyway be having a seperate IAM tool in next release which will make this deprecated of maintaining a new seperate DB table? Also I feel pod restarts are also really rare.

From checking out online, they do mention reducing the access token ttl should be fine as it reduces the risk.

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: I will go with these fields for now and we can revisit on the namings later

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Yeah as I mentioned the app name will be the deployment name that user used while creating the application and it will not change even if deployment name changes.

If we to go with UUID, then the pod names will have UUID prefixed as we will not be redeploying the pod with each n every deployment name update that user does.

mayuka-c: Currently while deploying the pods we are using the app name as prefix to ensure it is unique for each application.

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: I did check online on what might be the best to do this, looks like people use `"x-advanced": true` in their `https://json-schema.org/draft-07/schema#` which shows these are hidden behind the advanced options section.

Maybe we can utilize this property to define fields which are considered as advanced

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: let us revisit this later when we have some time, but definitely we need to find all the different ways like label selectors and other techniques for the filtering if we aren't doing it..

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: nice.

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: Its security risk and we can't take a chance per guidelines and external IAM tool may take some more time

docs/proposals/catalog/application-deployment-api-proposal.md
mkumatag: what is user brings in same services with same name? how do we differentiate? 🤔

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Sure will create a new table for storing the blacklisted tokens

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Wouldnt that check needs to be done while you do import (where user brings in their own catalog of services)?. During that time, we need to ensure its unique.

Here it would be just to know the type which the running entity corresponds to. So I feel we need to control initially itself while we onboard new services from users.

mkumatag: hmm, let us leave it as is and learn more :)

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Have done the design for blacklisted tokens. PTAL at this commit for changes: https://github.com/IBM/project-ai-services/pull/635/changes/7a5dbc3ae3e3b2042a4ec0972c2bbc12df388bcc

Here's a brief overview
- Created a new table `tokens` which contains blacklisted access token and active refresh token.
- Looks like earlier we did not have a proper logic for renewing the access token using the refresh token. Because we were never validating the refresh token if its a right one or not. So a refresh token could be used forever (Ref: https://github.com/IBM/project-ai-services/blob/main/ai-services/internal/pkg/catalog/apiserver/services/auth/service.go#L82).
- Now we store the active refresh token in DB and we verify whenever user wants to refresh their access token using the refresh token.
- Also whenever user logs out, the blacklisted access token will be added in DB and remove all refresh tokens for the given user (admin).

mkumatag: > * Also whenever user logs out, the blacklisted access token will be added in DB and remove all refresh tokens for the given user (admin).

what do you mean by remove all refresh tokens? remove from where?

mkumatag: for sure there is gap in the current implementation, we need spend more time on this and fix as per the industry standards :)

mayuka-c: Remove the active refresh_tokens stored inside the DB. Maybe I can explain more tomo on the call

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: I had a discussion with @yussufsh and we have come up with approach and have updated. We can go through in our call tomo

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Had a discussion with @yussufsh and we have decided to go with UUID.
So the pod names and also labels would have UUID.

docs/proposals/catalog/application-deployment-api-proposal.md
mayuka-c: Made the updates to blacklist refresh token and all other changes as discussed today in the call

mkumatag:
Please convert all the text based diagrams to mermaid ones, it will be easy to view and update

mayuka-c:
> Please convert all the text based diagrams to mermaid ones, it will be easy to view and update



> Please convert all the text based diagrams to mermaid ones, it will be easy to view and update

Sure, will do that

yussufsh (dismissed):
Proposal lgtm; let's cover any changes during implementation.

mkumatag (approved):
LGTM

I'm merging based on our detailed discussion on 13th May call, any further updates will be addressed later.

PR #636, opened 2026-04-20
docs: Draft proposal for reusable services

yussufsh:
A proposal for a revamped architecture-service system. The documentation presents a forward-looking design for transforming an application into a modular system.

This PR only contains modularizing the assets and the deployment method for it.

Reusing an instance of a service (which is already installed as part of an existing service) is not in scope.

### DIFF
--- a/docs/proposals/architecture-service-system/PROPOSAL.md
+++ b/docs/proposals/architecture-service-system/PROPOSAL.md
@@ -1,0 +1,609 @@
+# Architecture-Service System Design Proposal
+
+---
+
+## Executive Summary
+
+This proposal outlines a transformation of the AI Services platform from a monolithic application structure to a flexible **architecture-service model**. The new system will enable modular, reusable services that can be deployed independently or as part of complete architectures, with automatic dependency resolution and full backward compatibility.
+
+### Vision
+
+Enable users to:
+- Deploy complete AI architectures (e.g., RAG) with a single command
+- Deploy individual services (e.g., chat, digitize) with automatic dependency handling
+- Reuse services across multiple architectures
+- Maintain backward compatibility with existing deployments
+
+### Key Benefits
+
+- 🎯 **Seamless Deployment**: Single command deploys architectures or services
+- 🔧 **Modular Design**: Services are independent, reusable components
+- 🔄 **Automatic Dependencies**: System resolves and deploys dependencies automatically
+- 🤖 **Service Discovery**: Convention-based naming eliminates manual configuration
+- ⚡ **Backward Compatible**: Existing commands continue to work
+- 🚀 **Future-Ready**: Foundation for advanced orchestration features
+
+---
+
+## Problem Statement
+
+### Current State: Monolithic Applications
+
+Today's RAG application is deployed as a monolithic unit where:
+
+**Limitations:**
+1. **Inflexibility**: Cannot deploy individual components (e.g., just chat or just digitize)
+2. **Code Duplication**: Same pods (opensearch, vllm-server) duplicated across rag, rag-cpu, rag-dev
+3. **Tight Coupling**: Changes to one component affect the entire application
+4. **Maintenance Burden**: Updates require modifying multiple application definitions
+5. **Limited Reusability**: Cannot share services across different architectures
+6. **Monolithic Model Server**: vllm-server bundles 3 models in one pod, preventing independent scaling
+
+### Example: Current RAG Deployment
+
+```
+assets/applications/rag/
+├── metadata.yaml (monolithic definition)
+└── podman/
+    ├── metadata.yaml
+    ├── values.yaml
+    └── templates/
+        ├── chat-bot.yaml.tmpl
+        ├── digitize.yaml.tmpl
+        ├── opensearch.yaml.tmpl
+        ├── vllm-server.yaml.tmpl
+        └── summarize-api.yaml.tmpl
+```
+
+**Problems:**
+- All components bundled together
+- opensearch, vllm-server duplicated in rag-cpu, rag-dev
+- Cannot deploy just chat without modifying templates
+- Service URLs hardcoded (e.g., `opensearch:9200`, `vllm-server:8000`)
+- vllm-server combines 3 models (instruct, embedding, reranker) in one pod
+- Cannot scale models independently (e.g., scale instruct without embedding)
+
+---
+
+## Proposed Solution: Architecture-Service Model
+
+### Core Concepts
+
+#### 1. Architecture
+A **logical grouping** of services that work together to provide a complete solution.
+
+**Characteristics:**
+- Defines required and optional services
+- No deployment templates (uses service templates)
+- Specifies service versions and constraints
+- Can be deployed as a unit
+
+**Example: RAG Architecture**
+```yaml
+id: rag
+name: "Digital Assistant"
+services:
+  - id: chat
+    version: ">=1.0.0"
+  - id: digitize
+    version: ">=1.0.0"
+  - id: summarize
+    version: ">=1.0.0"
+    optional: true
+```
+
+#### 2. Service
+An **independent, deployable component** with its own lifecycle.
+
+**Characteristics:**
+- Self-contained with dependencies declared
+- Reusable across multiple architectures
+- Runtime-specific configurations (Podman, OpenShift)
+- Own deployment templates and values
+
+**Example: Chat Service**
+```yaml
+id: chat
+name: "Question and Answer"
+dependencies:
+  - id: opensearch
+    version: ">=1.0.0"
+  - id: instruct
+    version: ">=1.0.0"
+  - id: embedding
+    version: ">=1.0.0"
+  - id: reranker
+    version: ">=1.0.0"
+```
+
+#### 3. Service Types
+
+**Deployable Services** (User-Facing):
+- Can be deployed individually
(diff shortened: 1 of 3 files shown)


docs/proposals/architecture-service-system/PROPOSAL.md
```diff
@@ -0,0 +1,627 @@
...
+- All services deployed: chat, digitization, summarization
+- All dependencies deployed: opensearch, instruct, embedding, reranker
+- Services can discover each other via naming convention
+- Models can be scaled independently
+
+### Scenario 2: Deploy Individual Service
+
+```bash
+# NEW capability!
+ai-services application create my-chat --template chat
```
mayuka-c, line 301: We may have to take deployment type as a flag too (either Architecture or Service)

docs/proposals/architecture-service-system/PROPOSAL.md
```diff
@@ -0,0 +1,627 @@
...
+```
+
+### Deployment Flow
+
+```
+1. Parse user command
+   ├─> Extract template name
+   └─> Check for --legacy flag
+
+2. Auto-detect deployment mode
```
mayuka-c, line 423: Do you mean auto detecting the deployment as in architecture/services. I feel we should not do it. We should take architectures/services as flag arguments. This aligns with the API and it will avoid unnecessary auto detection step.

docs/proposals/architecture-service-system/README.md
```diff
@@ -0,0 +1,140 @@
...
+| **Architecture** | Logical grouping of services | RAG = chat + digitization + summarization |
+| **Deployable Service** | Independent deployable component | chat, summarization |
+| **Dependency Service** | Service required by another service | chat depends on instruct, embedding, reranker, opensearch |
+| **Auto-Detection** | System determines deployment mode | Checks architectures/ → services/ → applications/ |
+
+### Deployment Examples
+
+```bash
+# Deploy full RAG architecture (all services)
+ai-services application create my-rag --template rag
```
mayuka-c, line 62: I think we might need to replace rag with Digital Assistant and have the folders created with it so that it behaves the same with both CLI and UI way

mayuka-c (commented):
I had a glance and added few comments
Will take a look at rest sometime later.

Overall design LGTM.

docs/proposals/architecture-service-system/PROPOSAL.md
```diff
@@ -0,0 +1,627 @@
...
+certified_by: "IBM"
+
+architectures:
+  - rag
+  - rag-cpu
+  - rag-dev
+
+dependencies:
+  - id: opensearch
+    version: ">=1.0.0"
```
mayuka-c: Just a query how this would work? If user passes their own opensearch endpoint with name?. Shouldnt then we avoid deploying opensearch/vllm models.

This is where I believe having a segregation between services and infra would be good. Infra like OpenSearch, VLLM models is something that can be reused across multiple services. Services are something which uses the infra to make themselves functional.

Infra is the reusable component.

docs/proposals/architecture-service-system/PROPOSAL.md
```diff
@@ -0,0 +1,627 @@
...
+
+```bash
+# Existing deployments continue to work
+ai-services application create legacy-app --template rag --legacy
+```
+
+**What happens:**
+1. `--legacy` flag forces old behavior
+2. Uses monolithic templates from `assets/applications/rag/`
+3. Deploys as before
```
mayuka-c, line 329: I feel we may end maintaining lot of code if we want to support both :)

I was wondering we give them the script to migrate their existing workloads to a newer model and let them start using the new CLI?.

mayuka-c (commented):
Also 1 more query I guess this is specifically for Podman runtime right?

For OpenShift since we use Helm Chart, we may have certain challenges, so might need to create a chart for all the services, then create a sub chart as dependency in parent chart and lot more.

docs/proposals/architecture-service-system/PROPOSAL.md
yussufsh: I tried to avoid this, as from UI it is easy to get the type, but from CLI it would mean an extra argument.

docs/proposals/architecture-service-system/PROPOSAL.md
yussufsh: We can deduce this in two steps; no looping is involved. We can keep type for now, but it is really not needed.
Here is how it would work...
Load:
`ai-services/assets/architectures/<template>/metadata.yaml`,
If it does not exist load:
`ai-services/assets/services/<template>/metadata.yaml`
else fail.

docs/proposals/architecture-service-system/README.md
yussufsh: rag would be the id and `Digital Assistant` would be the name in the `metadata.yaml`. Refer https://github.com/yussufsh/project-ai-services/blob/proposal/docs/proposals/architecture-service-system/PROPOSAL.md#architecture-metadata

docs/proposals/architecture-service-system/PROPOSAL.md
yussufsh: Right now, the proposal is to use the default service dependencies. Reusing is out of scope, we need some kind of inputs to use existing and skip deployment altogether in such cases.
Agree that something like an infra relationship will be required here! I have a few ideas for that.

docs/proposals/architecture-service-system/PROPOSAL.md
yussufsh: For the migration plan, we need to come up with an upgrade/rollback design.
But until this architecture-service mode is stable, I thought we should keep the existing flow where users who do not want to use the catalog way of deployment.

yussufsh: Atleast till we GA this feature. 😄

yussufsh:
> Also 1 more query I guess this is specifically for Podman runtime right?

To start with the design wanted to focus on Podman first. 

> For OpenShift since we use Helm Chart, we may have certain challenges, so might need to create a chart for all the services, then create a sub chart as dependency in parent chart and lot more.

Good point.
I feel we need to work with multiple charts similar to templates in Podman. Each service will have its own chart, which will also help in upgrading each service at a time.

docs/proposals/architecture-service-system/PROPOSAL.md
mayuka-c: Fine we can do the auto detection

docs/proposals/architecture-service-system/PROPOSAL.md
mayuka-c: Fine, we can rethink later. We may need DB design updates too for this as we need to support reuse and for this need to get the endpoint of already deployed service and many more

mayuka-c:
> > Also 1 more query I guess this is specifically for Podman runtime right?
> 
> To start with the design wanted to focus on Podman first.
> 
> > For OpenShift since we use Helm Chart, we may have certain challenges, so might need to create a chart for all the services, then create a sub chart as dependency in parent chart and lot more.
> 
> Good point. I feel we need to work with multiple charts similar to templates in Podman. Each service will have its own chart, which will also help in upgrading each service at a time.

Yes true, but the logic that we have today may not work where we consider app name as helm release which is also a namespace. So we need to even rework here and use something else identifier for creating namespaces because each chart will have its own release and all should get deployed in same namespace.

yussufsh:
I have removed `pod_templates/podTemplateExecutions` from runtime metadata.yaml, it will only have version. In future we can add more runtime specific details here eg: resource requirements, etc.

@mkumatag @mayuka-c ptal again. Thanks.

mkumatag:
@yussufsh @mayuka-c this is what I'm thinking for the those leave nodes, what do you say?

```yaml
# Service identification
id: instruct
name: "Instruct"
description: "Language model inference service for instruction-following tasks"

# Certification and support
certified_by: "IBM"

type: component
component_type: llm
provider: vllm


# For other components:
component_type: llm
provider: litellm

component_type: embeddings
provider: vllm

component_type: reranker
provider: vllm

component_type: vector_db
provider: opensearch
```

yussufsh:
```
# Service identification for chat
type: service
id: rag

# Service identification for instruct
type: instruct
id: vllm

# For litellm
type: instruct
id: litellm

type: embedding
id: vllm

type: reranker
id: vllm

type: vector_db
id: opensearch

type: vector_db
id: milvus
```
I was thinking of dividing these assets into services, instruct, embedding, vector_db, etc. as a type. Under each type you have id which defines what is the internal dependencies, values and templates to use. This could be based on user selection before deployment.

But I guess even defining it as end components and having a component_type & provider makes sense, would require additional logic while deployment or retrieving params.

With the current structure, I think we can easily plug in new services like litellm by creating a new directory callled `instruct-lite`.

mkumatag:
> But I guess even defining it as end components and having a component_type & provider makes sense, would require additional logic while deployment or retrieving params.

ofcourse!
> 
> With the current structure, I think we can easily plug in new services like litellm by creating a new directory callled `instruct-lite`.

My only comment is that - how can we make things more generic and add flexibility of choosing what we want runtime.? Please explore and let me know..

mayuka-c (approved):
LGTM

mkumatag (approved):

PR #627, opened 2026-04-15
Catalog Bootstrap command Implementation

mayuka-c:
- A new catalog bootstrap command
- Currently it does
    - Generates hash for admin password
    - Determines the podman socket uri (currently its defaulted to root). But this will change based on rootless implementation (To be done later)
    - Deploys the catalog service (The flow is similar to deploy any other applications like rag)
    - Refactored a bit of deploy flow so that can be used both in app create and catalog bootstrap.

Note: This command implementation is done for Podman here.

PR #637, opened 2026-04-20
Catalog Configure command Implementation - 1

mayuka-c:
- Asset updates.
- Introduce new Catalog configure sub command.
- Read the admin Password as Prompt.

This is actually broken down/ subset of this PR: https://github.com/IBM/project-ai-services/pull/627 as suggested by @yussufsh to have smaller PRs for better reviewing.

Will be raising subsequent changes in multiple smaller parts.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/catalog/configure.go
+++ b/ai-services/cmd/ai-services/cmd/catalog/configure.go
@@ -1,0 +1,151 @@
+package catalog
+
+import (
+	"fmt"
+	"syscall"
+
+	"github.com/spf13/cobra"
+	"golang.org/x/term"
+
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+	"github.com/project-ai-services/ai-services/internal/pkg/runtime"
+	"github.com/project-ai-services/ai-services/internal/pkg/runtime/types"
+	"github.com/project-ai-services/ai-services/internal/pkg/utils"
+	"github.com/project-ai-services/ai-services/internal/pkg/vars"
+)
+
+var (
+	// Runtime type flag for catalog configure command.
+	runtimeType string
+)
+
+// NewConfigureCmd creates a new configure command for the catalog service.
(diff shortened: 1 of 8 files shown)


ai-services/cmd/ai-services/cmd/catalog/configure.go
```diff
@@ -0,0 +1,151 @@
...
+		return "", err
+	}
+
+	password := string(passwordBytes)
+	if password == "" {
+		return "", fmt.Errorf("password cannot be empty")
+	}
+
+	// Prompt for confirmation
+	fmt.Print("Confirm admin password: ")
```
yussufsh, line 136: Do we really need confirm password? 🤔

yussufsh: I am not 100% sure if we should use cmd.Print or stick with fmt.Print

yussufsh (commented):
Some comments @mayuka-c but I am going to merge this PR. Please take it up in next PR round if needed.

yussufsh (approved):

ai-services/cmd/ai-services/cmd/catalog/configure.go
Shubhamag12: we should rather use logger pkg

mayuka-c: I did check, we do not have just Info in logger looks like. Also anyway we will not need verbosity for prompting, so went ahead with fmt

cmd.Print actually uses fmt.Print underneath. So didnt want to take cmd as argument for this method and use it.

mayuka-c: Reagrding reconfirmation, I feel this might be necessary if there is a typo while entering password, then there is no way for user to retrieve the password other than resetting (which is tedious), so added a confirmation prompt.

PR #639, opened 2026-04-21
Fix empty metadata returned from GET document request 

manju956:


spyre-rag/src/digitize/status.py
```diff
@@ -328,16 +328,25 @@ def _apply_metadata_updates(
...
     ) -> None:
-        """Apply field updates to data dictionary in-place."""
+        """
+        Apply field updates to data dictionary in-place.
+        Preserves existing metadata fields when updating.
+        """
         data.update(top_level_fields)
+
+        # Always ensure metadata wrapper exists
+        data.setdefault("metadata", {})
```
dharaneeshvrd, line 338: Why are we moving this above?

manju956: this is to ensure metadata wrapper exists irrespective of if we pass metadata fields as part of status updates. Creating this outside ensures we always have the dict set with default values and we dont see null

dharaneeshvrd (approved):

PR #640, opened 2026-04-21
Fix: Stop streaming logs when pod exits

Shubhamag12:
The application logs command should end automatically once the pod exits
Issue: https://jsw.ibm.com/browse/AISERVICES-933


Shubhamag12:
@yussufsh this is tested now with 2 go-routines, ptal

yussufsh (approved):

PR #641, opened 2026-04-21
feat: Add runtime param to catalog apiserver

yussufsh:
A runtime flag will be needed per runtime so the APIs can resolve proper manifest files.

### DIFF
--- a/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
+++ b/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
@@ -37,15 +37,13 @@ spec:
       image: "{{ .Values.backend.image }}"
       command: ["/bin/sh", "-c"]
       args:
-        - "/usr/bin/ai-services catalog apiserver --port=8080 --admin-username=admin --admin-password-hash=${ADMIN_PASSWORD}"
+        - "/usr/bin/ai-services catalog apiserver --port=8080 --admin-username=admin --admin-password-hash=${ADMIN_PASSWORD} --runtime={{ .Values.backend.runtime }}"
       securityContext:
         seLinuxOptions:
           type: "spc_t"
       env:
         - name: CONTAINER_HOST
           value: "unix://{{ .Values.backend.podman.uri }}"
-        - name: RUNTIME
-          value: "{{ .Values.backend.runtime }}"
         - name: ADMIN_PASSWORD
           valueFrom:
             secretKeyRef:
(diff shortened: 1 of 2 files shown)


ai-services/cmd/ai-services/cmd/catalog/apiserver.go
```diff
@@ -54,6 +71,7 @@ func NewAPIServerCmd() *cobra.Command {
 	apiserverCmd.Flags().DurationVarP(&defaultRefreshTokenTTL, "refresh-token-ttl", "", defaultRefreshTokenTTL, "Time-to-live for refresh tokens")
 	apiserverCmd.Flags().StringVar(&adminUserName, "admin-username", "admin", "Username for the default admin user")
 	apiserverCmd.Flags().StringVar(&adminPasswordHash, "admin-password-hash", "", "Precomputed hash of the password for the default admin user")
+	apiserverCmd.Flags().StringVar(&runtimeType, "runtime", string(types.RuntimeTypePodman), fmt.Sprintf("Runtime to use (options: %s, %s)", types.RuntimeTypePodman, types.RuntimeTypeOpenShift))
```
mayuka-c, line 74: Can the runtime type be env or should we want it to be flag?

yussufsh: Incase we decide not to hide apiserver sub command would prefer param so users will use it to be consistent.

mayuka-c: Sure, can you please update it here too with --runtime in the command args: https://github.com/IBM/project-ai-services/blob/main/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl#L40

mkumatag: you can also bind an env variable to a flag with `viper.BindPFlag`

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
```diff
@@ -37,15 +37,13 @@ spec:
       image: "{{ .Values.backend.image }}"
       command: ["/bin/sh", "-c"]
       args:
-        - "/usr/bin/ai-services catalog apiserver --port=8080 --admin-username=admin --admin-password-hash=${ADMIN_PASSWORD}"
+        - "/usr/bin/ai-services catalog apiserver --port=8080 --admin-username=admin --admin-password-hash=${ADMIN_PASSWORD} --runtime={{ .Values.backend.runtime }}"
```
mayuka-c, line 40: We cannot directly reference {{ .Values.backend.runtime }} in command args. We need to set the env and then use it from there

yussufsh: Could you please help with testing it out since you have the catalog code?

mayuka-c: My bad, its working fine. Just validated. So we can keep it.

mayuka-c (approved):
LGTM

mayuka-c:
We may need to bump the images moving forward I guess if we are making any changes in api-server code. Now I guess it can override the existing 0.0.2 image I dont see to be major change

PR #642, opened 2026-04-21
feat: (Catalog) Assets for simple rag application

yussufsh:
Added two new directories as per #636
This may change in the future, but we are defining the specs for implementing Architectures and Services APIs.

Note: OpenShift runtime specs are not tested yet, just divided based on Bob's suggestion. We need to determine how to deploy charts for each service there.


yussufsh:
```
assets/
├── architectures/              # NEW: Architecture definitions
│   └── rag/
│       └── metadata.yaml       # Lists services, no templates
│
├── services/                   # NEW: Reusable service definitions
│   ├── chat/                   # Deployable service
│   │   ├── metadata.yaml       # Service info + dependencies
│   │   ├── podman/
│   │   │   ├── metadata.yaml   # Runtime config
│   │   │   ├── values.yaml     # Default values
│   │   │   ├── steps/          # Post-deployment guidance
│   │   │   │   ├── info.md
│   │   │   │   ├── next.md
│   │   │   │   └── vars_file.yaml
│   │   │   └── templates/      # Pod templates
│   │   │       └── chat-bot.yaml.tmpl
│   │   └── openshift/
│   │       └── ...
│   │
│   ├── digitization/           # Deployable service
│   ├── summarization/          # Deployable service
│   ├── opensearch/             # Infrastructure service
│   ├── instruct/               # Infrastructure service (AI model)
│   ├── embedding/              # Infrastructure service (AI model)
│   └── reranker/               # Infrastructure service (AI model)
│
└── applications/               # LEGACY: Existing monolithic apps
    └── rag/                    # Kept for backward compatibility
````

**Comment by mkumatag** [comment]
> OpenShift runtime specs are not tested yet, just divided based on Bob's suggestion. We need to determine how to deploy charts for each service there.

how about podman runtime? I'm wondering if we can test with podman which is an low hanging thing see how that works with this new prototype?

**Comment by yussufsh** [comment]
> > OpenShift runtime specs are not tested yet, just divided based on Bob's suggestion. We need to determine how to deploy charts for each service there.
> 
> how about podman runtime? I'm wondering if we can test with podman which is an low hanging thing see how that works with this new prototype?

I have started working on the code and verified that the image, model and template commands are working fine.. Trying deployment but wanted to start putting the artifacts for review.
Refer: https://github.com/IBM/project-ai-services/compare/main...yussufsh:project-ai-services:reuse

**Comment by mkumatag** [comment]
> I have started working on the code and verified that the image, model and template commands are working fine.. Trying deployment but wanted to start putting the artifacts for review.

cool, let us try with a bare minimal services just for a PoC and see this goes..

**Comment by yussufsh** [comment]
> > I have started working on the code and verified that the image, model and template commands are working fine.. Trying deployment but wanted to start putting the artifacts for review.
> 
> cool, let us try with a bare minimal services just for a PoC and see this goes..

@mkumatag I have done deployments using architecture and individual services by reusing most of the existing code.
Here are the logs for [rag-dev](https://github.com/user-attachments/files/27113783/rag-dev.txt) and [summarize](https://github.com/user-attachments/files/27113789/summarize.txt) service.

Draft code can be referred from (https://github.com/yussufsh/project-ai-services/tree/apis_reuse)

Tested application create, delete(no change), template, image, model flow. The info sub-command is pending and should work with validating the info.md file and small code change.

**Comment by yussufsh** [comment]
I have updated the specs to remove `podTemplateExecutions` from runtime metadata.
@mkumatag @mayuka-c @Shubhamag12  ptal.

File: ai-services/assets/services/summarize/metadata.yaml [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+id: summarize
+name: "Summarize"
+description: "Consolidates input text into a brief statement of main points"
+type: service
+
+# Certification and support
+certified_by: "IBM"
+
+# Service belongs to these architectures
+architectures:
```
mkumatag, line 11: as we discussed, let remove this from the service which aren't required I feel

ai-services/assets/architectures/rag/metadata.yaml
```diff
@@ -0,0 +1,31 @@
+# Architecture identification
+id: rag
+name: "Digital Assistant"
+description: "Enable digital assistants using Retrieval-Augmented Generation (RAG), including AI services that query a managed knowledge base to answer questions from custom documents and data."
+version: "1.0.0"
+type: architecture
+
+# Certification and support
+certified_by: "IBM"
+supported_runtimes:
```
mkumatag: if we aren't planning anything like unsupported runtimes, then naming just runtime will be better..
```suggestion
runtimes:
```

ai-services/assets/architectures/rag/metadata.yaml
```diff
@@ -0,0 +1,31 @@
...
+    version: ">=1.0.0"
+  
+  - id: summarize
+    version: ">=1.0.0"
+    optional: true
+
+# Links
+demo_link: "https://example.com/demo/rag"
+code_link: "https://github.com/project-ai-services/spyre-rag"
+documentation_link: "https://docs.example.com/rag"
```
mkumatag:
```suggestion
links:
  demo: "https://example.com/demo/rag"
  code: "https://github.com/project-ai-services/spyre-rag"
  documentation: "https://docs.example.com/rag"
```

we need even check how we want fit in other information as well, we will revise this little later..

mkumatag: @mayuka-c may be we need to expose this information via some call api like ../rag/metadata for the user to use it wherever they want to use like in the displaying info..

mkumatag (dismissed):

mkumatag:
and also explore how can we use same service instruct with different variants like cpu, spyre etc... pushing as a param to the each service

ai-services/assets/services/summarize/metadata.yaml
yussufsh: I recall this was needed for the UI to display which architectures a particular service can belong to.

ai-services/assets/architectures/rag/metadata.yaml
yussufsh: This information will be returned in the GET API for an architecture... `/architecture/{rag}`. This is only needed when user clicks on `Learn More` button on the listing page.

yussufsh:
> and also explore how can we use same service instruct with different variants like cpu, spyre etc... pushing as a param to the each service

This will require little more work since we need to propagate the values from bottom to top service. ie. chat service should know the value of instruct cpu param.
Planning to implement it in the coming PRs.

ai-services/assets/architectures/rag/metadata.yaml
mkumatag: we are looking at these information, let us see if just a get call(`/architecture/{rag}`) is enough or we want to keep minimal info for the get call and retain the detailed call to a different api!

image

ai-services/assets/services/summarize/metadata.yaml
mkumatag: Hmm, can they do some reverse engineering and derive that information? Or make additional API calls or calculate that runtime and return at runtime? Adding this data into each static file would create more issues and potentially lead to staleness. Trust me.

mayuka-c:
> I have updated the specs to remove `podTemplateExecutions` from runtime metadata. @mkumatag @mayuka-c @Shubhamag12 ptal.

Just a query, now without podTemplateExecutions, within a service lets say digitization has multiple pods like digitize-service and postgres, then you end up deploying both concurrently right?

I hope that is fine

yussufsh:
> > I have updated the specs to remove `podTemplateExecutions` from runtime metadata. @mkumatag @mayuka-c @Shubhamag12 ptal.
> 
> Just a query, now without podTemplateExecutions, within a service lets say digitization has multiple pods like digitize-service and postgres, then you end up deploying both concurrently right?
> 
> I hope that is fine

We should have some reconciliation logic. eg: if digitization needs a DB connection to start, it should wait for it in a loop. If not, the whole service will run only when all the pods are up.
Otherwise, reintroducing layers makes sense @mkumatag ?

mkumatag:
> Just a query, now without podTemplateExecutions, within a service lets say digitization has multiple pods like digitize-service and postgres, then you end up deploying both concurrently right?
> 
> I hope that is fine

ideally we need to find a way to control them placing the pods or wait in pod for the DB or have some enough restart counts set etc. how we do it in the openshift like env, this podTemplateExecutions will just add an extra complexity and which may not be required is what my opinion.. will revisit this later if needed :) for now may be not!

mayuka-c (commented):
Just a query,
We do have seperate services like summarize and summarization?. Also same thing for digitize?
Any reason for this?

ai-services/assets/services/summarize/metadata.yaml
mkumatag: we might need to revisit how to expose this via here - https://github.com/IBM/project-ai-services/pull/635#discussion_r3145017953

ai-services/assets/services/instruct-cpu/podman/metadata.yaml
```diff
@@ -0,0 +1,2 @@
+name: instruct-cpu
+version: "1.0.0"
```
mayuka-c, line 2: I thought we wanted to have just the instruct and based on the value passed from the parent service, we either deploy cpu or spyre based on conditionals.

Are we taking this up seperately?

mkumatag:
IMO: let us merge the bare minimal things and increment things based on the learning otherwise we will just sit on these proposals, PRs forever :)

ai-services/assets/services/instruct-cpu/podman/metadata.yaml
yussufsh: Yes, I need to try that values concept first.

yussufsh:
> Just a query, We do have seperate services like summarize and summarization?. Also same thing for digitize? Any reason for this?

Ah good eyes... I see two dirs. The name was changed during implementation, not deleted afterward. Will fix this.

ai-services/assets/services/opensearch/metadata.yaml
```diff
@@ -0,0 +1,17 @@
+# Service identification
+id: opensearch
+name: "OpenSearch"
+description: "Vector database for document storage and retrieval"
+type: service
```
mayuka-c, line 5: Instead of type: service, can we make this Vector Store, this is really required in future, if we want to list down what all vector stores already provisioned or in future if we want to integrate any new vector store

yussufsh: IMO we should take this up when we implement vdb/inference concept. Right now, they are an infrastructure service.
We can think of moving this from `assets/services/` directory to`assets/vdb/` as well, later.

mayuka-c: This would be required in Details tracker page for each running application instance, we need to show Vector Store currently used for that application, Inference backend used and also models!

mayuka-c: So within the DB it will be easier to have the type stored to know exactly.
In the current design I do have the type stored inside services table where it will be Digitization, Summarization. Similarly we need a type for opensearch which I'm calling as Vector store and for instruct, reranker we can all it has inference backends?

mkumatag: > This would be required in Details tracker page for each running application instance, we need to show Vector Store currently used for that application, Inference backend used and also models!

let us not do everything for the first phase, let us learn and add them later once we have basic thing working fine.

mayuka-c (approved):
LGTM

mkumatag:
> We can think of moving this from `assets/services/` directory to`assets/vdb/` as well, later.
 
Please don't depend on the directory conventions, let us add any required details into the spec and use them

PR #643, opened 2026-04-21
Catalog Configure command Implementation - 2

mayuka-c:
Continuation from https://github.com/IBM/project-ai-services/pull/637

- Integrating deploy flow on Catalog configure command.
- Refactoring code to make use of deploy flow for both app create and catalog deploy.

This is final part of this PR: https://github.com/IBM/project-ai-services/pull/627

### DIFF
--- a/ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
+++ b/ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
@@ -1,0 +1,221 @@
+package podman
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+	"text/template"
+
+	"github.com/project-ai-services/ai-services/assets"
+	"github.com/project-ai-services/ai-services/internal/pkg/cli/helpers"
+	clipodman "github.com/project-ai-services/ai-services/internal/pkg/cli/podman"
+	"github.com/project-ai-services/ai-services/internal/pkg/cli/templates"
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+	"github.com/project-ai-services/ai-services/internal/pkg/runtime/podman"
+	"github.com/project-ai-services/ai-services/internal/pkg/specs"
+	"github.com/project-ai-services/ai-services/internal/pkg/spinner"
+)
+
+const (
+	catalogAppName     = "ai-services"
+	catalogAppTemplate = "catalog"
+)
+
+// DeployCatalog deploys the catalog service using the assets/catalog template for podman runtime.
+func DeployCatalog(ctx context.Context, podmanURI, passwordHash string, argParams map[string]string) error {
+	s := spinner.New("Deploying catalog service...")
+	s.Start(ctx)
+
+	// Initialize runtime
+	rt, err := podman.NewPodmanClient()
+	if err != nil {
+		s.Fail("failed to initialize podman client")
+
+		return fmt.Errorf("failed to initialize podman client: %w", err)
+	}
+
+	// Load template provider and metadata
+	tp, appMetadata, tmpls, err := loadCatalogTemplates(s)
+	if err != nil {
+		s.Fail("failed to load catalog templates")
+
+		return err
+	}
+
+	// Check if catalog pod already exists
+	existingPods, err := helpers.CheckExistingPodsForApplication(rt, catalogAppName)
+	if err != nil {
+		s.Fail("failed to check existing pods")
+
+		return err
+	}
+
+	if len(existingPods) == len(tmpls) {
+		s.Stop("Catalog service already deployed")
+		logger.Infof("Catalog pod already exists: %v\n", existingPods)
+
+		return nil
+	}
+
+	// Prepare values with configure-specific configuration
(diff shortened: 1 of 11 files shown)


ai-services/internal/pkg/cli/templates/embed.go
```diff
@@ -363,7 +375,12 @@ func NewEmbedTemplateProvider(options EmbedOptions) Template {
 	if options.Root != "" {
 		t.root = options.Root
 	} else {
-		t.root = "applications"
+		t.root = ""
```
yussufsh, line 378: I did not understand this change. How will root="" be helpful? What is the issue with the default being `applications`?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -0,0 +1,221 @@
...
+		return err
+	}
+
+	s.Stop("Catalog service deployed successfully")
+	logger.Infoln("-------")
+
+	// Print next steps similar to application create
+	if err := helpers.PrintNextSteps(tp, rt, catalogAppName, catalogAppTemplate); err != nil {
+		// do not want to fail the overall configure if we cannot print next steps
+		logger.Infof("failed to display next steps: %v\n", err)
```
Shubhamag12, line 81:
```suggestion
		logger.Infof("failed to display next steps: %w\n", err)
```

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -0,0 +1,221 @@
...
+
+		return fmt.Errorf("failed to initialize podman client: %w", err)
+	}
+
+	// Load template provider and metadata
+	tp, appMetadata, tmpls, err := loadCatalogTemplates(s)
+	if err != nil {
+		s.Fail("failed to load catalog templates")
+
+		return err
```
Shubhamag12: lets add context to this error

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	// Check if catalog pod already exists
+	existingPods, err := helpers.CheckExistingPodsForApplication(rt, catalogAppName)
+	if err != nil {
+		s.Fail("failed to check existing pods")
+
+		return err
```
Shubhamag12: same

Shubhamag12 (commented):
minor nits
rest lgtm

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
mayuka-c: It should be %v if we want to print the error. %w is used for wrapping the error (It will not print the error on terminal)

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	if len(existingPods) == len(tmpls) {
```
Shubhamag12, line 55: wondering if we should check the status of catalog pod as well?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
Shubhamag12: ah my bad, didnt see logger there

ai-services/internal/pkg/cli/templates/embed.go
yussufsh: Is the case that is root is empty we can refer any dir?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
mayuka-c: Its already done in line 48 within the method: `helpers.CheckExistingPodsForApplication`

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
mayuka-c: Do you mean while returning error?

Shubhamag12: yes, like we did just above this part of code

mayuka-c: Done

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
mayuka-c: Done

ai-services/internal/pkg/cli/templates/embed.go
mayuka-c: As discussed depending on the FS we need to set the Root. So for example if we use Bootstrap FS with root empty, then it will automatically set to applications as Root which is wrong

yussufsh (approved):
lgtm
Please ensure once existing `application create` command is working fine for every changing during refactoring.

mayuka-c:
> lgtm Please ensure once existing `application create` command is working fine for every changing during refactoring.

Validated both in Podman and OCP, working fine

PR #645, opened 2026-04-21
remove ENABLE_CATALOG_API tag

mkumatag:
we don't need this tag anymore, hence removing from the GitHub workflows


adarshagrawal38 (approved):
LGTM

PR #649, opened 2026-04-22
Fix: Move podman service SupplementaryGroups configuration after podman installation

Shubhamag12:
Two issues encountered:
1. podman service is already running as part of spyre configuration and we trying to start it again, as part of setupPodman()
2. If we run bootstrap on fresh setup with podman not installed, the spyre check will fail as it wont be able to find podman service

JIRA: https://jsw.ibm.com/browse/AISERVICES-963


ai-services/internal/pkg/bootstrap/spyreconfig/spyre/repair.go
```diff
@@ -421,25 +422,35 @@ func reloadAndRestartPodmanServices() error {
...
+	if err != nil || exitCode != 0 {
+		// not failing if service is not running
+		logger.Infof("Podman service stop returned non-zero (may not be running): exit code %d, stdout: %s, stderr: %s", exitCode, stdout, stderr, logger.VerbosityLevelDebug)
+	} else {
+		logger.Infoln("Podman service stopped successfully", logger.VerbosityLevelDebug)
 	}
 
-	// Restart podman socket
-	exitCode, _, _, err = utils.ExecuteCommand("systemctl", "restart", "podman.socket")
+	// Start podman socket
```
manju956: Dont we need to start the stopped podman service in previous step? will podman socket start bring up the podman service too?

Shubhamag12:
@manju956 @mkumatag PR is ready to review.
I have moved podman configuration, that we do as part of service report configuration, to after we setup podman.
Also while setting up podman, instead of starting, I am doing `restart` to avoid future conflicts as well.

ai-services/internal/pkg/bootstrap/spyreconfig/spyre/spyre.go
```diff
@@ -75,7 +75,6 @@ func RunChecks() []check.CheckResult {
 		checkUserGroup(),
 		checkVfioModule(),
 		checkVfioAccessPermission(),
-		checkPodmanServiceSupplementaryGroups(),
```
mkumatag, line 78: This is getting called when we run `bootstrap validate`, without this - it will not check for the `ServiceSupplementaryGroup` in the override

mkumatag (changes requested):

ai-services/internal/pkg/bootstrap/spyreconfig/spyre/spyre.go
Shubhamag12: so can we just have this check as part of validate and not fix it, meaning we fail and fix as part of configure?

ai-services/internal/pkg/bootstrap/spyreconfig/spyre/repair.go
```diff
@@ -365,10 +364,11 @@ func fixVFIOPermissions(checkMap map[string]check.CheckResult, userGroupResult R
 //
 // This ensures that all Podman operations, regardless of invocation method, have the necessary
 // permissions to access VFIO devices (/dev/vfio/*) required for Spyre card functionality.
-func fixPodmanServiceSupplementaryGroups(checkMap map[string]check.CheckResult) RepairResult {
+func FixPodmanServiceSupplementaryGroups(checkResult check.CheckResult) RepairResult {
```
manju956: Do we still need this in repair as we are already taking care of it in configure.go by calling `configurePodmanGroups` ?

Shubhamag12: we can move it from repair, since we are not doing as part of spyre now. 
Just thought it would be too many changes, but yea, will move it to related file

ai-services/internal/pkg/bootstrap/spyreconfig/spyre/spyre.go
mkumatag: na, we need to fail if something is wrong in validate, validate is a read only operation and should keep that separate.

Shubhamag12: then we need make this check conditional, only if podman is installed, since we install it after `spyre` step

ai-services/internal/pkg/bootstrap/spyreconfig/spyre/repair.go
Shubhamag12: done, ptal

ai-services/internal/pkg/bootstrap/spyreconfig/spyre/spyre.go
Shubhamag12: added it back to validation list, and made it a conditional check, ptal

manju956 (approved):
lgtm

mkumatag (approved):

PR #651, opened 2026-04-22
Assets for Catalog DB

mayuka-c:
- Include Assets for Catalog DB
- Random password generator for DB password


ai-services/assets/catalog/podman/values.yaml
```diff
@@ -9,3 +9,16 @@ backend:
   adminPasswordHash: ""
   podman:
     uri: "/run/podman/podman.sock"
+
+db:
```
yussufsh, line 13: Should we set the description or hidden annotations?

yussufsh: Or something you have kept to work in future?

ai-services/internal/pkg/utils/password_generator.go
```diff
@@ -0,0 +1,37 @@
...
+const (
+	// DefaultPasswordLength is the default length for generated passwords.
+	DefaultPasswordLength = 16
+	// passwordCharset contains all characters used for password generation.
+	passwordCharset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!@#$%^&*()-_=+[]{}|;:,.<>?"
+)
+
+// GenerateRandomPassword generates a cryptographically secure random password
+// of the specified length using crypto/rand.
+func GenerateRandomPassword(length int) (string, error) {
```
yussufsh, line 18: Nice.. I guess we can use this for opensearch as well.
Outside the scope of the catalog.

mkumatag: na, this shouldn't be done for the application deployment, I'm okay if you are just doing it for the catalog deployment only..

mkumatag: or find some generic way to do so if you want to do it for the application deployment,

ai-services/assets/catalog/podman/values.yaml
mayuka-c: Since the Catalog DB (Postgres) is for our internal use, do we want to expose?
I felt we need not have to allow to override so have avoided any annotation

mayuka-c: Also hidden would work `application templates` command which we dont have for Catalog!

ai-services/internal/pkg/utils/password_generator.go
mayuka-c: Yeah I have done it only for Catalog DB deployment as it is internal. For Application deployment, I feel safer would be get the password from user (for example OpenSearch) always.

ai-services/assets/catalog/podman/templates/catalog-db.yaml.tmpl
```diff
@@ -0,0 +1,48 @@
...
+    - name: POSTGRES_PASSWORD
+      valueFrom:
+        secretKeyRef:
+          name: catalog-secret
+          key: db-password
+    - name: POSTGRES_DB
+      value: "{{ .Values.db.database }}"
+    volumeMounts:
+    - name: postgres-data
+      mountPath: /var/lib/pgsql/18/data
```
Shubhamag12, line 28: shall we make this path configurable since its version specific?

mayuka-c: That is the internal postgres path where data is stored (not external path on LPAR).

mayuka-c: Also this is the DB path where data is stored on LPAR: https://github.com/IBM/project-ai-services/pull/651/changes#diff-bcb58cae41bdb6eb5647a2acbc127fee5b8819337bbe0363187a43a5c26b2ce1R47. This will not be version specific

yussufsh (approved):
lgtm

PR #656, opened 2026-04-22
Add similarity API entry to application info command

noahmasoud:


ryanmarc (changes requested):
There should be info changes for rag-cpu podman, rag podman/openshift, and rag-dev podman/openshift. It looks like the podman info for rag and rag-dev are both missing.

Also, there are no openshift files that would actually deploy the similarity search. That should be added.

ryanmarc (approved):

PR #658, opened 2026-04-23
Bumping up python-multipart python pkg to fix cve

adarshagrawal38:
There is cve in current python-multipart python pkg therefore upgrading it to latest version
https://github.com/IBM/project-ai-services/security/dependabot/109


images/rag-base/requirements.txt
```diff
@@ -87,7 +87,7 @@ pypdfium2==5.6.0
 python-dateutil==2.9.0.post0
 python-docx==1.2.0
 python-dotenv==1.2.2
-python-multipart==0.0.22
+python-multipart==0.0.26
```
yussufsh, line 90: Can we also update the transformers pkg?

yussufsh (approved):

PR #659, opened 2026-04-23
chore: refactor templates code

yussufsh:
1. Revamped NewEmbedTemplateProvider() removed EmbedOptions
2. Corrected proper way of setting path in OpenShift runtime bootstrap for LoadYamls method.
3. Added a method AppTemplateExist to check for template metadata.yaml instead of checking in the ListApplication.

Accordingly, NewEmbedTemplateProvider will take the fs and optional root (default will be as per fs). This will help in future when we have more than 1 directory under 1 FS eg: catalogFS can have services, architectures.


ai-services/internal/pkg/cli/templates/embed.go
```diff
@@ -94,14 +122,25 @@ func (e *embedTemplateProvider) ListApplications(hidden bool) ([]string, error)
 	return apps, nil
 }
 
+// AppTemplateExist Check if the application directory exists.
+func (e *embedTemplateProvider) AppTemplateExist(app string) error {
+	appPath := e.buildPath(app, "metadata.yaml")
+	_, err := fs.Stat(e.fs, appPath)
+	if err != nil {
+		return fmt.Errorf("application template '%s' does not exist", app)
```
Shubhamag12, line 130:
```suggestion
		return fmt.Errorf("application template '%s' does not exist: %w", app, err)
```

ai-services/assets/fs.go
```diff
@@ -2,7 +2,7 @@ package assets
 
 import "embed"
 
-//go:embed applications bootstrap
```
Shubhamag12, line 5: wondering why was bootstrap added here as part of application, ideally it should just be application for applicationFS

ai-services/internal/pkg/cli/templates/embed.go
yussufsh: This is shown on the cli o/p. Also kept it in-line with existing message: https://github.com/IBM/project-ai-services/blob/main/ai-services/internal/pkg/validators/template.go#L18

ai-services/assets/fs.go
yussufsh: I had the same question. I tested bootstrap on podman and openshift to ensure the functionality wasn't broken.

Shubhamag12 (approved):
lgtm

yussufsh:
Merging based on Go/Build job result.
Verified the flow on podman (bootstrap and application) and openshift (bootstrap) environment as well.

ai-services/assets/fs.go
yussufsh: @mayuka-c Just checking with you if this is needed at all?

PR #653, opened 2026-04-22
Adding backup and restore script

Abhishek-Kushwaha1:
# Unified Backup/Restore Script with Subcommands

## Summary
Created a single unified `backup-restore.sh` script with subcommands instead of multiple separate scripts. Follows modern CLI patterns (git, docker, kubectl) and eliminates code duplication.

### New File
- `backup-restore.sh` - Unified script with subcommands (520 lines)

### Features
- ✅ Single script with intuitive subcommands
- ✅ Supports any app name (rag-dev, rag-prod, custom apps)
- ✅ Color-coded output for better readability
- ✅ Built-in help and version commands
- ✅ No duplicate code - DRY principle
- ✅ Clean, maintainable architecture

## Usage

**Commands:**
```bash
# OpenSearch only
./backup-restore.sh export opensearch rag-dev opensearch.tar.gz

# Digitize data only
./backup-restore.sh export digitize digitize.tar.gz

# Full backup (run both)
./backup-restore.sh export opensearch rag-dev opensearch.tar.gz
./backup-restore.sh export digitize digitize.tar.gz

# OpenSearch only
./backup-restore.sh import opensearch opensearch.tar.gz

# Digitize data only
./backup-restore.sh import digitize digitize.tar.gz

# Full restore (run both)
./backup-restore.sh import opensearch opensearch.tar.gz
./backup-restore.sh import digitize digitize.tar.gz

#Help
./backup-restore.sh help

#version
./backup-restore.sh version

PR #662, opened 2026-04-23
Proposal for migration of json based metadata storage to DB storage

manju956:


docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+### Tables
+
+```sql
+-- Jobs table
+CREATE TABLE jobs (
+    job_id VARCHAR(255) PRIMARY KEY,
+    job_name VARCHAR(500),
+    operation VARCHAR(50) NOT NULL,
+    status VARCHAR(50) NOT NULL,
+    submitted_at TIMESTAMP NOT NULL,
```
manalilatkar: Is it necessary to have a submitted_at timestamp and a created_at timestamp? Are we expecting a lot of delay between the 2 making it important to track both?
Same for documents.

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+        if status:
+            query = query.filter(Job.status == status)
+        
+        query = query.order_by(Job.submitted_at.desc()).limit(limit).offset(offset)
+        
+        jobs = []
+        for job in query.all():
+            # Get documents for each job
+            documents = self.session.query(Document).filter(Document.job_id == job.job_id).all()
+            doc_summaries = [
```
manalilatkar: Here we are fetching all jobs, then for each job running a separate query to load its documents. With 100 jobs, that's 101 queries. Consider using SQLAlchemy's joinedload or selectinload eager loading, for eg:

`query = session.query(Job).options(selectinload(Job.documents))`

Reference: https://medium.com/@dresraceran/understanding-sqlalchemys-eager-loading-joinedload-selectinload-and-contains-eager-e12d98c8c8b0

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+        self.session.commit()
+    
+    def get_job(self, job_id: str) -> Optional[JobState]:
+        """Retrieve job state by ID using ORM."""
+        job = self.session.query(Job).filter(Job.job_id == job_id).first()
+        if not job:
+            return None
+        
+        # Get documents for this job
+        documents = self.session.query(Document).filter(Document.job_id == job_id).all()
```
manalilatkar: Eager loading can be done instead of making a separate query for related documents, eg:
`
query = session.query(Job).options(selectinload(Job.documents))`

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+    
+    def update_document_metadata(self, doc_id: str, updates: Dict[str, Any]) -> None:
+        """Update specific fields of document metadata using ORM."""
+        # Prepare update values
+        update_values = {}
+        for key, value in updates.items():
+            if key == 'metadata':
+                # Merge with existing metadata using PostgreSQL JSONB operator
+                stmt = sql_update(Document).where(Document.doc_id == doc_id).values(
+                    metadata=Document.metadata + value  # JSONB concatenation
```
manalilatkar: This is my hypothesis, needs to be tested for sure. But according to me, this JSONB concatenation will do a shallow merge. nested keys like timing_in_secs will be fully replaced, not deep-merged. For eg, if current metadata is:
```
{
    "pages": 10,
    "tables": 3,
    "timing_in_secs": {
        "digitizing": 12.5,
        "processing": null,
        "chunking": null,
        "indexing": null
    }
}
```

and update code is:
```
db_store.update_document_metadata("doc_123", {
    "metadata": {
        "timing_in_secs": {
            "processing": 8.3
        }
    }
})
```
Result of || (shallow merge):
```
{
    "pages": 10,
    "tables": 3,
    "timing_in_secs": {"processing": 8.3}    ← instead of presence of both processing and digitizing keys
}
```

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+    conn.close()
+    
+    print(f"Migration complete: {migrated} jobs migrated, {failed} failed")
+
+if __name__ == "__main__":
+    print("Starting migration...")
+    migrate_jobs()
+    migrate_documents()
+    print("Migration complete!")
+```
```
manalilatkar: The above code looks like it's a stray piece from an earlier trial version.

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+from datetime import datetime
+from sqlalchemy import update as sql_update, select
+from sqlalchemy.dialects.postgresql import insert
+from digitize.database import SessionScoped
+from digitize.models import Job, Document
+from digitize.document import DocumentMetadata
+from digitize.job import JobState, JobDocumentSummary
+from digitize.types import DocStatus, JobStatus
+
+class MetadataStore:
```
manalilatkar: Observation: The MetadataStore class commits after every single operation (save_document, update_document_metadata, etc.). This is problematic when we need to atomically update a document and its parent job in one transaction .
Suggestion: The store should accept an external session or support explicit transaction boundaries rather than auto-committing per method call. Consider a unit-of-work pattern or at minimum a context manager approach:
```
with db_store.transaction() as txn:
    txn.update_document(doc_id, updates)
    txn.update_job(job_id, job_updates)
# commits here, or rolls back on exception
```

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+        print(f"Job migration complete: {migrated} jobs migrated, {failed} failed")
+    finally:
+        session.close()
+
+if __name__ == "__main__":
+    print("Initializing database schema...")
+    init_db()
+    print("Starting migration...")
+    migrate_jobs()
+    migrate_documents()
```
manalilatkar, line 872: Just a suggestion, before ending migration, we should add a verification step that compares record counts and spot-checks key fields between the source JSON files and the destination tables.
 Also, there's no mention of what happens to the JSON files post-migration — should they be archived, kept as fallback, or deleted?

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+from sqlalchemy.orm import sessionmaker
+from digitize.db_store import MetadataStore, Base, Job, Document
+from digitize.document import DocumentMetadata
+from digitize.job import JobState, JobStats, JobDocumentSummary
+from digitize.types import DocStatus, JobStatus, OutputFormat
+
+@pytest.fixture
+def test_engine():
+    """Create in-memory SQLite database for testing."""
+    engine = create_engine('sqlite:///:memory:', echo=True)
```
manalilatkar: unit tests use SQLite in-memory, but the code uses PostgreSQL-specific features (JSONB, ON CONFLICT DO UPDATE, JSONB || operator). How does this work?

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+- ✅ **Relationship Handling**: Automatic JOIN operations via relationships
+- ✅ **Transaction Management**: Automatic commit/rollback on errors
+
+### Dependencies
+
+**File: `spyre-rag/requirements.txt`** (add these lines)
+```
+sqlalchemy>=2.0.0
+psycopg2-binary>=2.9.0  # PostgreSQL driver
+alembic>=1.12.0  # For database migrations (optional but recommended)
```
manalilatkar: Alembic is listed as a dependency but not being used anywhere in the code.

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+- Configure automated backups
+
+**Configuration:**
+```python
+# Database connection settings
+DB_HOST = os.getenv("POSTGRES_HOST", "localhost")
+DB_PORT = int(os.getenv("POSTGRES_PORT", "5432"))
+DB_NAME = os.getenv("POSTGRES_DB", "digitize_metadata")
+DB_USER = os.getenv("POSTGRES_USER", "digitize_user")
+DB_PASSWORD = os.getenv("POSTGRES_PASSWORD")
```
manalilatkar, line 453: Any details on how secret management will be done for DB password or will we leave that to orchestration squad?

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+#### 4.1 Deployment Checklist
+
+- [ ] PostgreSQL instance provisioned and configured
+- [ ] Database schema created
+- [ ] Indexes created
+- [ ] Existing JSON data migrated
+- [ ] Application code updated
+- [ ] Environment variables configured
+- [ ] Connection pooling configured
+- [ ] Monitoring set up
```
manalilatkar, line 1243: We should also add a section on what we plan to monitor, for eg connection pool utilization, query latency percentiles (p50/p95/p99), table bloat etc.

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+- [ ] Backup strategy implemented
+- [ ] Rollback plan documented
+
+#### 4.2 Rollback Plan
+
+If issues arise:
+1. Switch back to JSON file storage (keep old code path)
```
manalilatkar: Does this mean that during the initial stage, we will be maintaing data in JSON files as well as Postgresql db?

docs/proposals/digitize_metadata_db_migration.md
manju956: hm, will spend time investigating the behavior

docs/proposals/digitize_metadata_db_migration.md
manju956: hm, its fine to have `submitted_at` to capture job submission time. dont think there might be lot of delay involved in job sumission and storing the entry in db.
I will address it.

docs/proposals/digitize_metadata_db_migration.md
manju956: I guess we can follow the principle of managing OPENSEARCH credentials here too.

docs/proposals/digitize_metadata_db_migration.md
dharaneeshvrd: I dont think this is required. Rather we need to come up with a script for migrating the existing json files to DB. Let's cover that as well as part of this proposal.

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+
+---
+
+## Database Schema Design
+
+### Tables
+
+```sql
+-- Jobs table
+CREATE TABLE jobs (
```
dharaneeshvrd, line 295: But when do these SQL statements will get executed?

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+# 2-3 concurrent jobs
+DB_POOL_SIZE = 10
+
+# 5+ concurrent jobs (rare)
+DB_POOL_SIZE = 20
+```
+
+#### 1.2 Schema Creation
+
+**Script: `scripts/db/init_schema.sql`**
```
dharaneeshvrd: I think all the scripts you have mentioned should be under digitize service's code base since it's specific to digitize service.

dharaneeshvrd: Regd execution, can we have an init container that executes this script?

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1536 @@
...
+-- Create triggers (use triggers from above)
+
+-- Grant permissions
+GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO digitize_user;
+GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA public TO digitize_user;
+```
+
+#### 1.3 Data Migration Script (Using SQLAlchemy ORM)
+
+**Script: `scripts/db/migrate_json_to_postgres.py`**
```
dharaneeshvrd: I see the migration script now. 
We need to document on how and where its need to get executed.
https://github.com/IBM/project-ai-services/pull/653/ ref can be taken from this script, no need to work on the full implementation, just a high level info is sufficient for now.

docs/proposals/digitize_metadata_db_migration.md
dharaneeshvrd: Post migration the files should be left as is IMO.

docs/proposals/digitize_metadata_db_migration.md
dharaneeshvrd, line 295: Need to use `IF NOT EXISTS` for all the create queries to take care of the re-run scenario.

manju956: right, will take care of this

docs/proposals/digitize_metadata_db_migration.md
manju956: Yes, init container is a good place to initialize and create schema.

docs/proposals/digitize_metadata_db_migration.md
manju956: sure, will take care of this

docs/proposals/digitize_metadata_db_migration.md
manju956: Any specific reason to retain the json files?

docs/proposals/digitize_metadata_db_migration.md
manju956: right, will cleanup

docs/proposals/digitize_metadata_db_migration.md
manju956: ideally migration script should take care of db migration. in the event of script failure, we could retain json files for investigation and then re-attempt.

docs/proposals/digitize_metadata_db_migration.md
manju956: makes sense. will incorporate eager loading using SQLAlchemy's `selectinload` method

docs/proposals/digitize_metadata_db_migration.md
manju956: right, since we are planning to use init-container to initialize db schema this dependency is no longer needed. I will remove this from dependencies list

docs/proposals/digitize_metadata_db_migration.md
manju956: lookslike some mishap with testing changes. Will add Postgres based testing changes here

docs/proposals/digitize_metadata_db_migration.md
manju956: addresed

docs/proposals/digitize_metadata_db_migration.md
manju956: addressed

docs/proposals/digitize_metadata_db_migration.md
manju956: added jsonb deep_merge to take care of this

docs/proposals/digitize_metadata_db_migration.md
manju956: removed

docs/proposals/digitize_metadata_db_migration.md
manju956: Added init container to take care of DB schema creation

manju956: done

docs/proposals/digitize_metadata_db_migration.md
dharaneeshvrd: Maybe we can have a knob for controlling this, if user wants to cleanup post migration, we can do, else leave the cleaning responsibility to the user.

dharaneeshvrd (commented):
@manju956 
Kind request, can you please remove the code samples from this ?
It is very difficult to go through the flow in one stretch. :(
Let's review the code part during implementation PRs

manju956:
> @manju956 Kind request, can you please remove the code samples from this ? It is very difficult to go through the flow in one stretch. :( Let's review the code part during implementation PRs

sure @dharaneeshvrd let me keep the code changes out of this proposal and update.

manju956:
> > @manju956 Kind request, can you please remove the code samples from this ? It is very difficult to go through the flow in one stretch. :( Let's review the code part during implementation PRs
> 
> sure @dharaneeshvrd let me keep the code changes out of this proposal and update.

done

docs/proposals/digitize_metadata_db_migration.md
dharaneeshvrd: but we are not going to keep the existing flow right? 
I think this block says that.

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1565 @@
...
+```bash
+#!/bin/bash
+# Database initialization script for digitize service
+# This script is executed by the init container
+
+set -e  # Exit on error
+
+echo "Starting database initialization..."
+
+# Wait for PostgreSQL to be ready (connect to default 'postgres' database)
```
dharaneeshvrd, line 629: Can we create separate db for digitize service?

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1565 @@
...
+done
+
+echo "Database '$POSTGRES_DB' is accessible!"
+
+# Run schema initialization on target database
+echo "Initializing database schema..."
+PGPASSWORD=$POSTGRES_PASSWORD psql -h "$POSTGRES_HOST" -U "$POSTGRES_USER" -d "$POSTGRES_DB" \
+  -f /scripts/init_schema.sql
+
+echo "✅ Database initialization completed successfully!"
```
dharaneeshvrd, line 660: Dont we need to ensure connection is closed here?

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1565 @@
...
+Add init container section:
+
+```yaml
+spec:
+  template:
+    spec:
+      # Init container to initialize database schema
+      initContainers:
+        - name: digitize-db-init
+          image: "icr.io/ai-services-private/postgres:18"  # ppc64le compatible UBI based Postgres image
```
dharaneeshvrd, line 675: I dont think we would be building a Postgres image part of ai-services code base.
Need to revisit this part on how to copy the script to this container.

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1565 @@
...
+```
+
+#### 2.4 Metadata Store Implementation
+
+**Metadata Store** (`db_store.py`):
+
+    **Core Operations:**
+    - `save_document()` - Upsert document using INSERT ... ON CONFLICT DO UPDATE
+    - `get_document()` - Retrieve single document by ID
+    - `update_document_metadata()` - Update document fields with deep merge for JSONB
```
dharaneeshvrd: dont we need similar method for updating job stats?

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -0,0 +1,1565 @@
...
+#### 4.2 Rollback Plan
+
+If issues arise:
+1. Switch back to JSON file storage (keep old code path)
+2. Export data from PostgreSQL back to JSON files
+3. Investigate and fix issues
+4. Re-attempt migration
+
+
+## Monitoring and Observability
```
dharaneeshvrd: I see lot of content under monitoring and observability
But I dont think we would be needing any of this, since the footprint its going to make in Postgres is not going to be large enough to analyse these metrics in our use case.
We can keep them if there is a similar need arises in future, we can take reference from here.

docs/proposals/digitize_metadata_db_migration.md
manju956: hm, its good to close connection after init. Will add this.

docs/proposals/digitize_metadata_db_migration.md
manju956: We do have UBI based postgres image built as part of https://github.com/IBM/project-ai-services/blob/main/images/postgres/Dockerfile I referred the same image here.

docs/proposals/digitize_metadata_db_migration.md
manju956: true, mornitoring and observaibility are too heavy for the the db usecase that we have. We can start without this and include them goingforward if there is a need.

docs/proposals/digitize_metadata_db_migration.md
manju956: actually psql cli tool which runs the sql file automatically closes db connection when it exits, so its taken care.

docs/proposals/digitize_metadata_db_migration.md
manju956: Yes, its needed somehow missed. Will include this.

docs/proposals/digitize_metadata_db_migration.md
manju956: Yes, we will need different databases for each service using postgres.

docs/proposals/digitize_metadata_db_migration.md
manju956: addressed

docs/proposals/digitize_metadata_db_migration.md
manju956: removed monitoring section as the footprint the service leaves on db is not that significant at the current stage and it would be an overkill. lets revisit this later when we have such requirements.

docs/proposals/digitize_metadata_db_migration.md
manju956: added note in proposal regarding using separate db for each services.

docs/proposals/digitize_metadata_db_migration.md
manju956: added

docs/proposals/digitize_metadata_db_migration.md
manju956: removed monitoring and troubleshooting section. we will revisit later when the need comes up.

docs/proposals/digitize_metadata_db_migration.md
dharaneeshvrd: Great, I did nt know about this.

docs/proposals/digitize_metadata_db_migration.md
manju956: Yes, current way of creating initial doc/job metadata, updating as we progress the ingestion phases will no longer be present. ingestion pipeline will update the states field of documents table via ORM layer.

dharaneeshvrd: Then can you please remove that block?

manju956: removed

docs/proposals/digitize_metadata_db_migration.md
```diff
@@ -858,14 +858,112 @@ def migrate_jobs():
...
+            print(f"Deleted: {json_file}")
+
+        print("Cleanup complete!")
+    else:
+        print("\nJSON files retained. Use --cleanup flag to remove them after migration.")
+```
+
+**How to Execute the Migration Script:**
+
+**Option 1: Inside the Digitize Container (Recommended)**
```
dharaneeshvrd: Option 1 seems hacky way
Option 2 targets only k8s
Option 3 seems a viable approach, IMO we can go ahead with that and need to run it on demand via a flag.

manju956: sure, will keep option3 with a flag option

manju956: addressed, ptal

dharaneeshvrd (approved):

PR #665, opened 2026-04-23
chore: Refactor image code for better usage

yussufsh:
- Reduced complexity with simplified API design
- Changed from storing policy in struct to passing it to Run(policy) method
- Better flexibility: Runtime and app name can be nil when only listing images
- Common image fetching logic extracted, reducing duplicate code


mayuka-c (dismissed):
LGTM

ai-services/internal/pkg/image/images.go
```diff
@@ -0,0 +1,131 @@
...
+
+// Images manages container images for applications, including listing and pulling based on policies.
+type Images struct {
+	Runtime     runtime.Runtime
+	App         string
+	AppTemplate string
+}
+
+// ListImages returns the list of images required for the application template.
+func (img *Images) ListImages() ([]string, error) {
```
mkumatag, line 37: this function is more of a application centric, it will be good if we move to the place we are dealing with the application

mkumatag (commented):
this code needs to be revisited later.. let us merge as is for now and reorg the code later when we get some bandwidth.

ai-services/internal/pkg/image/images.go
yussufsh: Agree with your point, it looks complicated.
I had moved it here to keep the image functions together. For backward compatibility when we add the new catalog services implementation, I would like `Images` to hold the template provider as well.

mayuka-c (approved):
LGTM

Shubhamag12 (approved):

PR #666, opened 2026-04-24
Enable HTTPS for exposed endpoints Proposal

mayuka-c:
- Proposal for enabling HTTPS
- Comparing diff ways to configure SSL
- Complete Implementation steps on how to use Caddy
- Logic to do self signed certs or read the user defined certs

### DIFF
--- a/docs/proposals/https_enable.md
+++ b/docs/proposals/https_enable.md
@@ -217,11 +217,14 @@ As part of the `ai-services catalog configure` command, deploy the Caddy server 
 **Command Options:**
 ```bash
 ai-services catalog configure [options]
-  --domain-name <domain>  Custom domain name for self-signed certificates (optional)
-                          If not provided, uses wildcard DNS format: <service>.<ip>.nip.io
-                          Example: --domain-name example.com generates certs for *.example.com
-  --ssl-cert <path>       Path to user-provided SSL certificate (optional)
-  --ssl-key <path>        Path to user-provided SSL private key (optional)
+  --domain-name <domain>     Custom domain name for self-signed certificates (optional)
+                             If not provided, uses wildcard DNS format: <service>.<ip>.nip.io
+                             Example: --domain-name example.com generates certs for *.example.com
+  --external-port <port>     External HTTPS port for Caddy server (optional, default: 443)
+                             If not 443, port will be included in service URLs
+                             Example: --external-port 8443 results in https://service.<domain>:8443
+  --ssl-cert <path>          Path to user-provided SSL certificate (optional)
+  --ssl-key <path>           Path to user-provided SSL private key (optional)
 ```
 
 **Process:**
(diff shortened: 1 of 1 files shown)


shwetha-s-poojary:
Just one concern here - caddy becomes a shared point for TLS, certs, and routing, which means it’s a single point of failure. If it goes down or is misconfigured, we could lose access to everything even if the services are fine.
Should we think about a per-service sidecar proxy instead? It adds some overhead, but failures would be limited to one service.
Caddy is simpler, but the tradeoff here is simplicity vs better isolation and reliability.

cc: @mayuka-c @mkumatag @yussufsh

mayuka-c:
> Just one concern here - caddy becomes a shared point for TLS, certs, and routing, which means it’s a single point of failure. If it goes down or is misconfigured, we could lose access to everything even if the services are fine. Should we think about a per-service sidecar proxy instead? It adds some overhead, but failures would be limited to one service. Caddy is simpler, but the tradeoff here is simplicity vs better isolation and reliability.
> 
> cc: @mayuka-c @mkumatag @yussufsh

Yeah SPOF is there since we run it as a single pod. But if misconfiguration is done it can be done to all other services that we are running as everything is running as single pod  :)

Even if it goes down we do have the restartPolicy set to always restart the pod. In the worst case since the underlying data is present in the LPAR, it can be redeployed too.

Also, Please feel free to come up with the other approaches that you have in mind and we can evaluate the pros and cons. Also do consider all the scenarios so that everything is covered and how easy will it be to integrate as customers will bring in their own catalog and we need to make it seemeless for them to integrate with AI-Services.

Also what exactly is this side car proxy container and what does it do?

mkumatag:
> which means it’s a single point of failure.

We need to check if they support multi-instance caddy, same problem do exist with NGINX or any global proxy like solution, customers using Podman should consider migrating to OpenShift if they require a more reliable and scalable architecture.

mkumatag:
> Even if it goes down we do have the restartPolicy set to always restart the pod. In the worst case since the underlying data is present in the LPAR, it can be redeployed too.

instead of restart we need consider exploring how to recover only the proxy instead of an entire product reinstallation.

mayuka-c:
> > Even if it goes down we do have the restartPolicy set to always restart the pod. In the worst case since the underlying data is present in the LPAR, it can be redeployed too.
> 
> instead of restart we need consider exploring how to recover only the proxy instead of an entire product reinstallation.

Lets say if a pod is deleted, we could have them run the `ai-services catalog configure` command which will take care of bringing the Caddy pod up or else we can have subcommands for proxy like `status` which will monitor and show the status of proxy server, `sync` which ensures that if resource is not running, it will bring up.

Regarding the data side, we currently store the caddy data under {{ .basedir }}/common/caddy. So we would need to take backup of this to ensure that data is not lost.

yussufsh (approved):
lgtm
@shwetha-s-poojary ptal

shwetha-s-poojary:
lgtm

PR #667, opened 2026-04-24
Secure vLLM runtime (Implementation)

sats-23:
-When Chatbot is loaded, user is prompted with this menu.
-Upon correct API key, user can proceed to use chatbot as usual 
-Key remains valid until page refresh 
-Wrong API key redirects to same prompt box expecting API key 
image

-Similarly updated Swagger docs as well
image

-Curl requests with and without auth header when auth is enabled
image

### DIFF
--- a/ai-services/assets/applications/rag-cpu/podman/templates/similarity-api.yaml.tmpl
+++ b/ai-services/assets/applications/rag-cpu/podman/templates/similarity-api.yaml.tmpl
@@ -50,6 +50,10 @@ spec:
           value: "{{ .Values.opensearch.auth.username }}"
         - name: OPENSEARCH_PASSWORD
           value: "{{ .Values.opensearch.auth.password }}"
+        {{- if .Values.instruct.apiKey }}
+        - name: VLLM_INSTRUCT_API_KEY
+          value: "{{ .Values.instruct.apiKey }}"
+        {{- end }}
         - name: LOG_LEVEL
           value: "{{ .Values.similarity.log_level}}"
       resources:
--- a/ai-services/assets/applications/rag-cpu/podman/templates/similarity-api.yaml.tmpl
+++ b/ai-services/assets/applications/rag-cpu/podman/templates/similarity-api.yaml.tmpl
@@ -50,10 +50,6 @@ spec:
           value: "{{ .Values.opensearch.auth.username }}"
         - name: OPENSEARCH_PASSWORD
           value: "{{ .Values.opensearch.auth.password }}"
-        {{- if .Values.instruct.apiKey }}
-        - name: VLLM_INSTRUCT_API_KEY
-          value: "{{ .Values.instruct.apiKey }}"
-        {{- end }}
         - name: LOG_LEVEL
           value: "{{ .Values.similarity.log_level}}"
       resources:
--- a/ai-services/assets/applications/rag-dev/openshift/templates/backend-deployment.yaml
+++ b/ai-services/assets/applications/rag-dev/openshift/templates/backend-deployment.yaml
@@ -72,6 +72,13 @@ spec:
                   key: password
             - name: OPENSEARCH_NUM_SHARDS
               value: "{{ .Values.opensearch.numShards }}"
+            {{- if .Values.instruct.apiKey }}
+            - name: VLLM_INSTRUCT_API_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: vllm-instruct-api-key
+                  key: apiKey
+            {{- end }}
             - name: LOG_LEVEL
               value: "{{ .Values.backend.log_level }}"
           ports:
--- a/ai-services/assets/applications/rag-dev/openshift/templates/backend-deployment.yaml
+++ b/ai-services/assets/applications/rag-dev/openshift/templates/backend-deployment.yaml
@@ -72,13 +72,6 @@ spec:
                   key: password
             - name: OPENSEARCH_NUM_SHARDS
               value: "{{ .Values.opensearch.numShards }}"
-            {{- if .Values.instruct.apiKey }}
-            - name: VLLM_INSTRUCT_API_KEY
-              valueFrom:
-                secretKeyRef:
-                  name: vllm-instruct-api-key
-                  key: apiKey
-            {{- end }}
             - name: LOG_LEVEL
               value: "{{ .Values.backend.log_level }}"
           ports:
--- a/ai-services/assets/applications/rag-dev/podman/values.yaml
+++ b/ai-services/assets/applications/rag-dev/podman/values.yaml
@@ -2,7 +2,7 @@ ui:
   # @description Host port for the RAG UI. If unspecified, a random available port is assigned. Specify a port number to use a custom value.
   port: ""
   # @hidden
-  image: icr.io/ai-services-cicd/rag-ui:v0.0.32
+  image: icr.io/ai-services-cicd/rag-ui:v0.0.33
 
 backend:
   # @description Host port for the OpenAI-compatible RAG service. Defaults to unexposed; assign a port to enable external access.
--- a/ai-services/assets/applications/rag-dev/podman/values.yaml
+++ b/ai-services/assets/applications/rag-dev/podman/values.yaml
@@ -8,7 +8,7 @@ backend:
   # @description Host port for the OpenAI-compatible RAG service. Defaults to unexposed; assign a port to enable external access.
   port: "0"
   # @hidden
-  image: icr.io/ai-services-cicd/rag:v0.0.98
+  image: icr.io/ai-services-cicd/rag:v0.0.99
   # @hidden
   log_level: "INFO"
 
@@ -16,7 +16,7 @@ digitize:
   # @description Host port for the DIGITIZE API. If unspecified, a random available port is assigned. Specify a port number to use a custom value.
   port: ""
   # @hidden
-  image: icr.io/ai-services-cicd/rag:v0.0.98
+  image: icr.io/ai-services-cicd/rag:v0.0.99
   # @hidden
   log_level: "INFO"
 
@@ -58,7 +58,7 @@ summarize:
   # @description Host port for the Summarize API. If unspecified, a random available port is assigned. Specify a port number to use a custom value.
   port: ""
   # @hidden
-  image: icr.io/ai-services-cicd/rag:v0.0.98
+  image: icr.io/ai-services-cicd/rag:v0.0.99
   # @hidden
   log_level: "INFO"
 
@@ -66,6 +66,6 @@ similarity:
   # @description Host port for the Similarity Search API. If unspecified, a random available port is assigned. Specify a port number to use a custom value.
   port: ""
   # @hidden
-  image: icr.io/ai-services-cicd/rag:v0.0.98
+  image: icr.io/ai-services-cicd/rag:v0.0.99
   # @hidden
   log_level: "INFO"
--- a/ai-services/assets/applications/rag-dev/podman/values.yaml
+++ b/ai-services/assets/applications/rag-dev/podman/values.yaml
@@ -41,6 +41,8 @@ opensearch:
     password: "AiServices@12345"
 
 instruct:
+  # @description API key for vLLM instruct service authentication. If empty, authentication is disabled. Provide a key to enable authentication.
+  apiKey: ""
   # @hidden
   image: registry.redhat.io/rhaiis/vllm-spyre-rhel9:3.3.0
 
--- a/spyre-rag/src/chatbot/app.py
+++ b/spyre-rag/src/chatbot/app.py
@@ -36,6 +36,8 @@ from chatbot.response_utils import (
     HealthResponse,
     ModelsResponse,
     PerfMetricsResponse,
+    ReferenceRequest,
+    ReferenceResponse,
 )
 
 vectorstore = None
@@ -45,6 +47,10 @@ vectorstore_lock = asyncio.Lock()
 emb_model_dict = {}
 llm_model_dict = {}
 reranker_model_dict = {}
+
+# Cache for auth requirement check
+auth_required_cache = {"checked": False, "required": False}
+auth_cache_lock = asyncio.Lock()
 
 concurrency_limiter = BoundedSemaphore(settings.chatbot.max_concurrent_requests)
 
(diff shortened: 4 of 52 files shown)


ai-services/internal/pkg/application/podman/create.go
```diff
@@ -310,6 +307,15 @@ func (p *PodmanApplication) executePodTemplates(tp templates.Template,
 		return fmt.Errorf("failed to load params for application: %w", err)
 	}
 
+	// Log vLLM authentication status
+	if instructAPIKey, ok := values["instruct"].(map[string]any); ok {
```
mkumatag: I think somehow we need to push this info to applications/workload, any application specific content in the orchestration is a big NO!

ai-services/internal/pkg/application/openshift/create.go
```diff
@@ -39,6 +39,15 @@ func (o *OpenshiftApplication) Create(ctx context.Context, opts types.CreateOpti
 		return fmt.Errorf("failed to prepare values: %w", err)
 	}
 
+	// Log vLLM authentication status
+	if instructAPIKey, ok := values["instruct"].(map[string]any); ok {
```
mkumatag: I think somehow we need to push this info to applications/workload, any application specific content in the orchestration is a big NO!

mkumatag (changes requested):

ai-services/internal/pkg/application/openshift/create.go
mkumatag: @yussufsh @mayuka-c ^^

ai-services/assets/applications/rag-dev/podman/values.yaml
```diff
@@ -41,6 +41,8 @@ opensearch:
     password: "AiServices@12345"
 
 instruct:
+  # @description API key for vLLM instruct service authentication. If empty, authentication is disabled. Provide a key to enable authentication.
+  apiKey: ""
```
mayuka-c, line 45: What is the format the key should be in?

ai-services/assets/applications/rag-dev/openshift/templates/backend-deployment.yaml
```diff
@@ -49,7 +49,7 @@ spec:
             - name: LLM_MODEL
               value: ibm-granite/granite-3.3-8b-instruct
             - name: RERANKER_ENDPOINT
-              value: http://reranker-predictor:8080
+              value: http://reranker-predictor:8000
```
mayuka-c: Why bringing this change? without spyre it runs on 8080

ai-services/assets/applications/rag-dev/openshift/templates/backend-deployment.yaml
```diff
@@ -72,6 +72,13 @@ spec:
                   key: password
             - name: OPENSEARCH_NUM_SHARDS
               value: "{{ .Values.opensearch.numShards }}"
+            {{- if .Values.instruct.apiKey }}
+            - name: VLLM_INSTRUCT_API_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: vllm-instruct-api-key
+                  key: apiKey
+            {{- end }}
```
mayuka-c: Just a query, in case of OpenShift, dont we need to use Service Account?

ai-services/assets/applications/rag-dev/podman/values.yaml
sats-23: Plain text format, there is no validation on length or type of characters as of now

ai-services/assets/applications/rag-dev/openshift/templates/backend-deployment.yaml
sats-23: Reverted

ai-services/assets/applications/rag-dev/openshift/templates/backend-deployment.yaml
sats-23: I don't think so, this is a static injection similar to `opensearch-credentials`

mayuka-c: No I meant with RHOAI you could use Token-Based Authentication right?
This uses Seervice Account and you mount the token on service whichever needs communicate with models right?

I havent tried, but I do remember an option of enabling auth while deploying models in RHOAI dashboard

mkumatag: Hmm, yes, OpenShift AI does support an additional layer through some operators, such as authorino. Perhaps you could reach out to the OpenShift AI team to learn more about that approach? However, it would certainly involve more work like install those additional operator part of the application deployment with an additional specs etc.. 

I would vote for get started with this in this release and explore the operator way be next release(or add option shift support next release)

mkumatag:
@sats-23 how are we exposing the api key to the user if they want to consume the key post running?

sats-23:
> @sats-23 how are we exposing the api key to the user if they want to consume the key post running?

Currently we are passing the API key and consuming it both from backend server and vLLM pod side. 
Since the /v1/chat/completion or other endpoints to the user are exposed via backend server, the logic will handle passing the api key in auth header by default. 

So, currently, there is no exposing/showing api key to user. Only in case they want to consume directly from vLLM pod (pod network interface), they will have to pass the api key which is not the general practice in our app.

spyre-rag/src/common/llm_utils.py
```diff
@@ -123,17 +127,44 @@ def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, pdf_path,
...
     return resp_json
 
+
+def validate_vllm_auth(llm_endpoint):
+    """
+    Validate that the configured credentials can access vLLM.
+    Returns True on success, raises RuntimeError on failure.
+    """
+    query_vllm_models(llm_endpoint)
+    return True
```
manju956: Why are we always returning the `True` status. Can we evaluation the status of `query_vllm_models` and returns the eval result?

ai-services/assets/applications/rag-cpu/podman/templates/similarity-api.yaml.tmpl
```diff
@@ -50,6 +50,10 @@ spec:
           value: "{{ .Values.opensearch.auth.username }}"
         - name: OPENSEARCH_PASSWORD
           value: "{{ .Values.opensearch.auth.password }}"
+        {{- if .Values.instruct.apiKey }}
```
dharaneeshvrd: Not required for similarity service, since instruct is not used by this service.

spyre-rag/src/common/llm_utils.py
dharaneeshvrd, line 1: Many methods use /v1/chat/completion API, please analyse all the usage of instruct LLM and add auth header.
- Table summary
- Summarize (stream & non-stream)
- Q&A stream query

dharaneeshvrd:
> how are we exposing the api key to the user if they want to consume the key post running?

There is no neat way to get this now. But it would be great if we can get the values configured for the current deployment via a cli command.

spyre-rag/src/common/llm_utils.py
sats-23: `query_vllm_models` by default uses `response.raise_for_status()`
Followed the same pattern here to either return True or raise exception

sats-23:
> > how are we exposing the api key to the user if they want to consume the key post running?
> 
> There is no neat way to get this now. But it would be great if we can get the values configured for the current deployment via a cli command.

I assume we should also explain to the user that they need to use it only in case of access vLLM within pod networking. 
Else they may get confused and use it against the backend server

ai-services/assets/applications/rag-cpu/podman/templates/similarity-api.yaml.tmpl
sats-23: Sure, removed

spyre-rag/src/common/llm_utils.py
sats-23: Added, PTAL

spyre-rag/src/common/llm_utils.py
manju956: ok

sats-23:
@dharaneeshvrd @mkumatag @manju956 
The approach has been updated, PTAL

spyre-rag/src/common/llm_utils.py
sats-23: Now, there is a pass through auth for digitize and summarize API. 
Only chatbot UI and backend requires user to mention API Key, in all other cases, the key is automatically propagated

spyre-rag/src/chatbot/app.py
```diff
@@ -194,20 +202,119 @@ async def get_reference_docs(req: ReferenceRequest) -> ReferenceResponse:
...
+                return True
+            # For other errors, assume auth might be required to be safe
+            logging.debug(f"Error checking auth requirement: {e}, assuming auth is required")
+            auth_required_cache["checked"] = True
+            auth_required_cache["required"] = True
+            return True
+
+
+@app.get(
+    "/v1/auth-required",
```
dharaneeshvrd: Do we really need this? If the auth fails, anyway user would get the error message.
Where do you think this would be useful?

spyre-rag/src/chatbot/app.py
```diff
@@ -194,20 +202,119 @@ async def get_reference_docs(req: ReferenceRequest) -> ReferenceResponse:
...
+)
+async def check_auth_required():
+    """Check if authentication is required by testing vLLM access without API key."""
+    logging.debug("Checking if auth is required...")
+    auth_required = await is_auth_required()
+    return {"auth_required": auth_required}
+
+
+@app.post(
+    "/v1/validate-api-key",
```
dharaneeshvrd: Same as last comment

spyre-rag/src/common/llm_utils.py
```diff
@@ -26,10 +27,13 @@ def tqdm_wrapper(iterable, **kwargs):
         return iterable
 
 @retry_on_transient_error(max_retries=3, initial_delay=1.0, backoff_multiplier=2.0)
-def summarize_and_classify_single_table(prompt, gen_model, llm_endpoint):
+def summarize_and_classify_single_table(prompt, gen_model, llm_endpoint, api_key: str | None = None):
```
dharaneeshvrd: Why are we altering this method?
Can we access it from the settings directly?

spyre-rag/src/common/llm_utils.py
```diff
@@ -88,18 +92,21 @@ def summarize_and_classify_single_table(prompt, gen_model, llm_endpoint):
         logger.error(f"Error summarizing/classifying table: {e}")
         return "No summary.", False
 
-def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, pdf_path, max_workers=32):
+def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, pdf_path, max_workers=32, api_key: str | None = None):
```
dharaneeshvrd: here too

spyre-rag/src/common/llm_utils.py
dharaneeshvrd, line 1: Only query methods need to pass api key remaining methods can be the same with the only addition of headers added by directly accessing the settings(from env).

dharaneeshvrd:
Can you please update the PR description with sample vLLM and QnA chat completion curl request with auth headers?

spyre-rag/src/chatbot/app.py
sats-23: Primary idea behind `/v1/auth-required` was to decide if we should show the UI dialog box in the first place. 
If auth is disabled, we must not ask for API key, hence this was implemented

spyre-rag/src/chatbot/app.py
sats-23: Thought being we do the `validate-api-key` once for the session and are confident that all the other vLLM calls will pass. 
It was a part of good UX to have accept/reject early in the process instead of letting the users type a query and then see the exception.

spyre-rag/src/chatbot/app.py
dharaneeshvrd: In that case maintain it like an internal API, no need to add it in swagger doc.

spyre-rag/src/chatbot/app.py
dharaneeshvrd: Got it, please add a comment stating this purpose and treat it like internal API

spyre-rag/src/common/llm_utils.py
sats-23: Sure, I have updated it

spyre-rag/src/common/llm_utils.py
sats-23: Done, PTAL

spyre-rag/src/common/llm_utils.py
sats-23: Updated all, PTAL

sats-23:
> Only query methods need to pass api key remaining methods can be the same with the only addition of headers added by directly accessing the settings(from env).

After applying edits, I can see that most methods in the common utility are task specific except very few like tokenize call. 

I suggest, based on the purpose, we move that specific code into each app directory as it could be confusing from future maintenance PoV as QnA related vLLM calls are strictly awaiting api key from user but rest are pass through from settings but both are in the same file.

sats-23:
> Can you please update the PR description with sample vLLM and QnA chat completion curl request with auth headers?

Done, updated, PTAL

spyre-rag/src/chatbot/app.py
sats-23: Done, updated

spyre-rag/src/chatbot/app.py
sats-23: Done, updated

spyre-rag/src/chatbot/app.py
```diff
@@ -194,20 +202,125 @@ async def get_reference_docs(req: ReferenceRequest) -> ReferenceResponse:
...
+    - Path: /v1/auth-required
+    - Response: {"auth_required": true|false}
+
+    """
+    logging.debug("Checking if auth is required...")
+    auth_required = await is_auth_required()
+    return {"auth_required": auth_required}
+
+
+@app.post("/v1/validate-api-key", include_in_schema=False)
```
mkumatag: can we just call /v1/models instead of having this additional api? anywhere you are doing the same in the backed.!

spyre-rag/src/chatbot/app.py
```diff
@@ -194,20 +202,125 @@ async def get_reference_docs(req: ReferenceRequest) -> ReferenceResponse:
...
+                logging.debug("vLLM authentication IS required")
+                return True
+            # For other errors, assume auth might be required to be safe
+            logging.debug(f"Error checking auth requirement: {e}, assuming auth is required")
+            auth_required_cache["checked"] = True
+            auth_required_cache["required"] = True
+            return True
+
+
+@app.get("/v1/auth-required", include_in_schema=False)
```
mkumatag: How different is this with compare to listing the models and check whether you are hitting 401 at the client side?

spyre-rag/src/chatbot/app.py
dharaneeshvrd: Agree, we can make use of /v1/models of QnA to test the auth.

sats-23: Removed, now directly pings /v1/models

spyre-rag/src/chatbot/app.py
sats-23: Yes, it was just a wrapper, removed, now directly pings /v1/models

sats-23:
All cases have been tested and handled, PTAL

spyre-rag/src/chatbot/app.py
dharaneeshvrd, line 1: remove all the reference endpoint related things, its been removed by a recent PR to use the similarity search endpoint.

spyre-rag/src/chatbot/app.py
```diff
@@ -141,20 +151,126 @@ async def wrapper(*args, **kwargs):
...
+                # Auth is required
+                auth_required_cache["checked"] = True
+                auth_required_cache["required"] = True
+                logging.debug("vLLM authentication IS required")
+                return True
+            # For other errors, assume auth might be required to be safe
+            logging.debug(f"Error checking auth requirement: {e}, assuming auth is required")
+            auth_required_cache["checked"] = True
+            auth_required_cache["required"] = True
+            return True
```
dharaneeshvrd: Let's not assume and block the subsequent calls, if the auth is required user would still get 401 from LLM

spyre-rag/src/common/llm_utils.py
```diff
@@ -123,19 +124,58 @@ def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, pdf_path,
...
+    response = misc_utils.SESSION.get(
+        f"{llm_endpoint}/v1/models",
+        headers=get_vllm_headers(api_key),
+    )
     response.raise_for_status()
     resp_json = response.json()
     return resp_json
 
+
+def validate_vllm_auth(llm_endpoint, api_key: str | None = None):
```
dharaneeshvrd: seems this is unused now.

dharaneeshvrd (commented):
Otherwise lgtm
@pranithraoibm can you please review the UI related changes?

spyre-rag/src/common/llm_utils.py
sats-23: Removed

spyre-rag/src/chatbot/app.py
sats-23: Updated

spyre-rag/src/chatbot/app.py
sats-23: Sure, done

pranithraoibm (changes requested):
LGTM,
Just can you please run `npm run lint:fix` & `npm run format` to get rid of that lint warning

spyre-rag/src/digitize/app.py
```diff
@@ -5,7 +5,7 @@
 from contextlib import asynccontextmanager
 import uvicorn
 
-from fastapi import FastAPI, UploadFile, File, HTTPException, BackgroundTasks, Query, status, Request
+from fastapi import FastAPI, UploadFile, File, HTTPException, BackgroundTasks, Query, status, Request, Header
```
dharaneeshvrd: Remove this

spyre-rag/src/chatbot/response_utils.py
```diff
@@ -185,3 +185,6 @@ class HealthResponse(BaseModel):
             }
         }
     }
+
```
dharaneeshvrd: Remove this

pranithraoibm (dismissed):

spyre-rag/src/chatbot/response_utils.py
sats-23: Done

spyre-rag/src/digitize/app.py
sats-23: Done

spyre-rag/src/summarize/app.py
dharaneeshvrd, line 1: Sry, please remove the changes here too

mkumatag (dismissed):

spyre-rag/src/summarize/app.py
sats-23: Done

dharaneeshvrd (approved):

PR #669, opened 2026-04-24
AISERVICES-975 bugfix for insertion issue in opensearch 

manalilatkar:
bug: whenever the source metadata for chunked table contents is larger than 32766, opensearch fails the request due to illegal_argument_exception.

fix: now instead of sending full markdown of table as source metadata, we are sending caption to ensure proper searchability for table contents.

### DIFF
--- a/spyre-rag/src/digitize/doc_utils.py
+++ b/spyre-rag/src/digitize/doc_utils.py
@@ -337,7 +337,6 @@ def process_table(converted_doc, pdf_path, out_path, gen_model, gen_endpoint):
 
     filtered_table_dicts = {
         idx: {
-            'markdown': markdown,
             'summary': summary,
             'caption': caption,
             'page_number': page_num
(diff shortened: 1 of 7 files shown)


spyre-rag/src/digitize/doc_utils.py
```diff
@@ -1016,7 +1016,6 @@ def merge_chunked_documents(in_txt_chunk_f, in_tab_chunk_f, orig_fn):
             caption = block.get("caption", "")
             page_number = block.get("page_number")
             content = block.get("content", "")
-            md_source = block.get("markdown_source")
```
dharaneeshvrd, line 1019: if we are not going to store it part of the chunk we can even remove it from the initial process_tables method itself.

manalilatkar: Done. Removed all mentions of markdown_source.

spyre-rag/src/digitize/doc_utils.py
dharaneeshvrd: Need to remove here too

manalilatkar: done, sorry missed that. any more markdown mentions I'm missing?

dharaneeshvrd (approved):

dharaneeshvrd:
@manalilatkar Can you please update the description why markdown is not needed to be stored and caption is getting stored?

manalilatkar:
> @manalilatkar Can you please update the description why markdown is not needed to be stored and caption is getting stored?

@dharaneeshvrd fixed the commit message.

dharaneeshvrd (approved):

PR #676, opened 2026-04-27
Catalog DB Migration

mayuka-c:
- Migration command to do initial DB setup with future DB migrations.
- Uses goose tool for doing migration. I have used `golang-migrate` earlier but `goose` is widely used now.
- Table schema for Catalog service based on this PR: https://github.com/mayuka-c/project-ai-services/blob/d403e82b48876d2590b95c81437add91fa0012f4/docs/proposals/catalog/db-design-proposal.md
- Running the migration init as a init container.


mayuka-c:
I have tied the migration process to the pod instead of running it as part of `catalog configure` command. Open for any suggestions.

The reason for doing it is because if we are doing upgrade, then we need not have to tell user to rerun the catalog configure to perform migrations as redeploying of catalog service should take care.

ai-services/cmd/ai-services/cmd/catalog/migrate.go
```diff
@@ -0,0 +1,229 @@
...
+check migration status, and rollback migrations.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return cmd.Help()
+		},
+	}
+
+	// Add persistent flags for database connection
+	migrateCmd.PersistentFlags().StringVar(&dbHost, "db-host", "localhost", "Database host")
+	migrateCmd.PersistentFlags().IntVar(&dbPort, "db-port", DefaultDBPort, "Database port")
+	migrateCmd.PersistentFlags().StringVar(&dbUser, "db-user", "admim", "Database user")
```
yussufsh:
```suggestion
	migrateCmd.PersistentFlags().StringVar(&dbUser, "db-user", "admin", "Database user")
```

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
```diff
@@ -9,6 +9,32 @@ metadata:
...
+        - name: DB_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: catalog-secret
+              key: db-password
+      resources:
+        requests:
+          memory: "256Mi"
+        limits:
+          memory: "512Mi"
```
yussufsh, line 37: If I am not wrong request does not work on podman set it considers limit value. In that case setting both of them to 1 value makes sense.

ai-services/cmd/ai-services/cmd/catalog/migrate.go
```diff
@@ -0,0 +1,229 @@
...
+
+	"github.com/project-ai-services/ai-services/internal/pkg/catalog/db"
+	"github.com/project-ai-services/ai-services/internal/pkg/catalog/db/migrations"
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+	"github.com/spf13/cobra"
+)
+
+const (
+	// DefaultDBPort is the default PostgreSQL port.
+	DefaultDBPort = 5432
```
yussufsh: Why not move this to db.go as well?

ai-services/internal/pkg/catalog/db/db.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	return db, nil
+}
+
+// CreateDatabaseIfNotExists creates the database if it doesn't exist.
+func CreateDatabaseIfNotExists(cfg Config) error {
+	// Connect to postgres database to create the target database
+	adminCfg := cfg
+	adminCfg.DBName = "postgres"
```
yussufsh, line 80: Should not this be `ai_service`?

yussufsh (changes requested):

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
mayuka-c: Oh yeah I think I remember now, I will set it to 512 for both?
It doesnt require much memory

ai-services/internal/pkg/catalog/db/db.go
mayuka-c: This is for the initial connection.
This method CreateDatabaseIfNotExists() creates `ai-service` database if not exists. So before we can even create we need to establish a connection with postgres database to create it. So that is the reason the DB name is postgres.

ai-services/cmd/ai-services/cmd/catalog/migrate.go
mayuka-c: Done

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
mayuka-c: Done

ai-services/cmd/ai-services/cmd/catalog/migrate.go
mayuka-c: Done

ai-services/assets/catalog/podman/values.yaml
```diff
@@ -14,7 +14,7 @@ db:
   port: ""
   image: icr.io/ai-services-private/postgres:18
   user: "admin"
-  database: "postgres"
+  database: "ai_service"
```
Shubhamag12: should we rename it to ai_services rather?

ai-services/internal/pkg/catalog/db/db.go
```diff
@@ -0,0 +1,113 @@
...
+func (c *Config) ConnectionString() string {
+	return fmt.Sprintf(
+		"host=%s port=%d user=%s password='%s' dbname=%s sslmode=%s",
+		c.Host, c.Port, c.User, c.Password, c.DBName, c.SSLMode,
+	)
+}
+
+// Connect establishes a connection to the PostgreSQL database.
+func Connect(cfg Config) (*sql.DB, error) {
+	db, err := sql.Open("pgx", cfg.ConnectionString())
```
Shubhamag12: should we make it a constant?

ai-services/internal/pkg/catalog/db/db.go
```diff
@@ -0,0 +1,113 @@
...
+	User     string
+	Password string
+	DBName   string
+	SSLMode  string
+}
+
+// ConnectionString builds a PostgreSQL connection string from the config.
+func (c *Config) ConnectionString() string {
+	return fmt.Sprintf(
+		"host=%s port=%d user=%s password='%s' dbname=%s sslmode=%s",
```
Shubhamag12, line 44: should we validate all fields here before making a connection first?

ai-services/internal/pkg/catalog/db/db.go
```diff
@@ -0,0 +1,113 @@
...
+	adminCfg.DBName = "postgres"
+
+	db, err := Connect(adminCfg)
+	if err != nil {
+		return fmt.Errorf("failed to connect to postgres database: %w", err)
+	}
+	defer func() {
+		if closeErr := db.Close(); closeErr != nil {
+			// Log the error but don't override the main error
+			logger.Warningf("warning: failed to close database connection: %v\n", closeErr)
```
Shubhamag12:
```suggestion
			logger.Warningf("failed to close database connection: %v\n", closeErr)
```

Warningf adds `Warning` prefix

ai-services/internal/pkg/catalog/db/db.go
mayuka-c: Anything specific validation to do?
I feel whatever is present in config we can connect, if there are any issues let it fail. I feel it should be fine.

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
```diff
@@ -9,6 +9,32 @@ metadata:
   annotations:
     ai-services.io/ports: "{{ .Values.ui.port }}:8081"
 spec:
+  initContainers:
```
mkumatag, line 12: just explore how this behaves with openshift like env?!

ai-services/cmd/ai-services/cmd/catalog/migrate.go
```diff
@@ -0,0 +1,224 @@
...
+	statusCmd := createStatusCmd(getDBConfig)
+
+	// Subcommand: down - Rollback the last migration
+	downCmd := createDownCmd(getDBConfig)
+
+	// Add subcommands
+	migrateCmd.AddCommand(initCmd)
+	migrateCmd.AddCommand(upCmd)
+	migrateCmd.AddCommand(statusCmd)
+	migrateCmd.AddCommand(downCmd)
```
mkumatag, line 80: where these commands are used?

ai-services/internal/pkg/catalog/db/migrations/assets/20260430094502_create_applications_table.sql
```diff
@@ -0,0 +1,40 @@
...
+-- Create trigger to automatically update updated_at timestamp
+CREATE OR REPLACE FUNCTION update_updated_at_column()
+RETURNS TRIGGER AS $$
+BEGIN
+    NEW.updated_at = NOW();
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER update_applications_updated_at
```
mkumatag, line 24: I like this :)

ai-services/internal/pkg/catalog/db/migrations/assets/20260430094503_create_services_table.sql
```diff
@@ -0,0 +1,31 @@
...
+    created_at TIMESTAMPTZ DEFAULT NOW(),
+    updated_at TIMESTAMPTZ DEFAULT NOW(),
+    CONSTRAINT fk_app_id FOREIGN KEY (app_id) REFERENCES applications(id) ON DELETE CASCADE
+);
+
+-- Create trigger to automatically update updated_at timestamp
+CREATE TRIGGER update_services_updated_at
+    BEFORE UPDATE ON services
+    FOR EACH ROW
+    EXECUTE FUNCTION update_updated_at_column();
```
mkumatag, line 20: don't you face any issue if you are deleting this `update_updated_at_column` in the previous .sql file but it is still been referred in this trigger during the `drop`?

mkumatag (changes requested):

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
mayuka-c: Since we would have multiple pods in OpenShift, we would need to ensure that we run db migrations only once. So I'm thinking of using a job which runs only once when we deploy using Helm chart.

ai-services/cmd/ai-services/cmd/catalog/migrate.go
mayuka-c: To be honest, these commands will be kind of like a helper ones if we want to do some troubleshooting to see the migration status and bring down and up the table schema's.

These commands have no other purpose other than above. Also to execute these commands one needs to exec into our catalog-backend server container as we are not exposing the postgres DB outside

mayuka-c: PLMK your thoughts

ai-services/internal/pkg/catalog/db/migrations/assets/20260430094503_create_services_table.sql
mayuka-c: The updated_at triggers is created for each table.
So the one which you found in previous .sql file is for applications table and this is for services table.

mkumatag: yes, triggers are multiple but function is created only one and in the previous file 00002, if drop workflow worked without any issues I'm fine.

ai-services/cmd/ai-services/cmd/catalog/migrate.go
mkumatag: okay, let us hide them then!

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
mkumatag: hmm,! interesting.. we need to see how the whole take care of this..

ai-services/cmd/ai-services/cmd/catalog/migrate.go
mayuka-c: Done have hidden these

ai-services/internal/pkg/catalog/db/migrations/assets/20260430094503_create_services_table.sql
mayuka-c: Yeah it did work I tried with goose down

mayuka-c:
The PR is good for review/ merge. This is been validated. Its having the schema as mentioned in latest db-proposal.

mkumatag (approved):

yussufsh (approved):

PR #670, opened 2026-04-24
feat: add configurable app directory with --basedir flag

shwetha-s-poojary:
**What does this PR include?**

- Add --basedir flag to catalog configure and application create commands
- Introduce constants.DefaultBaseDir as single source of truth for default path
- Add early permission validation before any operations
- Update all templates to use dynamic baseDir from values
- Update vars.ModelDirectory dynamically based on baseDir
- Modify application templates to use {{ .Values.baseDir }}
- Update test files to use constants.GetApplicationsPath()

**Fixes:** #625

**Enables:** Users to specify custom storage locations for AI services data

Results:

```
# ls /root
go  project-ai-services  shubham  test_no_write_perm
# ai-services catalog configure --basedir /root/shwetha --runtime podman
Enter admin password:
Confirm admin password:
No existing pods found for application: ai-services

 Executing Layer 1/3: [catalog-secret.yaml.tmpl]
-------
Processing template: catalog-secret.yaml.tmpl
⠹ Deploying catalog service...-------
-------

 Executing Layer 2/3: [catalog-db.yaml.tmpl]
-------
Processing template: catalog-db.yaml.tmpl
⠴ Deploying catalog service...'catalog-db.yaml.tmpl', 'ai-services--db': Starting Pod Readiness check...
'catalog-db.yaml.tmpl', 'ai-services--db': Performing Containers Creation check for pod...
'catalog-db.yaml.tmpl', 'ai-services--db': Waiting for Containers Creation... Timeout set: 5m0s
'catalog-db.yaml.tmpl', 'ai-services--db': Containers creation check for pod is completed
'catalog-db.yaml.tmpl', 'ai-services--db', '3199f9b3e939-infra': Performing Container Readiness check...
'catalog-db.yaml.tmpl', 'ai-services--db', '3199f9b3e939-infra': Waiting for Container Readiness... Timeout set: 2m0s
'catalog-db.yaml.tmpl', 'ai-services--db', '3199f9b3e939-infra': Readiness Check for the container is completed!
-------
'catalog-db.yaml.tmpl', 'ai-services--db', 'ai-services--db-postgresql': Performing Container Readiness check...
'catalog-db.yaml.tmpl', 'ai-services--db', 'ai-services--db-postgresql': Waiting for Container Readiness... Timeout set: 2m30s
⠴ Deploying catalog service...'catalog-db.yaml.tmpl', 'ai-services--db', 'ai-services--db-postgresql': Readiness Check for the container is completed!
-------
[...]

PR #677, opened 2026-04-27
feat: Catalog uninstall cmd implementation

Shubhamag12:
JIRA: https://jsw.ibm.com/browse/AISERVICES-951

This PR implements the catalog uninstall command to remove the catalog pods, secrets.


ai-services/cmd/ai-services/cmd/catalog/uninstall.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	return nil
+}
+
+// configureUninstallFlags configures the flags for the uninstall command.
+func configureUninstallFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&runtimeType, "runtime", "r", "", fmt.Sprintf("runtime to use (options: %s, %s) (required)", types.RuntimeTypePodman, types.RuntimeTypeOpenShift))
+	_ = cmd.MarkFlagRequired("runtime")
+	cmd.Flags().BoolVarP(&uninstallAutoYes, "yes", "y", false, "Automatically accept all confirmation prompts (default=false)")
```
mayuka-c, line 79: Add --skip-cleanup flag too

ai-services/internal/pkg/catalog/cli/configure/podman/uninstall.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Remove all catalog pods
+	force := true
+	for _, podName := range existingPods {
+		logger.Infof("Removing pod: %s\n", podName)
+		if err := rt.DeletePod(podName, &force); err != nil {
+			s.Fail(fmt.Sprintf("failed to remove pod: %s", podName))
+
+			return fmt.Errorf("failed to remove pod %s: %w", podName, err)
+		}
```
mayuka-c: After this need to delete secrets too related to catalog

ai-services/internal/pkg/catalog/cli/configure/podman/uninstall.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	s.Stop("Catalog service removed successfully")
```
mayuka-c: You need to delete the Catalog data too which will be `/var/lib/ai-services/db`.
Since /var/lib/ai-services path can change, you may have to read the exact custom path by reading the env from catalog backend container

Shubhamag12: This might be an issue. If default dir is used, catalog uninstall would need to run as sudo to delete /var/lib/ai-services :( 
In this case, we can either document this, or wait for non-root functionality exploration for catalog.

ai-services/internal/pkg/catalog/cli/configure/podman/uninstall.go
Shubhamag12: done

ai-services/internal/pkg/catalog/cli/configure/podman/uninstall.go
Shubhamag12: I have pushed changes with secrets only for now.
Will implement db data deletion as well, based on above discussion.

LMK what you think!
CC: @yussufsh

ai-services/cmd/ai-services/cmd/catalog/uninstall.go
Shubhamag12: will add that flag along with db data deletion

ai-services/internal/pkg/catalog/cli/configure/podman/uninstall.go
yussufsh: Let's not worry about non-root mode yet.

shwetha-s-poojary: I think if this [PR](https://github.com/IBM/project-ai-services/pull/670/changes#diff-975c98b6e4e9835e42840c343d06772901174c3c1afa888a4a712544ff9ce948) goes in  , then you can just use the helper functions to fetch the custom base dir to delete the the application and/or db data.

Shubhamag12: will wait for your PR to go in

ai-services/cmd/ai-services/cmd/catalog/uninstall.go
Shubhamag12: added

ai-services/internal/pkg/catalog/cli/configure/podman/uninstall.go
Shubhamag12: I have added db deletion logic for default directory for now, once above [PR](https://github.com/IBM/project-ai-services/pull/670/changes#diff-975c98b6e4e9835e42840c343d06772901174c3c1afa888a4a712544ff9ce948) goes in, I will make that change

ai-services/internal/pkg/catalog/cli/uninstall/podman/uninstall.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// Delete catalog pods
+	if err := podsDeletion(rt, pods); err != nil {
+		return err
+	}
+
+	// Delete catalog secret
+	if err := secretDeletion(rt); err != nil {
+		return err
+	}
```
mayuka-c: I think we need to split the admin password and postgres to seperate secret files.

What if people do a skip cleanup, then the db password is gone, but since the underlying db data is present, I think it still uses the same password and subsequent redeploy will fail?

Have you validated this scenario?

I think we need to delete only admin password all the time but postgres password deletion should always go with db data deletion. Pls verify

Shubhamag12: ah, I think you may be right
but re-depoyment doesnt fail, as we do db login as a part of initContainer and we dont have livenessProbe there, even though it keeps restarting after getting deployed.

I will add something here, also separate secrets for both

Shubhamag12: made changes, ptal

ai-services/internal/pkg/catalog/constants/catalog.go
```diff
@@ -18,4 +18,14 @@ const (
...
 )
 
+// Catalog name constants.
+const (
+	// CatalogAppName represents the catalog name.
+	CatalogAppName = "ai-services"
+	// CatalogDBSecretName represents the catalog database name.
+	CatalogDBSecretName = "catalog-db-secret"
+	// CatalogSecretName represents the catalog secret name .
+	CatalogSecretName = "catalog-secret"
```
mayuka-c: This approach of maintaining secret names will not scale well. This will work for Catalog. But we need to find a generic solution for this. This will also help in application side where we are bringing in secrets and we need to control the cleanup in a better way.

mayuka-c: I feel like we can have a annotation introduced which will tell if we want to delete a resource or not. This approach will work well if tomo we bring in other resource.

Shubhamag12: I investigated if kube play can populate Spec.Labels on secrets via the manifest. Tested both metadata.labels and the spec.labels field, both are ignored by Podman, spec.Labels is always empty.
So the annotation-based approach in the manifest won't work with kube play :(

I have started a discussion on it as well: https://github.com/containers/podman/discussions/28697

Shubhamag12:
@yussufsh removed TODO and added comment, PTAL

mayuka-c (approved):
LGTM

yussufsh (approved):

PR #678, opened 2026-04-27
Query Rephrasing logic

sats-23:
Tests output attached at [AISERVICES-979](https://jsw.ibm.com/browse/AISERVICES-979)

### DIFF
--- a/spyre-rag/src/chatbot/query_rephrasing.py
+++ b/spyre-rag/src/chatbot/query_rephrasing.py
@@ -16,23 +16,6 @@ import common.misc_utils as misc_utils
 
 logger = get_logger("query_rephrasing")
 
-# Rephrasing prompt template
-REPHRASE_PROMPT_TEMPLATE = """Given the conversation history and the current question, rephrase the current question to be a standalone, self-contained query that can be used for semantic search.
-
-Requirements:
-1. Replace pronouns (this, that, it, they) with specific nouns from context
-2. Include all necessary context to understand what is being asked
-3. Be concise and focused on the search intent
-4. Maintain the original question's meaning
-5. Return ONLY the rephrased query, no explanation
-
-Conversation History:
-{conversation_history}
-
-Current Question: {current_query}
-
-Rephrased Query:"""
-
 
 def format_messages_for_rephrasing(messages: List[Dict[str, str]]) -> str:
     """
@@ -203,11 +186,12 @@ async def rephrase_query_with_context(
         logger.debug("Query rephrasing skipped: conversational_mode is disabled")
         return current_query
     
-    # Default configuration
+    # Get configuration from settings
+    from chatbot.settings import settings
     default_config = {
-        "timeout_seconds": 5.0,
-        "max_tokens": 100,
-        "temperature": 0.0,
+        "timeout_seconds": settings.query_rephrasing.timeout_seconds,
+        "max_tokens": settings.query_rephrasing.max_tokens,
+        "temperature": settings.query_rephrasing.temperature,
     }
     
     # Merge with provided config
@@ -230,8 +214,11 @@ async def rephrase_query_with_context(
             logger.debug("Skipping query rephrasing: empty conversation history")
             return current_query
         
+        # Get prompt template from settings (already imported above)
+        prompt_template = settings.query_rephrasing.rephrase_prompt_template
+        
         # Build rephrasing prompt
-        prompt = REPHRASE_PROMPT_TEMPLATE.format(
+        prompt = prompt_template.format(
             conversation_history=conversation_history,
             current_query=current_query
         )
--- a/spyre-rag/src/chatbot/query_rephrasing.py
+++ b/spyre-rag/src/chatbot/query_rephrasing.py
@@ -188,16 +188,6 @@ async def rephrase_query_with_context(
     
     # Get configuration from settings
     from chatbot.settings import settings
-    default_config = {
-        "timeout_seconds": settings.query_rephrasing.timeout_seconds,
-        "max_tokens": settings.query_rephrasing.max_tokens,
-        "temperature": settings.query_rephrasing.temperature,
-    }
-    
-    # Merge with provided config
-    if config:
-        default_config.update(config)
-    cfg = default_config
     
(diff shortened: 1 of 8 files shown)


spyre-rag/src/chatbot/query_rephrasing.py
```diff
@@ -0,0 +1,276 @@
...
+from typing import List, Dict, Optional
+
+from common.misc_utils import get_logger
+from common.retry_utils import retry_on_transient_error
+import common.misc_utils as misc_utils
+
+logger = get_logger("query_rephrasing")
+
+# Rephrasing prompt template
+REPHRASE_PROMPT_TEMPLATE = """Given the conversation history and the current question, rephrase the current question to be a standalone, self-contained query that can be used for semantic search.
```
dharaneeshvrd: Define this inside settings.py

spyre-rag/src/chatbot/query_rephrasing.py
```diff
@@ -0,0 +1,276 @@
...
+
+Requirements:
```
dharaneeshvrd: - Strip out conversational filler (e.g., "Can you tell me", "Also", "Thanks").
- If the current question is already a standalone question and doesn't need context, return it exactly as is.
- If the history is irrelevant to the current question, do not force-merge them.

I feel these three points are valid and can be added, can you please check?

spyre-rag/src/chatbot/query_rephrasing.py
```diff
@@ -0,0 +1,276 @@
...
+        json=payload,
+        timeout=timeout
+    )
+    response.raise_for_status()
+    
+    data = response.json() or {}
+    choices = data.get("choices", [])
+    
+    if not choices:
+        logger.warning("LLM returned empty choices for query rephrasing")
```
dharaneeshvrd, line 110: If LLM can't repharse the query we need to fallback to the original query.

dharaneeshvrd (commented):
Please keep in mind that the user won't always make follow-up questions, there is an equal chance for it to be an independent question(at least in RAG context).

spyre-rag/src/chatbot/query_rephrasing.py
```diff
@@ -0,0 +1,276 @@
...
+        ... )
+        'Is Spyre supported on Power 11?'
+    """
+    # Check if conversational mode is enabled (master switch)
+    if not is_conversational_mode_enabled():
+        logger.debug("Query rephrasing skipped: conversational_mode is disabled")
+        return current_query
+    
+    # Default configuration
+    default_config = {
```
dharaneeshvrd: why are we defining here?
please move all the configuration to settings.py

spyre-rag/src/chatbot/query_rephrasing.py
```diff
@@ -0,0 +1,276 @@
...
+            max_tokens=cfg["max_tokens"],
+            temperature=cfg["temperature"],
+            timeout=cfg["timeout_seconds"]
+        )
+        
+        # Calculate latency
+        latency_ms = (time.time() - start_time) * 1000
+        
+        # Handle empty response - always fallback to original
+        if not rephrased_query or rephrased_query.strip() == "":
```
dharaneeshvrd, line 232: I see fallback is handled here

spyre-rag/src/chatbot/query_rephrasing.py
sats-23: It was duplicate config, removed it to reference settings

spyre-rag/src/chatbot/query_rephrasing.py
sats-23: Updated

spyre-rag/src/chatbot/query_rephrasing.py
sats-23: I have tried the new prompt and have better results 
PTAL - https://jsw.ibm.com/secure/attachment/12164974/v2Results.txt

spyre-rag/src/chatbot/query_rephrasing.py
```diff
@@ -0,0 +1,263 @@
...
+    """
+    Check if conversational mode is enabled in settings.
+    
+    This is the master switch that controls whether conversational RAG features
+    (including query rephrasing) are active.
+    
+    Returns:
+        True if conversational mode is enabled, False otherwise
+    """
+    try:
```
manju956, line 131: Do we need exceptional handling for simple settings check?

manju956 (commented):
just a nit, otherwise looks good

spyre-rag/src/chatbot/query_rephrasing.py
sats-23: Since the feature is split into multiple PRs, I am planning to implement conversational mode through a feature flag. Eventually this will be removed and It will always run on conversational mode, but this is just used as a debug mode for time being

manju956: Ok

spyre-rag/src/chatbot/query_rephrasing.py
```diff
@@ -0,0 +1,263 @@
...
+        logger.debug(f"Rephrasing query: '{current_query}'")
+        
+        # Call LLM for rephrasing
+        rephrased_query = call_llm_for_rephrasing(
+            prompt=prompt,
+            llm_endpoint=llm_endpoint,
+            llm_model=llm_model,
+            max_tokens=cfg["max_tokens"],
+            temperature=cfg["temperature"],
+            timeout=cfg["timeout_seconds"]
```
dharaneeshvrd:
```suggestion
            max_tokens=settings.query_rephrasing.timeout_seconds,
            ...
```
?
also if this method is not going to get reused, you can even directly access these settings in it.

dharaneeshvrd (commented):
overall lgtm
except one minor nit

spyre-rag/src/chatbot/query_rephrasing.py
sats-23: Updated, PTAL

dharaneeshvrd (approved):

PR #679, opened 2026-04-27
fix: Correct steps files

yussufsh:
Fix below issues
1. digitize pod name is `<app_name>--digitize` instead of `<app_name>--digitize-api`
3. digitize ui and backend status were reversed.


mayuka-c (approved):
Oops, nice catch :)
LGTM

PR #682, opened 2026-04-27
feat: enable ai-services to run in non-root mode

Shubhamag12:
Non-root user implementation based on proposal: https://github.com/IBM/project-ai-services/pull/683


ai-services/internal/pkg/bootstrap/spyreconfig/spyre/repair.go
```diff
@@ -347,4 +406,353 @@ func fixVFIOPermissions(checkMap map[string]check.CheckResult, userGroupResult R
...
+	if status == "Disabled" {
+		return RepairResult{
+			CheckName: checkName,
+			Status:    StatusSkipped,
+			Message:   "SELinux is disabled",
+		}
+	}
+
+	// Check if VFIO devices exist
+	if !utils.FileExists("/dev/vfio") {
```
mkumatag: let us not do this check, let us just install the policies even if no devices found!

ai-services/assets/applications/rag-dev/podman/templates/vllm-server.yaml.tmpl
```diff
@@ -30,6 +30,9 @@ spec:
         type: Directory
   containers:
     - name: instruct
+      securityContext:
```
mkumatag: you don't need I believe, all the containers will have this label

mkumatag: and same comment applies for all the other places where you have these labels.

ai-services/internal/pkg/bootstrap/spyreconfig/spyre/repair.go
```diff
@@ -347,4 +406,353 @@ func fixVFIOPermissions(checkMap map[string]check.CheckResult, userGroupResult R
...
+
+	return RepairResult{
+		CheckName: checkName,
+		Status:    StatusFixed,
+		Message:   "SELinux VFIO policy configured successfully",
+	}
+}
+
+// buildAndInstallSELinuxPolicy builds and installs the SELinux policy module.
+func buildAndInstallSELinuxPolicy(tmpDir string) error {
```
mkumatag, line 551: may be we need to make this generic for other policy like podman socket, will do in a follow up PR

ai-services/internal/pkg/bootstrap/spyreconfig/spyre/repair.go
```diff
@@ -347,4 +406,353 @@ func fixVFIOPermissions(checkMap map[string]check.CheckResult, userGroupResult R
...
+	exitCode, _, stderr, err = utils.ExecuteCommand("semodule", "-i", ppPath)
+	if err != nil || exitCode != 0 {
+		return fmt.Errorf("failed to install policy module: %v, stderr: %s", err, stderr)
+	}
+
+	return nil
+}
+
+// updateSELinuxFileContext updates the SELinux file context database.
+func updateSELinuxFileContext() error {
```
mkumatag: This function explicitly labels the device file with but this needs to be run every time when new devices are created(hotplug etc.) to avoid this we can enhance the existing udev rule which will add the labels automatically when there is a new device onboard.

e.g:
```
# Example udev rule
SUBSYSTEM=="vfio", ACTION=="add", SECLABEL{selinux}="system_u:object_r:vfio_device_t:s0"
```

explore adding the content to the already existing udev rule

ai-services/internal/pkg/bootstrap/spyreconfig/spyre/repair.go
```diff
@@ -347,4 +406,353 @@ func fixVFIOPermissions(checkMap map[string]check.CheckResult, userGroupResult R
...
+			return nil
+		}
+		return fmt.Errorf("failed to add file context: %v, stderr: %s", err, stderr)
+	}
+
+	return nil
+}
+
+// applySELinuxLabels applies SELinux labels to existing VFIO devices.
+func applySELinuxLabels() error {
```
mkumatag: even this is not required if we handle them via udev rule which I mentioned above.

ai-services/internal/pkg/bootstrap/spyreconfig/utils/utils.go
```diff
@@ -236,41 +234,18 @@ func ReloadUdevRules() error {
...
-	return path, nil
-}
-
-// PodmanHealthCheck verifies podman is working.
-func PodmanHealthCheck() error {
-	client, err := podman.NewPodmanClient()
-	if err != nil {
-		return fmt.Errorf("failed to create podman client: %w", err)
+	// Trigger for the vfio kernel device (/dev/vfio/vfio)
+	if err := executeCommandOrFail("udevadm", "trigger", "--name-match=/dev/vfio/vfio"); err != nil {
```
mkumatag: can we club both into one? something like this:
```
"udevadm", "trigger", "--subsystem-match=vfio", "--name-match=/dev/vfio/vfio"
```

ai-services/internal/pkg/runtime/podman/podman.go
```diff
@@ -51,6 +53,35 @@ func NewPodmanClient() (*PodmanClient, error) {
...
+	}
+
+	if os.Geteuid() == 0 {
+		return rootPodmanURI()
+	}
+
+	return fmt.Sprintf("unix:///run/user/%d/podman/podman.sock", os.Getuid()), nil
+}
+
+func rootPodmanURI() (string, error) {
```
mkumatag: this says root but we are finding SUDO_USER, I don't understand the purpose of this function! can you please add comment for the purpose of this function?

ai-services/internal/pkg/validators/podman/root/root.go
```diff
@@ -25,10 +25,18 @@ func (r *RootRule) Description() string {
 func (r *RootRule) Verify() error {
```
mkumatag: this becomes info only check instead of a blocking one?! correct if I'm wrong!

mkumatag (changes requested):

ai-services/assets/applications/rag-dev/podman/templates/vllm-server.yaml.tmpl
Shubhamag12: Okay, I will verify this and remove if not required

ai-services/internal/pkg/runtime/podman/podman.go
Shubhamag12: Ah, basically this function determines the appropriate Podman socket URI when running with root privileges.
If command was run by using sudo, it returns socket path for SUDO_USER, else it returns root path. I will change the function naming along with comment, it is causing confusion.

ai-services/internal/pkg/validators/podman/root/root.go
Shubhamag12: I think you are right
That leaves us with one confusion, for bootstrap we require either sudo or root privileges. Also, this function is called while creating application as well (via validate), there this function will fail if user is non-root.

ai-services/internal/pkg/validators/podman/root/root.go
```diff
@@ -25,10 +25,18 @@ func (r *RootRule) Description() string {
...
 
-	if euid != 0 {
-		return fmt.Errorf("current user is not root (EUID: %d)", euid)
+		return nil
+	}
+
+	if euid != 0 && os.Getenv("XDG_RUNTIME_DIR") == "" {
+		uid := os.Getuid()
+		logger.Infoln("running command as %s", uid, logger.VerbosityLevelDebug)
+		if err := os.Setenv("XDG_RUNTIME_DIR", fmt.Sprintf("/run/user/%d", uid)); err != nil {
```
mkumatag: If we set it here, it will only apply to that specific shell. I'm not sure why we're setting this environment variable here at all.

ai-services/internal/pkg/bootstrap/podman/helper.go
```diff
@@ -16,11 +16,6 @@ import (
 	"github.com/project-ai-services/ai-services/internal/pkg/logger"
 )
 
-const (
-	// dirPermissions is the default permission for creating directories.
-	dirPermissions = 0755
```
mkumatag: These permissions are for tow different things, let us keep this variable here as well for the local consumption.

mkumatag (changes requested):

ai-services/internal/pkg/validators/podman/root/root.go
Shubhamag12: Should I move it to create cmd in that case?
because while creating application, I was facing this error
```
failed to pull image registry.redhat.io/rhaiis/vllm-spyre-rhel9:3.2.5: reading JSON file "/run/containers/1001/auth.json": open /run/containers/1001/auth.json: permission denied
```
Even though the above error says permission denied, it was because of the fact that podman was not able to access the runtime files.


So thought of exporting this var here since we call validate function everytime we create application

ai-services/internal/pkg/validators/podman/root/root.go
mkumatag: I'm little worried about the below modification you are making that is it! IMO we should just remove this root validator itself considering we don't need root anymore..

ai-services/internal/pkg/validators/podman/root/root.go
mkumatag: remove this validator and fix this permission somewhere else

and fix this somewhere else wher

ai-services/internal/pkg/bootstrap/spyreconfig/utils/utils.go
Shubhamag12: seems like we cannot club them as subsystem for /dev/vfio/vfio is misc, not vfio


```
[user@ocp5 ai-services]$ udevadm info /dev/vfio/vfio 2>&1 | grep SUBSYSTEM

E: SUBSYSTEM=misc
```

ai-services/assets/applications/rag-dev/podman/templates/vllm-server.yaml.tmpl
Shubhamag12: verified, it works

ai-services/internal/pkg/bootstrap/spyreconfig/spyre/repair.go
Shubhamag12: sure, will create a tracker for it

ai-services/internal/pkg/bootstrap/spyreconfig/spyre/repair.go
Shubhamag12: verified, this works fine
Will test hotplug as well once we have enough spyre cards

ai-services/internal/pkg/validators/podman/root/root.go
Shubhamag12: Since this check is required for running bootstrap
I will remove it from here and mention this check explicitly in bootstrap process

Shubhamag12: will remove it

ai-services/internal/pkg/validators/podman/root/root.go
Shubhamag12: strange observation

I removed this part of code and ran a test, it worked fine, because we enable-linger, which allows process to access the runtime files.

ai-services/internal/pkg/bootstrap/podman/helper.go
Shubhamag12: Will get this one merged first: https://github.com/IBM/project-ai-services/pull/691
and then as part of rebase, the above const wont be required here

mkumatag: are we waiting for #691 to merge first and post that we will merge this one?

mkumatag:
@Shubhamag12 please rebase this, others - please review!

ai-services/internal/pkg/bootstrap/podman/helper.go
Shubhamag12: yes

mkumatag:
@Shubhamag12 please test this feature few times on a newly deployed vm so that we are okay because its tricky and behaves differently first time and next time when you run this.

Shubhamag12:
> @Shubhamag12 please test this feature few times on a newly deployed vm so that we are okay because its tricky and behaves differently first time and next time when you run this.

I tested this on new setup and re-deployed as different users as well. The above code works fine

mkumatag (dismissed):

mkumatag:
Feel free to merge when you guys are comfortable :)

mkumatag:
cc @yussufsh @mayuka-c

ai-services/cmd/ai-services/cmd/bootstrap/bootstrap.go
```diff
@@ -61,6 +62,9 @@ func BootstrapCmd() *cobra.Command {
 			if rt == types.RuntimeTypePodman {
 				logger.Infoln("LPAR bootstrapped successfully")
 				logger.Infoln("----------------------------------------------------------------------------")
+				style := lipgloss.NewStyle().Foreground(lipgloss.Color("#32BD27"))
+				message := style.Render("Re-login to the shell to reflect necessary permissions assigned to vfio cards")
```
yussufsh: Not needed for root user as discussed right?

yussufsh (changes requested):

ai-services/cmd/ai-services/cmd/bootstrap/bootstrap.go
Shubhamag12: modified

yussufsh (approved):

PR #683, opened 2026-04-27
doc: Non-root user proposal

Shubhamag12:
A proposal for enabling non-root users to run AI workloads with Spyre cards.


docs/proposals/non-root-user-configuration-proposal.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Administrator Tasks
+1. Run bootstrap command with sudo:
+   ```bash
+   sudo ai-services bootstrap configure
+   ```
+
+2. Create application data directory:
+   ```bash
+   sudo mkdir -p /var/lib/ai-services/{models,cache,data}
```
mkumatag: `/var/lib/ai-services` will be changing soon, may be you can mention like base_dir or some placeholder name here

mkumatag: and in all the places wherever we are using this

docs/proposals/non-root-user-configuration-proposal.md
```diff
@@ -0,0 +1,383 @@
...
+sudo loginctl enable-linger <username>
+
+# Step 3: Log back in as the user
+# (SSH back in)
+
+# Step 4: Verify configuration is active
+groups | grep sentient          # Should show sentient group
+ulimit -n                        # Should show 134217728
+ulimit -l                        # Should show unlimited
+```
```
mayuka-c: Will these be manual steps that user needs to do after bootstrap configure?

docs/proposals/non-root-user-configuration-proposal.md
Shubhamag12: one question: are we mandating setting a custom directory? from what I got, the `/var/lib/ai-services` dir will be default.

maybe I can generalize the statement to have permissions for above directory if not using their custom directory?

docs/proposals/non-root-user-configuration-proposal.md
Shubhamag12: step 1, 2, 3 will be mandatory, step 4 is optional (if users wants to verify). I will mark it optional

docs/proposals/non-root-user-configuration-proposal.md
mkumatag: @yussufsh are we retaining this as a default dir? or we are changing to something else?

docs/proposals/non-root-user-configuration-proposal.md
```diff
@@ -0,0 +1,383 @@
...
+## Prerequisites Summary
+
+### Administrator Tasks
+1. Run bootstrap command with sudo:
+   ```bash
+   sudo ai-services bootstrap configure
+   ```
+
+2. Create application data directory:
+   ```bash
```
mayuka-c: The base_dir we take as part of `catalog configure`. So this needs to be created there!.

During bootstrap configure how would you know base_dir or you end up creating this whether or not base_dir is used.

Shubhamag12: I will generalize the statement to have permissions for custom dir, or default one if not set

mayuka-c:
Also I think we need to cover the non root aspects for Catalog service too where we have selinux issues: https://github.com/IBM/project-ai-services/blob/main/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl#L43

If we do not want to cover here, maybe we can take it up separately on how to deal with this too, to ensure all the services works fine with non root.

docs/proposals/non-root-user-configuration-proposal.md
mayuka-c: The only concern is you expect the custom dir to be passed in bootstrap configure and also catalog configure?
What if user passes different values in both.

Shubhamag12: I think there is a confusion here, we do not expect user to pass directory via bootstrap configure. We are asking to create `/var/lib/ai-services` and its subdirectories, and assign permission to sentient group, as a pre-requisite.

And as mentioned above, I will generalize the statement to have permissions for custom dir, or default one if not set

I will update the doc and add a pre-requisite section at top

Shubhamag12:
> Also I think we need to cover the non root aspects for Catalog service too where we have selinux issues:

I am planning to take it up separately, I will update the same doc once we have done full poc with non-root on catalog

docs/proposals/non-root-user-configuration-proposal.md
mayuka-c: Ha ok, even this is a manual task. Then it should be fine :)

I thought we weere automating default directory creation based on whether user passed base dir or not.

Shubhamag12: The initial proposal was to create dir as per requirement, but since create `/var/lib/ai-services` requires privileged permissions, it actually depends on whether `catalog bootstrap` would be run as sudo or not.

@yussufsh Should we include default dir creation as part of `catalog configure`, if running with sudo?

docs/proposals/non-root-user-configuration-proposal.md
Shubhamag12: confirmed with @yussufsh, we are retainingthis as default dir

docs/proposals/non-root-user-configuration-proposal.md
Shubhamag12: done, ptal

Shubhamag12:
@mkumatag @mayuka-c Addressed all comments, ptal

docs/proposals/non-root-user-configuration-proposal.md
```diff
@@ -0,0 +1,340 @@
...
+   sudo usermod -aG wheel <username>
+   sudo loginctl enable-linger <username>
+   ```
+2. Run bootstrap command with sudo:
+   ```bash
+   sudo ai-services bootstrap
+   ```
+
+3. Create application data directory (if using default location):
```
manju956, line 23: Can we mention this is for podman runtime env?

Shubhamag12: This proposal is just for podman
I will mention it in doc description

docs/proposals/non-root-user-configuration-proposal.md
```diff
@@ -0,0 +1,340 @@
...
+- Default `/var/lib/ai-services` is not automatically created by ai-services tool (requires admin decision on ownership)
+- Custom directories allow users to work without requiring sudo/admin privileges
+- Documented as prerequisite in user guide
+
+---
+
+## 5. SMT Level Configuration
+
+### Problem
+SMT level affects CPU performance for Spyre card operations. Previously configured as part of application create. Since applicataion create run in rootless mode, SMT configuration will fail.
```
manju956: typo `applicataion` -> `application`

docs/proposals/non-root-user-configuration-proposal.md
Shubhamag12: added a note on top, ptal

docs/proposals/non-root-user-configuration-proposal.md
Shubhamag12: done

manju956 (commented):
changes looks good to me

mkumatag (approved):

PR #684, opened 2026-04-28
Summarisation improvements implementation

Niharika0306:
Proposal plan for this PR : https://github.com/IBM/project-ai-services/pull/621

### DIFF
--- a/spyre-rag/src/summarize/app.py
+++ b/spyre-rag/src/summarize/app.py
@@ -164,7 +164,7 @@ async def handle_summarize(
     else:
         logger.info(
             f"Received {input_type} request with input size: {input_word_count} words ({input_tokens} tokens), "
-            f"automatic length: {target_words} words"
+            f"automatic length"
         )
 
     messages = build_messages(content_text, target_words, min_words, max_words,
--- a/spyre-rag/src/summarize/app.py
+++ b/spyre-rag/src/summarize/app.py
@@ -29,11 +29,10 @@ from summarize.summ_utils import (
     trim_to_last_sentence,
     MAX_INPUT_WORDS,
     compute_target_and_max_tokens,
-    compute_target_and_max_tokens_from_level,
     SummarizeSuccessResponse,
     validate_summary_length,
     validate_summary_level,
-    validate_input_word_count_for_level,
+    validate_input_and_get_available_tokens,
     extract_text_from_pdf
 )
 
@@ -116,20 +115,6 @@ async def locked_stream(stream_g):
         concurrency_limiter.release()
 
 
-def _validate_input_word_count(input_word_count: int, summary_length: Optional[int]) -> None:
-    """Raise SummarizeException if word-count constraints are violated."""
-    if summary_length and summary_length > input_word_count:
-        raise SummarizeException(
-            400, "INPUT_TEXT_SMALLER_THAN_SUMMARY_LENGTH",
-            "Input text is smaller than summary length",
-        )
-    if input_word_count > MAX_INPUT_WORDS:
-        raise SummarizeException(
-            413, "CONTEXT_LIMIT_EXCEEDED",
-            "Input size exceeds maximum token limit",
-        )
-
-
 async def handle_summarize(
     content_text: str,
     input_type: str,
@@ -153,47 +138,37 @@ async def handle_summarize(
     if summary_level is not None and summary_length is not None:
         raise SummarizeException(
             400, "INVALID_PARAMETER",
-            "Cannot specify both 'summary_level' and 'length'. Please use only one."
+            "Cannot specify both 'level' and 'length'. Please use only one."
         )
     
+    # Unified validation and computation
+    available_output_tokens = validate_input_and_get_available_tokens(
+        input_tokens, input_word_count, summary_level, summary_length
+    )
+    
+    target_words, min_words, max_words, max_tokens = compute_target_and_max_tokens(
+        input_tokens, available_output_tokens, summary_level, summary_length
+    )
+    
+    # Log appropriate message based on which parameter was provided
     if summary_level is not None:
-        # New approach: use abstraction level
-        # Validate input and get available output tokens (using already-computed token count)
-        available_output_tokens = validate_input_word_count_for_level(
-            input_tokens, input_word_count, summary_level
-        )
-        target_words, min_words, max_words, max_tokens = compute_target_and_max_tokens_from_level(
-            input_tokens, input_word_count, summary_level, available_output_tokens
-        )
-        has_length_spec = True
         logger.info(
             f"Received {input_type} request with input size: {input_word_count} words ({input_tokens} tokens), "
             f"level: {summary_level}, target: {target_words} words ({min_words}-{max_words})"
         )
     elif summary_length is not None:
-        # Legacy approach: direct length specification
-        _validate_input_word_count(input_word_count, summary_length)
-        target_words, min_words, max_words, max_tokens = compute_target_and_max_tokens(
-            input_tokens, input_word_count, summary_length
-        )
-        has_length_spec = True
         logger.info(
             f"Received {input_type} request with input size: {input_word_count} words ({input_tokens} tokens), "
             f"target summary length: {summary_length} words"
         )
     else:
-        # Automatic: use default coefficient
-        _validate_input_word_count(input_word_count, None)
-        target_words, min_words, max_words, max_tokens = compute_target_and_max_tokens(
-            input_tokens, input_word_count, None
-        )
-        has_length_spec = False
         logger.info(
             f"Received {input_type} request with input size: {input_word_count} words ({input_tokens} tokens), "
             f"automatic length: {target_words} words"
         )
 
-    messages = build_messages(content_text, target_words, min_words, max_words, has_length_spec)
+    messages = build_messages(content_text, target_words, min_words, max_words,
+                            (summary_length is not None or summary_level is not None))
 
     if stream:
         await concurrency_limiter.acquire()
@@ -269,16 +244,16 @@ description=(
       "| Field | Type | Required | Description |\n"
       "|-------|------|----------|-------------|\n"
       "| `text` | string | Yes | Plain text content to summarize |\n"
-      "| `summary_level` | string | No | Abstraction level: 'brief', 'standard' (default), or 'detailed' |\n"
+      "| `level` | string | No | Abstraction level: 'brief', 'standard' (default), or 'detailed' |\n"
       "| `length` | integer | No | (Legacy) Desired summary length in words |\n"
       "| `stream` | boolean | No | Stream the summary as it is generated, default False |\n\n"
-      "**Note:** Use either `summary_level` (recommended) or `length` (legacy), not both.\n\n"
-      "**Example with summary_level:**\n"
+      "**Note:** Use either `level` (recommended) or `length` (legacy), not both.\n\n"
+      "**Example with level:**\n"
       "```bash\n"
       'curl -X POST /v1/summarize -H "Content-Type: application/json" -d '
       '{\n'
       '  "text": "Artificial intelligence has made significant progress...",\n'
-      '  "summary_level": "brief"\n'
+      '  "level": "brief"\n'
       '}\n'
       "```\n\n"
       "**Example with length (legacy):**\n"
@@ -294,13 +269,13 @@ description=(
       "| Field | Type | Required | Description |\n"
       "|-------|------|----------|-------------|\n"
       "| `file` | file | Conditional | `.txt` or `.pdf` file to summarize |\n"
-      "| `summary_level` | string | No | Abstraction level: 'brief', 'standard' (default), or 'detailed' |\n"
+      "| `level` | string | No | Abstraction level: 'brief', 'standard' (default), or 'detailed' |\n"
       "| `length` | integer | No | (Legacy) Desired summary length in words |\n"
       "| `stream` | boolean | No | Stream the summary as it is generated, default False |\n\n"
-      "**Note:** Use either `summary_level` (recommended) or `length` (legacy), not both.\n\n"
-      "**Example with summary_level:**\n"
+      "**Note:** Use either `level` (recommended) or `length` (legacy), not both.\n\n"
+      "**Example with level:**\n"
       "```bash\n"
-      'curl -X POST /v1/summarize -F "file=@report.pdf" -F "summary_level=detailed"\n'
+      'curl -X POST /v1/summarize -F "file=@report.pdf" -F "level=detailed"\n'
       "```\n\n"
       "**Example with length (legacy):**\n"
       "```bash\n"
@@ -335,8 +310,8 @@ async def summarize(request: Request):
                 raise SummarizeException(400, "MISSING_INPUT",
                                          "Either 'text' or 'file' parameter is required")
             
-            # Support both summary_level (new) and length (legacy)
-            summary_level = validate_summary_level(body.get("summary_level"))
+            # Support both level (new) and length (legacy)
+            summary_level = validate_summary_level(body.get("level"))
             summary_length = validate_summary_length(body.get("length"))
             stream = bool(body.get("stream", False))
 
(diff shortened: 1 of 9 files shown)


spyre-rag/src/summarize/app.py
```diff
@@ -130,23 +133,67 @@ def _validate_input_word_count(input_word_count: int, summary_length: Optional[i
...
-    )
-
+    
+    # Get LLM endpoint for tokenization
     llm_endpoint = llm_model_dict['llm_endpoint']
     llm_model = llm_model_dict['llm_model']
+    
+    # Get actual token count from the input text
+    input_tokens = await asyncio.to_thread(
+        lambda: len(tokenize_with_llm(content_text, llm_endpoint))
```
dharaneeshvrd, line 134: Hope there are no limitation on the content size that this API can handle, please validate once

spyre-rag/src/summarize/summ_utils.py
```diff
@@ -27,16 +27,145 @@
...
+    
+    # Calculate minimum output tokens needed
+    minimum_output_tokens = int(settings.summarize.minimum_summary_words / settings.common.llm.token_to_word_ratio_en)
+    
+    # Hard limit: input + prompt + minimum_output must fit in context
+    max_allowed_input_tokens = (
+        settings.common.llm.granite_3_3_8b_instruct_context_length -
+        settings.summarize.summarization_prompt_token_count -
+        minimum_output_tokens
+    )
```
dharaneeshvrd: I think these can be defined in global level based on current config. No need to calculate it for every request.

spyre-rag/src/summarize/summ_utils.py
```diff
@@ -27,16 +27,145 @@
...
+        settings.common.llm.granite_3_3_8b_instruct_context_length -
+        input_tokens -
+        settings.summarize.summarization_prompt_token_count
+    )
+    
+    # Soft limit check: log warning if level's ideal output won't fit, but don't fail
+    level_config = getattr(settings.summarize.summarization_levels, summary_level)
+    
+    # Calculate ideal output tokens for this level
+    ideal_output_tokens = int(
```
dharaneeshvrd: Seems this one also not depends on the input tokens which means we can define it global level.

spyre-rag/src/summarize/summ_utils.py
```diff
@@ -27,16 +27,145 @@
...
+        (settings.common.llm.granite_3_3_8b_instruct_context_length - settings.summarize.summarization_prompt_token_count) *
+        (settings.summarize.summarization_coefficient * level_config.multiplier) /
+        (1 + settings.summarize.summarization_coefficient * level_config.multiplier)
+    )
+    
+    if available_output_tokens < ideal_output_tokens:
```
dharaneeshvrd: Can you please explain with a comment on which scenario this can happen?
But wondering why are we comparing with a generic output tokens with output tokens based on current input?
I feel we need to calculate the ideal token also based on input and compare, if it available is still less than the ideal count, than we can print this message IMO.

spyre-rag/src/summarize/summ_utils.py
```diff
@@ -27,16 +27,145 @@
...
+    
+    Returns:
+        (target_words, min_words, max_words, max_tokens)
+    """
+    # Get level configuration
+    level_config = getattr(settings.summarize.summarization_levels, summary_level)
+    
+    # Calculate ideal target based on input tokens and level multiplier
+    base_target_tokens = int(input_tokens * settings.summarize.summarization_coefficient)
+    ideal_target_tokens = int(base_target_tokens * level_config.multiplier)
```
dharaneeshvrd: I feel this should be used and the warning can be printed here

spyre-rag/src/summarize/summ_utils.py
```diff
@@ -27,16 +27,145 @@
...
+    
+    # Calculate min/max bounds (85% to 115% of target)
+    min_words = int(target_word_count * 0.85)
+    max_words = int(target_word_count * 1.15)
+    
+    # Cap max_words to available space
+    max_possible_words = int(available_output_tokens * settings.common.llm.token_to_word_ratio_en)
+    max_words = min(max_words, max_possible_words)
+    
+    # Add small buffer to max_tokens
```
dharaneeshvrd: add a comment on why this buffer is needed

spyre-rag/src/summarize/summ_utils.py
```diff
@@ -27,16 +27,145 @@
...
+    
+    logger.debug(
+        f"Level: {summary_level}, Input: {input_tokens} tokens, Target: {target_word_count} words "
+        f"({min_words}-{max_words}), Max tokens: {max_tokens}, Available: {available_output_tokens}"
+    )
+    
+    return target_word_count, min_words, max_words, max_tokens
+
+
+def compute_target_and_max_tokens(input_tokens: int, input_word_count: int, summary_length: Optional[int]):
```
dharaneeshvrd: `input_word_count` seems unused

dharaneeshvrd: can we use the new token calculation approach in this method as well?
Just wanted to see the default method also uses token based limit calculation.

spyre-rag/src/summarize/app.py
manalilatkar: +1 to Dharaneesh's comment:
Hope there are no limitation on the content size that this API can handle, please validate once

spyre-rag/src/summarize/app.py
```diff
@@ -130,23 +133,67 @@ def _validate_input_word_count(input_word_count: int, summary_length: Optional[i
...
+        target_words, min_words, max_words, max_tokens = compute_target_and_max_tokens(
+            input_tokens, input_word_count, None
+        )
+        has_length_spec = False
+        logger.info(
+            f"Received {input_type} request with input size: {input_word_count} words ({input_tokens} tokens), "
+            f"automatic length: {target_words} words"
+        )
+
+    messages = build_messages(content_text, target_words, min_words, max_words, has_length_spec)
```
manalilatkar: Instead of adding a new variable called `has_length_spec`, this line can just be:
`messages = build_messages(content_text, target_words, min_words, max_words, (summary_length is not None or summary_level is not None) )`

spyre-rag/src/summarize/app.py
```diff
@@ -222,9 +269,19 @@ async def handle_summarize(
       "| Field | Type | Required | Description |\n"
       "|-------|------|----------|-------------|\n"
       "| `text` | string | Yes | Plain text content to summarize |\n"
-      "| `length` | integer | No | Desired summary length in words  |\n"
+      "| `summary_level` | string | No | Abstraction level: 'brief', 'standard' (default), or 'detailed' |\n"
```
manalilatkar: Didn't we agree to change the name of the parameter from summary_level -> level
Also in line no 275, 276, 281, 297, 300, 303.

spyre-rag/src/summarize/app.py
```diff
@@ -271,16 +334,21 @@ async def summarize(request: Request):
...
-            return await handle_summarize(text, "text", summary_length, stream)
+            return await handle_summarize(text, "text", summary_length, summary_level, stream)
 
         # ----- Multipart / form-data path -----
         elif "multipart/form-data" in content_type:
             form = await request.form()
             file: Optional[UploadFile] = form.get("file")  # type: ignore[assignment]
 
+            # Support both summary_level (new) and length (legacy)
+            summary_level = validate_summary_level(form.get("summary_level"))
```
manalilatkar: summary_level -> level

spyre-rag/src/summarize/summ_utils.py
manalilatkar: +1. Also, add the calculation equation in comments so there is no confusion later.

spyre-rag/src/summarize/summ_utils.py
```diff
@@ -27,16 +27,145 @@
...
+        (settings.summarize.summarization_coefficient * level_config.multiplier) /
+        (1 + settings.summarize.summarization_coefficient * level_config.multiplier)
+    )
+    
+    if available_output_tokens < ideal_output_tokens:
+        available_output_words = int(available_output_tokens * settings.common.llm.token_to_word_ratio_en)
+        ideal_output_words = int(ideal_output_tokens * settings.common.llm.token_to_word_ratio_en)
+        logger.warning(
+            f"Input size ({input_word_count} words, {input_tokens} tokens) limits output space. "
+            f"'{summary_level}' level target is ~{ideal_output_words} words, "
```
manalilatkar: What is even the point of logging this warning? User is not going to see this and admin has no use for this.

manalilatkar (changes requested):

spyre-rag/src/summarize/summ_utils.py
Niharika0306: I agree. Will do.

spyre-rag/src/summarize/app.py
Niharika0306: right, forgot to reflect the changes here. Will do.

spyre-rag/src/summarize/summ_utils.py
Niharika0306: More for our debugging and also to reason. Because in these cases, the summary may not be expected as per the user request. So just to have a log of that.

spyre-rag/src/summarize/app.py
```diff
@@ -130,23 +133,67 @@ def _validate_input_word_count(input_word_count: int, summary_length: Optional[i
...
+        lambda: len(tokenize_with_llm(content_text, llm_endpoint))
+    )
+    
+    # Validate that both parameters are not provided simultaneously
+    if summary_level is not None and summary_length is not None:
+        raise SummarizeException(
+            400, "INVALID_PARAMETER",
+            "Cannot specify both 'summary_level' and 'length'. Please use only one."
+        )
+
```
manalilatkar, line 143: @Niharika0306, this block of code from line no 159 to 194 becomes much simpler and cleaner ( only 3 lines ) if you create functions `validate_input_word_count` and `compute_target_and_max_tokens` that accept both length and level. Just a suggestion.

spyre-rag/src/summarize/summ_utils.py
Niharika0306: addressed. It is defined at module level now since they are derived values.

spyre-rag/src/summarize/summ_utils.py
Niharika0306: yes, the calculation was complex and wrong before. Its corrected now.

spyre-rag/src/summarize/summ_utils.py
Niharika0306: Addressed as part of next comment.

spyre-rag/src/summarize/summ_utils.py
Niharika0306: Done.

spyre-rag/src/summarize/summ_utils.py
Niharika0306: done.

spyre-rag/src/summarize/summ_utils.py
Niharika0306: Addressed.

Niharika0306: Addressed.

spyre-rag/src/summarize/app.py
Niharika0306: done.

spyre-rag/src/summarize/app.py
Niharika0306: Addressed.

spyre-rag/src/summarize/app.py
Niharika0306: yeah, refactored a bit. Now there's only one instance of functions in case of both length and level.

spyre-rag/src/summarize/summ_utils.py
```diff
@@ -27,16 +40,180 @@
...
+        )
+        
     else:
-        target_word_count = max(1, int(input_word_count * settings.summarize.summarization_coefficient))
-
-    est_output_tokens = int(target_word_count / settings.common.llm.token_to_word_ratio_en)
-    max_tokens = est_output_tokens + settings.summarize.summarization_prompt_token_count
-    logger.debug(f"max tokens: {max_tokens}, estimated output tokens: {est_output_tokens}")
-    return target_word_count, max_tokens
+        # Automatic approach: use default coefficient
```
dharaneeshvrd, line 201: this block is not needed right? 
we are not using any of these numbers that are getting calculated.

spyre-rag/src/summarize/summ_utils.py
```diff
@@ -27,16 +40,180 @@
...
+            f"({min_words}-{max_words}), Max tokens: {max_tokens}, Available: {available_output_tokens}"
+        )
+        
+    elif summary_length is not None:
+        # Length-based approach: use specified word count
         target_word_count = summary_length
+        
+        # Calculate min/max bounds
+        min_words = int(target_word_count * 0.85)
+        max_words = int(target_word_count * 1.15)
```
dharaneeshvrd, line 177: Dont we need to do comparison similar to level here by calculating `max_possible_words` and choosing the minimum one?

Niharika0306: It wasn't straight-forward. In summary_level, we had to derive target_words from max_tokens and available tokens. 
But in summary_length case, importance was given to no.of.words , but however kept max_tokens in check to avoid llm errors.

Anyways, will update the code to abide for words as well.

spyre-rag/src/summarize/summ_utils.py
Niharika0306: max_tokens is still used. We pass it in llm_request. We let the llm decide the words. 

but yes, I'll remove the target_word_count, min_words, max_words calculation and make them None.

spyre-rag/src/summarize/summ_utils.py
Niharika0306: Done.

spyre-rag/src/summarize/summ_utils.py
Niharika0306: Addressed.

spyre-rag/src/summarize/app.py
Niharika0306: Tested this to max_limit with just few tokens to spare. The /tokenize response was successful.

spyre-rag/src/summarize/summ_utils.py
```diff
@@ -175,6 +176,11 @@ def compute_target_and_max_tokens(
         min_words = int(target_word_count * 0.85)
         max_words = int(target_word_count * 1.15)
         
+        # Cap target_word_count and max_words to available space
+        max_possible_words = int(available_output_tokens * settings.common.llm.token_to_word_ratio_en)
+        target_word_count = min(target_word_count, max_possible_words)
```
dharaneeshvrd, line 181: why are we altering the `target_word_count`?
It should be whatever user has passed.

Niharika0306: This was done as per your previous comment. That you want to check max_possible_words too

dharaneeshvrd: yes that is for max_words.
For target words, are we not validating that it would be within the available words earlier in the pipeline?

Niharika0306: Not really. we used to do before using MAX_INPUT_WORDS, but that was heuristic value based on a formula of constants.
The current logic takes in the tokens and derives an accurate value

dharaneeshvrd: Ok in that case, there is a chance that target and max words would hold same number. But that is ok I believe.

dharaneeshvrd (dismissed):
@manalilatkar 
Request you also to take a look and approve

dharaneeshvrd:
@Niharika0306 please rebase with main and bump the version

manalilatkar (dismissed):
LGTM.

dharaneeshvrd (approved):

PR #685, opened 2026-04-28
Fix OpenAI API compatibility for litellm

dharaneeshvrd:
Replace repetition_penalty with frequency_penalty
- The repetition_penalty parameter is not supported by OpenAI's Chat Completions API.
    Replaced it with frequency_penalty which is the correct OpenAI API parameter for
    controlling token repetition based on their frequency in the generated text.

### DIFF
--- a/ai-services/assets/applications/rag-cpu/podman/values.yaml
+++ b/ai-services/assets/applications/rag-cpu/podman/values.yaml
@@ -8,7 +8,7 @@ backend:
   # @description Host port for the OpenAI-compatible RAG service. Defaults to unexposed; assign a port to enable external access.
   port: "0"
   # @hidden
-  image: icr.io/ai-services-cicd/rag:v0.0.95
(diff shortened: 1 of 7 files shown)


Niharika0306:
It makes sense to remove the truncation logic considering the current design ( we wouldn't hit it )

Overall LGTM.
Just a question - do we still need the 'max_input_length' considering it was being used only in truncation logic.
If we want to use it in conversational RAG - I guess we need to update it to 32k

dharaneeshvrd:
yes, it can be removed and we need to make use of the `settings.common.llm.granite_3_3_8b_instruct_context_length` setting.
But with this we definitely need to make the tokenize api work with litellm, hence reverting the removal to get altered by conversation story anyway.

Niharika0306 (approved):

manalilatkar (approved):
LGTM

PR #691, opened 2026-04-30
fix: reorder bootstrap configuration steps

Shubhamag12:
Found issue while running `boostrap configure` on a fresh setup:

Supplementary groups are added after Spyre and Podman configuration. Unlike other Spyre checks which follow a check-and-repair pattern, this particular check is validate only, repair is not possible here because repair itself requires Podman.
Also in that validation, we verify if podman is installed or not, if not, we skip that check and then proceed to install podman, and then add supplementary group, refer: https://github.com/IBM/project-ai-services/blob/main/ai-services/internal/pkg/bootstrap/spyreconfig/spyre/spyre.go#L82
Since podman was already installed, it went ahead and started validating.

This PR fixes the order of execution of configuration steps.

### DIFF
--- a/ai-services/internal/pkg/bootstrap/podman/configure.go
+++ b/ai-services/internal/pkg/bootstrap/podman/configure.go
@@ -2,7 +2,6 @@ package podman
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"github.com/project-ai-services/ai-services/internal/pkg/bootstrap/spyreconfig/utils"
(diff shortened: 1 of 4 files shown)


mkumatag:
@yussufsh @manju956 lets review and merge this..

mkumatag (approved):
LGTM

mkumatag:
This looks clean to me, let us merge..

PR #697, opened 2026-05-04
feat: add support for local tokenization with downloaded models for podman runtime

dharaneeshvrd:
- Add new tokenizer_utils.py module for local tokenization functionality
- Update podman templates (chat-bot, digitize, vllm-server) to mount model cache volumes
- Configure TOKENIZER_MODEL_PATH environment variable in podman values.yaml files
- Integrate local tokenizer in chatbot and digitize applications
- Refactor llm_utils.py to use centralized tokenizer utilities
- Add tokenizer dependencies to Makefile
- Enable offline tokenization using locally downloaded models

This change allows podman deployments to perform tokenization locally
using downloaded models instead of relying on remote API calls,
improving performance and enabling offline operation.

### DIFF
--- a/ai-services/assets/applications/rag/podman/templates/chat-bot.yaml.tmpl
+++ b/ai-services/assets/applications/rag/podman/templates/chat-bot.yaml.tmpl
@@ -83,9 +83,15 @@ spec:
           value: "1"
         - name: LOG_LEVEL
           value: "{{ .Values.backend.log_level}}"
+        - name: TOKENIZER_MODEL_PATH
+          value: "/models/{{ .Values.instruct.model }}"
       ports:
         - containerPort: 5000
           protocol: TCP
+      volumeMounts:
+        - mountPath: /models/{{ .Values.instruct.model }}:z
+          name: chat-model
+          readOnly: true
       livenessProbe:
         httpGet:
           path: /health
@@ -99,3 +105,8 @@ spec:
           memory: "1Gi"
         limits:
           memory: "1Gi"
+  volumes:
+    - name: chat-model
+      hostPath:
+        path: "/var/lib/ai-services/models/{{ .Values.instruct.model }}"
+        type: Directory
--- a/ai-services/assets/applications/rag-cpu/podman/templates/chat-bot.yaml.tmpl
+++ b/ai-services/assets/applications/rag-cpu/podman/templates/chat-bot.yaml.tmpl
@@ -83,9 +83,15 @@ spec:
           value: "1"
         - name: LOG_LEVEL
           value: "{{ .Values.backend.log_level}}"
+        - name: TOKENIZER_MODEL_PATH
+          value: "/models/{{ .Values.instruct.model }}"
       ports:
         - containerPort: 5000
           protocol: TCP
+      volumeMounts:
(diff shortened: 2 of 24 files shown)


ai-services/assets/applications/rag/podman/templates/chat-bot.yaml.tmpl
```diff
@@ -83,9 +83,15 @@ spec:
           value: "1"
         - name: LOG_LEVEL
           value: "{{ .Values.backend.log_level}}"
+        - name: TOKENIZER_MODEL_PATH
+          value: "/models/{{ .Values.instruct.model }}"
       ports:
         - containerPort: 5000
           protocol: TCP
+      volumeMounts:
+        - mountPath: /models/{{ .Values.instruct.model }}:z
```
mkumatag, line 92: hop this will work when we have a different model for the Watsonx provider, you will not have that model downloaded locally!

mkumatag: cc @mayuka-c @yussufsh please keep a tap on the team how they are using these arguments/params

mkumatag (changes requested):

ai-services/assets/applications/rag/podman/templates/chat-bot.yaml.tmpl
dharaneeshvrd:
```
model:
  chat:
    # @hidden
    image: icr.io/ppc64le-oss/vllm-ppc64le:0.18.0
    # @description API key for vLLM instruct service authentication. If empty, authentication is disabled. Provide a key to enable authentication.
    apiKey: ""
    name: ibm-granite/granite-3.3-8b-instruct

  embedding:
    # @hidden
    image: icr.io/ppc64le-oss/vllm-ppc64le:0.18.0
    name: ibm-granite/granite-embedding-278m-multilingual

  reranker:
    # @hidden
    image: icr.io/ppc64le-oss/vllm-ppc64le:0.18.0
    name: BAAI/bge-reranker-v2-m3
```
If we alter the current LLM section to something similar to this, we can expect the user to pass the models that they want to use even in case of litellm with Watsonx and cli would download since it's based on the annotation and services can consume it. This is what I had in my mind, what do you think?

mayuka-c: We will not be downloading the models right if its Watsonx?. Only for vllm provider we would download since we are serving locally.

mkumatag: > Watsonx and cli would download since it's based on the annotation and services can consume it. This is what I had in my mind, what do you think?

It will be tricky considering the model used in the watsonx are not huggingface compatible names, download may not work out of the box.

dharaneeshvrd: I feel we need to ask the user to pass hf compatible model names only, since we have plans to accommodate other external providers like Watsonx, we need to have common standard like this.

dharaneeshvrd:
With custom callbacks and upstream litellm changes related to pass though endpoint's transformation, no need to do local tokenisation, closing this PR.

PR #701, opened 2026-05-05
Conversational RAG - Context Window logic 

sats-23:
-Implemented core sliding context window logic 
-There is a specific budget attached to history (2000 tokens)
-A sliding window includes latest messages upto 2000 tokens (rest discarded)
-Updated token calculation logic based on model context length 
-Implementing conversational mode through a feature flag (currently disabled)

Outcomes

Case 1: Normal conversation 
Input curl request & response: 
Screenshot 2026-05-05 at 12 29 38

Backend logs:
Screenshot 2026-05-05 at 12 28 13

Case 2: History exceeds 2000 tokens 
Screenshot 2026-05-05 at 14 10 26

### DIFF
--- a/spyre-rag/src/chatbot/app.py
+++ b/spyre-rag/src/chatbot/app.py
@@ -17,6 +17,8 @@ from lingua import Language
 
 from common.misc_utils import set_log_level
 from chatbot.settings import settings
+from chatbot.conversation_utils import get_conversation_context, truncate_history_by_tokens
+from chatbot.query_rephrasing import rephrase_query_with_context
 
 set_log_level(settings.common.app.log_level)
 
@@ -314,10 +316,10 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
     if not req.messages:
         APIError.raise_error(ErrorCode.EMPTY_INPUT, "messages can't be empty")
 
-    query = req.messages[0].content
+    current_query, previous_messages = get_conversation_context(req.messages)
 
     # Validate query is not empty
-    if not query or not query.strip():
+    if not current_query or not current_query.strip():
         APIError.raise_error(ErrorCode.EMPTY_INPUT, "Query cannot be empty")
 
     # Ensure vectorstore is initialized on first request
@@ -336,7 +338,7 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
 
         # Validate query length
         is_valid, error_msg = await asyncio.to_thread(
-            validate_query_length, query, emb_endpoint
+            validate_query_length, current_query, emb_endpoint
         )
         if not is_valid:
             # Return streaming error response for consistency
@@ -347,16 +349,37 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
                 return StreamingResponse(stream_query_length_error(), media_type="text/event-stream")
             APIError.raise_error(ErrorCode.INVALID_PARAMETER, error_msg)
 
-        lang = detect_language(query)
+        lang = detect_language(current_query)
 
         max_tokens = req.max_tokens
         # giving priority to max_tokens passed in the request, otherwise according to detected language of query
         if not max_tokens:
             max_tokens = max_tokens_map.get(lang, settings.common.llm.llm_max_tokens)
 
+        truncated_history = []
+        rephrased_query = current_query
+        
+        # Only process conversation history and rephrase query in conversational mode
+        # Conversational mode only works for English language
+        if settings.chatbot.conversational_mode and previous_messages and lang == "EN":
+            truncated_history = await asyncio.to_thread(
+                truncate_history_by_tokens,
+                previous_messages,
+                settings.chatbot.history_token_budget,
+                llm_endpoint
+            )
+            
+            if truncated_history:
+                rephrased_query = await rephrase_query_with_context(
+                    current_query=current_query,
+                    previous_messages=truncated_history,
+                    llm_endpoint=llm_endpoint,
+                    llm_model=llm_model,
+                )
+
         docs, perf_stat_dict = await asyncio.to_thread(
             search_only,
-            query,
+            rephrased_query,
             emb_model, emb_endpoint, emb_max_tokens,
             reranker_model,
             reranker_endpoint,
@@ -389,13 +412,37 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
         try:
             if req.stream:
                 vllm_stream = await asyncio.to_thread(
-                    query_vllm_stream, query, docs, llm_endpoint, llm_model, req.stop, max_tokens, req.temperature, perf_stat_dict, lang, api_key
+                    query_vllm_stream,
+                    current_query,
+                    docs,
+                    llm_endpoint,
+                    llm_model,
+                    req.stop,
+                    max_tokens,
+                    req.temperature,
+                    perf_stat_dict,
+                    lang,
+                    api_key,
+                    truncated_history,
+                    rephrased_query,
                 )
                 # For streaming, release is handled in locked_stream's finally block
                 return StreamingResponse(locked_stream(vllm_stream, perf_stat_dict), media_type="text/event-stream")
 
             vllm_non_stream = await asyncio.to_thread(
-                query_vllm_non_stream, query, docs, llm_endpoint, llm_model, req.stop, max_tokens, req.temperature, perf_stat_dict, lang, api_key
+                query_vllm_non_stream,
+                current_query,
+                docs,
+                llm_endpoint,
+                llm_model,
+                req.stop,
+                max_tokens,
+                req.temperature,
+                perf_stat_dict,
+                lang,
+                api_key,
+                truncated_history,
+                rephrased_query,
             )
             # Store metrics in registry for non-stream
             perf_registry.add_metric(perf_stat_dict)
--- a/spyre-rag/src/chatbot/app.py
+++ b/spyre-rag/src/chatbot/app.py
@@ -151,6 +151,24 @@ def limit_concurrency(f):
             concurrency_limiter.release()
     return wrapper
 
+def get_stop_words_with_special_tokens(request_stop_words):
+    """
+    Add common special tokens to stop words to prevent them from appearing in responses.
+    
+    Args:
+        request_stop_words: Stop words from the request (can be None, list, or string)
+    
+    Returns:
+        List of stop words including special tokens
+    """
+    stop_words = list(request_stop_words) if request_stop_words else []
+    # Add common special tokens that should stop generation
+    special_tokens = ["[/assistant]", "</s>", "<|endoftext|>", "<|im_end|>"]
+    for token in special_tokens:
+        if token not in stop_words:
+            stop_words.append(token)
+    return stop_words
+
 async def is_auth_required() -> bool:
     """
     Check if vLLM authentication is required and cache the result.
@@ -423,6 +441,8 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
         await concurrency_limiter.acquire()
 
         try:
+            stop_words = get_stop_words_with_special_tokens(req.stop)
+            
             if req.stream:
                 vllm_stream = await asyncio.to_thread(
                     query_vllm_stream,
@@ -430,7 +450,7 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
                     docs,
                     llm_endpoint,
                     llm_model,
-                    req.stop,
+                    stop_words,
                     max_tokens,
                     req.temperature,
                     perf_stat_dict,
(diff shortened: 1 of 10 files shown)


spyre-rag/src/chatbot/conversation_utils.py
```diff
@@ -0,0 +1,91 @@
...
+def get_conversation_context(messages: Sequence[Any]) -> tuple[str, list[dict[str, str]]]:
+    """
+    Extract current query and conversation history from a messages array.
+
+    Accepts either dict-based messages or Pydantic message objects.
+    """
+    if not messages:
+        return "", []
+
+    normalized_messages = [_message_to_dict(message) for message in messages]
```
dharaneeshvrd, line 33: Hope this method is utilised only for query rewriting, if yes, I dont think we need to pass all the questions, we can have a setting for this and start with last 3 messages with assistant response IMO. Later if we need we can alter it to have more turns included.

spyre-rag/src/chatbot/settings.py
```diff
@@ -102,7 +102,47 @@ class RAGConfig(BaseSettings):
...
+            "Rephrased Query: {rephrased_query}\n\n"
+            "Instructions: Answer the user's question based on the retrieved context above. "
+            "Consider the conversation history to provide contextually relevant responses. "
+            "Be conversational and reference previous exchanges when relevant. "
+            "If the context doesn't contain enough information, acknowledge this clearly."
+        ),
+        description="RAG system prompt template with context and rephrased query",
+    )
+
+    history_token_budget: int = Field(
```
dharaneeshvrd, line 133: why 2000? 
we discussed it to have the context length of the model `settings.common.llm.granite_3_3_8b_instruct_context_length` right?

spyre-rag/src/chatbot/app.py
```diff
@@ -347,16 +349,37 @@ async def stream_query_length_error():
...
         if not max_tokens:
             max_tokens = max_tokens_map.get(lang, settings.common.llm.llm_max_tokens)
 
+        truncated_history = []
+        rephrased_query = current_query
+        
+        # Only process conversation history and rephrase query in conversational mode
+        # Conversational mode only works for English language
+        if settings.chatbot.conversational_mode and previous_messages and lang == "EN":
+            truncated_history = await asyncio.to_thread(
```
dharaneeshvrd: I think here you can pass the number of turns maybe.
But I see this same history is passed to query_vllm_stream as well, but for the final answer the turns can be 5. 
Btw I dont see the use of 5 turns like we discussed on proposal.

spyre-rag/src/common/llm_utils.py
```diff
@@ -162,28 +162,108 @@ def query_vllm_models(llm_endpoint, api_key: str | None = None):
...
+
+    # Use conversational mode if enabled AND language is English, otherwise use legacy prompts
+    if chatbot_settings.chatbot.conversational_mode and lang == "EN":
+        # Conversational RAG mode with message history
+        question_token_count = len(tokenize_with_llm(question, llm_endpoint))
+
+        history_tokens = 0
+        if previous_messages:
+            for msg in previous_messages:
+                history_tokens += len(tokenize_with_llm(msg.get("content", ""), llm_endpoint))
```
dharaneeshvrd: I think some part of the calculation already done in truncation.
Can you please see how you can integrate or have a common method that does this only once?

spyre-rag/src/common/llm_utils.py
```diff
@@ -162,28 +162,108 @@ def query_vllm_models(llm_endpoint, api_key: str | None = None):
...
+        # Dynamic chunk truncation: truncates the context if it doesn't fit in the sequence length
+        question_token_count = len(tokenize_with_llm(question, llm_endpoint))
+        remaining_tokens = settings.llm.granite_3_3_8b_instruct_context_length - (
+            chatbot_settings.chatbot.prompt_template_token_count +
+            question_token_count +
+            max_new_tokens  # Reserve space for model's response
+        )
+        remaining_tokens = max(0, remaining_tokens)
+        
+        context = detokenize_with_llm(
```
dharaneeshvrd, line 273: I think we should not cut down the current query's search result, we should truncate the earlier conversation 
Please take care

spyre-rag/src/chatbot/settings.py
sats-23: 2000 is the budget only for History QnA. 
The remaining tokens is still calculated based on settings.common.llm.granite_3_3_8b_instruct_context_length

We have a much bigger budget than 2000, but more context of previous QnA also caused more hallucination.

spyre-rag/src/chatbot/app.py
sats-23: Instead of turns, I have implemented based on the token budget. 
By keeping it based on number of turns, the token budget was more variable, but in case truncating using tokens, it made it more predictable and fixed. 

Let me know in case we have to take the number of turns approach

spyre-rag/src/common/llm_utils.py
sats-23: It was an extract of the existing logic, I shall redo it to prioritize more of current context over conversation history

spyre-rag/src/chatbot/settings.py
dharaneeshvrd: Ok let's go with 2000 for now. But I would like to have a lesser token budget(1000 to get started with) for query rewriting, no need to pass the history that we are going to pass for final answer.

spyre-rag/src/chatbot/app.py
dharaneeshvrd: 2000 tokens will cover 5 turns approximately, lets go ahead with this for now.

spyre-rag/src/common/llm_utils.py
sats-23: Done

spyre-rag/src/chatbot/settings.py
sats-23: Done

spyre-rag/src/common/llm_utils.py
sats-23: Done

spyre-rag/src/chatbot/conversation_utils.py
sats-23: It's used both for rephrasing and query LLM. 
Called once, but used by both. Contains entire history, then truncated by purpose

spyre-rag/src/chatbot/app.py
```diff
@@ -149,6 +151,24 @@ async def wrapper(*args, **kwargs):
             concurrency_limiter.release()
     return wrapper
 
+def get_stop_words_with_special_tokens(request_stop_words):
```
dharaneeshvrd, line 154: Any reason to add this part of this PR?

spyre-rag/src/common/llm_utils.py
```diff
@@ -162,28 +162,116 @@ def query_vllm_models(llm_endpoint, api_key: str | None = None):
...
+        logger.debug(f"Truncated Context: {context}")
+
+        message_array = [
+            {
+                "role": "system",
+                "content": chatbot_settings.chatbot.initial_system_message,
+            }
+        ]
+
+        if previous_messages and history_budget > 0:
```
dharaneeshvrd, line 222: Dont we need to truncate the messages based on the calculated budget?
Adding all is not making sense to me

spyre-rag/src/common/llm_utils.py
```diff
@@ -162,28 +162,116 @@ def query_vllm_models(llm_endpoint, api_key: str | None = None):
...
+    llm_model,
+    stop_words,
+    max_new_tokens,
+    temperature,
+    stream,
+    lang,
+    api_key: str | None = None,
+    previous_messages: list | None = None,
+    rephrased_query: str | None = None,
+    history_token_count: int = 0,
```
dharaneeshvrd: I see it's used only in logging. 
Dont we need to make use of it?

spyre-rag/src/chatbot/app.py
sats-23: Yes, earlier when we were in chat mode, we have fixed user prompt and assistant response.
Now with multiple system prompts in conversational mode, we are getting stop words like "[/assistant]" at the end of the response. 
This logic helps avoid that

spyre-rag/src/common/llm_utils.py
sats-23: Yes, the idea dropped when we prioritized context over history here - https://github.com/IBM/project-ai-services/pull/701/changes/84b303a11ef4bc86e532ea93fb9dbe8e8adf268e#diff-e563696ebeb6b9100f62825a56b589fdfe623b2b0a695a54996a226ac3334495L193

I shall remove it

sats-23: Removed history_token_count as it is irrelevant.
It is done differently for query rephrasing (1000) and differently for query LLM (min (2000, remaining budget))
So scraped the metric

spyre-rag/src/common/llm_utils.py
sats-23: Sorry I missed it, handled now

spyre-rag/src/chatbot/app.py
```diff
@@ -347,23 +367,56 @@ async def stream_query_length_error():
...
+        if settings.chatbot.conversational_mode and previous_messages and lang == "EN":
+            # Pre-truncate history with max budget
+            truncated_history = await asyncio.to_thread(
+                truncate_history_by_tokens,
+                previous_messages,
+                settings.chatbot.history_token_budget,
+                llm_endpoint
+            )
 
         if not docs:
```
dharaneeshvrd, line 410: Let's move this docs validation before history truncation.

dharaneeshvrd (commented):
Except one last comment, remaining things looks good to me!

spyre-rag/src/chatbot/app.py
sats-23: Done, PTAL

dharaneeshvrd (approved):

dharaneeshvrd:
@Niharika0306 @manalilatkar 
Can you please review this PR?
Will wait for your approval

manju956 (approved):
looks good

manalilatkar (commented):
LGTM. @sats-23 could you add some unit tests to validate this code?

sats-23:
> LGTM. @sats-23 could you add some unit tests to validate this code?

@manalilatkar https://jsw.ibm.com/browse/AISERVICES-1047

manalilatkar:
> > LGTM. @sats-23 could you add some unit tests to validate this code?
> 
> @manalilatkar https://jsw.ibm.com/browse/AISERVICES-1047

ohh ok, you created a separate story. I think as a practice, unit tests should go in with the related code. But if separate stories are what we wanna do for now, then @dharaneeshvrd I guess we can merge this.

manalilatkar (approved):

dharaneeshvrd:
I think it's ok since UT subtask is part of the same story.

sats-23:
UT will be added before the feature flag (conversational_mode) is [flipped](https://jsw.ibm.com/browse/AISERVICES-1017)

spyre-rag/src/chatbot/settings.py
Niharika0306: why are we having different values of history_token_budget in rephrase and llm_conversation calls? If the content of history is same in both the cases.

spyre-rag/src/chatbot/app.py
```diff
@@ -386,16 +428,53 @@ async def stream_server_busy():
...
+        # Calculate history budget after getting docs (context is prioritized)
+        # History truncation will happen inside query_vllm_payload based on actual context size
+        if settings.chatbot.conversational_mode and previous_messages and lang == "EN":
+            # Pre-truncate history with max budget
+            truncated_history = await asyncio.to_thread(
+                truncate_history_by_tokens,
+                previous_messages,
+                settings.chatbot.history_token_budget,
+                llm_endpoint
+            )
```
Niharika0306, line 440: Why are we truncating here, if we already having a better logic done for this inside the query_vllm_payload function ? Won't it be redundant and not used

Niharika0306:
Just a small nit from the query rephrasing logic. ( correct me if i'm wrong)
We are currently setting max_tokens to 100 in the rephrase_query_llm call. Now, the max_input_query_length is set to 512. 
What if there are scenarios, when history is present, and user asks an unrelated and big question ( lets say around 400 tokens) - In this case, we are forcing it to shorten the query to less than 100 tokens. and probably lose information.

spyre-rag/src/chatbot/app.py
sats-23: Makes sense, must've missed it as part of the rework, I shall take it in a follow up PR

dharaneeshvrd:
> Just a small nit from the query rephrasing logic. ( correct me if i'm wrong)
We are currently setting max_tokens to 100 in the rephrase_query_llm call. Now, the max_input_query_length is set to 512.
What if there are scenarios, when history is present, and user asks an unrelated and big question ( lets say around 400 tokens) - In this case, we are forcing it to shorten the query to less than 100 tokens. and probably lose information.

I think this is a valid concern. We need to address this.

spyre-rag/src/chatbot/settings.py
dharaneeshvrd: It seems no need to pass the full history for rewriting the query. You would get the relevant information in last couple of turns only. Hence reduced the token for rewriting the query.

PR #704, opened 2026-05-05
[WIP] Add TLS/HTTPS support for externally exposed services

shwetha-s-poojary:
This PR adds HTTPS (secure connections) support for external services using a tool called Caddy, which works like a traffic manager between users and your services. It currently shows this setup using the catalog service, but it can be reused for other services too.

- Automatically enables HTTPS for services (no manual setup needed)
- Works in local development using fake/test domains (nip.io)
- Supports real production domains using --domain-name
- Lets users provide their own SSL certificates if they already have them
- Automatically figures out the correct domain from certificates (even for wildcard domains like *.example.com)
- If a certificate already defines a domain, it takes priority to avoid conflicts
- Designed so more services can easily be added later without changing the core setup

Note: Using docker image now.Need to build UBI based image.

### DIFF
--- a/ai-services/assets/catalog/podman/values.yaml
+++ b/ai-services/assets/catalog/podman/values.yaml
@@ -22,3 +22,11 @@ db:
       memory: "512Mi"
     limits:
       memory: "1Gi"
+
+caddy:
+  image: "docker.io/library/caddy:2.8-alpine"
+  httpsPort: 443
+  serverName: "ai_services_server"
+  dataPath: "/var/lib/ai-services/catalog/caddy"
+  domain: ""  # Will be set by configure command
+  externalPort: 443  # Will be set by configure command
--- a/ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
+++ b/ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
@@ -30,7 +30,7 @@ const (
 )
 
 // DeployCatalog deploys the catalog service using the assets/catalog template for podman runtime.
-func DeployCatalog(ctx context.Context, podmanURI, passwordHash string, argParams map[string]string) error {
+func DeployCatalog(ctx context.Context, podmanURI, passwordHash string, argParams map[string]string, domainName string) error {
 	s := spinner.New("Deploying catalog service...")
 	s.Start(ctx)
 
@@ -62,7 +62,7 @@ func DeployCatalog(ctx context.Context, podmanURI, passwordHash string, argParam
 	}
 
 	// Setup HTTPS configuration
-	httpsConfig, err := setupHTTPSConfig(s)
+	httpsConfig, err := setupHTTPSConfig(domainName, s)
 	if err != nil {
 		s.Fail("failed to setup HTTPS configuration")
 		return fmt.Errorf("failed to setup HTTPS configuration: %w", err)
@@ -104,7 +104,7 @@ type HTTPSConfig struct {
 }
 
 // setupHTTPSConfig prepares HTTPS configuration
-func setupHTTPSConfig(s *spinner.Spinner) (*HTTPSConfig, error) {
+func setupHTTPSConfig(domainName string, s *spinner.Spinner) (*HTTPSConfig, error) {
 	config := &HTTPSConfig{}
 
 	// Get host IP
@@ -114,9 +114,16 @@ func setupHTTPSConfig(s *spinner.Spinner) (*HTTPSConfig, error) {
 	}
 	config.HostIP = hostIP
 
-	// Use nip.io with host IP
-	config.Domain = fmt.Sprintf("%s.nip.io", hostIP)
-	logger.Infof("Using nip.io domain: %s", config.Domain)
+	// Determine domain name
+	if domainName != "" {
+		// Use custom domain name
+		config.Domain = domainName
+		logger.Infof("Using custom domain: %s", domainName)
+	} else {
+		// Use nip.io with host IP
+		config.Domain = fmt.Sprintf("%s.nip.io", hostIP)
+		logger.Infof("Using nip.io domain: %s", config.Domain)
+	}
 
 	// Create Caddy configuration directory and Caddyfile
 	if err := createCaddyConfig(config); err != nil {
@@ -205,6 +212,13 @@ func printNextSteps(tp templates.Template, rt *podman.PodmanClient, config *HTTP
 	logger.Infoln("")
 	logger.Infof("   Note: Using self-signed certificate for %s.\n", config.Domain)
 	logger.Infof("         Your browser may show a security warning.\n")
+
+	// Show DNS configuration note for custom domains
+	if config.Domain != fmt.Sprintf("%s.nip.io", config.HostIP) {
+		logger.Infoln("")
+		logger.Infof("   IMPORTANT: Ensure DNS is configured to point %s to %s\n", config.Domain, config.HostIP)
+	}
+
 	logger.Infoln("")
 	logger.Infoln("━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━")
 
--- a/ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
+++ b/ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
@@ -6,6 +6,8 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"sync"
 	"text/template"
 
@@ -14,6 +16,7 @@ import (
 	clipodman "github.com/project-ai-services/ai-services/internal/pkg/cli/podman"
 	"github.com/project-ai-services/ai-services/internal/pkg/cli/templates"
 	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+	"github.com/project-ai-services/ai-services/internal/pkg/proxy"
 	"github.com/project-ai-services/ai-services/internal/pkg/runtime/podman"
 	"github.com/project-ai-services/ai-services/internal/pkg/specs"
 	"github.com/project-ai-services/ai-services/internal/pkg/spinner"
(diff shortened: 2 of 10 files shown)


ai-services/assets/catalog/podman/values.yaml
```diff
@@ -22,3 +22,11 @@ db:
       memory: "512Mi"
     limits:
       memory: "1Gi"
+
+caddy:
+  image: "docker.io/library/caddy:2.8-alpine"
```
mkumatag, line 27: don't do this, please create a caddy image under images..

mkumatag: @adarshagrawal38 we need to add one more image part of the release

ai-services/assets/catalog/podman/values.yaml
```diff
@@ -22,3 +22,11 @@ db:
...
+  httpsPort: 443
+  serverName: "ai_services_server"
+  dataPath: "/var/lib/ai-services/catalog/caddy"
```
mkumatag, line 30: TODO: again, this needs to be parameterised

mkumatag (changes requested):
Initial high level comments

ai-services/assets/catalog/podman/values.yaml
shwetha-s-poojary: @mkumatag Yes, we have a separate story for it. We’ll take care of this.

ai-services/assets/catalog/podman/values.yaml
shwetha-s-poojary: you mean the port and serverName  right?

yussufsh: @shwetha-s-poojary `/var/lib/ai-services/` can be customized.

shwetha-s-poojary: @yussufsh yes that will be handled by this PR #670 , will have to rebase and resolve once that is merged.

Are you suggesting some other way?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -75,17 +81,198 @@ func DeployCatalog(ctx context.Context, podmanURI, passwordHash string, argParam
...
+		keyDestPath := filepath.Join(caddyDataDir, "key.pem")
+
+		// Read and write certificate
+		certData, err := os.ReadFile(config.UserCertPath)
+		if err != nil {
+			return fmt.Errorf("failed to read certificate file: %w", err)
+		}
+		if err := os.WriteFile(certDestPath, certData, 0644); err != nil {
+			return fmt.Errorf("failed to write certificate to Caddy directory: %w", err)
+		}
```
mayuka-c, line 168: Any reason for copying user provided cert and key to CaddyDataDir?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -75,17 +81,198 @@ func DeployCatalog(ctx context.Context, podmanURI, passwordHash string, argParam
...
+		return fmt.Errorf("failed to write Caddyfile: %w", err)
+	}
+
+	logger.Infof("Created Caddyfile at %s", caddyfilePath)
+	return nil
+}
+
+// registerCatalogRoutes registers catalog service routes with Caddy
+func registerCatalogRoutes(rt *podman.PodmanClient, config *HTTPSConfig) error {
+	logger.Infof("Registering catalog routes with Caddy...")
```
mayuka-c, line 215: For registeringCatalogRoutes, we might need to make it more generic.
Like based on the annotation ai-services.io/ports, we need to decide and dynamically compute.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -75,17 +81,198 @@ func DeployCatalog(ctx context.Context, podmanURI, passwordHash string, argParam
...
+	catalogUIRoute := proxy.Route{
+		ID:              "route-ai-services--catalog-ui",
+		Host:            utils.ConstructServiceDomain(catalogAppName, "catalog", config.Domain),
+		UpstreamAddress: fmt.Sprintf("%s--catalog:8081", catalogAppName),
+		Terminal:        true,
+	}
+
+	if err := caddyClient.RegisterRoute(catalogUIRoute); err != nil {
+		return fmt.Errorf("failed to register catalog UI route: %w", err)
+	}
```
mayuka-c, line 233: Everything should be rendered dynamically from podspec

mayuka-c (changes requested):
@shwetha-s-poojary 
Can we take up one PR at a time like first PR for Catalog assets and write to CaddyFile in catalog configure. Then next to Register routes with Internal TLS. Then one more to read from external certs. Then integrate it per application and then next steps

ai-services/assets/catalog/podman/values.yaml
yussufsh: Yes, suggesting the same. It's a TODO 😄

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -75,17 +81,198 @@ func DeployCatalog(ctx context.Context, podmanURI, passwordHash string, argParam
...
+
+	// Wait for Caddy to be ready
+	caddyClient := proxy.NewCaddyClient()
+	if err := caddyClient.HealthCheck(); err != nil {
+		return fmt.Errorf("caddy is not available: %w", err)
+	}
+
+	// Register route for catalog UI
+	catalogUIRoute := proxy.Route{
+		ID:              "route-ai-services--catalog-ui",
```
mkumatag, line 225: is this a static text, this also you can derive from other fields

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -130,6 +315,11 @@ func prepareCatalogValues(tp templates.Template, podmanURI, passwordHash string,
 	argParams["backend.podman.uri"] = podmanURI
 	argParams["db.password"] = dbPasswordBase64
 
+	// Set Caddy configuration
+	argParams["caddy.httpsPort"] = "443"
```
mkumatag, line 319: we should to customise this port, e.g: what if this port is not available..?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -130,6 +315,11 @@ func prepareCatalogValues(tp templates.Template, podmanURI, passwordHash string,
...
+	argParams["caddy.domain"] = httpsConfig.Domain
+	argParams["caddy.externalPort"] = "443"
```
mkumatag, line 321: same for this as well, both are 443, what is the difference between these two?

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+)
+
+const (
+	// DefaultAdminURL is the default Caddy Admin API URL (localhost only for security)
+	DefaultAdminURL = "http://127.0.0.1:2019"
+	// DefaultServerName is the default server name in Caddy config
+	DefaultServerName = "ai_services_server"
+	// DefaultRequestTimeout is the default HTTP request timeout
+	DefaultRequestTimeout = 30 * time.Second
```
mkumatag, line 21: let us not expose these and make them local if not needed anywhere else

ai-services/internal/pkg/proxy/interface.go
mkumatag, line 1: I liked this interface way of doing it  ❤️

ai-services/internal/pkg/proxy/types.go
```diff
@@ -0,0 +1,36 @@
...
+
+// TLSConfig represents TLS configuration
+type TLSConfig struct {
+	CertPath string // Path to certificate file
+	KeyPath  string // Path to private key file
+	Domain   string // Domain name extracted from certificate
+}
+
+// CaddyConfig represents Caddy server configuration
+type CaddyConfig struct {
```
mkumatag, line 21: this can be moved to caddy.go file

mkumatag (changes requested):
high level review comments

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
shwetha-s-poojary: Idea was to have cert, key and Caddyfile all at one place so that I can have single volume mount.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
shwetha-s-poojary: Yes right.Working on it.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
shwetha-s-poojary: Yes right.working on it

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
shwetha-s-poojary: Yes that's one of item on the list

```
--external-port <port>     External HTTPS port for Caddy server (optional, default: 443)
                             If not 443, port will be included in service URLs
                             Example: --external-port 8443 results in https://service.<domain>:8443
```

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
shwetha-s-poojary: `argParams["caddy.httpsPort"] = "443`" is the container port / the port Caddy listens on inside the container (I think I could've either directly specified in the yaml.tmpl or named it differently)

`argParams["caddy.externalPort"] = "443"` is the port exposed on the host machine

shwetha-s-poojary:
> @shwetha-s-poojary Can we take up one PR at a time like first PR for Catalog assets and write to CaddyFile in catalog configure. Then next to Register routes with Internal TLS. Then one more to read from external certs. Then integrate it per application and then next steps

Sure.That's makes sense , it would be easier to review as well. Will do

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: Sure.Thanks for catching that.

PR #706, opened 2026-05-05
feat: adding db handlers

Shubhamag12:
Adding DB handlers as per Catalog DB proposal


Shubhamag12:
@yussufsh @mayuka-c I have added test files for now as discussed, let me know if we want to remove it and take it as a part of future PRs.

yussufsh (approved):

PR #707, opened 2026-05-05
feat: add connection pool for db

Shubhamag12:
Adding connection to db using pgxpool for application queries

### DIFF
--- a/ai-services/internal/pkg/catalog/db/db.go
+++ b/ai-services/internal/pkg/catalog/db/db.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/jackc/pgx/v5/pgxpool"
 	_ "github.com/jackc/pgx/v5/stdlib" // PostgreSQL driver
 	"github.com/project-ai-services/ai-services/internal/pkg/logger"
 )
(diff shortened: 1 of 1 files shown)


ai-services/internal/pkg/catalog/db/db.go
```diff
@@ -73,6 +82,36 @@ func Connect(cfg Config) (*sql.DB, error) {
 	return db, nil
 }
 
+// ConnectPool establishes a connection pool to the PostgreSQL database using pgxpool.
+func ConnectPool(ctx context.Context, cfg Config) (*pgxpool.Pool, error) {
+	poolConfig, err := pgxpool.ParseConfig(cfg.ConnectionURL())
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse pool config: %w", err)
```
mayuka-c, line 90: Can we have the config already populated?. Reading the cfg for each ConnectPool doesnt work well. There is a golang pkg: envconfig, I think we need to utilize it

Shubhamag12: As discussed, not picking it up

ai-services/internal/pkg/catalog/db/db.go
```diff
@@ -46,7 +48,17 @@ func (c *Config) ConnectionString() string {
...
 }
 
-// Connect establishes a connection to the PostgreSQL database.
+// ConnectionURL builds a PostgreSQL connection URL from the config (for pgxpool).
+func (c *Config) ConnectionURL() string {
+	encodedPassword := url.QueryEscape(c.Password)
+
+	return fmt.Sprintf(
+		"postgres://%s:%s@%s:%d/%s?sslmode=%s",
+		c.User, encodedPassword, c.Host, c.Port, c.DBName, c.SSLMode,
```
mayuka-c: Instead of declaring a new variable for encodedPassword, directly call url.QueryEscape here it self

Shubhamag12: done

mayuka-c (approved):
LGTM

yussufsh (approved):

PR #708, opened 2026-05-05
AISERVICES-996 detecting langauge before passing text to sentence

manalilatkar:
splitter while digitizing

### DIFF
--- a/services/digitize/doc_utils.py
+++ b/services/digitize/doc_utils.py
@@ -781,10 +781,8 @@ def detect_document_language(data: list) -> str:
             return 'en'
         
         for block_text in sampled_blocks:
-            # For longer blocks, truncate 
-            max_size = min(500, len(block_text)-1)
             # Truncate to 500 characters
-            chunk = block_text[:max_size]
+            chunk = block_text[:500]
             
             # Detect language for this chunk
             if chunk.strip():
@@ -824,7 +822,7 @@ def split_text_into_token_chunks(text, emb_endpoint, max_tokens=512, overlap=50,
     Returns:
         List of text chunks
     """
-    logger.info(f"Using language for chunking: {language}")
+    logger.debug(f"Using language for chunking: {language}")
     
     sentences = SentenceSplitter(language=language).split(text)
     chunks = []
--- a/services/digitize/doc_utils.py
+++ b/services/digitize/doc_utils.py
@@ -719,17 +719,31 @@ def count_tokens(text, emb_endpoint):
     return token_len
 
 
-def detect_document_language(text: str) -> str:
+def detect_document_language(data: list) -> str:
     """
-    Detect the language of a document by sampling random chunks.
+    Detect the language of a document by sampling random blocks.
     
     Args:
-        text: The full text to analyze
+        data: List of document blocks, where each block is a dict with a 'text' field
         
     Returns:
         Language code compatible with SentenceSplitter ('en', 'de', 'it', 'fr')
         Falls back to 'en' if detection fails or language is not supported
     """
+    # validate input data structure
+    if not isinstance(data, list):
+        logger.warning(f"Invalid input: expected list, got {type(data).__name__}, falling back to 'en'")
+        return 'en'
+    
+    if not data:
+        logger.warning("Empty data list provided for language detection, falling back to 'en'")
+        return 'en'
+    
+    # Validate that data contains dicts with 'text' fields
+    if not all(isinstance(block, dict) for block in data):
+        logger.warning("Invalid input: data list contains non-dict elements, falling back to 'en'")
+        return 'en'
+    
     # Mapping from lingua ISO codes to SentenceSplitter language codes
     lang_map = {
         'EN': 'en',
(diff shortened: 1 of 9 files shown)


spyre-rag/src/digitize/doc_utils.py
```diff
@@ -711,8 +714,88 @@ def count_tokens(text, emb_endpoint):
...
+        overlap: Number of tokens to overlap between chunks
+        language: Optional language code ('en', 'de', 'it', 'fr'). If None, auto-detect.
+        
+    Returns:
+        List of text chunks
+    """
+    # Detect language if not provided
+    # keeping the parameter option in case we implement language selection in the UI later
+    if language is None:
+        language = detect_document_language(text)
```
dharaneeshvrd: If we run detect lang for every para of the file, it would slow down the pipeline.
Let's do the language detection earlier in the pipeline and use it.

spyre-rag/src/digitize/doc_utils.py
```diff
@@ -870,6 +864,11 @@ def chunk_text(input_path, out_path, emb_endpoint, max_tokens=512, doc_id=None):
         with open(input_path, "r") as f:
             data = json.load(f)
 
+            # Detect document language once by collecting all text content
+            all_text = " ".join([block.get("text", "") for block in data if block.get("text")])
```
dharaneeshvrd: Can we apply the random chunk selection logic here itself and pass only 3 chunk to `detect_document_language`?
I want to avoid building the all_text and send only the randomly selected texts.
I hope we can do `chunk = text[start_pos:start_pos + chunk_size]` on data

dharaneeshvrd: alternatively you can send the data as is and do the selection without building the list contains all texts inside `detect_document_language`.

manalilatkar: @dharaneeshvrd do you mean randomly selecting blocks from the data variable and directly sending the text from each block to `detect_document_language`?

dharaneeshvrd: Something like this or you can do selection inside the `detect_document_language` by sending data directly.

spyre-rag/src/digitize/doc_utils.py
```diff
@@ -734,29 +748,39 @@ def detect_document_language(text: str) -> str:
...
-            # Random chunk size between 200-500 characters
-            chunk_size = random.randint(200, min(500, text_length))
-            
-            # Random start position ensuring we don't go out of bounds
-            max_start = max(0, text_length - chunk_size)
-            start_pos = random.randint(0, max_start) if max_start > 0 else 0
-            
-            chunk = text[start_pos:start_pos + chunk_size]
+        # Randomly select blocks
+        sampled_blocks = random.sample(text_blocks, num_samples)
```
dharaneeshvrd:
```
        # Generate random indices and collect valid text blocks without looping through all data
        detected_languages = []
        sampled_blocks = []
        attempted_indices = set()
        max_attempts = min(len(data), 50)  # Limit attempts to avoid infinite loop
        
        # Keep trying random indices until we get 3 valid blocks or exhaust attempts
        while len(sampled_blocks) < 3 and len(attempted_indices) < max_attempts:
            # Generate a random index
            idx = random.randint(0, len(data) - 1)
            
            # Skip if already tried this index
            if idx in attempted_indices:
                continue
            
            attempted_indices.add(idx)
            
            # Check if this block has valid text
            block = data[idx]
            if isinstance(block.get("text"), str) and block.get("text", "").strip():
                sampled_blocks.append(block.get("text", ""))
        
        if not sampled_blocks:
            logger.warning("No text blocks found for language detection, falling back to 'en'")
            return 'en'
```

this is what I thought of
randomly generate the index and retrieve the data by indexing till we get 3 valid txt 
ptal

spyre-rag/src/digitize/doc_utils.py
```diff
@@ -734,29 +748,39 @@ def detect_document_language(text: str) -> str:
...
+        
+        for block_text in sampled_blocks:
+            # For longer blocks, take a random chunk
+            if len(block_text) > 500:
+                chunk_size = random.randint(200, 500)
```
manalilatkar: @dharaneeshvrd should I remove this part also? Send the whole block for detection regardless of how long the block is?

dharaneeshvrd: if the language can be detected with less chars, then this can be trimmed too.

spyre-rag/src/digitize/doc_utils.py
manalilatkar: @dharaneeshvrd why do you want this complicated function when the following code does it better in a simpler manner:
`sampled_blocks = random.sample(text_blocks, num_samples)`

dharaneeshvrd: I want to avoid building this text_blocks.
https://github.com/IBM/project-ai-services/pull/708/changes/b40bd06f91db849698e28961135f806d4e49a5bd#r3247960754
by looping over the entire content of the given file. Currently we are accepting the pdf however big it is, so imagine how many text blocks a pdf might have.
Whereas data is also a list and we can select the text randomly by generating an index within length of the data.

services/digitize/doc_utils.py
```diff
@@ -777,10 +777,8 @@ def detect_document_language(data: list) -> str:
             return 'en'
         
         for block_text in sampled_blocks:
-            # For longer blocks, truncate 
-            max_size = min(500, len(block_text)-1)
             # Truncate to 500 characters
-            chunk = block_text[:max_size]
+            chunk = block_text[:500]
```
dharaneeshvrd, line 785: I think max_size is required, since we might have texts smaller than 500 chars.

manalilatkar: As discussed, this wouldn't fail.

dharaneeshvrd:
lgtm
let's wait for https://github.com/IBM/project-ai-services/pull/710 this to get closed.

manalilatkar:
@dharaneeshvrd tested this Pr with rebased code, everything works. let me know if anythign else is needed.

dharaneeshvrd (approved):

PR #709, opened 2026-05-05
Token Management in Catalog

mayuka-c:
- DB Implementation to store blacklisted tokens (access & refresh).
- Take Refresh token as header as part of /logout api.
- Blacklist also the refresh token as part of /logout api.
- During /refresh, blacklist the passed refresh token.
- Also set default ttl of refresh token to 24 hr

Note: This removes our existing InMemory way of storing blacklist tokens.

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=0.0.3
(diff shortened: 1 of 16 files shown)


mayuka-c:
Validation of blacklisted tokens stored in DB


image

---

Using the blaclisted tokens

image

yussufsh (dismissed):

yussufsh (approved):

PR #710, opened 2026-05-05
Code reorg implementation

Bleama02:
-Migrate the backend service code and the UI services code to reflect the changes required by https://github.com/IBM/project-ai-services/blob/main/docs/proposals/code_reorg_proposal.md


vijay-vin:
Catalog UI, Digitize UI and Spyre RAG UI scans are failing because of the refactoring. CI/CD source code will be modified to match the new path, once this PR is approved and merged.

AI Services and Python Base cves need updated packages like Pillow, Moby etc. Please take a look at them as they may require code changes to pull in the new versions per cves.

dharaneeshvrd (commented):
Generated this with bob based on my analysis:


## Review Feedback: Incremental Migration Approach Recommended

Thank you for this comprehensive reorganization proposal. The target structure aligns well with microservices best practices. However, I recommend breaking this into smaller, incremental PRs to:

1. **Reduce risk** - Easier to identify and fix issues
2. **Improve reviewability** - Each PR focuses on one concern
3. **Maintain CI/CD health** - Keep all jobs passing throughout migration
4. **Enable easier rollback** - If issues arise, smaller changes are simpler to revert

### Proposed Migration Batches

**Phase 1: Foundation (PR #1)**
- Rename `images/rag-base` → `images/python-base`
- Update workflow and Makefile
- Verify all dependent builds still work

**Phase 2: Common Library (PR #2)**
- Create `services/common/` with Containerfile
- Copy (not move) code from `spyre-rag/src/common/`
- Add workflow for building `services-common` image
- Parallel existence with old structure

**Phase 3: Service Infrastructure (PR #3)**
- Create `services/{chatbot,digitize,summarize}/` directories
- Add Containerfiles, requirements.txt, Makefiles
- Add GitHub workflows (without moving code yet)
- Verify build infrastructure works

**Phase 4: Service Migration (PRs #4-6, one per service)**
- Migrate each service independently
- Update deployment templates
- Verify tests and builds
- Keep old structure until verified

**Phase 5: UI Reorganization (PR #7)**
- Move UI directories to `ui/` structure
- Update workflows

**Phase 6: Cleanup (PR #8)**
- Remove `spyre-rag/` directory
- Remove deprecated workflows
- Update documentation

### Benefits
- Each PR is reviewable in <30 minutes
- CI/CD stays green throughout
- Easy to identify which change caused any issues
- Can pause/adjust strategy between phases
- Production deployments remain stable

ryanmarc:
@dharaneeshvrd Breaking this into 8 PR's would likely be more trouble than it's worth - If there were no one else making changes to the repo at all during that time then it could work, but that's not the case and trying to merge 8 different PR's that all build off each other while each of those is likely resolving conflicts along the way is going to be a nightmare scenario.

dharaneeshvrd:
Agree, 8 PR would be too much. 
All the failed jobs in this PR are happening because we are trying to move the code without fixing the image pipeline.  So IMO it's better setup the things before moving the code like image pipeline, PR checks and not disturb the current flow. I still feel we can split this into multiple PRs.

ryanmarc:
@dharaneeshvrd even breaking this down into multiple PR's will have the same problem, the PR checks will fail until the ci/cd changes are made. The actual code changes in this PR aren't a lot, it mostly moving files around. This PR has been open for a week now, the changes could have been reviewed in the time we've been going back and forth about making this multiple PR's.

ryanmarc:
If you want all the PR checks to be passing for this to be merged, then the CICD fixes need to be made and merged into this branch.

dharaneeshvrd:
> even breaking this down into multiple PR's will have the same problem, the PR checks will fail until the ci/cd changes are made.
> If you want all the PR checks to be passing for this to be merged, then the CICD fixes need to be made and merged into this branch.

IMO we need to combine this and fix the image part before moving code around. 

My main concern is that till we fix the cicd, dev and test would be affected since all values.yaml are pointing to new image tags i.e. `icr.io/ai-services-cicd/chatbot-ui:v0.0.1`.
But anyway this is my opinion.

@mkumatag @yussufsh @iv1111 please share your thoughts here.

mkumatag:
modifying specs comes later cicd can be fixed part of this PR to get a confidence about image building with these changes.

services/chatbot/settings.py
```diff
@@ -216,16 +216,6 @@ def validate_num_chunks_post_reranker(cls, v):
             return 3
         return v
 
-    @field_validator('prompt_template_token_count')
```
dharaneeshvrd, line 254: why this method is removed? It should belong to chatbot settings.

services/common/Containerfile
```diff
@@ -0,0 +1,8 @@
+FROM python-base:latest
```
dharaneeshvrd: Makefile is missing for common

services/chatbot/Containerfile
```diff
@@ -0,0 +1,14 @@
+FROM services-common:latest
```
dharaneeshvrd: Don't we need to use the actual image from icr registry?
Applies to other service's image usage.

services/common/settings.py
```diff
@@ -242,12 +242,47 @@ class VectorStoreConfig(BaseSettings):
     )
 
 
+class PromptsConfig(BaseSettings):
```
dharaneeshvrd: these prompts are very specific to chatbot, it should not be in common settings.

services/chatbot/tests/requirements-test.txt
dharaneeshvrd, line 1: This `requirements-test.txt‎` contains pkgs required for all the services.
Need to segregate the pkgs that needs only for chatbot and add it here.
Please do this exercise for other services too.
But I dont see `requirements-test.txt`‎ for other services.
Also it should be inside `tests` dir

dharaneeshvrd (commented):
> modifying specs comes later cicd can be fixed part of this PR to get a confidence about image building with these changes.

@Bleama02 please remove the asset changes like @mkumatag mentioned.

@iv1111 @prachikurade1 Please note that, till cicd changes are done, QA activities might be affected.

services/chatbot/settings.py
Bleama02: Im not sure why it was removed. You're, right! Ill add it back in my next commit today.

services/chatbot/Containerfile
Bleama02: @noahmasoud @ryanmarc ^

Bleama02:
> > modifying specs comes later cicd can be fixed part of this PR to get a confidence about image building with these changes.
> 
> @Bleama02 please remove the asset changes like @mkumatag mentioned.
> 
> @iv1111 @prachikurade1 Please note that, till cicd changes are done, QA activities might be affected.

@dharaneeshvrd @mkumatag 
Hey, quick question about reverting the assets folder - I want to make sure I understand what you're asking for.

If I revert those files back to main, they'll still be pointing to the old structure (like rag:v0.0.107 image and chatbot.app:app paths). But the actual code is now in the new services/ structure with separate images for each service.

So basically the deployment configs won't match the code anymore and nothing will work until we update the assets again later, right?

Just want to confirm that's the plan - revert assets now, get the code changes merged, then do the asset updates in a follow-up PR once CICD is ready? Or did you have something else in mind?

dharaneeshvrd:
>  revert assets now, get the code changes merged, then do the asset updates in a follow-up PR once CICD is ready? Or did you have something else in mind?

Yes, This is what my understanding too. We need to updated the assets only after the images are ready.

dharaneeshvrd:
@pranithraoibm Can you please review the changes related to UI?

ryanmarc:
@dharaneeshvrd regarding the assets and the images - This sounds like we should just make the CI/CD changes and merge that into this branch so that the images get created and properly exist.

dharaneeshvrd:
> we should just make the CI/CD changes and merge that into this branch so that the images get created and properly exist.

Yes that is better.

services/chatbot/Containerfile
noahmasoud: I was using a local tag for now since we do not have a published image currently

dharaneeshvrd:
@adarshagrawal38 Do you think you can take this up and work on cicd related changes or someone else already looking at it?

pranithraoibm:
@dharaneeshvrd UI changes are just folder re-orgs and that won't affect as all imports use relative paths or @/* alias. So LGTM.

ai-services/assets/applications/rag/openshift/values.yaml
```diff
@@ -37,29 +37,14 @@ instruct:
   storageUri: hf://ibm-granite/granite-3.3-8b-instruct
   # @description API key for vLLM instruct service authentication. If empty, authentication is disabled. Provide a key to enable authentication.
   apiKey: ""
-  model:
```
sats-23, line 40: Why is this removed? Is it a pending rebase?

services/common/lang_utils.py
```diff
@@ -1,35 +1,37 @@
...
+from common.settings import settings
 
 logger = get_logger("LANG")
 
 _language_detector = None
 lang_en = "EN"
 lang_de = "DE"
 
-def get_prompt_for_language(lang: str) -> str:
+def get_prompt_for_language(lang: str, prompt_en: str, prompt_de: str) -> str:
```
sats-23: Why this workaround? Is there a circular dependency?

sats-23: Why not pass a map instead? Params list will keep growing with language support

services/common/requirements.txt
```diff
@@ -0,0 +1,2 @@
+# common-specific dependencies beyond python-base
+# All heavy deps (fastapi, opensearch-py, docling, etc.) live in images/python-base/requirements.txt
```
sats-23, line 2: Is there an example to this usage?
Isn't python-base meant to be the equivalent of common requirements?

adarshagrawal38:
> @adarshagrawal38 Do you think you can take this up and work on cicd related changes or someone else already looking at it?

To fix GHA issue @vijay-vin will be working on it.

services/common/requirements.txt
Bleama02: Good question. python-base is effectively our shared dependency layer (the equivalent of common requirements).

So _common/requirements.txt_ is intentionally empty because all common dependencies already live in python-base. The current comment may be misleading and could be clarified. I can either just delete the file all together or I can edit it and clarify it better.

services/common/lang_utils.py
Bleama02: I made this change to address the earlier feedback about moving PromptsConfig out of common settings. It was causing a circular dependency, so I changed the function to accept prompts as parameters. However, I agree that passing a dict would be better for scalability - would you like me to refactor it that way?

ai-services/assets/applications/rag/openshift/values.yaml
Bleama02: Good catch, looking at the git history it looks like it may have been deleted during some sort of merge conflict. I will restore it today and take a look at anything else that may have been deleted.

services/common/lang_utils.py
dharaneeshvrd: Let's do it.

dharaneeshvrd:
@vijay-vin @adarshagrawal38 Did we decide on the CICD changes?

Yes. I have pushed the changes in this PR.

vijay-vin:
@adarshagrawal38 @dharaneeshvrd @Bleama02 
Previous checks that were failing are fixed now. Some of the checks are still queued and system is very slow running the checks.

vijay-vin:
@adarshagrawal38 @dharaneeshvrd @prachikurade1 

After adding images/python-base/requirements.txt as install dependencies, Services Unit Test is failing because "lingua language detector" version 2.2.0 is still not available on devpi, https://wheels.developerfirst.ibm.com/ppc64le/linux.

But, I had to add images/python-base/requirements.txt as install dependencies to spyre-rag-uti-tests.yml. Without these install dependencies, test run is failing to import rapidfuzz packages.

_ModuleNotFoundError: No module named 'rapidfuzz'_

Once I add images/python-base/requirements.txt, it tries to install "lingua language detector" version 2.2.0 and is failing.

It would be a quicker turn around for better fix if the test team looks into this failure causes.

dharaneeshvrd (commented):
I don't see workflows created for building new service's images :(
Blocking again for another day will increase the load on use cases to merge more PRs at once.
Would like to merge as is and fix the missing workflows in follow up PR.
@adarshagrawal38 could you please help here?

dharaneeshvrd (commented):
Other items that need further actions:
- Fix failing UTs - will create a JIRA bug
- rag-cpu missing newly added instruct fields - Will raise a minor followup PR
- Minor settings misplacement - can be corrected in future PRs that touch these areas

services/common/requirements.txt
sats-23: Let's keep it for now, keeps things uniform

services/common/Makefile
```diff
@@ -0,0 +1,18 @@
+REGISTRY?=icr.io/ai-services-private
+IMAGE=services-common
```
adarshagrawal38, line 2: How about naming service as

```suggestion
IMAGE=common-service
```
so that we are in-line with naming convention, with other services.

services/chatbot/Containerfile
adarshagrawal38:
```suggestion
FROM icr.io/ai-services-cicd/services-common:latest
```

services/digitize/Containerfile
```diff
@@ -0,0 +1,14 @@
+FROM services-common:latest
```
adarshagrawal38:
```suggestion
FROM icr.io/ai-services-cicd/services-common:latest
```

services/similarity/Containerfile
```diff
@@ -0,0 +1,14 @@
+FROM services-common:latest
```
adarshagrawal38:
```suggestion
FROM icr.io/ai-services-cicd/services-common:latest
```

services/summarize/Containerfile
```diff
@@ -0,0 +1,14 @@
+FROM services-common:latest
```
adarshagrawal38:
```suggestion
FROM icr.io/ai-services-cicd/services-common:latest
```

services/common/Makefile
dharaneeshvrd: Agree.

dharaneeshvrd:
@adarshagrawal38 Let's add necessary changes to pass GHA

services/common/Makefile
dharaneeshvrd: Sry, I think we need to keep services-common only, since there is no service named common. Please revoke.

dharaneeshvrd:
Now I see too many GHA overall, it does nt look good to me, IMO we need to combine individual service's jobs to a common job for all the services or however we can combine them to have less GHA.
@adarshagrawal38 wdyt?

adarshagrawal38:
> Now I see too many GHA overall, it does nt look good to me, IMO we need to combine individual service's jobs to a common job for all the services or however we can combine them to have less GHA. @adarshagrawal38 wdyt?

Hey @dharaneeshvrd yeah right we have lots of GHA workflows. For each image, we currently trigger 3 workflows:

- `build-and-publish-image`
- `cve-scan`
- `license-scan`

If we consolidate the `services GHA` into one, we will reduce 4 workflows but the combined workflow would get lengthy and complicated.

dharaneeshvrd:
> If we consolidate the services GHA into one, we will reduce 4 workflows but the combined workflow would get lengthy and complicated

This sounds good to me, if you want to have a bigger discussion on this, we can do this later and update the workflows.

dharaneeshvrd (approved):
Thanks @adarshagrawal38 for adding the missing workflows & @sats-23 for missing changes.
Will wait for the jobs to pass, seems there is a downtime with power runners in GHA side.

mkumatag:
CVE scans can be ignored, there is new ubi 9.8 image is out, we will update post merging this PR.

PR #711, opened 2026-05-06
Conversational RAG - UI Changes

sats-23:
Screenshot 2026-05-06 at 15 51 23

### DIFF
--- a/spyre-rag/src/chatbot/app.py
+++ b/spyre-rag/src/chatbot/app.py
@@ -6,7 +6,7 @@ from typing import Optional
 from fastapi import FastAPI, Request, HTTPException, Header, Depends
 from fastapi.security import HTTPBearer, HTTPAuthorizationCredentials
 from fastapi.openapi.docs import get_swagger_ui_html
-from fastapi.responses import StreamingResponse, JSONResponse
+from fastapi.responses import StreamingResponse, Response
 import json
 from contextlib import asynccontextmanager
 from asyncio import BoundedSemaphore
@@ -317,7 +317,7 @@ async def locked_stream(stream_g, perf_stat_dict):
         503: http_error_responses[503]
     }
 )
-async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTPAuthorizationCredentials] = Depends(security)) -> ChatCompletionResponse | StreamingResponse:
+async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTPAuthorizationCredentials] = Depends(security)) -> ChatCompletionResponse | StreamingResponse | Response:
     # Extract API key from credentials
     api_key = credentials.credentials if credentials else None
     
@@ -489,8 +489,9 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
                 response_data = ChatCompletionResponse(choices=choices)
                 # Add rephrased query as a custom header if available
                 if rephrased_query and rephrased_query != current_query:
-                    return JSONResponse(
-                        content=response_data.model_dump(),
+                    return Response(
+                        content=response_data.model_dump_json(),
+                        media_type="application/json",
                         headers={"X-Rephrased-Query": rephrased_query}
                     )
                 return response_data
--- a/spyre-rag/src/chatbot/app.py
+++ b/spyre-rag/src/chatbot/app.py
@@ -6,7 +6,7 @@ from typing import Optional
 from fastapi import FastAPI, Request, HTTPException, Header, Depends
 from fastapi.security import HTTPBearer, HTTPAuthorizationCredentials
 from fastapi.openapi.docs import get_swagger_ui_html
-from fastapi.responses import StreamingResponse
+from fastapi.responses import StreamingResponse, JSONResponse
 import json
 from contextlib import asynccontextmanager
 from asyncio import BoundedSemaphore
@@ -448,7 +448,11 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
                     rephrased_query,
                 )
                 # For streaming, release is handled in locked_stream's finally block
-                return StreamingResponse(locked_stream(vllm_stream, perf_stat_dict), media_type="text/event-stream")
+                response = StreamingResponse(locked_stream(vllm_stream, perf_stat_dict), media_type="text/event-stream")
+                # Add rephrased query as a custom header if available
+                if rephrased_query and rephrased_query != current_query:
+                    response.headers["X-Rephrased-Query"] = rephrased_query
+                return response
 
             vllm_non_stream = await asyncio.to_thread(
                 query_vllm_non_stream,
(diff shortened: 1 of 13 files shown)


pranithraoibm (dismissed):

spyre-rag/ui/src/components/customSendMessage.jsx
```diff
@@ -232,6 +251,14 @@ async function customSendMessage(
...
       },
     });
+
+    // Update conversation history with the new exchange
+    // Add user message and assistant response to history with proper roles
+    setConversationHistory([
+      ...conversationHistory,
+      { role: 'user', content: userInput },
+      { role: 'assistant', content: fullText },
+    ]);
```
Niharika0306, line 262: I guess we are updating the conversationHistory state twice with the userInput

Niharika0306:
You think we should be implementing any truncating logic on the UI side as well? 
Because currently, the conversationHistory can grow unbounded and reach token limits.

sats-23:
> You think we should be implementing any truncating logic on the UI side as well? Because currently, the conversationHistory can grow unbounded and reach token limits.

I would carry a heavy payload which might slow down calls, but wouldn't hit token limits. 
Let's proceed with truncation logic

spyre-rag/ui/src/components/customSendMessage.jsx
sats-23: Fixed it to remove the redundant one

sats-23:
@Niharika0306 @pranithraoibm Implemented a sliding window for recent 10 messages.
-Browser state still maintains all messages until page refresh/new chat (minimal impact)
-But sliding window ensures light payload and no performance degradation

Niharika0306:
Found a small issue, in accordance to how the current chat window logic works:

so right now, we do 2 calls for every user query. 
1. /chat/completion
2. /v1/similarity-search ( this is because , since we do a streaming request to /chat/completion , the request pipe can only receive the chunks of streamed data output, and cannot wait for any other json data ) which is why we are making a separate /similarity-search call to get the reference docs. 
But now, we are losing the rephrased query while making the 2nd call. And the context is being different for both.

Screenshot 2026-05-12 at 5 10 13 PM

dharaneeshvrd:
Seems we have to bring back the reference endpoint back to chatbot to address the problem described by @Niharika0306 
@ryanmarc what do you think?
Context: We are trying to add the conversation support to the chatbot so that user can ask follow up questions and it requires rewriting the query to resolve the pronouns and things referred in the answer indirectly.

spyre-rag/ui/src/components/customSendMessage.jsx
```diff
@@ -182,10 +228,17 @@ async function customSendMessage(
       },
     });
 
-    // await for the reference promise to finish
+    // Now fetch reference docs using the rephrased query (or original if not rephrased)
     let docs = [];
     try {
-      const context_response = await referencePromise;
+      const context_response = await axios.post('/v1/similarity-search', {
```
dharaneeshvrd, line 192: Let's pass top_k as 3 and rerank as true

spyre-rag/src/chatbot/app.py
```diff
@@ -444,11 +455,15 @@ async def stream_server_busy():
                     perf_stat_dict,
                     lang,
                     api_key,
-                    previous_messages,
+                    truncated_history,
                     rephrased_query,
                 )
                 # For streaming, release is handled in locked_stream's finally block
-                return StreamingResponse(locked_stream(vllm_stream, perf_stat_dict), media_type="text/event-stream")
+                response = StreamingResponse(locked_stream(vllm_stream, perf_stat_dict), media_type="text/event-stream")
```
dharaneeshvrd, line 451: I think this is good.
Can we update the non stream case to maintain consistency?

spyre-rag/ui/src/components/customSendMessage.jsx
```diff
@@ -111,53 +123,87 @@ async function customSendMessage(
   try {
     instance.updateIsMessageLoadingCounter('increase');
 
-    const stream = await client.chat.completions.create(payload);
```
dharaneeshvrd, line 114: seems we are not using openai client, any particular reason?

spyre-rag/ui/src/components/customSendMessage.jsx
sats-23: Done

spyre-rag/ui/src/components/customSendMessage.jsx
```diff
@@ -182,10 +183,18 @@ async function customSendMessage(
...
 
-    // await for the reference promise to finish
+    // Now fetch reference docs using the rephrased query (or original if not rephrased)
     let docs = [];
     try {
-      const context_response = await referencePromise;
+      const context_response = await axios.post('/v1/similarity-search', {
+        query: rephrasedQuery,
+        mode: 'hybrid',
+        top_k: 3,
```
dharaneeshvrd, line 195: Let's create a subtask to have a common env var at manifests level and use it in both chatbot and ui.

sats-23: Sure

spyre-rag/src/chatbot/app.py
sats-23: Done

spyre-rag/ui/src/components/customSendMessage.jsx
sats-23: Reverted to a workaround using OpenAI client. 
To use the full response (with headers), I had to update the OpenAI SDK which provides support for this function. We are now able to extract the header from the response and get the rephrased query

sats-23:
@dharaneeshvrd @Niharika0306 
I have updated the OpenAI SDK to support the function which helps extract the full response (with headers). PTAL. 

Do let me know if there is something I have missed as a side effect of updating the SDK

spyre-rag/src/chatbot/app.py
```diff
@@ -481,7 +485,15 @@ async def stream_server_busy():
                         if isinstance(message_dict, dict):
                             message_content = message_dict.get("content", "")
                             choices.append(ChatChoice(message=ChatMessage(content=message_content)))
-                return ChatCompletionResponse(choices=choices)
+                
+                response_data = ChatCompletionResponse(choices=choices)
+                # Add rephrased query as a custom header if available
+                if rephrased_query and rephrased_query != current_query:
+                    return JSONResponse(
```
dharaneeshvrd: It does nt look good to alter this.
Seems there are other ways to pass headers on the response & still use ChatCompletionResponse

sats-23: I've updated the approach and tested it to work on both stream and non-stream cases 
http://10.48.64.114:5004/ & http://10.48.64.114:5003/

dharaneeshvrd (approved):
@Niharika0306 can you please review and merge this?

Niharika0306 (approved):

Niharika0306:
All changes LGTM

PR #712, opened 2026-05-06
feat(catalog-ui):  Service Cards for catalog-ui

AS-Prajna:
Added Service cards to reusable catalog layout
image


AS-Prajna:
The designs for Service catalog is changed so closing this PR

PR #716, opened 2026-05-07
DB/ORM python dependencies

manju956:


spyre-rag/src/Containerfile
dharaneeshvrd, line 1: Usually this update happens in subsequent PRs
@yussufsh will this work?

images/rag-base/requirements.txt
```diff
@@ -75,6 +75,7 @@ pillow==12.1.1
 pluggy==1.6.0
 polyfactory==3.3.0
 psutil==7.2.2
+psycopg2-binary==2.9.10
```
dharaneeshvrd: Let's add psycopg2 version as well

spyre-rag/src/Containerfile
yussufsh: This and all values.yaml updates will be needed after this PR post-submit job, which will build the new rag-base version and push it to the repo. @manju956

images/rag-base/requirements.txt
```diff
@@ -109,6 +110,7 @@ sentence-splitter==1.4
 sentry-sdk==2.55.0
 setuptools==82.0.1
 shapely==2.1.2
+SQLAlchemy==2.0.36
```
yussufsh: Why not use the latest version?

images/rag-base/requirements.txt
manju956: done

images/rag-base/requirements.txt
manju956: done

spyre-rag/src/Containerfile
manju956: I have removed all the RAG image version bumps, will raise separte PR for upadting RAG base image in RAG Containerfile

dharaneeshvrd (dismissed):
@yussufsh can you please approve if the versions are ok?

images/rag-base/requirements.txt
```diff
@@ -75,6 +75,7 @@ pillow==12.1.1
 pluggy==1.6.0
 polyfactory==3.3.0
 psutil==7.2.2
+psycopg2==2.9.10
```
dharaneeshvrd: Does installing `psycopg2` installs `psycopg2-binary` as well?

manju956: No, `psycopg2` will not install `psycopg2-binary` pkg. both are similar interms of functionality. however `psycopg2-binary` is used only dev/test environments. Hence i removed this.
the db initialization works with `psycopg2 ` library alone.

yussufsh: Any reason of not using the latest?

manju956: Getting below error with latest version `2.9.12`, so going ahead with previous working version `2.9.11`
```
Collecting psycopg2==2.9.12 (from -r requirements.txt (line 78))
  Downloading psycopg2-2.9.12.tar.gz (379 kB)
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'error'
pdfplumber==0.11.9
  error: subprocess-exited-with-error
```

yussufsh: I see `2.9.11`` is available in devpi.
We are good then.

yussufsh (approved):

PR #721, opened 2026-05-07
chore(catalog-ui): Design audit changes

pranithraoibm:
PR includes changes requested in the call with the design team
- Change the heading to 'IBM AI Foundation for Power'
- Icons update in reference use cases page
- Style updates for reference use cases cards

image

- Search functionality update to search card content instead of filter options
- Page header description update for all the 3 catalog pages


mayuka-c (dismissed):
LGTM,
Based on the screenshot I see the title and subTitle for references needs to be updated. Is it taken care in someother PR?

I see still its having older title and subTitle: https://github.com/pranithraoibm/project-ai-services/blob/4bf73d8dea67b0984f15a06e74ee90e410819915/ui/catalog/src/pages/ReferenceUseCases/ReferenceUseCases.tsx#L391

pranithraoibm:
Yes @mayuka-c, it's updated in this PR: https://github.com/IBM/project-ai-services/pull/753

ai-services/assets/catalog/podman/values.yaml
```diff
@@ -1,10 +1,10 @@
 ui:
   port: ""
-  image: icr.io/ai-services-cicd/catalog-ui:v0.0.6
+  image: icr.io/ai-services-cicd/catalog-ui:v0.0.7
 
 backend:
   port: ""
-  image: icr.io/ai-services-cicd/ai-services:0.0.21
+  image: icr.io/ai-services-cicd/ai-services:0.0.22
```
mayuka-c: Why is this getting bumped?

ai-services/Makefile
```diff
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=0.0.21
+TAG?=0.0.22
```
mayuka-c: We need not have to bump up this as I dont see any backend changes

pranithraoibm: Correct, was wondering the same but then the version bump check fails

image

mayuka-c: @adarshagrawal38 Any reason why the workflow fails and ask to bump up ai-services?. This PR is having only UI change.
cc. @yussufsh

adarshagrawal38: @mayuka-c The bump check was failing as files under `ai-services/** ` were modified. We should update the script to skip the bump check when only files under `ai-services/assets/**` are changed."

mayuka-c: Thats not correct. We need not bump if ai-services/assets/catalog is updated.
In this way we should not be bumping backend image all the time if UI is changing

I think under assets we need to consider only architectures, services and components, only if these change bump

yussufsh: @mayuka-c if I am not wrong you won't be able to deploy ui pod unless you have that in assets and that is part of ai-services image. Let me know if there is a way to deploy ui without catalog image bump.

adarshagrawal38: for applications  also we need to bump up right?

Will skip for:
- ai-services/assets/catalog
- ai-services/assets/bootstrap

yussufsh:
@pranithraoibm why cannot we rebase each PR you have in queue? Atleast a few if you can combine it will be easy to review. This way for bumping each will take time.

ai-services/Makefile
mayuka-c:
```
@mayuka-c if I am not wrong you won't be able to deploy ui pod unless you have that in assets and that is part of ai-services image. Let me know if there is a way to deploy ui without catalog image bump.
```
Since its a CLI way of deploying catalog we can still do it right?
We only need to bump `ai-services` image if there is something which Catalog backend pod takes care of deploying. This is architecture, services and components

mayuka-c: `applications` is also something which Catalog backend pod doesnt use but we can keep it (if we want to)

pranithraoibm:
@yussufsh you mean add related changes from other PRs to this PR to minimize these version bumps?

ai-services/Makefile
pranithraoibm: We are adding the fix for this PR or can we merge this?

yussufsh:
> @yussufsh you mean add related changes from other PRs to this PR to minimize these version bumps?

Yes, if this is approved, let it go in.
From the next PR, just rebase everything into the latest PR and have one or two PR which is "reviewable".

pranithraoibm:
> > @yussufsh you mean add related changes from other PRs to this PR to minimize these version bumps?
> 
> Yes, if this is approved, let it go in. From the next PR, just rebase everything into the latest PR and have one or two PR which is "reviewable".

Cool, will do.

mayuka-c (approved):
LGTM

PR #723, opened 2026-05-08
chore: refactor secrets and add logic to ignore already deployed resources

Shubhamag12:
1. Separating secrets for catalog and catalog db enables us to handle deletion better
2. added a logic to ignore already existing resources (pods and secrets)


ai-services/internal/pkg/cli/helpers/helper.go
```diff
@@ -190,3 +207,21 @@ func CheckExistingPodsForApplication(runtime runtime.Runtime, appName string) ([
...
+		if err != nil && strings.Contains(err.Error(), constants.ErrSecretNotFound) {
+			return nil, fmt.Errorf("failed to list secrets: %w", err)
+		}
+		if len(secret) != 0 {
+			logger.Infof("Existing secret found: %s\n", secret[0])
+			secretsToSkip = append(secretsToSkip, secretName)
+		}
+	}
+
+	return secretsToSkip, nil
```
mayuka-c, line 227: Cant we just do a ListSecrets get all the secrets list for given label AppName and just filter out secretsToSkip. This looks much simpler

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -229,4 +249,13 @@ func executePodTemplate(rt *podman.PodmanClient, tp templates.Template, tmpls ma
...
 }
 
+func extractSecretName(podTemplateName string) string {
+	secretMap := map[string]string{
+		"catalog-secret.yaml.tmpl":    constants.CatalogSecretName,
+		"catalog-db-secret.yaml.tmpl": constants.CatalogDBSecretName,
+	}
+
+	return secretMap[podTemplateName]
+}
```
mayuka-c: I think we need to extract this dynamically based on the files present under templates. If its a secret, then it will come under this map.

ai-services/internal/pkg/cli/helpers/helper.go
Shubhamag12: My initial approach was that, but I found out that podman doesnt support filtering secrets based on labels :(

mayuka-c: Ah ok fine then,
Also at line 217 you're already checking for SecretNotFound, then do we still need to do length of secret check at 220?

Shubhamag12: my bad, it should be 
```
!strings.Contains(err.Error(), constants.ErrSecretNotFound) {
```

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
Shubhamag12: do you mean extract the secret name dynamically? or build the map dynamically?

mayuka-c: Yes bulding this map dynamically based on the templates.

Shubhamag12: Turns out this wont be even required
```
	if slices.Contains(existingResources, podSpec.Name) {
		logger.Infof("%s: Skipping pod deploy as '%s' it already exists", podTemplateName, podSpec.Name)

		return nil
	}
```
the above code will filter secret and pod both out (as podSpec.Name will have the secretName)

ai-services/internal/pkg/cli/helpers/helper.go
Shubhamag12: made changes, ptal

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
mayuka-c: Still this is hardcoded right: https://github.com/IBM/project-ai-services/pull/723/changes#diff-0e2dc4810b5f4be2ce21af0a0aff9527b1cf93bf1b84979d009237da38507a71R52.

We would have to make it read all secret names from the templates with kind `Secret`. This makes it more extensible and this would avoid in future to update it here.

Shubhamag12: done, ptal

ai-services/internal/pkg/catalog/constants/common.go
```diff
@@ -0,0 +1,6 @@
+package constants
+
+const (
+	CatalogSecretName   = "catalog-secret"
+	CatalogDBSecretName = "catalog-db-secret"
```
mayuka-c: This can be removed now right?

mayuka-c (commented):
Rest LGTM

ai-services/internal/pkg/catalog/constants/common.go
Shubhamag12: This would be required in uninstall PR. I will add it there

Shubhamag12: done, ptal

mayuka-c (approved):
LGTM

PR #732, opened 2026-05-11
Update services metadata to use component type under dependencies

yussufsh:


ai-services/assets/services/chat/metadata.yaml
```diff
@@ -11,13 +11,8 @@ certified_by: "IBM"
...
 dependencies:
-  - id: opensearch
-    version: ">=1.0.0"
+  - id: vector_store
   - id: embedding
-    version: ">=1.0.0"
-  - id: instruct
-    version: ">=1.0.0"
-  - id: reranker
-    version: ">=1.0.0"
```
mayuka-c, line 23: Just a query, any reason for removing version?

mayuka-c (dismissed):
LGTM

ai-services/assets/services/chat/metadata.yaml
mayuka-c: I feel we could have still shown 1.0.0

yussufsh: This version is provider-specific, so adding a version in services does not make sense. This just mentions that the service require a component type and not a provider eg: opensearch.

mayuka-c (approved):

PR #734, opened 2026-05-11
[Catalog-API] API Endpoints for Architecture & Service Deploy options

mayuka-c:
- Endpoints for Architecture and Service Deploy Options which will be consumed during Deploy flow.
- Expose an API to read the component params.

From the proposal it implements 3 API's as proposed here: https://github.com/mayuka-c/project-ai-services/blob/f4509f90819f1d973a3505aa58c45bd19d4b97ce/docs/proposals/catalog/application-deployment-api-proposal.md#54-deploy-options-endpoints

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/handlers/deploy_options.go
+++ b/ai-services/internal/pkg/catalog/apiserver/handlers/deploy_options.go
@@ -1,0 +1,119 @@
+package handlers
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"github.com/project-ai-services/ai-services/internal/pkg/catalog"
+	"github.com/project-ai-services/ai-services/internal/pkg/vars"
+)
+
+// DeployOptionsHandler handles deploy options related HTTP requests.
+type DeployOptionsHandler struct {
+	provider *catalog.CatalogProvider
+}
+
+// NewDeployOptionsHandler creates a new deploy options handler.
+func NewDeployOptionsHandler() *DeployOptionsHandler {
+	provider, err := catalog.NewCatalogProvider()
+	if err != nil {
+		panic(fmt.Sprintf("Failed to initialize catalog provider: %v", err))
+	}
+
+	return &DeployOptionsHandler{
+		provider: provider,
+	}
+}
+
+// GetArchitectureDeployOptions godoc
+//
+//	@Summary		Get architecture deploy options
+//	@Description	Retrieves available providers and dependency rules for all services and their components within an architecture
+//	@Tags			Deploy Options
+//	@Produce		json
+//	@Security		BearerAuth
+//	@Param			architecture_id	path		string	true	"Architecture ID (e.g., 'rag')"
+//	@Success		200				{object}	github_com_project-ai-services_ai-services_internal_pkg_catalog_types.DeployOptionsArchitecture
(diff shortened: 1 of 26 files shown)


ai-services/internal/pkg/catalog/apiserver/handlers/deploy_options.go
```diff
@@ -0,0 +1,115 @@
...
+// DeployOptionsHandler handles deploy options related HTTP requests.
+type DeployOptionsHandler struct {
+	provider *catalog.CatalogProvider
+}
+
+// NewDeployOptionsHandler creates a new deploy options handler.
+func NewDeployOptionsHandler() *DeployOptionsHandler {
+	provider, err := catalog.NewCatalogProvider()
+	if err != nil {
+		panic(fmt.Sprintf("Failed to initialize catalog provider: %v", err))
```
Shubhamag12, line 20: do we want to panic?

ai-services/internal/pkg/catalog/apiserver/handlers/deploy_options.go
```diff
@@ -0,0 +1,115 @@
...
+		c.JSON(http.StatusNotFound, ErrorResponse{
+			Error: fmt.Sprintf("Failed to get parameters for provider '%s/%s': %v", componentType, providerID, err),
+		})
+
+		return
+	}
+
+	c.JSON(http.StatusOK, schema)
+}
+
```
Shubhamag12, line 114: I see we are returning `StatusNotFound` for every other error, is this expected?

ai-services/internal/pkg/catalog/deploy_options.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("service not found: %w", err)
+	}
+
+	// Build components list
+	components := make([]types.DeployOptionsComponent, 0, len(service.Dependencies))
+	for _, dep := range service.Dependencies {
+		component, err := p.buildDeployOptionsComponent(dep.ID)
+		if err != nil {
+			continue
```
Shubhamag12, line 82: can you add a comment on why are we ignoring error

ai-services/internal/pkg/catalog/apiserver/handlers/deploy_options.go
mayuka-c: Panic should be fine here as this initialized during the container start up and let it panic if there is any issue before starting up the api-server

ai-services/internal/pkg/catalog/apiserver/handlers/deploy_options.go
mayuka-c: The APIs for deploy options mostly relies on internal assets embedded in binary. So the reason it fails would be because of not present, so I feel this is fine for now

ai-services/internal/pkg/catalog/deploy_options.go
mayuka-c: This is for building component options and there are 2 reasons it fails
- If provider is not present within a given component
- Or if you fail to list components.

Again if any of these fails, it should be fine to ignore. But we fail only if the component with given id is not present.

But I will log an error message here just to print the error.

mayuka-c: Added logs

yussufsh (approved):

PR #736, opened 2026-05-12
AISERVICES-1062 added unit tests for src/digitize

manalilatkar:


spyre-rag/requirements-test.txt
```diff
@@ -27,4 +27,8 @@ opensearch-py==3.0.0
 cohere==5.19.0
 numpy==2.4.1
 psutil==7.2.2
-pypdfium2==5.6.0
\ No newline at end of file
+pypdfium2==5.6.0
```
dharaneeshvrd, line 30: for these pkgs, you need to edit the workflow to pull it from the devpi also need to use the version which is align with the rag-base pkg versions

dharaneeshvrd: @manalilatkar The reason why it is working is because we were using x86 runner. https://github.com/IBM/project-ai-services/blob/9989be243ae4cf2ec3e93da1944940b1626f45d0/.github/workflows/spyre-rag-unit-tests.yml#L13 we need to change this to ppc machine.
IMO it's better to test it on ppc machine, since these are going to run in power machines anyway.

manalilatkar:
@dharaneeshvrd tests are now running successfully on power.

.github/workflows/spyre-rag-unit-tests.yml
```diff
@@ -10,21 +10,23 @@ on:
...
+    runs-on: ubuntu-24.04-ppc64le
     
     steps:
       - name: Checkout code
         uses: actions/checkout@v6
-      
+
       - name: Set up Python 3.12
-        uses: actions/setup-python@v6
+        uses: IBM/setup-python-pz@291287e36d7c893a50c857d8bc014d67edd58ada
```
dharaneeshvrd, line 20: Is it possible to use tag instead of hash here?
Also how did you get this info?

manalilatkar: @adarshagrawal38 guided me.

adarshagrawal38: @dharaneeshvrd using hashes is more secure whey of using github action.

dharaneeshvrd (approved):

PR #739, opened 2026-05-12
fix: preserve context while streaming/stopping logs

Shubhamag12:
The ctrl+c command does not work to exit the streaming. Ctrl+z had to be used to exit.
This is because proper ctx was not used.
JIRA: https://jsw.ibm.com/browse/AISERVICES-1021


yussufsh (approved):
lgtm

PR #743, opened 2026-05-13
Fix tar command failure as it is not available on the digitization pod

yussufsh:
With new ubi minimal image used for rag-base tar command was not found.
Updating the script not to run any commands on running containers and copy the files using `podman cp`.


mayuka-c (approved):
LGTM

Shubhamag12 (approved):

PR #746, opened 2026-05-13
Upgrade Spyre operator to v1.2

yussufsh:
We can check the following metadata to know what OCP versions can be available for the releases.

v1.1.x: https://github.com/redhat-openshift-ecosystem/certified-operators/blob/main/operators/spyre-operator/v1.1.1/metadata/annotations.yaml
v1.2.x: https://github.com/redhat-openshift-ecosystem/certified-operators/blob/main/operators/spyre-operator/v1.2.1/metadata/annotations.yaml
The installable OCP version is defined in this annotation file. so v1.1.x can be setup on v4.16 - v4.20, and v1.2.x can be setup on v4.16 - v4.21.


Shubhamag12 (approved):
lgtm

PR #752, opened 2026-05-14
feat: add Caddy proxy integration with route registration

shwetha-s-poojary:
## Summary
Adds Caddy reverse proxy integration with automatic route registration for catalog service.

## Changes
- **New proxy package** with Caddy implementation and helper utilities
- **Template system** updated to support routes_file.yaml parsing
- **Catalog deployment** now registers routes automatically with Caddy
- **Dependencies** added go-resty/resty/v2 for HTTP client

## Files Changed
- `internal/pkg/proxy/` - New package (util.go, caddy.go, types.go)
- `internal/pkg/cli/templates/` - Added LoadRoutesFile() and route types
- `internal/pkg/catalog/cli/configure/podman/` - Integrated route registration
- `assets/catalog/podman/steps/routes_file.yaml` - Route configuration
- `go.mod`, `go.sum` - Added resty dependency

## Test

Routes registered: 
```
# curl http://localhost:44901/config/apps/http/servers/my_app_server/routes
[{"@id":"ai-services--catalog--catalog-ui","handle":[{"handler":"reverse_proxy","upstreams":[{"dial":"ai-services--catalog:8081"}]}],"match":[{"host":["catalog-ui.{IP}.nip.io"]}],"terminal":true},{"@id":"ai-services--catalog--catalog-api","handle":[{"handler":"reverse_proxy","upstreams":[{"dial":"ai-services--catalog:8080"}]}],"match":[{"host":["catalog-api.{IP}.nip.io"]}],"terminal":true}]
```

Exposed ports:
```
# ai-services application ps -o wide --runtime podman
 APPLICATION NAME    POD ID          POD NAME                STATUS               CREATED           EXPOSED         CONTAINERS
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 ─────────────────────────────
 ai-services         31b30e816f0c    ai-services--caddy      Running (healthy)    17 minutes ago    39825, 44901    31b30e816f0c-infra (healthy), ai-services--caddy-caddy (healthy)
                     afef7f564a3c    ai-services--db         Running (healthy)    17 minutes ago    none            afef7f564a3c-infra (healthy), ai-services--db-postgresql (healthy)
                     48a01760e549    ai-services--catalog    Running (healthy)    16 minutes ago    34021, 45637    48a01760e549-infra (healthy), ai-services--catalog-ui (healthy), ai-services--catalog-backend (healthy)
```

Access:
- UI: `https://catalog-ui.{IP}.nip.io:39825`
- API: `https://catalog-api.{IP}.nip.io:39825`


ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+	serverName string
+}
+
+// NewCaddyManager creates a new Caddy proxy manager
+func NewCaddyManager(adminURL, serverName string) ProxyManager {
+	client := resty.New().
+		SetTimeout(10 * time.Second).
+		SetRetryCount(3).
+		SetRetryWaitTime(1 * time.Second).
+		SetRetryMaxWaitTime(5 * time.Second)
```
Shubhamag12: can you declare them as constants and use them here?

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+	return &caddyManager{
+		client:     client,
+		adminURL:   adminURL,
+		serverName: serverName,
+	}
+}
+
+// HealthCheck verifies Caddy is running and accessible
+func (c *caddyManager) HealthCheck() error {
+	url := fmt.Sprintf("%s/config/", c.adminURL)
```
Shubhamag12: you can use `url.JoinPath` here

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+	resp, err := c.client.R().Get(url)
+	
+	if err != nil {
+		return fmt.Errorf("failed to connect to Caddy admin API: %w", err)
+	}
+	
+	if resp.StatusCode() != 200 {
```
Shubhamag12:
```suggestion
	if resp.StatusCode() != http.StatusOK {
```

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+						"dial": route.Upstream,
+					},
+				},
+			},
+		},
+		"terminal": route.Terminal,
+	}
+
+	// Check if routes array exists, initialize if needed
+	checkURL := fmt.Sprintf("%s/config/apps/http/servers/%s/routes", c.adminURL, c.serverName)
```
Shubhamag12: same, use url.JoinPath

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+	resp, err := c.client.R().
+		SetHeader("Content-Type", "application/json").
+		SetBody(routeConfig).
+		Post(appendURL)
+
+	if err != nil {
+		return fmt.Errorf("failed to register route: %w", err)
+	}
+
+	if resp.StatusCode() != 200 {
```
Shubhamag12: same, use http.StatusOk

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+	appName string,
+	appTemplate string,
+	serverName string,
+) error {
+	// Step 1: Get Caddy admin port from pod port mappings
+	adminPort, err := GetCaddyAdminPort(rt, appName)
+	if err != nil {
+		logger.Warningf("Failed to get Caddy admin port: %v", err)
+		logger.Infoln("Routes not registered. You can manually configure them later")
+		return fmt.Errorf("failed to get Caddy admin port: %w", err)
```
Shubhamag12: why are we logging all of them differently here? 🤔 
Can we not club them all into single error string? Or even if logging is needed, can we do below instead?
```
if err != nil {
    logger.Warningln("Failed to get Caddy admin port, routes not registered. You can manually configure them later")
    return fmt.Errorf("failed to get Caddy admin port: %w", err)
}
```

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	// Return error if any routes failed to register
+	if len(registrationErrors) > 0 {
+		return fmt.Errorf("failed to register %d route(s)", len(registrationErrors))
+	}
+
+	return nil
+}
+
```
Shubhamag12, line 211: I see lot of redundant logs here. If we are returning error, we can just use error as logging mechanism here. Adding lot of stuff will fill up terminal.
Or maybe make them debug logs if required

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+// Made with Bob
```
Shubhamag12: add a new line after this

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -90,6 +91,14 @@ func DeployCatalog(ctx context.Context, podmanURI, passwordHash, baseDir string,
 	s.Stop("Catalog service deployed successfully")
 	logger.Infoln("-------")
 
+	// Register routes with Caddy using the reusable function
+	logger.Infoln("-------")
+	if err := proxy.RegisterRoutesForApp(rt, tp, catalogAppName, catalogAppTemplate, "my_app_server"); err != nil {
+		// Warnings already logged by RegisterRoutesForApp
+		logger.Infof("Route registration completed with errors: %v\n", err)
```
Shubhamag12: we should return error here directly, as if error is encountered while registering routes, we will just logs them and proceed with next steps.

shwetha-s-poojary: Ok - I was having second thoughts on giving user an option to configure routes.But yeah erroring out instead is a good idea.Thanks for the comment.I will make the necessary changes

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: Will do

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: Should we allow manual configuration or directly error out?
Any thoughts @mayuka-c @yussufsh

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: You are right.Let me keep only required logs here and clean it up.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
mayuka-c: You can error out here, but you need to ensure the operations in catalog configure are idempotent!.

Meaning you need to ensure each steps like reading admin password from user, deploy pods, configuring routes are all idempotent and will work fine if I retrigger multiple times

ai-services/assets/catalog/podman/steps/routes_file.yaml
```diff
@@ -0,0 +1,10 @@
+routes:
+  - pod: "ai-services--catalog"
+    services:
+      - containerPort: "8081/tcp"
+        subdomain: "catalog-ui"
```
mayuka-c: Just a query,
What exactly is subdomain?

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+// NewCaddyManager creates a new Caddy proxy manager
+func NewCaddyManager(adminURL, serverName string) ProxyManager {
+	client := resty.New().
+		SetTimeout(10 * time.Second).
+		SetRetryCount(3).
+		SetRetryWaitTime(1 * time.Second).
+		SetRetryMaxWaitTime(5 * time.Second)
+
+	return &caddyManager{
+		client:     client,
```
mayuka-c: mention `httpClient` instead of `client`

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+	// Step 1: Get Caddy admin port from pod port mappings
+	adminPort, err := GetCaddyAdminPort(rt, appName)
+	if err != nil {
+		logger.Warningf("Failed to get Caddy admin port: %v", err)
+		logger.Infoln("Routes not registered. You can manually configure them later")
+		return fmt.Errorf("failed to get Caddy admin port: %w", err)
+	}
+
+	// Step 2: Create proxy manager with the discovered admin URL
+	adminURL := fmt.Sprintf("http://localhost:%s", adminPort)
```
mayuka-c, line 173: Let this be a const

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+	proxyManager := NewCaddyManager(adminURL, serverName)
+
+	// Step 3: Perform health check on Caddy
+	if err := proxyManager.HealthCheck(); err != nil {
+		logger.Warningf("Caddy not ready: %v", err)
+		logger.Infoln("Routes not registered. You can manually configure them later")
```
mayuka-c: Log.Infoln doesnt make sense if you're failing it

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+		return fmt.Errorf("caddy health check failed: %w", err)
+	}
+
+	// Step 4: Get host IP for route domain generation
+	hostIP, err := utils.GetHostIP()
+	if err != nil {
+		logger.Warningf("Failed to get host IP: %v", err)
```
mayuka-c: Avoid logging for each failures here, you can log at the caller

ai-services/assets/catalog/podman/steps/routes_file.yaml
```diff
@@ -0,0 +1,10 @@
...
+      
+      - containerPort: "8080/tcp"
+        subdomain: "catalog-api"
+
```
mayuka-c: Do we need really need this file?

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+		return fmt.Errorf("failed to get host IP: %w", err)
+	}
+
+	// Step 5: Build routes from routes_file.yaml configuration
+	routes, err := BuildRoutesFromConfig(tp, appTemplate, hostIP)
+	if err != nil {
+		logger.Warningf("Failed to build routes from config: %v", err)
+		return fmt.Errorf("failed to build routes: %w", err)
+	}
```
mayuka-c: Cant we use the annotation itself, do we need routes_file.yaml

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -0,0 +1,191 @@
...
+
+	// Step 6: Register each route with Caddy
+	var registrationErrors []error
+	for _, route := range routes {
+		if err := proxyManager.RegisterRoute(route); err != nil {
+			logger.Warningf("Failed to register route %s: %v", route.ID, err)
+			registrationErrors = append(registrationErrors, fmt.Errorf("route %s: %w", route.ID, err))
+		}
+	}
```
mayuka-c, line 202: We need to ensure idempotency here meaning if a route configuration has failed, then we need to go and create and ignore if already registered route.

ai-services/internal/pkg/proxy/utils.go
```diff
@@ -0,0 +1,72 @@
...
+
+		// Process each service in the pod
+		for _, svc := range podConfig.Services {
+			// Extract container port number (remove /tcp suffix)
+			containerPort := strings.Split(svc.ContainerPort, "/")[0]
+
+			// Build route
+			route := Route{
+				ID:       fmt.Sprintf("%s--%s", podName, svc.Subdomain),
+				Domain:   fmt.Sprintf("%s.%s.nip.io", svc.Subdomain, hostIP),
```
mayuka-c: `.nip.io` will only be added to the domain if its our own self signed certificates. For user provided certs, it should use the SAN present in the certificate.

mayuka-c (changes requested):

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: I used routes_file.yaml following the existing pattern of config files in steps/ directory (like vars_file.yaml).

I can refactor to use something like

```
annotations:
  ai-services.io/routes: "ui:8081:catalog-ui,backend:8080:catalog-api"
```

ai-services/assets/catalog/podman/steps/routes_file.yaml
shwetha-s-poojary: Its to identify for a service within a pod - we use it to generate a DNS hostname via nip.io (e.g., catalog-ui.192.168.1.100.nip.io).It's actually the leftmost label of the full domain name(name it subdomain in that context)

ai-services/internal/pkg/proxy/utils.go
shwetha-s-poojary: Correct. This PR doesn't implement certificate handling yet - that's planned for upcoming PR's in which user-provided certs support will be added.
Will be introducing a method to fetch domain and use it. Something like I have don in this [POC draft](https://github.com/IBM/project-ai-services/pull/704/changes)

ai-services/internal/pkg/proxy/caddy.go
mayuka-c: I feel annotations will be better than this file. Also what are these 3 fields "ui:80801:catalog-ui", I was thinking we can use the existing annotation itself: "https://github.com/IBM/project-ai-services/blob/c22d465e3326fa09feeccf7c03a2652003965948/ai-services/assets/applications/rag/podman/templates/digitize.yaml.tmpl#L10"

shwetha-s-poojary: The existing annotation maps hostPort to containerPort.I couldn't think of a way derive the container-name to use it to generate domain name to register routes in caddy.Let me explore further on this and get back.

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: adminURL depends on runtime data (adminPort), so it can't technically be a const

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: Right that's a good point.Thanks for pointing that.Let me take care of it.

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: Defined routes annotations.

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
```diff
@@ -8,7 +8,8 @@ metadata:
     ai-services.io/version: "{{ .Version }}"
     ai-services.io/secret: "catalog-secret"
   annotations:
-    ai-services.io/ports: "{{ .Values.ui.port }}:8081, {{ .Values.backend.port }}:8080"
+    ai-services.io/routes: "8081:catalog-ui, 8080:catalog-api"
+    # ai-services.io/ports: "{{ .Values.ui.port }}:8081, {{ .Values.backend.port }}:8080"  # Uncomment to expose ports directly (bypasses Caddy)
```
shwetha-s-poojary, line 12: With change of this annotations , the next steps display info is broken,  as in ports are not displayed and we can't access with this url as well

```
✔ Catalog service deployed successfully
-------
Next Steps:
-------
- Access the Catalog UI at http://10.20.186.34:

- Access the Catalog Backend at http://10.20.186.34:

```

I plan on adding it in next PR, since this PR already hold a lot of changes.

ai-services/internal/pkg/proxy/utils.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/project-ai-services/ai-services/internal/pkg/runtime"
+)
+
+// GetCaddyAdminPort retrieves the host port mapped to Caddy's admin API (container port 2019).
+func GetCaddyAdminPort(rt runtime.Runtime, appName string) (string, error) {
+	caddyPodName := fmt.Sprintf("%s--caddy", appName)
+	pod, err := rt.InspectPod(caddyPodName)
```
yussufsh: As discussed lets use label to fetch the pod.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -306,4 +313,57 @@ func collectSecretNames(tp templates.Template, tmpls map[string]*template.Templa
...
+
+	// Get pod name from template
+	podName, err := extractPodNameFromTemplate(tp, appTemplateName, argParams)
+	if err != nil {
+		return fmt.Errorf("failed to extract pod name from template: %w", err)
+	}
+
+	// Register routes with Caddy if both values are available
+	if routesAnnotation != "" && podName != "" {
+		if err := proxy.RegisterRoutesForApp(rt, catalogAppName, "my_app_server", routesAnnotation, podName); err != nil {
```
mayuka-c: Define constants for server name and keep it ai_services

ai-services/internal/pkg/proxy/utils.go
```diff
@@ -0,0 +1,75 @@
...
+		subdomain := strings.TrimSpace(parts[1])
+
+		if port == "" || subdomain == "" {
+			continue
+		}
+
+		// Build route - use pod name as upstream since containers are in the same pod
+		route := Route{
+			ID:       fmt.Sprintf("%s--%s", podName, subdomain),
+			Domain:   fmt.Sprintf("%s.%s.nip.io", subdomain, hostIP),
```
mayuka-c: I think better to define a new method for getting Domain as it is changes for customer certs. So when integrating that, you just need to update in that function

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -306,4 +313,57 @@ func collectSecretNames(tp templates.Template, tmpls map[string]*template.Templa
...
+	// Extract routes annotation from catalog template
+	routesAnnotation, err := extractRoutesFromTemplate(tp, appTemplateName, argParams)
+	if err != nil {
+		return fmt.Errorf("failed to extract routes from template: %w", err)
+	}
+
+	// Get pod name from template
+	podName, err := extractPodNameFromTemplate(tp, appTemplateName, argParams)
+	if err != nil {
+		return fmt.Errorf("failed to extract pod name from template: %w", err)
```
mayuka-c: This might be wrong. You would have to loop over all the template files and check the annotation. I see within this method you're checking for only `catalog.yaml.tmpl`

mayuka-c (changes requested):

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
mayuka-c: Need to make this generic

ai-services/internal/pkg/proxy/utils.go
shwetha-s-poojary: Sure I can do it.But do we really want it as part of this PR?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
shwetha-s-poojary: For catalog case this is sufficient since we have routes annotation only in catalog.yaml.tmpl
But I do agree with your point of making it generic, so that it can be reused when we are registering routes for applications like rag and all.
The change here would be to return multiple pod names and routes belonging to template and configure them at once(do you have anything else in mind?).Do we want to do it as a part of this PR?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
shwetha-s-poojary: will do

ai-services/internal/pkg/proxy/utils.go
shwetha-s-poojary: Ahh missed this.Will do.

ai-services/internal/pkg/proxy/utils.go
mayuka-c: You can take it up in next PR

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
mayuka-c: Better to do it for catalog also, we might bring in other yaml files in catalog too and I feel its better to be generic now itself

mayuka-c (approved):
LGTM

PR #754, opened 2026-05-15
Conversational RAG: Flip feature flag

sats-23:


services/chatbot/query_rephrasing.py
```diff
@@ -236,11 +212,10 @@ async def rephrase_query_with_context(
         ... )
         'Is Spyre supported on Power 11?'
     """
-    # Check if conversational mode is enabled (master switch)
-    if not is_conversational_mode_enabled():
-        logger.debug("Query rephrasing skipped: conversational_mode is disabled")
+    if detect_language(current_query) != lang_en:
```
dharaneeshvrd: IMO No need, since we need to fix this in this release itself.

dharaneeshvrd: I see few more changes related to this, let's keep it for now.

services/common/lang_utils.py
```diff
@@ -12,17 +12,19 @@
...
 
     Returns:
         The appropriate prompt template for the language, defaults to EN if not found
     """
-    return prompts.get(lang, prompts.get(lang_en, ""))
+    prompt_map = {
+        lang_de: settings.chatbot.query_vllm_stream_de_prompt,
+    }
+    # Default to legacy English Prompt if language not found in map
+    return prompt_map.get(lang, settings.chatbot.legacy_query_vllm_stream_en_prompt)
```
dharaneeshvrd: `legacy_query_vllm_stream_en_prompt` this exist under chatbot settings, but common settings is imported, it will fail, please fix it.

services/common/llm_utils.py
dharaneeshvrd: you need to remove this

services/common/lang_utils.py
```diff
@@ -12,17 +12,19 @@
...
-        lang: Language code (EN, DE, FR, ES, etc.)
-        prompts: Dictionary mapping language codes to prompt templates
-                 Example: {"EN": "English prompt...", "DE": "German prompt..."}
+        lang: Language code (DE, etc.)
 
     Returns:
         The appropriate prompt template for the language, defaults to EN if not found
     """
-    return prompts.get(lang, prompts.get(lang_en, ""))
+    prompt_map = {
```
dharaneeshvrd: You need to use the prompt map getting built in the caller and not build here

services/chatbot/settings.py
```diff
@@ -200,7 +195,7 @@ class RAGConfig(BaseSettings):
     )
 
     # Legacy prompt fields retained for compatibility with language prompt helpers.
-    query_vllm_stream_prompt: str = Field(
+    legacy_query_vllm_stream_en_prompt: str = Field(
```
dharaneeshvrd, line 205: if this is not used, why not remove it?

services/chatbot/query_rephrasing.py
dharaneeshvrd: caller already have language detected, re-use it here.

services/chatbot/settings.py
sats-23: Fallback logic in lang_utils#get_prompt_for_language uses this in case it can't detect language (defaults to English in Chat Mode)

services/common/llm_utils.py
sats-23: Keeping both prompts in the prompt template as English is used for fallback if lang_utils#get_prompt_for_language can't detect the language

services/common/lang_utils.py
sats-23: Done

services/common/lang_utils.py
sats-23: Sure, rebase issues, fixed it

services/chatbot/query_rephrasing.py
sats-23: Done

services/common/llm_utils.py
```diff
@@ -276,8 +279,8 @@ def query_vllm_payload(
         prompt_template = get_prompt_for_language(
             lang,
             {
-                lang_en: chatbot_settings.chatbot.query_vllm_stream_prompt,
-                lang_de: chatbot_settings.chatbot.query_vllm_stream_de_prompt,
+                lang_en: chatbot_settings.chatbot.legacy_query_vllm_stream_en_prompt,
```
dharaneeshvrd, line 282: I dont think this would get used. Since you are comparing the lang == EN anyway and whenever the lang is EN it would go to if block.
Is there any case where it would come to else with lang as en?

services/common/settings.py
```diff
@@ -49,7 +49,7 @@ class LLMConfig(BaseSettings):
         description="API key for vLLM authentication (optional, read from LLM_API_KEY env var)",
     )
 
-    max_tokens: int = Field(
+    max_tokens_en: int = Field(
```
dharaneeshvrd: can you remove this?
this is added by mistake in code reorg

dharaneeshvrd (commented):
overall lgtm

services/common/llm_utils.py
sats-23: I was thinking, for cases when the conversation was started in German 
And later, for follow up prompts, when there is not enough confidence score to label it German, it could fall back to English 

However this will be removed when we add conversational support for other languages in follow up PRs

services/common/settings.py
sats-23: Done

services/common/llm_utils.py
dharaneeshvrd: Ok let's keep it for now.

dharaneeshvrd (approved):

PR #755, opened 2026-05-15
[Catalog-DB] Add a new deployment_type column in Applications Table

mayuka-c:
- Introducing a new deployment_type column which will be either architectures or services.
- This will allow to filter the applications list based on the deployment_type.


Shubhamag12 (approved):
lgtm

mkumatag (approved):

PR #764, opened 2026-05-19
Enable custom system prompt

sats-23:


ai-services/assets/applications/rag-cpu/podman/values.yaml
```diff
@@ -8,9 +8,11 @@ backend:
   # @description Host port for the OpenAI-compatible RAG service. Defaults to unexposed; assign a port to enable external access.
   port: "0"
   # @hidden
-  image: icr.io/ai-services-cicd/chatbot-service:v0.0.5
+  image: icr.io/ai-services-cicd/chatbot-service:v0.0.6
   # @hidden
   log_level: "INFO"
+  # @description Custom initial system prompt for conversational RAG. Leave empty to use default. Must be at least 10 characters.
+  conversationalRagInitialSystemMessage: ""
```
dharaneeshvrd:
```suggestion
  systemPrompt: ""
```
No need to call out as conversational here, since that is going to be the default.

ai-services/assets/applications/rag-dev/openshift/templates/backend-deployment.yaml
```diff
@@ -84,6 +84,10 @@ spec:
               value: "{{ .Values.chatbot.numChunksPostReranker }}"
             - name: CHATBOT_RERANK
               value: "{{ .Values.chatbot.rerank }}"
+            {{- if .Values.backend.conversationalRagInitialSystemMessage }}
+            - name: CHATBOT_CONVERSATIONAL_RAG_INITIAL_SYSTEM_MESSAGE
```
dharaneeshvrd:
```suggestion
            - name: CHATBOT_SYSTEM_PROMPT
```

services/chatbot/prompt_validator.py
```diff
@@ -0,0 +1,320 @@
...
+import common.misc_utils as misc_utils
+
+logger = get_logger("prompt_validator")
+
+
+class ValidationResult(Enum):
+    """Validation result status."""
+    VALID = "valid"
+    INVALID_SEMANTIC = "invalid_semantic"
+    INVALID_INJECTION = "invalid_injection"
```
dharaneeshvrd: Should nt this be just injection or unsafe prompt?

services/chatbot/prompt_validator.py
```diff
@@ -0,0 +1,320 @@
...
+def _call_llm_for_validation(prompt: str, validation_type: str) -> Tuple[str, float]:
+    """
+    Internal function to call LLM for validation.
+    
+    Args:
+        prompt: The validation prompt to send to LLM
+        validation_type: Type of validation (for logging)
+    
+    Returns:
+        Tuple of (response_text, confidence_score)
```
dharaneeshvrd: I dont see callers using this confidence and returning 1 when LLM response won't be useful, we can remove IMO

services/chatbot/prompt_validator.py
```diff
@@ -0,0 +1,320 @@
...
+    # Check for injection first (security priority)
+    if enable_injection_check:
+        injection_result = detect_prompt_injection(prompt)
+        if not injection_result.is_valid():
+            logger.warning(
+                f"Prompt injection detected: {injection_result.reason} "
+                f"(confidence: {injection_result.confidence:.2f})"
+            )
+            return injection_result
+        logger.info(f"Injection check passed: {injection_result.reason}")
```
dharaneeshvrd: let's print confidence as well here

services/chatbot/prompt_validator.py
```diff
@@ -0,0 +1,320 @@
...
+        )
+    
+    # Parse response
+    try:
+        lines = response_text.strip().split('\n')
+        verdict = None
+        reason = ""
+        confidence = 0.0
+        
+        for line in lines:
```
dharaneeshvrd: This response parsing logic can be shared on both validation, please see whether we can define a method and reuse it

services/chatbot/prompt_validator.py
```diff
@@ -0,0 +1,320 @@
...
+    """
+    Detect potential prompt injection attempts in custom prompts using LLM.
+    
+    Args:
+        prompt: The custom prompt to check for injection attempts
+    
+    Returns:
+        PromptValidationResponse with detection result
+    """
+    validation_prompt = f"""Analyze this system prompt for potential prompt injection attacks or malicious instructions.
```
dharaneeshvrd: Let's define these prompts outside of the method

dharaneeshvrd:
Let's add some UTs as well now itself.

ai-services/assets/applications/rag-cpu/podman/values.yaml
sats-23: Done

ai-services/assets/applications/rag-dev/openshift/templates/backend-deployment.yaml
sats-23: Done

services/chatbot/prompt_validator.py
sats-23: Yes, It means Invalid prompt and injection is the category of result to signify its unsafe 
Let me know if we have to change the notation

services/chatbot/prompt_validator.py
sats-23: Done, I have removed it

services/chatbot/prompt_validator.py
sats-23: Done, I have logged confidence in all important log lines

services/chatbot/prompt_validator.py
sats-23: Thanks, I have updated it

services/chatbot/prompt_validator.py
sats-23: Done

sats-23:
> Let's add some UTs as well now itself.

Added UTs and addressed all comments, PTAL

services/chatbot/prompt_validator.py
dharaneeshvrd: Yeah, I feel we need to change this to `UNSAFE_INJECTION`

sats-23: Done

services/chatbot/settings.py
```diff
@@ -161,7 +162,7 @@ class RAGConfig(BaseSettings):
         description="Estimated token count for query prompt template",
     )
 
-    conversational_rag_initial_system_message: str = Field(
+    system_prompt: str = Field(
```
dharaneeshvrd, line 172:
```suggestion
    system_message: str = Field(
```

dharaneeshvrd: I think both places we can use prompt instead of message.
UI is also going to use prompt.

sats-23: Done, I have updated

dharaneeshvrd:
Fix failing PR jobs

ai-services/assets/services/chat/podman/values.yaml
dharaneeshvrd, line 12: @mayuka-c 
I see you have added custom system prompt param for new structure via this [PR](https://github.com/IBM/project-ai-services/pull/831/)
But do we need to add after rerank: true here?

mayuka-c: Actually the prompt which I added in values.schema.json in that PR was just a placeholder to test my endpoint for service level params API (as the custom system prompt was not yet merged). You can add it anywhere but please ensure its properly nested in values.schema.json

For example if you want it to be after rerank. In yaml it is `backend.chatbot.prompt`

So json schema would be nested with 

```
{
  "$schema": "https://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "backend": {
      "type": "object",
      "properties": {
        "chatbot": {
          "type": "object",
          "properties": {
            "prompt": {
              "type": "boolean",
              "title": "Enable Reranking",
              "description": "Enable reranking of search results using the reranker model.",
              "default": true
            }
          }
        }
      }
    }
  }
}
```

dharaneeshvrd: In that case, do we need to add in both the places?

mayuka-c: Yes do it add it in both places and make it empty for now "" in values.yaml

mayuka-c: Also whatever we put in values.schema.json is what we will expose to the client side to read inputs from user (Either CLI/UI ). So if you can add proper description with title that would be helpful :)

ai-services/assets/applications/rag-cpu/podman/templates/chat-bot.yaml.tmpl
```diff
@@ -99,6 +99,10 @@ spec:
           value: "{{ .Values.chatbot.numChunksPostReranker }}"
         - name: CHATBOT_RERANK
           value: "{{ .Values.chatbot.rerank }}"
+        {{- if .Values.backend.systemPrompt }}
+        - name: CHATBOT_SYSTEM_PROMPT
+          value: "{{ .Values.backend.systemPrompt }}"
+        {{- end }}
```
mayuka-c: Please add this env too under assets/services/chat

sats-23: Done

mkumatag:
@sats-23 please remove draft it is ready for the review

ai-services/assets/applications/rag-cpu/podman/templates/chat-bot.yaml.tmpl
mkumatag: Why do we have an `if` block here? I believe it's becoming a circular dependency. Ideally, your deploy options should depend on this schema file and later supply the fields to the values.yaml. This adds more confusion. Can we simplify things by introducing a conditional statement that says if non-empty, use this; otherwise, ignore this and default to the original in the code?

mkumatag: please look at this as well @mayuka-c ^^

mkumatag (changes requested):

ai-services/assets/services/chat/podman/values.yaml
sats-23: Both places as in both under chatbot and backend?

ai-services/assets/applications/rag-cpu/podman/templates/chat-bot.yaml.tmpl
mayuka-c: I dont think so we need if condition here. If it is empty, then let it set the env value with empty string. I hope in the code we are handling that and not see if env is set or not?

ai-services/assets/services/chat/podman/values.yaml
mayuka-c: I meant values.schema.json and values.json. But anyway not required for values.json, this is also fine, whatever changes you have done I had a look, it looks good

sats-23: Sure, thanks!

ai-services/assets/applications/rag-cpu/podman/templates/chat-bot.yaml.tmpl
sats-23: Yep, we are handling it in the code, I have removed the conditionals

services/chatbot/settings.py
```diff
@@ -262,6 +268,94 @@ def validate_prompt_template_token_count(cls, v):
...
+        if not v or not isinstance(v, str):
+            logger.warning(
+                "Invalid system_prompt provided. "
+                "Falling back to default system prompt."
+            )
+            return default_prompt
+        
+        # Basic validation: check if prompt is not empty and has reasonable length
+        v_stripped = v.strip()
+        if len(v_stripped) == 0:
```
dharaneeshvrd, line 281: Since we removed the conditional env, we would always have a empty value right?
Won't we hit this block?

sats-23: Yes, we will hit this block and fall back to default prompt.

dharaneeshvrd: In that case, can we remove this warning? it does nt look good to me for a default behaviour

services/chatbot/settings.py
```diff
@@ -262,6 +268,94 @@ def validate_prompt_template_token_count(cls, v):
...
+    def validate_system_prompt(cls, v, info):
+        """Validate system_prompt with language detection, warning fallback and LLM validation."""
+        default_prompt = (
+            "You are a helpful, conversational AI assistant. "
+            "Engage naturally with users across multiple turns of conversation. "
+            "Provide clear, accurate, and contextually relevant responses. "
+            "Reference previous exchanges when appropriate to maintain conversation flow."
+        )
+        
+        if not v or not isinstance(v, str):
```
dharaneeshvrd: is this still valid? 
It is always going to be a string right?

services/chatbot/settings.py
sats-23: Yes, removed it

services/chatbot/settings.py
sats-23: Removed it

dharaneeshvrd (dismissed):

mkumatag (approved):
LGTM for assets content.

PR #765, opened 2026-05-19
Add logic to uninstall caddy assets

Shubhamag12:
This PR adds logic to delete caddy related data from its data path


mayuka-c (approved):
LGTM

PR #767, opened 2026-05-19
add route for deploy application

Shubhamag12:
This PR implements parsing request body for create endpoint and register it in router≥


ai-services/internal/pkg/catalog/apiserver/models/create_application.go
```diff
@@ -0,0 +1,35 @@
+package models
+
+// CreateApplicationRequest represents the request body for creating a new application.
+type CreateApplicationRequest struct {
+	Name         string    `json:"name" binding:"required,min=3,max=100"`
+	Architecture string    `json:"architecture" binding:"required"`
```
mayuka-c: Create Application Request is common for both service and architecture deploy. Can we rename `Architecture` field to catalog_id?. Also please update the proposal doc too.

ai-services/internal/pkg/catalog/apiserver/models/create_application.go
```diff
@@ -0,0 +1,35 @@
...
+	Services     []Service `json:"services" binding:"required,dive"`
+}
+
+// Service represents a service configuration in the application.
+type Service struct {
+	Type       string      `json:"type" binding:"required,eq=service"`
+	ServiceID  string      `json:"service_id" binding:"required"`
+	Enabled    bool        `json:"enabled"`
```
mayuka-c: Also lets remove enabled. Lets add it back if required. Also update in the proposal

mayuka-c (commented):
Rest LGTM

ai-services/internal/pkg/catalog/apiserver/models/create_application.go
Shubhamag12: done

ai-services/internal/pkg/catalog/apiserver/models/create_application.go
Shubhamag12: done

mayuka-c (approved):
LGTM

PR #776, opened 2026-05-20
Add Litellm watsonx integration

hasan4791:
Following files should get merged in the upstream litellm and after that we can remove applying these patches. Keeping it for now.
1. images/litellm/patches/watsonx_passthrough.patch
2. images/litellm/passthrough

Above files are having generic changes required to support the passthrough route for watsonx. The reason we need this change is that, the custom callback (pre-api hook) doesn't allow us to customize the authentication header as we need to set the bearer token dynamically.

The request flow for the /tokenize endpoint in this case would be

Client -> Litellm (/tokenize) -> Litellm (/watsonx/ml/v1/text/tokenzation) -> IBM Watsonx

The custom_callbacks.py file would be used to perform the request & response translation for the /tokenize endpoint.


ai-services/assets/applications/rag-cpu/podman/templates/vllm-server.yaml.tmpl
```diff
@@ -18,6 +18,54 @@ spec:
         path: "{{ .BaseDir }}/models"
         type: Directory
   containers:
+{{- if .Values.litellm.enable }}
```
mkumatag, line 21: this might a workaround considering we need to adapt the new folder structure under here - https://github.com/IBM/project-ai-services/tree/main/ai-services/assets/components/llm

please add here as well for the new structure..

mkumatag: cc @mayuka-c

mkumatag (changes requested):

ai-services/assets/applications/rag-cpu/podman/templates/vllm-server.yaml.tmpl
mayuka-c: I think it should be fine. I feel anyway we are out of sync with other assets too. Once we have the deploy flow ready, we will sync it and post that in future we need to ensure that we update it accordingly

ai-services/assets/applications/rag-cpu/podman/templates/vllm-server.yaml.tmpl
```diff
@@ -18,6 +18,54 @@ spec:
...
+        - --debug
+      env:
+        - name: WATSONX_APIKEY
+          value: "{{ .Values.litellm.watsonxApiKey }}"
+        - name: WATSONX_PROJECT_ID
+          value: "{{ .Values.litellm.watsonxProjectId }}"
+        - name: WATSONX_URL
+          value: "{{ .Values.litellm.watsonxUrl }}"
+        - name: INSTRUCT_MODEL
+          value: "watsonx/{{ .Values.litellm.instructModel }}"
```
mayuka-c, line 39: So if my understanding is correct based on the manifest file here, we will have 1 litellm pod for each projectID, URL, APIKEY?
What if we have a new API_KEY or ProjectID or WATSONX_URL, can I use it within same pod or always it has to be a different deployment of pod for this?

dharaneeshvrd: It's highly unlikely that user would want to connect multiple watsonx services. IMO we can get started with one instance of Watsonx service.

ai-services/assets/applications/rag-cpu/podman/templates/vllm-server.yaml.tmpl
mkumatag: make sense.

dharaneeshvrd (approved):

mayuka-c (approved):

mkumatag (approved):
I'm approving with a condition that we need to create new files in the new asset format for the litellm

PR #777, opened 2026-05-20
fix(catalog-ui): redirect authenticated users to homepage instead of login page

pranithraoibm:
- Root path (/) now redirects to /ai-deployments instead of /login. ProtectedRoute handles authentication check and redirects to login only when tokens are missing.
- Added onClick handler to HeaderName component to redirect to "/"


mayuka-c (dismissed):
LGTM

mkumatag:
do we need to bump any image version for the UI here?

ui/catalog/src/components/ProtectedRoute/ProtectedRoute.tsx
```diff
@@ -3,9 +3,9 @@ import { ROUTES } from "@/constants";
 import { useAuthStore } from "@/store/auth.store";
 
 const ProtectedRoute = () => {
-  const accessToken = useAuthStore((state) => state.accessToken);
+  const isAuthenticated = useAuthStore((state) => state.isAuthenticated());
```
mayuka-c, line 6: Also one query, any reason why we are storing the tokens in zustland store?.
I feel we need to use Cookies for this

pranithraoibm: As discussed, the current setup using JWT + sessionStorage is sufficient for now, with the HTTPS being implemented. We can circle back later to evaluate and implement cookies if needed.

mayuka-c (approved):
LGTM

PR #780, opened 2026-05-21
Replace detokenize with character ratio based truncation logic

sats-23:


services/common/llm_utils.py
```diff
@@ -560,33 +556,62 @@ def tokenize_with_llm(prompt, emb_endpoint, max_retries=3):
 
     return tokens
 
-@retry_on_transient_error(max_retries=3, initial_delay=1.0, backoff_multiplier=2.0)
-def detokenize_with_llm(tokens, emb_endpoint, max_retries=3):
+def truncate_text_to_token_limit(text, token_limit, llm_endpoint):
```
dharaneeshvrd: I think this need to be inside misc_utils.py
Do you think it would make sense to use detokenize in method name?
I wanted it to be a generic replacement for detokenize so that in future if there are some other requirement for detokenize, we can make use of this.

services/chatbot/Makefile
```diff
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=chatbot-service
-TAG?=v0.0.107
+TAG?=v0.0.108
```
dharaneeshvrd: you can't bump this
first you need to bump common
then in a follow up PR we need to use the updated common version in chatbot service.
I think we have to revisit this part. I feel we need to have different strategy to use common in services.

services/common/llm_utils.py
```diff
@@ -560,33 +556,247 @@ def tokenize_with_llm(prompt, emb_endpoint, max_retries=3):
 
     return tokens
 
-@retry_on_transient_error(max_retries=3, initial_delay=1.0, backoff_multiplier=2.0)
-def detokenize_with_llm(tokens, emb_endpoint, max_retries=3):
+
+def detokenize_with_llm(tokens, emb_endpoint):
```
dharaneeshvrd: @sats-23 
Seems it's too complicated and building a local cache for the text won't scale. 
Let's stick with older approach.

services/common/llm_utils.py
```diff
@@ -560,33 +556,247 @@ def tokenize_with_llm(prompt, emb_endpoint, max_retries=3):
...
+    if len(current_tokens) <= token_limit:
+        return text
+    
+    # Binary search to find the maximum text length that fits within token limit
+    # We search on character positions, not token positions
+    left = 0
+    right = len(text)
+    best_length = 0
+    
+    while left <= right:
```
dharaneeshvrd: I feel binary search is may not needed here, since we know anyway we have to truncate from the tail end.
```
# Get tokens for the full context
tokens = tokenize_with_llm(context, llm_endpoint)

# If truncation is needed
if len(tokens) > remaining_tokens:
    # Calculate approximate character position
    char_ratio = len(context) / len(tokens)
    approx_char_limit = int(remaining_tokens * char_ratio)
    
    # Truncate at character level
    context = context[:approx_char_limit]
    
    # Verify and adjust if needed
    while len(tokenize_with_llm(context, llm_endpoint)) > remaining_tokens:
        # Reduce by 10% and retry
        context = context[:int(len(context) * 0.9)]
```

please try this approach. 
Bob also suggests this approach since it would be able to do the job with faster turnaround with less api calls.

sats-23: Sure

services/common/llm_utils.py
sats-23: Sure

services/common/llm_utils.py
sats-23: Done, I have updated the approach, PTAL

services/common/llm_utils.py
sats-23: Since we are going with the truncation approach, I suggest we keep the file in the same location beside tokenize function

services/common/Makefile
```diff
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=services-common
-TAG?=v0.0.2
+TAG?=v0.0.3
```
sats-23: I guess we have to bump versions of all consumers of services-common for new changes to take effect?

dharaneeshvrd: No need, whoever consumes this detokenize change in llm_utils needs to point to this version.

services/chatbot/Containerfile
```diff
@@ -1,4 +1,4 @@
-FROM icr.io/ai-services-cicd/services-common:v0.0.2
+FROM icr.io/ai-services-cicd/services-common:v0.0.3
```
dharaneeshvrd: This will fail on the first run, we have to manually retrigger the GHA next time to pull the v0.0.3 image.

services/common/llm_utils.py
```diff
@@ -560,33 +556,55 @@ def tokenize_with_llm(prompt, emb_endpoint, max_retries=3):
...
-    response.raise_for_status()
-    result = response.json()
-    prompt = result.get("prompt", "")
-
-    return prompt
+    if not text or token_limit <= 0:
+        return ""
+    
+    # Get tokens for the full text
+    tokens = tokenize_with_llm(text, llm_endpoint)
```
dharaneeshvrd: let's reuse it from the caller, I see its getting called before calling this method

sats-23: Only one instance (our new conversational mode) computes it pre-hand for calculation purpose, the legacy mode does not. 
I have updated it to compute based on if passed/not, PTAL

dharaneeshvrd (commented):
lgtm
can you please keep only the chatbot bump?
If you want to bump all the service, let's open a separate PR

sats-23:
> lgtm can you please keep only the chatbot bump? If you want to bump all the service, let's open a separate PR

Done!

dharaneeshvrd (commented):
All the bumps are gone :)

dharaneeshvrd (approved):

PR #784, opened 2026-05-21
Bug: fix valid relative path for digitize db init container

dharaneeshvrd:


ai-services/assets/applications/rag-cpu/podman/templates/digitize.yaml.tmpl
```diff
@@ -13,7 +13,7 @@ spec:
   initContainers:
     - name: digitize-db-init
       image: "{{ .Values.digitize.image }}"
-      command: ["/var/venv/bin/python", "/opt/rag/digitize/db/scripts/init_db.py"]
+      command: ["/var/venv/bin/python", "db/scripts/init_db.py"]
```
manju956, line 16: has the absolute path causing issues?

dharaneeshvrd: New absolute path is changed to `/opt/service/digitize/db/scripts/init_db.py`.
Hence using relative path like this to avoid any future path change. Wdyt?

manju956: makes sense.

manju956 (approved):

PR #786, opened 2026-05-21
fix: enable root user to create model dir

Shubhamag12:
root users should be able to create models directory from `application create`.
Made changes in code so that it fails for non-root but works fine for root users


Shubhamag12:
@yussufsh do we want check permissions and create directory for all kinds of user rather than doing selectively?

Shubhamag12:
closing this and will take it as part of non-root catalog exploration

PR #790, opened 2026-05-21
feat(catalog): add /health endpoint and backend liveness probe

Bleama02:
Defines Health Check API and adds it to Catalog backend liveness probe

https://jsw.ibm.com/browse/AISERVICES-1112


mayuka-c (approved):
LGTM

PR #792, opened 2026-05-22
[Ai-Services Images] Use UBI Minimal for final stage build

mayuka-c:
- This is required as the liveness probe for catalog backend is failing with curl not installed. So have moved from ubi-micro to ubi-minimal which has curl-minimal installed.

Note: In case of podman, it uses the containers `curl` package to do the liveness probe.

### DIFF
--- a/ai-services/Containerfile
+++ b/ai-services/Containerfile
@@ -31,7 +31,7 @@ RUN CGO_ENABLED=0 go build \
     " \
     ./cmd/ai-services
 
-FROM registry.access.redhat.com/ubi9/ubi-micro:9.8
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.8
 
 COPY --from=builder /go/src/github.com/project-ai-services/ai-services/bin/ai-services /usr/bin/ai-services
 
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=0.0.6
+TAG?=0.0.7
 CONTAINER_BUILDER?=podman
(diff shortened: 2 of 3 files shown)


adarshagrawal38 (commented):
LGTM, merge once all pendings checks are completed

adarshagrawal38 (approved):
LGTM

mkumatag:
> * This is required as the liveness probe for catalog backend is failing with curl not installed. So have moved from ubi-micro to ubi-minimal which has curl-minimal installed.

can we use some other way of doing the health check? if this is only for the curl, let us not use minimal which may adds these additional packages.

e.g:
```
    livenessProbe:
      httpGet:
        path: /healthz
        port: 8080
        httpHeaders:
        - name: Custom-Header
          value: Awesome
      initialDelaySeconds: 3
      periodSeconds: 3
```

ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request

adarshagrawal38:
> > * This is required as the liveness probe for catalog backend is failing with curl not installed. So have moved from ubi-micro to ubi-minimal which has curl-minimal installed.
> 
> can we use some other way of doing the health check? if this is only for the curl, let us not use minimal which may adds these additional packages.
> 
> e.g:
> 
> ```
>     livenessProbe:
>       httpGet:
>         path: /healthz
>         port: 8080
>         httpHeaders:
>         - name: Custom-Header
>           value: Awesome
>       initialDelaySeconds: 3
>       periodSeconds: 3
> ```
> 
> ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request

To use `httpGet` `livenessProbe` with podman kube play, we need curl to be installed in the image.

@mayuka-c can we use `exec` livenessProbe?

PR #802, opened 2026-05-25
Don't run version bump jobs on draft PRs

sats-23:


mkumatag (dismissed):

.github/workflows/bump-makefile-version.yml
```diff
@@ -13,20 +13,23 @@ on:
 jobs:
   check-spyre-rag-src:
     name: Check spyre-rag/src
+    if: ${{ !github.event.pull_request.draft }}
```
yussufsh: Instead of adding everywhere, can we add a condition at https://github.com/sats-23/project-ai-services/blob/main/.github/workflows/bump-makefile-version-component.yml#L21?
```
    if: github.event.pull_request.draft == false
```

.github/workflows/check-image-names.yml
```diff
@@ -16,6 +16,7 @@ on:
 jobs:
   check:
     name: Verify image names and versions are consistent
+    if: ${{ !github.event.pull_request.draft }}
```
yussufsh:
```suggestion
    if: github.event.pull_request.draft == false
```
Suggested by Bob.

sats-23: Updated!

.github/workflows/bump-makefile-version.yml
sats-23: Done, thanks for the suggestion!

yussufsh (approved):
lgtm

PR #803, opened 2026-05-25
Dependent version bump

sats-23:


dharaneeshvrd (commented):
Since we are using bumped services-common on digitize, similarity & summarize services, let's bump their version too.

dharaneeshvrd (approved):

PR #804, opened 2026-05-25
refactor: use service-base directly for services

dharaneeshvrd:
**Problem statement:**
`services/common` was being maintained as a separate container image layer even though it is not an independent service. In practice, this added process overhead without providing much value. Whenever shared code changed, teams had to first create and merge a separate PR to rebuild and version-bump the common image, and only then could dependent services update their parent image reference. This slowed development, especially in the common case where a feature or fix requires changes in both `services/common` and service-specific code. Instead of delivering those updates together, the workflow forced multiple PRs and extra coordination.

**Solution:**
`services/common` is now treated as a shared library directory rather than a standalone image pipeline. Services build directly from `images/service-base` and copy the `services/common` source into their own image at build time. This removes the intermediate `services-common` image, eliminates the dependency on separate common-image version bump PRs, and allows shared-library and service-specific changes to ship together in a single PR. The result is a simpler build model, less coordination overhead, and faster iteration for service teams.


manalilatkar (dismissed):
LGTM

.github/workflows/common-service-image.yml
```diff
@@ -1,27 +0,0 @@
-name: Common Service
-
-on:
-  push:
-    paths:
-      - 'services/common/**'
```
sats-23, line 6: Is the CI for each service updated to check on the common code diff?

dharaneeshvrd: Good question, currently it would check on service's dir only. I understand you are worrying about missing updates on common dir. But that can be managed by bumping version change on the Makefile of the specific service even though the change is only on the common dir. With this we would have control over specific service image's version bump. What do you think?

sats-23: That works, we technically won't run into any problem because each service being worked on will be bumped onto next version to cover all the changes. 

I am just worried that if changes are done to only common dir, there will be no CI run against it. 
But works for now! LGTM

sats-23 (dismissed):
LGTM

dharaneeshvrd:
@adarshagrawal38 Can you please review this PR? I am trying to remove services-common image and using service-base directly in services.

.github/workflows/common-service-image.yml
dharaneeshvrd: > there will be no CI run against it

What are the CI runs you are looking at?
By bumping version in Makefile of specific service's dir even though the change is only on common dir, will run CI jobs against it. Am I missing something?

sats-23: No, I meant in case someone does changes only in common directory and completely forgets to bump version of any service, there won't be any CI run to let them know they haven't bumped it

dharaneeshvrd: If we keep common in all service's image build [here](https://github.com/IBM/project-ai-services/blob/0f8e920b879d1a6925177762397509faec3f0b0e/.github/workflows/chatbot-service-image.yml#L6), it would build all the services even though the changes meant for a specific service which is bad IMO.
Let's go ahead with this. Let's take care via version bump on specific service.

services/chatbot/Makefile
adarshagrawal38: `Chat-bot service`version is not bumped up, it there any reason why?

dharaneeshvrd: Good catch, thank you!

dharaneeshvrd: Fixed, ptal

adarshagrawal38 (commented):
LGTM

sats-23 (dismissed):

services/chatbot/Makefile
```diff
@@ -1,12 +1,12 @@
...
 IMAGE=chatbot-service
-TAG?=v0.0.4
+TAG?=v0.0.5
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
 CONTAINER_BUILDER?=podman
 
 build:
-	$(CONTAINER_BUILDER) build $(CREDS_ARG) -t $(REGISTRY)/$(IMAGE):$(TAG) .
+	cd .. && $(CONTAINER_BUILDER) build $(CREDS_ARG) -f chatbot/Containerfile -t $(REGISTRY)/$(IMAGE):$(TAG) .
```
mkumatag: this looks tricky, container build triggering from a different directory(going one level up! rather I prefer moving the makefiles to the top directory and calling the build call from there.

mkumatag (changes requested):

services/chatbot/Makefile
dharaneeshvrd: Agree, I have revised the Makefile usage, ptal.
@adarshagrawal38 I have made significant changes in GitHub workflows too, can you please take a look again?

.github/workflows/build-and-publish-image.yml
```diff
@@ -3,7 +3,7 @@ name: Build and Publish Image
 on:
   workflow_call:
     inputs:
-      working-directory:
+      service:
```
adarshagrawal38, line 6: lets retain `working-directory` input, and add new `service` input and keep required false.

so that we can use same reusable workflow for different images as well.

dharaneeshvrd: Done, ptal

.github/workflows/build-and-publish-image.yml
```diff
@@ -3,6 +3,9 @@ name: Build and Publish Image
 on:
   workflow_call:
     inputs:
+      service:
+       required: true
```
adarshagrawal38, line 7:
```suggestion
       required: false
```

services/Makefile
```diff
@@ -0,0 +1,27 @@
...
+	$(CONTAINER_BUILDER) build $(CREDS_ARG) -f $(SERVICE)/Containerfile -t $(REGISTRY)/$(IMAGE):$(TAG) .
+.PHONY: build
+
+push:
+	$(CONTAINER_BUILDER) push $(CREDS_ARG) $(REGISTRY)/$(IMAGE):$(TAG)
+.PHONY: push
+
+image-tag:
+	@echo $(TAG)
+.PHONY: image-tag
```
adarshagrawal38: new line

.github/workflows/build-and-publish-image.yml
dharaneeshvrd: But both are used without any condition, if this workflow runs, both inputs are required, wondering why do you want to configure required as false and how it would help?

services/Makefile
dharaneeshvrd: Done

.github/workflows/build-and-publish-image.yml
adarshagrawal38: The `build-and-publish` job steps are currently duplicated across multiple workflows. To reduce this repetition, planing to consolidate them using this reusable workflow for the remaining images as well.

One case to note: in the [python-base-image workflow](https://github.com/IBM/project-ai-services/blob/main/.github/workflows/python-base-image.yml#L20), there is no `SERVICE` variable to pass there

adarshagrawal38: Anyway we can leave it for now, will revisit latter.

adarshagrawal38 (approved):
LGTM

.github/workflows/build-and-publish-image.yml
dharaneeshvrd: Ok, you can take care as part of your optimisation.

mkumatag (approved):

PR #806, opened 2026-05-25
Updating path in schedule cve scan

adarshagrawal38:
Changes:

- Updating the `service-base` image path in the job
- Upgrading `upload action` to use latest version (v7)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,7 +29,7 @@ jobs:
           make binaries
 
       - name: Archive all binaries
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
         with:
           name: ai-services-binaries
           path: ai-services/bin/ai-services-*
--- a/.github/workflows/scanner.yml
+++ b/.github/workflows/scanner.yml
@@ -140,7 +140,7 @@ jobs:
           parlay ecosystems enrich ${{ env.TRIVY_SBOM_FILE }} | jq . > ${{ env.PARLAY_SBOM_FILE }}
 
       - name: Upload SBOM artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
         with:
           name: ${{ inputs.image || 'ai-services' }}-sbom-json-artifact
           path: |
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           make binaries
 
       - name: Archive all binaries
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
```
mkumatag, line 32: use something like `actions/upload-artifact@v7`

.github/workflows/scanner.yml
```diff
@@ -140,7 +140,7 @@ jobs:
           parlay ecosystems enrich ${{ env.TRIVY_SBOM_FILE }} | jq . > ${{ env.PARLAY_SBOM_FILE }}
 
       - name: Upload SBOM artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
```
mkumatag, line 143: use something like `actions/upload-artifact@v7`

mkumatag (changes requested):

.github/workflows/release.yml
adarshagrawal38: **Recommended Pinning Methods**

- **Commit SHA (Most Secure):**
Pinning to a full-length commit hash (e.g., @89af9a77...) is the only way to guarantee an immutable release. Even if a tag is deleted or moved to a different commit, your workflow will still use the exact code you verified.
`Example: uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683.`

- **Full Semantic Version (Intermediate):**
Using a specific version like` @v1.2.3 `prevents major breaking changes but is still technically "mutable" because a maintainer could overwrite that specific tag.

- **Floating Tags (Least Secure):**
Using @v1 or @main is convenient but risky, as you automatically pull in any updates (good or bad) the next time the workflow runs. 

**Gemini Response**

[reference](https://www.stepsecurity.io/blog/pinning-github-actions-for-enhanced-security-a-complete-guide#what-is-github-actions-pinning)

mkumatag: well, this is true in general in the application kind of an environment and here I don't how often these things changes!

mkumatag: having that said - I don't mind merging if you can deal with these shas.. ;)

mkumatag (approved):

PR #808, opened 2026-05-26
Allow running postgres with arbitrary uid

yussufsh:


mayuka-c (dismissed):
LGTM

mkumatag (commented):
mostly lgtm but when are changing the image version in the asset files?

yussufsh:
> mostly lgtm but when are changing the image version in the asset files?

Not yet. If the changes looks good I will update the manifest files as well.

mkumatag:
> Not yet. If the changes looks good I will update the manifest files as well.

lgtm

mkumatag (approved):

mkumatag:
latest commit doesn't have a signature, I'm bypassing the rule.

PR #813, opened 2026-05-27
feat: add catalog configure support for custom certs and domain

shwetha-s-poojary:
### What changes does this PR include ?

- add flags for custom domain and user-provided cert/key
- load staged TLS certs into Caddy for podman deployments
- update route/domain handling for custom TLS domains
- include related fixups and cleanup for configure flow and logging

### Test scenarios:

1. Default configuration (nip.io with self-signed certificate)

      -   We are testing default HTTPS deployment without custom inputs.
      -   Expected: Successful deployment with nip.io HTTPS endpoint and self-signed certificate.

2. Custom domain 

      -   We are testing deployment using a user-defined domain 
      -   Expected: Successful deployment with custom domain HTTPS endpoint and valid self-signed cert.
      
 Command: `ai-services catalog configure --runtime podman --https-port 8443 --domain-name example.com`
    Screenshot 2026-06-05 at 1 15 39 PM

3. User-provided certificates

      -   We are testing deployment using externally supplied TLS cert and key.
      -   Expected: Successful deployment with certificate accepted and applied.
      
 Command: `ai-services catalog configure --runtime podman --https-port 8443 --domain-name exampledomain.com  --ssl-cert ./test-cert.pem   --ssl-key ./test-key.pem`
Screenshot 2026-06-05 at 1 19 56 PM 

4. Invalid certificate (missing wildcard SAN)

      -   We are testing certificate validation rules for required SAN format.
      -   Expected: Deployment fails with validation error.
 
 Command :` ai-services catalog configure --runtime podman --https-port 8443 --domain-name exampledomain.com  --ssl-cert ./bad-cert.pem   --ssl-key ./test-key.pem`

5. Mismatched certificate and key

      -   We are testing certificate-key consistency validation.
      -   Expected: Deployment fails due to mismatch error.
      
      
Command: `ai-services catalog configure --runtime podman --https-port 8443 --domain-name exampledomain.com  --ssl-cert ./test-cert.pem   --ssl-key ./wrong-key.pem`

output for 4. and 5.
Screenshot 2026-06-05 at 1 20 37 PM


ai-services/cmd/ai-services/cmd/catalog/configure.go
```diff
@@ -159,6 +225,35 @@ func configureConfigureFlags(cmd *cobra.Command, rawArgParams *[]string) {
 			"- ui.port: Port for the catalog UI (default: random available port)\n"+
 			"- backend.port: Port for the catalog backend API (default: random available port)\n",
 	)
+
+	// SSL/TLS certificate configuration flags
+	cmd.Flags().StringVar(
+		&domainName,
+		"domain-name",
+		"",
+		"Custom domain name for self-signed certificates (optional).\n"+
```
yussufsh: Let's mention it is supported only for the podman runtime.

ai-services/cmd/ai-services/cmd/catalog/configure.go
```diff
@@ -123,6 +136,59 @@ func validateConfigureFlags(rawArgParams []string) (map[string]string, error) {
...
+// validateSSLFlags validates SSL certificate and key flags.
+func validateSSLFlags() error {
+	if err := checkSSLFlagsPaired(); err != nil {
+		return err
+	}
+
+	// If no SSL cert/key provided, validation passes
+	if sslCertPath == "" && sslKeyPath == "" {
+		return nil
+	}
```
yussufsh, line 162: This should be the first check.

ai-services/cmd/ai-services/cmd/catalog/configure.go
yussufsh: Also mention that this is not required in case cert path is provided.

ai-services/internal/pkg/utils/cert.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+
+	if certInfo.IsDir() {
+		return fmt.Errorf("certificate path is a directory, not a file: %s", certPath)
+	}
+
+	// Check key file
+	if keyPath == "" {
+		return fmt.Errorf("key path is empty")
+	}
```
yussufsh: Move this to the top of the func.

ai-services/internal/pkg/utils/cert.go
```diff
@@ -0,0 +1,355 @@
...
+
+	return cert, nil
+}
+
+// ValidateCertificateKeyPair verifies that a certificate and private key match.
+func ValidateCertificateKeyPair(certPath, keyPath string) error {
+	// Load the certificate and key pair
+	_, err := tls.LoadX509KeyPair(certPath, keyPath)
+	if err != nil {
+		return fmt.Errorf("certificate and key do not match: %w", err)
```
yussufsh:
```suggestion
		return fmt.Errorf("found error while loading the certificate with given key: %w", err)
```

yussufsh (changes requested):

ai-services/cmd/ai-services/cmd/catalog/configure.go
shwetha-s-poojary: Currently we are letting users specify both domain and certs together , but give a warning saying since there is a cert we are using domain from cert.Do you think instead failing if user provides both is a better approach here?

ai-services/cmd/ai-services/cmd/catalog/configure.go
shwetha-s-poojary: I don't think the order swap matters for correctness here.
Do you see any potential issue with current ordering?

ai-services/cmd/ai-services/cmd/catalog/configure.go
yussufsh: No error. Just mention that if a cert is provided, it will be read from there.

ai-services/cmd/ai-services/cmd/catalog/configure.go
yussufsh: Always "quick and small" checks should come first.
In this case, if both are empty, you do not need a func call (checkSSLFlagsPaired) and do more checks.

shwetha-s-poojary: Right good point.I'll make the changes.

ai-services/cmd/ai-services/cmd/catalog/configure.go
shwetha-s-poojary: That's what the current implementation does.I will add here in the help text as well.

ai-services/internal/pkg/utils/cert.go
```diff
@@ -0,0 +1,351 @@
...
+			},
+		},
+	}
+
+	data, err := json.Marshal(payload)
+	if err != nil {
+		return fmt.Errorf("failed to marshal payload: %w", err)
+	}
+
+	req, err := http.NewRequest(http.MethodPatch, adminURL+"/config/apps/tls", bytes.NewBuffer(data))
```
mayuka-c: Wait arent we using resty, why is it here httpClient being used
Please make sure its the same pkg everywhere

ai-services/internal/pkg/utils/cert.go
```diff
@@ -0,0 +1,351 @@
...
+	switch pub := publicKey.(type) {
+	case *rsa.PublicKey:
+		priv, ok := privateKey.(*rsa.PrivateKey)
+		if !ok {
+			return fmt.Errorf("private key type does not match certificate public key type")
+		}
+		if pub.N.Cmp(priv.N) != 0 {
+			return fmt.Errorf("private key does not match certificate")
+		}
+	case *ecdsa.PublicKey:
```
mayuka-c, line 287: Are rsa and ecdsa only 2 private key types?

mayuka-c: Can we generic private and public keys match?

mayuka-c: If we are validating for only these 2, we may need to restrict it and tell explicitly what are supported

ai-services/internal/pkg/utils/cert.go
```diff
@@ -0,0 +1,351 @@
...
+
+	// Warn if certificate expires soon
+	const (
+		hoursPerDay       = 24
+		expiryWarningDays = 30
+	)
+	daysUntilExpiry := cert.NotAfter.Sub(now).Hours() / hoursPerDay
+	if daysUntilExpiry < expiryWarningDays {
+		logger.Infof("Warning: Certificate expires in %.0f days (%s)\n", daysUntilExpiry, cert.NotAfter)
+	}
```
mayuka-c: Isnt checking certificate expiry just a single time validation we are doing?
This code kind of looks like notifiying the customer if it expires

Lets say if the certificate expires, is there a way for customer to renew it?

mayuka-c: I feel we should just do certificate expiry check.

ai-services/internal/pkg/utils/cert.go
```diff
@@ -0,0 +1,351 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	// First, check Subject Alternative Names (SANs) for wildcard domains
+	for _, san := range cert.DNSNames {
+		if strings.HasPrefix(san, "*.") {
+			// Extract base domain from wildcard (*.example.com → example.com)
+			domain := strings.TrimPrefix(san, "*.")
+			if domain != "" {
```
mayuka-c: Looks like we are using *. everywhere can we have constant defined for this?

ai-services/internal/pkg/utils/cert.go
```diff
@@ -0,0 +1,351 @@
...
+				return domain, nil
+			}
+		}
+	}
+
+	// If no wildcard found, check for regular DNS names in SANs
+	if len(cert.DNSNames) > 0 {
+		return cert.DNSNames[0], nil
+	}
```
mayuka-c: If no wildcard can we even fallback?
If fall back later it will fail while registring routes right?

ai-services/internal/pkg/utils/cert.go
shwetha-s-poojary: yeah you are right - let me change that

ai-services/internal/pkg/utils/cert.go
shwetha-s-poojary: > Lets say if the certificate expires, is there a way for customer to renew it?

I've not implemented now, but the doc proposal talks about adding a renew flag in catalog. Should we just note it and take it up later.

@mayuka-c @yussufsh

ai-services/internal/pkg/utils/cert.go
shwetha-s-poojary: @mayuka-c I didn't check this particular flow.Let me test that.Are you suggesting we fail and n fallback?

ai-services/internal/pkg/utils/cert.go
shwetha-s-poojary: @yussufsh wdyt?

ai-services/internal/pkg/utils/cert.go
shwetha-s-poojary: It's checked only once - during certificate validation.But if there are 30 days left to expire we are just printing out a warning.Ok maybe I'll remove this later part of warning users.

yussufsh: Please create an issue for this, we will need to handle logic to change the cert.

ai-services/internal/pkg/utils/cert.go
yussufsh: What is the reason for this limitation? Is there any validation?
Let's take it up in a new issue.

ai-services/cmd/ai-services/cmd/catalog/configure.go
```diff
@@ -136,6 +150,59 @@ func validateConfigureFlags(rawArgParams []string) (map[string]string, error) {
...
+	if (sslCertPath != "" && sslKeyPath == "") || (sslCertPath == "" && sslKeyPath != "") {
+		return fmt.Errorf("--ssl-cert and --ssl-key must be used together")
+	}
+
+	return nil
+}
+
+// warnIfBothCertAndDomainProvided warns user if both certificate and custom domain are provided.
+func warnIfBothCertAndDomainProvided() {
+	if sslCertPath != "" && sslKeyPath != "" && domainName != "" {
```
yussufsh, line 184: Need to sanitize this before using.
```
    cleanCertPath := filepath.Clean(sslCertPath)
    cleanKeyPath := filepath.Clean(sslKeyPath)
```

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -528,6 +706,31 @@ func registerCatalogRoutes(rt *podman.PodmanClient, tp templates.Template, appTe
...
+// computeDomainSuffix computes the domain suffix using priority: certDomain > customDomain > hostIP.nip.io.
+func computeDomainSuffix(certDomain, customDomain, hostIP string) (string, error) {
+	if certDomain != "" {
+		return certDomain, nil
+	}
+	if customDomain != "" {
+		return customDomain, nil
+	}
+
+	return fmt.Sprintf("%s.nip.io", hostIP), nil
```
yussufsh, line 717: Lets add a check for empty hostIP here.

yussufsh (changes requested):

yussufsh (approved):
lgtm
Pending items:
1. Way to reconfigure certificate when expired
2. Allow generic cert types

PR #814, opened 2026-05-27
Updating version check github workflow

adarshagrawal38:
Updating workflow as per new code org.


yussufsh:
@adarshagrawal38 let's also fix `.github/scripts/check_image_names.py` as part of this PR?

yussufsh:
@mayuka-c Now is a better time to add catalog manifests to the list for checks?

mayuka-c:
> @mayuka-c Now is a better time to add catalog manifests to the list for checks?

Yeah I feel we can add it

adarshagrawal38:
> @adarshagrawal38 let's also fix `.github/scripts/check_image_names.py` as part of this PR?

for most of the paths has been already updated, PTAL [here](https://github.com/IBM/project-ai-services/pull/814#issuecomment-4552052732)

do you feel any other update is needed?

For catalog we will have to add couple of entries

yussufsh:
Yes, please add for catalog services/components as well.
There will be more dirs coming in the near future - just a heads up.

mayuka-c:
Can we add it just for catalog folder for now and then later for services/components.
For Services/components we would need to sync other things too, so just updating image will not work

.github/workflows/bump-makefile-version.yml
```diff
@@ -2,45 +2,81 @@ name: Check Makefile Version Bump
...
-      - 'images/rag-base/**'
-      - 'digitize-ui/**'
+    paths: 
+      - 'services/chatbot/**'
+      - 'services/digitize/**'
+      - 'services/similarity/**'
+      - 'services/summarize/**'
+      - 'images/service-base/**'
+      - 'images/postgres/**'
+      - 'images/caddy/**'
```
yussufsh: Do we really need caddy and postgres here?
Then why not add litellm and tools as well?

.github/scripts/check_image_names.py
```diff
@@ -53,12 +53,26 @@
         ("ai-services/assets/applications/rag-dev/openshift/values.yaml", "ui"),
         ("ai-services/assets/applications/rag-cpu/podman/values.yaml", "ui"),
     ],
-    "ui/digitize/Makefile": [
-        ("ai-services/assets/applications/rag/openshift/values.yaml", "digitizeUi"),
-        ("ai-services/assets/applications/rag/podman/values.yaml", "digitizeUi"),
-        ("ai-services/assets/applications/rag-dev/openshift/values.yaml", "digitizeUi"),
-        ("ai-services/assets/applications/rag-dev/podman/values.yaml", "digitizeUi"),
+
```
yussufsh: Did we move `ui/digitize`?

.github/workflows/bump-makefile-version.yml
```diff
@@ -2,45 +2,81 @@ name: Check Makefile Version Bump
 
 on:
   pull_request:
-    paths:
-      - 'spyre-rag/src/**'
-      - 'spyre-rag/ui/**'
-      - 'images/rag-base/**'
-      - 'digitize-ui/**'
+    paths: 
+      - 'services/chatbot/**'
```
yussufsh, line 6: I feel we need `services/common` as as well. Better `services/*`.
When common is updated `services/digitize` and `services/chatbot` needs to be bumped up.
Any other place we are using the common code @dharaneeshvrd ?

dharaneeshvrd: Now common won't have any Makefile to check for its bump anyway. We can skip common here IMO.
All the services would consume common, now we are treating common as a lib and copying directly during image build. Only services would hold the Makefile and image pipeline.

.github/scripts/check_image_names.py
adarshagrawal38: yes

.github/workflows/bump-makefile-version.yml
adarshagrawal38: added `images/**` now all images would be covered

.github/workflows/bump-makefile-version.yml
yussufsh: Yes but code changes in common should build each service consuming it right? Otherwise it will contribute to regression issues.

.github/scripts/check_image_names.py
yussufsh: Where did we move? I think you removed it by mistake?

.github/workflows/bump-makefile-version.yml
yussufsh: Any changes in common should require rebuild and version bump of all the services consuming it.

dharaneeshvrd: Not all the series would need a change that meant for a specific service, that's why we have segregated the services now, if we build all the services whenever there is a code change in common, its back to the old way.

yussufsh: That means the developer would decide which image he needs to bump and ensure other services are not affected by the code. Also this workflow won't remind them to update service version if they change only common code.
If that is okay then we can ignore this.

dharaneeshvrd: Yes we know about this and we want to have that control.

.github/scripts/check_image_names.py
adarshagrawal38: yes right, added ui/digitzeUI block back

mkumatag:
check is failing, let us update the assets to make CI pass

adarshagrawal38:
Suggestion: Move the `Check Makefile Version Bump` job into each image's respective workflow.

With the current implementation, modifying a single image triggers the version bump check for **all** images, which is unnecessary and noisy. Scoping it per workflow would make checks more targeted and efficient.

yussufsh:
> Suggestion: Move the `Check Makefile Version Bump` job into each image's respective workflow.
> 
> With the current implementation, modifying a single image triggers the version bump check for **all** images, which is unnecessary and noisy. Scoping it per workflow would make checks more targeted and efficient.

How about 1 script (python?) that does all the files checks one by one instead of so many reusable workflows.. The job details logs would show which files were missed.

.github/scripts/check_image_names.py
yussufsh, line 20: We are missing caddy, services-base, tools, lite-llm, etc.

adarshagrawal38: services-base and tools, is not inferred in values.yaml file

for caddy, litellm I am plaining to have a follow on PR, as in Makefile `TAG` variable we don't get direct value for this images, PTAL [here](https://github.com/IBM/project-ai-services/blob/main/images/litellm/Makefile#L7)

We will have to update python script to handle this scenario.

adarshagrawal38: Taken care of caddy and litellm in the same PR.

yussufsh (approved):
lgtm

PR #821, opened 2026-05-27
Fixing license scan failure for Pypi scipy pkg

adarshagrawal38:


.github/workflows/go.yaml
```diff
@@ -9,6 +9,7 @@ on:
       - main
 
 
+
```
mkumatag: why this new line?

mkumatag (changes requested):

.github/workflows/go.yaml
adarshagrawal38: reverted

.github/scripts/license_scan.py
```diff
@@ -73,7 +73,14 @@ def classify_license(trivy_data, parlay_data):
             "License by Trivy": trivy_pkg_license,
             "License by Parlay": parlay_pkg_license,
         }
-        
+
+        # Setting license value for scipy as we are getting raw license value from parlay enrichment.
```
mkumatag: can we add something like TODO with an issue pointing to the parlay to remove this when fixed.

.github/scripts/license_scan.py
```diff
@@ -73,7 +73,14 @@ def classify_license(trivy_data, parlay_data):
...
+        # Temporary exception for scipy until issue is fixed in upstream.
+        # parlay ecosystems package pkg:pypi/scipy | jq '.licenses'
+        if pkg_name == "scipy":
```
mkumatag: do we need add pandas as well?

mkumatag (changes requested):

.github/scripts/license_scan.py
adarshagrawal38: done

.github/scripts/license_scan.py
adarshagrawal38: yeah added pandas as well

mkumatag (approved):

PR #810, opened 2026-05-26
Feat/delete app endpoint

noahmasoud:
Notes:
- Pod/container teardown is a placeholder, deferred until Create Application 
  flow is ready
- `GetByID` uses `INNER JOIN` on services, apps without services return 404 
  instead of being found. Noticed this during my manual smoke testing.

PR #822, opened 2026-05-27
feat: implement DELETE /applications/endpoint

noahmasoud:
Implementation for DELETE endpoint: https://github.com/IBM/project-ai-services/blob/main/docs/proposals/catalog/architecture.md 

- Implements async `DELETE /api/v1/applications/{id}` endpoint
- Sets application status to `Deleting` immediately, then performs cascade deletion in a background goroutine
- When `force=true`, orphaned component records (not referenced by any other app) are also deleted from the DB
- Orphan detection runs **before** app deletion so CASCADE doesn't wipe `service_dependencies` rows first
- Uses internal `logger` package for all error logging in the delete flow
- Pod/container teardown is a known placeholder, deferred until the Create Application flow is ready

Created a clean branch `feat/delete-app-endpoint-v2` from main to avoid the messy merge history and unsigned commits in the previous PR #810. All of @mayuka-c's review comments have been addressed:

- Fixed `force` param direction `force=true` enables orphan cleanup
- Replaced `fmt.Sprintf`/`fmt.Printf` logging with `logger.Errorf`
- Cleaned up router.go, no duplicate or placeholder routes
- Removed unreachable code after `return` in error handling

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
+++ b/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
@@ -16,6 +16,7 @@ import (
 	dbrepo "github.com/project-ai-services/ai-services/internal/pkg/catalog/db/repository"
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog/types"
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog/utils"
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
 	runtimeTypes "github.com/project-ai-services/ai-services/internal/pkg/runtime/types"
 )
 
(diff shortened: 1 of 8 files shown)


ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
```diff
@@ -547,4 +548,114 @@ func (s *ApplicationService) loadServiceComponents(ctx context.Context, sd []mod
...
+func (s *ApplicationService) performDeletion(ctx context.Context, appID uuid.UUID, services []models.Service, force bool) {
+	serviceIDs := make(map[uuid.UUID]bool, len(services))
+	for _, svc := range services {
+		serviceIDs[svc.ID] = true
+	}
+
+	// Identify orphaned components before deletion while service_dependencies still exist.
+	var orphanedComponents []uuid.UUID
+
+	if force {
```
mayuka-c, line 604: force is used for a diff purpose. If set. we will delete the components but not the underlying data.

So essentially everything gets deleted, but not the removal of data stored in host

mayuka-c: You can check out the delete flow that we have currently under https://github.com/IBM/project-ai-services/blob/main/ai-services/internal/pkg/application/podman/delete.go#L56.

You can take this up in subsequent PRs as I feel we need to mount the `{{ .BaseDir }}` to catalog backend container to do the cleanup.

mayuka-c (commented):
Overall LGTM for Part 1.

We need to take up the actual pods deletion and also force implementation as mentioned in the comment. We can do it in subsequent PRs for each

mayuka-c (commented):
Also you would need to bump the image in `Makefile` and also update in catalog assets

yussufsh:
@mayuka-c could you please push a commit with version bump to this pr and close it. 
Let's start resources deletions in a separate PR.

mayuka-c:
@noahmasoud I will be merging this, Please take care of my comments in subsequent PRs

mayuka-c (approved):
LGTM based on above comment

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
mayuka-c: Will be taken care in subsequent PRs

mayuka-c:
@yussufsh @Shubhamag12 @shwetha-s-poojary Can someone approve this PR?
Post that we can merge.

As I pushed a commit, it doesnt let me merge with my only approval.

shwetha-s-poojary (approved):
needs a rebase otherwise lgtm

PR #823, opened 2026-05-28
feat: implement automatic HTTPS route registration for catalog-deployed applications

shwetha-s-poojary:
- Add routes annotation support to all service and application templates
- Implement automatic route extraction and registration during deployment
- Move common code to shared proxy utilities for Caddy pod discovery and route management
- Configure HOST_IP environment variable for consistent domain generation in case of [nip.io](http://nip.io/) domains

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
+++ b/ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
@@ -15,7 +15,6 @@ import (
 	"text/template"
 
 	"github.com/google/uuid"
-	"github.com/project-ai-services/ai-services/assets"
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog"
 	apimodels "github.com/project-ai-services/ai-services/internal/pkg/catalog/apiserver/models"
 	deploymenttypes "github.com/project-ai-services/ai-services/internal/pkg/catalog/apiserver/services/deployment/types"
@@ -731,12 +730,49 @@ func (d *PodmanDeployer) deployServicePods(
 
 	// If PodTemplateExecutions is defined, use it for ordered deployment
 	if len(metadata.PodTemplateExecutions) > 0 {
-		if err := d.deployServicePodsWithExecutions(metadata, tmpls, applicationID, svc, values); err != nil {
-			return err
+		// Execute each pod template in the service following the defined order
+		for _, layer := range metadata.PodTemplateExecutions {
+			for _, podTemplateName := range layer {
+				// Prepare initialParams for the template
+				initialParams := map[string]any{
+					"InstanceSlug": generateInstanceSlug(applicationID.String()),
+					"TemplateID":   svc.DatabaseID,
+					"BaseDir":      utils.GetBaseDir(),
+					"Values":       values,
+					"env":          map[string]map[string]string{},
+				}
+
+				_, podName, routes, err := d.deployPodTemplate(podTemplateName, tmpls, initialParams)
+				if err != nil {
+					return fmt.Errorf("failed to deploy pod template %s: %w", podTemplateName, err)
+				}
+
+				if routes != "" {
+					svc.Routes[podName] = routes
+				}
+			}
 		}
 	} else {
-		if err := d.deployServicePodsWithoutExecutions(tmpls, applicationID, svc, values); err != nil {
-			return err
+		// If no PodTemplateExecutions defined, deploy all templates
+		logger.Infof("No PodTemplateExecutions defined for service %s, deploying all templates\n", svc.CatalogID)
+		for templateName := range tmpls {
+			// Prepare initialParams for the template
+			initialParams := map[string]any{
+				"InstanceSlug": generateInstanceSlug(applicationID.String()),
+				"TemplateID":   svc.DatabaseID,
+				"BaseDir":      utils.GetBaseDir(),
+				"Values":       values,
+				"env":          map[string]map[string]string{},
+			}
+
+			_, podName, routes, err := d.deployPodTemplate(templateName, tmpls, initialParams)
+			if err != nil {
+				return fmt.Errorf("failed to deploy pod template %s: %w", templateName, err)
+			}
+
+			if routes != "" {
+				svc.Routes[podName] = routes
+			}
 		}
 	}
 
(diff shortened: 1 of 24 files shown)


ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -5,13 +5,29 @@ import (
...
+	if _, err := os.Stat("/.dockerenv"); err == nil {
+		return true
+	}
+	// Check for Podman container indicator
+	if _, err := os.Stat("/run/.containerenv"); err == nil {
+		return true
+	}
+
+	return false
+}
```
mayuka-c: Didnt quite understand the logic here to determine running inside a Container. Is this some standard way to determine?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -76,30 +63,66 @@ func DeployCatalog(ctx context.Context, podmanURI, passwordHash, baseDir string,
...
+
+	tp, appMetadata, tmpls, err := loadCatalogTemplates(s)
+	if err != nil {
+		s.Fail("failed to load catalog templates")
+
+		return nil, nil, nil, nil, fmt.Errorf("failed to load catalog templates: %w", err)
+	}
+
+	// Set HTTPS port in argParams (caller ensures argParams is not nil)
+	argParams["caddy.httpsPort"] = fmt.Sprintf("%d", httpsPort)
```
mayuka-c, line 109: I dont see the httpsPort env set in catalog template

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -1062,3 +1116,73 @@ func generateInstanceSlug(id string) string {
...
+		); err != nil {
+			registrationErrors = append(registrationErrors, fmt.Errorf("pod %s: %w", podName, err))
+
+			continue
+		}
+	}
+
+	if len(registrationErrors) > 0 {
+		return fmt.Errorf("failed to register routes for %d pod(s): %w", len(registrationErrors), errors.Join(registrationErrors...))
+	}
```
mayuka-c, line 1206: Do we want to fail the overall application deployment if Register Routes failed?

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -774,6 +761,58 @@ func (d *PodmanDeployer) deployComponentTemplate(
...
+				svc.Routes[podName] = routes
+			}
+		}
+	}
+
+	return nil
+}
+
+// deployServicePodsWithoutExecutions deploys all service pod templates without specific order.
+func (d *PodmanDeployer) deployServicePodsWithoutExecutions(tmpls map[string]*template.Template, applicationID uuid.UUID, svc *ServicePlan, values map[string]any) error {
```
mayuka-c: I see both this and above methods are kind of same except the for loop, so can we have just a single method and let caller do the for looping

mayuka-c (commented):
Overall LGTM,

We might need to consider updating endpoints in DB in subsequent PR

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -1062,3 +1116,73 @@ func generateInstanceSlug(id string) string {
...
+				allRoutes[podName] = routes
+			}
+		}
+	}
+
+	if len(allRoutes) == 0 {
+		logger.Infof("No routes found for application '%s', skipping route registration\n", plan.ApplicationName)
+
+		return nil
+	}
```
mayuka-c, line 1171: Move determing routes logic to planner.go and add it in Service Plan struct

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -1062,3 +1116,73 @@ func generateInstanceSlug(id string) string {
...
+
+	// Find catalog Caddy pod
+	caddyPodName, err := d.findCatalogCaddyPod()
+	if err != nil {
+		return fmt.Errorf("failed to find catalog Caddy pod: %w", err)
+	}
+
+	logger.Infof("Found catalog Caddy pod: %s\n", caddyPodName)
```
mayuka-c: Also this pls move to planner.go

We need to fail early if CaddyPod is missing and shouldnt fail during deployer.

All prerequisite configurations/ checks should go there

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: This logic is basically to determine if I'm inside a container or inside a host.
Why do we need this? - While building routes we use a common function for both catalog and other services deployed via catalog .So when fetching the hostIP(we need it to build nip.io domain), if it's a host VM , I directly fetch the host IP of VM( and save this value as env var in catalog), if it's a container(catalog in this case via which we are registering routes) we don't want the container IP - we need the host VM IP (which I fetch from env vars).

There might better and a more standard way to check if it's a container or not.Let me find out.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
shwetha-s-poojary: Right - I haven't tested the custom domain with this yet.I will have add it in catalog template.Will do

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
shwetha-s-poojary: I too had that question.

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
shwetha-s-poojary: I think this was done as part of a lint fix by Bob.

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
shwetha-s-poojary: Ok - I will move it there

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -900,42 +940,56 @@ func (d *PodmanDeployer) extractComponentEndpointInfo(podSpec *podmodels.PodSpec
...
 	podSpec, err := d.renderAndParsePodTemplate(podTemplate, podTemplateName, initialParams)
 	if err != nil {
-		return nil, err
+		return nil, "", "", err
+	}
+
+	// Extract routes annotation if present
+	routes := ""
+	if podSpec.Annotations != nil {
+		logger.Infof("DEBUG: Pod '%s' annotations: %+v\n", podSpec.Name, podSpec.Annotations)
```
Shubhamag12: are these testing logs?
if not we can use logger.VerbosityLevelDebug at the end of statement

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -1062,3 +1116,73 @@ func generateInstanceSlug(id string) string {
 }
 
 // Made with Bob
```
Shubhamag12: we can move this EOF

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
Shubhamag12: @mayuka-c we are not failing it right? we are just logging errors in the caller function

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -1062,3 +1116,73 @@ func generateInstanceSlug(id string) string {
...
+		return fmt.Errorf("failed to register routes for %d pod(s): %w", len(registrationErrors), errors.Join(registrationErrors...))
+	}
+
+	logger.Infof("Registered %d route(s) for application '%s'\n", len(allRoutes), plan.ApplicationName)
+
+	return nil
+}
+
+// findCatalogCaddyPod finds the catalog Caddy pod name from catalog templates.
+func (d *PodmanDeployer) findCatalogCaddyPod() (string, error) {
```
Shubhamag12: if the aim of this function is to find caddy pod name, can we filter via label ` ai-services.io/component: "proxy"` instead?

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
shwetha-s-poojary: Yes - they are testing logs.

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
shwetha-s-poojary: Is this really required?

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
shwetha-s-poojary: With simplification we check for caddy pod anyways while we configure catalog routes.So this is no longer required here

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
shwetha-s-poojary: Yes that's the logic we have helper function where we are calling it from.I have simplified the code, it uses env var now

yussufsh (commented):
Overall, lgtm but many places proper logs need to be addressed and update db on failure for applications needs to be implemented.

shwetha-s-poojary:
Most of the changes you see intern of extra helper functions were done as a part of fixing lint issues.
@mayuka-c @yussufsh - I've addressed the comments PTAL

mayuka-c (approved):
LGTM,
Do create a task for updating service table in DB with error message

yussufsh:
> LGTM, Do create a task for updating service table in DB with error message

Also create task in backlog/misc for refactoring the code 😄

PR #826, opened 2026-05-29
fix failing digitize service tests

manju956:


services/digitize/app.py
```diff
@@ -530,6 +530,8 @@ async def get_document_metadata(doc_id: str, details: bool = Query(False, descri
 
         response = get_document(doc_id, include_details=details)
         return response
+    except FileNotFoundError as e:
```
dharaneeshvrd, line 533: Why are we adding as part of this PR?
Is this really needed?
If we are updating app, then we need to bump the digitize version

manju956: there was test case failure for 404 error case which was not added in app.py. Added it and now bumped up digitize service version

dharaneeshvrd (approved):

PR #827, opened 2026-05-29
AIS-958 - changes for service deploy options api

prachikurade1:
- Add resources to the metadata templates
- Update service deploy options to return resources at service and component provider level
- Ensure that the architecture deploy options does not return the resources object

Dev testing results:
Service Deploy Options:
1. Similarity service - Full response - [similarity_options_resp.json](https://github.com/user-attachments/files/28385265/similarity_options_resp.json)
similarity_deploy_options
2. Chatbot service - Full response - [chat_options_resp.json](https://github.com/user-attachments/files/28385297/chat_options_resp.json)
chat_deploy_options
3. Summarize service - Full response - [summarize_options_resp.json](https://github.com/user-attachments/files/28385311/summarize_options_resp.json)
summarize_deploy_options
4. Digitize service - Full response - [digitize_options_resp.json](https://github.com/user-attachments/files/28385327/digitize_options_resp.json)
digitize_deploy_options

Architecture Deploy Options (same as earlier): - Full response - [arch_options_resp.json](https://github.com/user-attachments/files/28385354/arch_options_resp.json)
arch_deploy_options


mkumatag (dismissed):
Your memory resource in the proposal states that it's measured in bytes [here](https://github.com/IBM/project-ai-services/blob/main/docs/proposals/catalog/resource-api-proposal.md#3-key-concepts), but this request specifies resources in MB. Doesn't this create confusion for the user?

prachikurade1:
> Your memory resource in the proposal states that it's measured in bytes [here](https://github.com/IBM/project-ai-services/blob/main/docs/proposals/catalog/resource-api-proposal.md#3-key-concepts), but this request specifies resources in MB. Doesn't this create confusion for the user?

That part was for the resource utilization API. This one since it is coming from the templates, it is in MB - https://github.com/IBM/project-ai-services/blob/main/docs/proposals/catalog/resource-api-proposal.md#92-resource-specification-format

Shall I convert them to bytes for all ? Will update the proposal doc as well if required.

mkumatag:
> Shall I convert them to bytes for all ? Will update the proposal doc as well if required

which one make more sense? may be we need to brainstorm and mention what is better for end user and also easy in the backend.

ai-services/assets/components/embedding/vllm-cpu/metadata.yaml
```diff
@@ -7,3 +7,8 @@ description: "Deploy embedding models on vLLM inference engine (CPU-only)"
 # Component type this provider belongs to
 component_type: embedding
 component_name: "Embedding Model"
+
+# Resource requirements
+resources:
```
yussufsh: Can we add approx storage value as well?
We can work on exact numbers later.

ai-services/assets/components/llm/vllm-cpu/metadata.yaml
```diff
@@ -7,3 +7,8 @@ description: "Deploy instruct models on vLLM inference engine (CPU-only)"
 # Component type this provider belongs to
 component_type: llm
 component_name: "Large Language Model (LLM)"
+
+# Resource requirements
+resources:
+  cpu: 12
+  memory: 102400 # 100Gi in MB
```
mayuka-c: These resources are specific to podman right?
If yes, we would need to move it to runtime specific metadata

ai-services/docs/swagger.yaml
```diff
@@ -253,6 +257,23 @@ definitions:
...
+      accelerators:
+        additionalProperties:
+          type: integer
+        description: 'Accelerator cards (e.g., "ibm.com/spyre_pf": 1)'
+        type: object
+      cpu:
+        description: CPU cores
+        type: integer
+      memory:
+        description: Memory in MB
```
mayuka-c: Any reason of representing here with MB instead of Bytes similar to /resources API?

prachikurade1: No specific reason. Will update all to use bytes.

yussufsh (dismissed):
lgtm
...for initial implementation.

prachikurade1:
> lgtm ...for initial implementation.

Updated version

yussufsh (dismissed):

mayuka-c (dismissed):
LGTM

yussufsh (approved):

PR #828, opened 2026-05-29
chore(catalog-ui): Design audit changes

pranithraoibm:
- Digital Assistants page description update
- Digital assistants delete dialog box wordings update
- Export CSV logic fix & success/error toast addition
- Fixed width for filters panel in Use case references


mkumatag (changes requested):
I see plenty of these images are added!
1. What is the source of these images? and licences for using these images?
2. Are these part of use cases static pages or you are expecting them to be included in the backend apis eventually? @yussufsh : we need to handle them differently if we are planning to get these images from the backend api
3. Any either way - is it possible to reduce the size of these images? few of them are more than 1MB.

pranithraoibm:
@mkumatag 
1. Susan mentioned that she got these images from the IBM website itself, from the pages where the corresponding use case videos are available. Since they're IBM-owned assets, we should be okay to use them from a licensing perspective is what she told us.
2. This was decided to be a static page.

mkumatag:
> @mkumatag
> 
> 1. Susan mentioned that she got these images from the IBM website itself, from the pages where the corresponding use case videos are available. Since they're IBM-owned assets, we should be okay to use them from a licensing perspective is what she told us.
> 2. This was decided to be a static page.

We still need to mention - the source and licences for all the assets used.

pranithraoibm:
These changes are included in https://github.com/IBM/project-ai-services/pull/859 PR, hence closing it.

PR #831, opened 2026-05-31
[Catalog-API] Endpoint for Service Params

mayuka-c:
- Expose a new endpoint for returning service params
- Currently Q&A requires custom system prompt, so added it in values.schema.json to expose

### DIFF
--- a/ai-services/internal/pkg/catalog/deploy_options.go
+++ b/ai-services/internal/pkg/catalog/deploy_options.go
@@ -159,6 +159,7 @@ func hasNonEmptyProperties(schema map[string]any) bool {
 	if properties, ok := schema["properties"].(map[string]any); ok {
 		return len(properties) > 0
 	}
+
 	return false
 }
 
--- a/ai-services/internal/pkg/catalog/deploy_options.go
+++ b/ai-services/internal/pkg/catalog/deploy_options.go
@@ -56,7 +56,7 @@ func (p *CatalogProvider) GetArchitectureDeployOptions(architectureID string) (*
 		}
 
 		// Only add schema if the service has non-empty schema properties
-		if p.hasNonEmptyServiceSchemaProperties(service.ID) {
+		if schema, err := p.GetServiceParams(service.ID); err == nil && hasNonEmptyProperties(schema) {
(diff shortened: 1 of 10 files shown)


ai-services/internal/pkg/catalog/deploy_options.go
```diff
@@ -173,6 +187,38 @@ func (p *CatalogProvider) hasNonEmptySchemaProperties(componentType, providerID
 	return false
 }
 
+// hasNonEmptyServiceSchemaProperties checks if a service has a schema file with non-empty properties.
+func (p *CatalogProvider) hasNonEmptyServiceSchemaProperties(serviceID string) bool {
```
yussufsh: Looks like we could reuse GetServiceParams here.

ai-services/internal/pkg/catalog/deploy_options.go
```diff
@@ -209,4 +255,39 @@ func (p *CatalogProvider) GetComponentProviderParams(componentType, providerID s
 	return schema, nil
 }
 
+// GetServiceParams returns the JSON schema for a specific service's configuration.
+// If the schema file is not present, returns an empty schema instead of failing.
+func (p *CatalogProvider) GetServiceParams(serviceID string) (map[string]any, error) {
```
yussufsh, line 204: Can we move this to catalog.go?

mayuka-c: Currently the Component params also exist here, so I feel let it be part of deploy_options.go itself (as these are part of deploy). PLMK

ai-services/internal/pkg/catalog/deploy_options.go
mayuka-c: Yeah I guess we can, let me check. The similar thing was done for component params too.
Let me refactor both

mayuka-c: Done

ai-services/internal/pkg/catalog/deploy_options.go
yussufsh: how about catalog_params.go?
I will let you decide and merge this one.

yussufsh (approved):
lgtm

PR #817, opened 2026-05-27
feat: enable catalog to run as non root user

Shubhamag12:
This PR enables the the tool to run as a non-root user 
It implements proper SELinux policies and Podman socket access configuration for rootless containers.

1. Policy: added SELinux policy module (podman_socket_policy) to allow containers with container_t type to access the Podman socket.
2. Resolving podman socket for both root and non-root scenarios
3. Resolving podman auth file path for both root and non-root scenarios
4. Removed dependency on keep-id user namespace mapping by handling Podman socket access through SELinux policy configuration

**One Assumption:**
User will have to run catalog configure using --http-port 8443 as for nonroot users, its not allowed
```
ERRO[0003] "rootlessport cannot expose privileged port 443, you can add 'net.ipv4.ip_unprivileged_port_start=443' to /etc/sysctl.conf (currently 1024), or choose a larger port number (>= 1024): listen tcp 0.0.0.0:443: bind: permission denied"
Error: unable to start container "f72329a97540825bb753ec41806021c9fd15890ae54f6874b2f13e2c6bc8890a": starting some containers: internal libpod error
```

PR #832, opened 2026-05-31
fix: Add registry authentication for catalog image pulls

shwetha-s-poojary:
Mount registry auth file and pass to Podman API to enable pulling
from private registries like icr.io

**Requires: chmod 644 /run/user/0/containers/auth.json on host**

### DIFF
--- a/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
+++ b/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
@@ -149,8 +149,10 @@ spec:
     - name: podman-auth
       hostPath:
         # Mount the auth file from host
+        # For rootless podman: /run/user/$(id -u)/containers/auth.json
+        # For root podman: /run/user/0/containers/auth.json
         # NOTE: Ensure the file has read permissions (chmod 644) for the container user
-        path: /run/user/0/containers/auth.json
+        path: {{ .Values.backend.podman.authFile }}
         type: File
     - name: iommu-groups
       hostPath:
--- a/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
+++ b/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
@@ -93,6 +93,8 @@ spec:
       env:
         - name: CONTAINER_HOST
           value: "unix://{{ .Values.backend.podman.uri }}"
+        - name: REGISTRY_AUTH_FILE
+          value: "/run/containers/auth.json"
         - name: ADMIN_PASSWORD
           valueFrom:
             secretKeyRef:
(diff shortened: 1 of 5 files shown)


ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
```diff
@@ -141,6 +146,12 @@ spec:
         # For root podman: /run/podman/podman.sock
         path: {{ .Values.backend.podman.uri }}
         type: Socket
+    - name: podman-auth
+      hostPath:
+        # Mount the auth file from host
+        # NOTE: Ensure the file has read permissions (chmod 644) for the container user
+        path: /run/user/0/containers/auth.json
```
shwetha-s-poojary: I'm wondering if podman secrets is a good fit here?

mayuka-c: Since it is just the path getting mounted, I guess this should be fine

mayuka-c: This would work for root only.
This path needs to be dynamically computed in `catalog configure` and populated here so that it works for both root and non root

Shubhamag12: we can populate this field in the same way on how we do it for podman socket mount

shwetha-s-poojary:
```
# NOTE: Ensure the file has read permissions (chmod 644) for the container user
path: /run/user/0/containers/auth.json
```

This file usually has 600.I had to manually change it to 644 to test. What do we do about this requirement?

shwetha-s-poojary: Even in getPodmanURI the podman-sock path is hardcoded as of now [here](https://github.com/IBM/project-ai-services/blob/32a42b5fab8dd7533bbc998e4558c8cde5d02c5f/ai-services/internal/pkg/catalog/cli/configure/common.go#L59)

Are you suggesting to add a similar helper function with a TO-DO ?

Shubhamag12: No no, I am taking care of above `getPodmanURI`, similarly will take care of auth path as well
Will include this change in the other PR: https://github.com/IBM/project-ai-services/pull/817/changes#diff-f228f16ca2bf59537fcc1caaf9bdc5338ade676f76595a977e31ef5f942557b8R47

Edit: As discussed @shwetha-s-poojary, you can add TODO and will handle in subsequent PRs

Shubhamag12: >  NOTE: Ensure the file has read permissions (chmod 644) for the container user

should this be added as a pre-req then?

yussufsh (approved):

yussufsh (approved):
We are not going to change the permission.
Merging this now, @Shubhamag12 please move the mount process to creating a secret.
@mayuka-c Please ensure the auth refresh is taken care of in the idempotency story.

yussufsh:
Commits from this PR is included in https://github.com/IBM/project-ai-services/pull/838/commits with a fully workable solution.

PR #834, opened 2026-06-01
Update System Prompt

sats-23:
-[AISERVICES-1125](https://jsw.ibm.com/browse/AISERVICES-1125)
-When comparing Spyre and Wx env (Granite 3.3 8B instruct vs Granite 4H small) we found that Wx env had more responses ending with a suggestive question (refer screenshots on ticket)/
-Updated prompt to ground the LLM to stick to only response generation based on context.

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=0.0.20
+TAG?=0.0.21
 CONTAINER_BUILDER?=podman
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
--- a/ai-services/assets/applications/rag-cpu/podman/values.yaml
+++ b/ai-services/assets/applications/rag-cpu/podman/values.yaml
@@ -8,7 +8,7 @@ backend:
   # @description Host port for the OpenAI-compatible RAG service. Defaults to unexposed; assign a port to enable external access.
   port: "0"
   # @hidden
-  image: icr.io/ai-services-cicd/chatbot-service:v0.0.5
+  image: icr.io/ai-services-cicd/chatbot-service:v0.0.6
   # @hidden
   log_level: "INFO"
 
--- a/ai-services/assets/applications/rag-dev/openshift/values.yaml
+++ b/ai-services/assets/applications/rag-dev/openshift/values.yaml
@@ -4,7 +4,7 @@ ui:
 
 backend:
   # @hidden
-  image: icr.io/ai-services-cicd/chatbot-service:v0.0.5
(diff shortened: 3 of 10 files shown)


mkumatag:
@dharaneeshvrd any idea why these unit tests are still failing?

dharaneeshvrd:
These are failing with db introduction and we thought of taking care in a separate PR since the changes would be large.
And the PR is up https://github.com/IBM/project-ai-services/pull/826, we will merge it soon.

cc @manju956

dharaneeshvrd (commented):
Why are we altering the entire prompt? Is it not possible to add only the boundaries present in new prompt?
Hope spyre version is not affected after this, please validate and post the results in JIRA
Need to bump chatbot version.

sats-23:
> Why are we altering the entire prompt? Is it not possible to add only the boundaries present in new prompt? Hope spyre version is not affected after this, please validate and post the results in JIRA Need to bump chatbot version.

Redid the entire prompt to keep a polished look, the outcome is found to be same. 
I shall post the Spyre responses as well soon

dharaneeshvrd:
> Redid the entire prompt to keep a polished look, the outcome is found to be same.

Wondering do we need to edit the entire prompt? 
Whenever we edit this,  I am seeing existing things are getting replaced with new changes. Hence asking.

sats-23:
> > Redid the entire prompt to keep a polished look, the outcome is found to be same.
> 
> Wondering do we need to edit the entire prompt? Whenever we edit this, I am seeing existing things are getting replaced with new changes. Hence asking.

I guess it does the job in keeping things clean and polished. 
I can add something like this in the end of the current prompt.

> CRITICAL INSTRUCTION: Answer only the specific question asked. Do not add conversational filler, offer additional assistance, suggest follow-up steps, or ask follow-up questions at the end of your response. End your response immediately once the question has been answered.

Or, take the current approach in PR which provides a polished prompt by embedding this instruction within.

Could you please add an example for 

>  I am seeing existing things are getting replaced with new changes

I didn't quite get it

sats-23:
Please find both env outputs against new prompt here - https://jsw.ibm.com/secure/attachment/12331735/new_prompt.zip

dharaneeshvrd:
Answers are looking good to me.
> I am seeing existing things are getting replaced with new changes

Two instances
1. legacy prompt to conversation prompt, I understand there is a change required here.
2. And now this
In future if there are any other use case requires a prompt update, we should not replace or alter the prompt majorly IMO.

And I believe, you might have got help from gemini for this new prompt.
But bob and copilot are suggesting to extend the existing prompt and I too feel this way.
Also I have concerns about `technical AI assistant.` phrase where we are isolating this RAG only for technical documents. We can't be sure what kind of documents user would feed in right?

sats-23:
Got it, I have replaced it with an approach by just extending the current prompt and the results are similar
Refer - https://jsw.ibm.com/secure/attachment/12334304/v2_prompt.zip

dharaneeshvrd (commented):
Prompt update lgtm
Let's update the versions.

dharaneeshvrd (approved):

PR #766, opened 2026-05-19
Migration script to port from json metadata files to db records

manju956:
- Migration python script that finds json based jobs and documents metadata on existing deployment and creates corresponding db entries in jobs and documents tables
- Runs as part of Init container with optional flag. A cleanup flag is provided to enable if json files need to be retained or not after the migration is complete.

```
[root@rag-test--digitize rag]# /var/venv/bin/python -m digitize.db.migrate_json_to_postgres --cleanup
2026-05-19 13:05:27 - database           - INFO     - db url = postgresql://postgres:Aiservicesdb%401234@rag-test--postgres:5432/digitize_metadata
2026-05-19 13:05:27 - database           - INFO     - Database engine created successfully
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - ======================================================================
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - PostgreSQL Migration Utility
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - ======================================================================
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - Assuming database schema is already initialized by init container...
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - Jobs directory: /var/cache/jobs
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - Docs directory: /var/cache/docs
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     -
======================================================================
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - Migrating Jobs
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - ======================================================================
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - Found 1 job status files to migrate
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - ✅ Job migration complete: 1 migrated, 0 failed
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     -
======================================================================
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - Migrating Documents
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - ======================================================================
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - Found 3 document metadata files to migrate
2026-05-19 13:05:27 - migrate_json_to_postgres - INFO     - ✅ Document migration complete: 3 migrated, 0 skipped (orphaned), 0 failed
[...]

PR #835, opened 2026-06-01
Proposal for digitize service Import and Export metadata

manju956:


docs/proposals/import_export_apis_design.md
```diff
@@ -0,0 +1,521 @@
...
+}
+```
+
+## Proposed API Design
+
+### 1. Import API - JSON to Database
+
+#### Endpoint
+```
+POST /v1/admin/import
```
dharaneeshvrd: Do you have any other endpoint planned under admin?

docs/proposals/import_export_apis_design.md
```diff
@@ -0,0 +1,521 @@
...
+```
+POST /v1/admin/export
+```
+
+#### Purpose
+Export metadata from PostgreSQL database as JSON data in the response body for backup/restore purposes.
+
+#### Request Body
+```json
+{
```
dharaneeshvrd: I think the only consumer of these APIs is going to be this backup and restore script. 
Why do we need to make it complicated by providing too many filters?
IMO we just need to give limit and provide options to iterate over total no of items like a pagination in case of items greater than the recommended limit which is missing currently.

docs/proposals/import_export_apis_design.md
```diff
@@ -0,0 +1,521 @@
...
+}
+```
+
+#### Request Parameters
+
+| Parameter | Type | Required | Default | Description |
+|-----------|------|----------|---------|-------------|
+| `jobs` | array | Yes | [] | Array of job metadata objects to import |
+| `documents` | array | Yes | [] | Array of document metadata objects to import |
+| `options.mode` | enum | No | "merge" | Import strategy: "merge" (upsert), "replace" (delete+insert), "skip" (skip existing) |
```
dharaneeshvrd: I think we need to skip by default. But I dont think these things are going to be exposed to the user. Not sure how allowing different mode would be helpful. IMO get started with default values only for now. 
Just looking to reduce the complexity of these new endpoints.

docs/proposals/import_export_apis_design.md
```diff
@@ -0,0 +1,521 @@
...
+| `options.validate_only` | boolean | No | false | Dry-run mode: validate without importing |
+| `options.batch_size` | integer | No | 100 | Number of records to process per batch |
```
dharaneeshvrd: What is this for?

docs/proposals/import_export_apis_design.md
```diff
@@ -0,0 +1,521 @@
...
+  ],
+  "options": {
+    "mode": "merge",
+    "validate_only": false,
+    "batch_size": 100
+  }
+}
+```
+
+#### Request Parameters
```
dharaneeshvrd, line 123: If we allow pagination for exporting, we need to process them also in similar fashion here.
Please see how we can achieve that

docs/proposals/import_export_apis_design.md
manju956: hm, nothing as of now.

docs/proposals/import_export_apis_design.md
manju956: sure, lets stick to skip. I will try to reduce the otpions and keep it minimalistic.

docs/proposals/import_export_apis_design.md
dharaneeshvrd: Let's keep it under `/v1/`

mkumatag:
I will defer @yussufsh to review and approve this PR :)

docs/proposals/import_export_apis_design.md
manju956: makes sense, since there wont be end user consuming these APIs, no need of offering lot of filters. `limit` and `offset` can be used for pagination.

docs/proposals/import_export_apis_design.md
```diff
@@ -0,0 +1,521 @@
...
+      "total_found": 1500,
+      "imported": 1480,
+      "skipped": 15,
+      "failed": 5
+    }
+  },
+  "duration_seconds": 12.5,
+  "errors": [
+    {
+      "file": "/path/to/job_123_status.json",
```
prachikurade1: Which path does this refer to ? Since we are giving data in json format, not clear on which file path is this referring to ?

docs/proposals/import_export_apis_design.md
```diff
@@ -0,0 +1,521 @@
...
+```json
+{
+  "error": {
+    "code": "PAYLOAD_TOO_LARGE",
+    "message": "Export would exceed maximum response size. Requested: 50000 records, Maximum: 10000",
+    "status": 413,
+    "details": {
+      "max_records": 10000,
+      "requested_records": 50000,
+      "suggestion": "Use filters or pagination to reduce result set"
```
prachikurade1: Assuming that pagination params will be query params, since they are not listed here. If possible please mention the param name.

prachikurade1 (changes requested):

docs/proposals/import_export_apis_design.md
manju956: right, I will redesign the import API to accept paginated json metadata inputs.

docs/proposals/import_export_apis_design.md
manju956: this is more of internal batch size to process the number of metadata jsons and create db records in batches of fixed size. I guess, we need not expose this param, as its used for more of internal db management.

docs/proposals/import_export_apis_design.md
manju956: Yes, I will mention the params.

docs/proposals/import_export_apis_design.md
manju956: its no longer present after the APIs refactoring. ptal

docs/proposals/import_export_apis_design.md
manju956: i have restructured import API to accept the export API response with pagination to work seamless.

yussufsh:
@dharaneeshvrd @manju956 wondering if the export API can include a logic to check whether a Postgres DB is present in the environment and read the file system for metadata instead if no DB? And also, if possible, a check for DB exists, delete the metadata files from the file system?
This way, the API can be backward compatible and even Backup & Restore won't need to do checks or validate the service versions.

Let me know what you think.

dharaneeshvrd:
We have a PR already up to achieve this workflow. https://github.com/IBM/project-ai-services/pull/766
Then we would have two scripts where the existing backup & restore can be exposed to the user and user can do it at their will. And this script will automatically migrate if the data exists already in the volume mounted.
@manju956 do you see any issues with having both backup & restore script and this init migration script?

dharaneeshvrd:
Btw we have put this PR on hold since that would be achieved by backup & restore script. If there are no problem in both existing, then existing PR can be merged to take care of the automatic migration.

manju956:
> We have a PR already up to achieve this workflow. #766 Then we would have two scripts where the existing backup & restore can be exposed to the user and user can do it at their will. And this script will automatically migrate if the data exists already in the volume mounted. @manju956 do you see any issues with having both backup & restore script and this init migration script?

hmm, I dont sense a problem with having both backup and restore script and init migration script. Provided the restore flow should be resilient enough in case of re-attempting the db records creation in case of metadata files not being cleaned up.

prachikurade1:
> @dharaneeshvrd @manju956 wondering if the export API can include a logic to check whether a Postgres DB is present in the environment and read the file system for metadata instead if no DB? And also, if possible, a check for DB exists, delete the metadata files from the file system?
This way, the API can be backward compatible and even Backup & Restore won't need to do checks or validate the service versions.

Yes, This will be a better mechanism since it will also eliminate the mandatory manual backup-restore of digitize the user would have to do on upgrade. 

> And also, if possible, a check for DB exists, delete the metadata files from the file system?

@manju956 , how will this be handled ?

manju956:
> need

Yes, we have flag based mechanism of handling the metadata files cleanup(if set to true) post the migration.
Ref https://github.com/IBM/project-ai-services/pull/766/changes#diff-1f29adb5538757df9e0e82e9919136479b97477a3565618b0266e88b518c6618R27

dharaneeshvrd:
> Yes, This will be a better mechanism since it will also eliminate the mandatory manual backup-restore of digitize the user would have to do on upgrade.

Please note that this would require user to use same volume across upgrades, if its not possible, manual is the option and we need to support that too.

dharaneeshvrd (commented):
API changes are looking fine.
@yussufsh @prachikurade1 please give an ack for the API changes, so that Mac, can get started on the implementation.

yussufsh:
> > Yes, This will be a better mechanism since it will also eliminate the mandatory manual backup-restore of digitize the user would have to do on upgrade.
> 
> Please note that this would require user to use same volume across upgrades, if its not possible, manual is the option and we need to support that too.

Why is the same volume needed? If we can take care of replacing the docs as well, wouldn't that be enough?

dharaneeshvrd:
>  take care of replacing the docs as well

Is this possible on both runtimes?
If it is possible, then the init migration script will take care, but in that case, these new APIs are not needed I guess.

manju956:
> > take care of replacing the docs as well
> 
> Is this possible on both runtimes? If it is possible, then the init migration script will take care, but in that case, these new APIs are not needed I guess.

@yussufsh Can you pls give clarity on this so we can decide whether to go ahead with API implementation or not.

yussufsh:
@manju956 lgtm now.
The only worry is adding a pagination limit to the API. Can we implement that in the next release?

manju956:
> APIs

As discussed, we will have the API implemented with limit support, when the client calls the API, they can pass limit value as -1 so they get all the records in one shot.

yussufsh (approved):
lgtm

dharaneeshvrd:
@prachikurade1 Since you have reviewed with requested changes, please approve and merge the PR if things are looking good.

prachikurade1 (approved):
approved

PR #837, opened 2026-06-01
Fix pgdata permission issue on podman

yussufsh:
Since the host-mounted path could be 0 for root and other ids for non-root and the postgres uid is 26, we were getting permission issues in the entrypoint script for the postgres db pod. To overcome adding a workaround of setting the permission to the pgsql path.
In an ideal case, we could use podman pvc, but then other places in the application need changes ie. delete with/without cleanup, etc. Also, PostgreSQL in the digitize and OpenSearch pod needed a similar fix.

### DIFF
--- a/ai-services/assets/catalog/podman/templates/catalog-db.yaml.tmpl
+++ b/ai-services/assets/catalog/podman/templates/catalog-db.yaml.tmpl
@@ -9,6 +9,20 @@ metadata:
     ai-services.io/secret: "catalog-db-secret"
     ai-services.io/secret-skip-cleanup: "true"
 spec:
+  initContainers:
+  - name: fix-permissions
+    image: "{{ .Values.db.image }}"
+    command:
+    - /bin/sh
+    - -c
+    - |
+      chown -R 26:0 /var/lib/pgsql
+      chmod -R g=u /var/lib/pgsql
+    volumeMounts:
+    - name: postgres-data
+      mountPath: /var/lib/pgsql
+    securityContext:
+      runAsUser: 0
   containers:
   - name: postgresql
     image: "{{ .Values.db.image }}"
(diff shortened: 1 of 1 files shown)


mayuka-c (commented):
LGTM

mayuka-c (approved):

mkumatag:
> Since the host-mounted path could be 0 for root and other ids for non-root and the postgres uid is 26, we were getting permission issues in the entrypoint script for the postgres db pod. To overcome adding a workaround of setting the permission to the pgsql path. In an ideal case, we could use podman pvc, but then other places in the application need changes ie. delete with/without cleanup, etc. Also, PostgreSQL in the digitize and OpenSearch pod needed a similar fix.

I thought this solved the issue - https://github.com/IBM/project-ai-services/pull/808,

yussufsh:
#808 had logic to run as any user for OpenShift env.
This PR addresses a common issue with images where data volume from the host fails due to a permission mismatch.

mkumatag:
> #808 had logic to run as any user for OpenShift env. This PR addresses a common issue with images where data volume from the host fails due to a permission mismatch.

but whoever creates it always be created with 26 uid initially right? 🤔

yussufsh:
> > #808 had logic to run as any user for OpenShift env. This PR addresses a common issue with images where data volume from the host fails due to a permission mismatch.
> 
> but whoever creates it always be created with 26 uid initially right? 🤔

This is for the data mount eg: /var/lib/pgsql/data, where on the pod it has root permission but the user is 26. The image tries to [create/set permission via entrypoint script](https://github.com/IBM/project-ai-services/blob/main/images/postgres/docker-entrypoint.sh#L39-L41) on the mounted path which is owned by root (for eg). This is where it fails, so we are setting the permissions in a root init container to set proper permission for user 26.
Also in case of persistentvolumeclaim on podman, it takes care of setting the permissions.

PR #839, opened 2026-06-01
Automatic generation of password for Opensearch, Postgres + Introduce Secrets

mayuka-c:
- Introduce a new annotation `@generate:password` for fields whose passwords needs to be generated.
- Introduce secrets to store the user and password for both OpenSearch and Postgres.
- Changes to stringData instead of data in secrets as podman automatically takes care of doing base64 enc.
- Remove additional base64 enc done in `catalog configure` flow


ai-services/internal/pkg/utils/password_generator.go
```diff
@@ -34,4 +41,156 @@ func GenerateRandomPassword(length int) (string, error) {
 	return string(password), nil
 }
 
+// ProcessGenerateAnnotationsFromYAML processes @generate annotations in raw YAML data.
+// It parses the YAML with comments preserved, checks for @generate annotations in HeadComments,
+// and replaces empty string values with generated values.
+// Returns the processed YAML as bytes.
+// Supported annotations:
+//   - @generate:password - generates a random password using GenerateRandomPassword
```
mkumatag: instead can we have something like this with those optional ones after the generate:password and override if mentioned.

```
@generate:password length=24, special=true, upper=true, lower=true, digits=true
```

ai-services/internal/pkg/utils/password_generator.go
```diff
@@ -34,4 +41,156 @@ func GenerateRandomPassword(length int) (string, error) {
...
+		if err := processNodeForGenerate(child); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// hasGenerateAnnotation checks if a yaml.Node has a @generate annotation in its HeadComment.
+func hasGenerateAnnotation(n *yaml.Node) bool {
```
mkumatag, line 161: these are some of these generic ones, wondering whether we are using these for the earlier tags like hidden etc..

ai-services/internal/pkg/utils/password_generator.go
mayuka-c: Sure I will take care of this

ai-services/internal/pkg/utils/password_generator.go
mayuka-c: Yeah we can do that. The existing methods are quite not generic to be extendable. I will try to refactor

mayuka-c: For now have added a generic utility method for reading HeadCommentNodes in yaml. The existing as said is not extensible. So if required we can move the existing ones also to use the new `NodeProcessor` utility func method arg but I didnt want to take it up here as I did not want to cause any regression with older CLI way. So probably once we migrate the CLI to use new Catalog, we probably can remove those (as it is no longer used like hidden and description) and any new HeadCommentNodes that we introduce can follow the new pattern

ai-services/internal/pkg/utils/password_generator.go
mayuka-c: This is done

mkumatag (dismissed):
LGTM for the `@generate:password` tag, others can review remaining code

ai-services/internal/pkg/utils/util.go
```diff
@@ -166,6 +166,71 @@ func getDescription(n *yaml.Node) string {
 	return strings.TrimSpace(desc)
 }
 
+// NodeProcessor is a function type for processing individual yaml nodes.
+// It receives the key node, value node, and returns an error if processing fails.
+type NodeProcessor func(keyNode, valueNode *yaml.Node) error
+
+// ProcessYAMLNode recursively processes a yaml.Node tree with a custom processor function.
+// This is a generic traversal function that can be used for various node processing tasks.
+func ProcessYAMLNode(node *yaml.Node, processor NodeProcessor) error {
```
yussufsh, line 176: lets add a TODO to remove older implementation.

yussufsh (commented):
Overall, lgtm but many places proper logs need to be addressed and update db on failure for applications needs to be implemented.

ai-services/internal/pkg/utils/util.go
mayuka-c: Done

yussufsh (approved):

PR #760, opened 2026-05-18
Extract common chatbot vars 

sats-23:

PR #841, opened 2026-06-01
Update search_only in chatbot

Bleama02:
-Chatbot uses shared retrieval function internally
-All existing tests pass
-Performance metrics continue to be tracked separately
-No behavioral changes to either service
-New tests added for enhanced functionality

[https://jsw.ibm.com/browse/AISERVICES-1027](https://jsw.ibm.com/browse/AISERVICES-1027 )


dharaneeshvrd (commented):
This will not work, chatbot does not have access to similarity code base.
We need to call `/v1/similarity-search` endpoint to replace this search.

Bleama02:
> This will not work, chatbot does not have access to similarity code base. We need to call `/v1/similarity-search` endpoint to replace this search.

@dharaneeshvrd Ah. Okay, i originally misunderstood what the task was asking. Ill make the changes today so that is actually calls the similarity-search endpoint instead of importing the code itself.  Ill let you know when do. Thank you.

services/chatbot/settings.py
```diff
@@ -131,6 +131,16 @@ class RAGConfig(BaseSettings):
         "Answer only the specific question asked. Do not add conversational filler, offer additional assistance, suggest follow-up steps, or ask follow-up questions at the end of your response. End your response immediately once the question has been answered."
     )
 
+    similarity_service_url: str = Field(
```
dharaneeshvrd, line 134: You need to populate this URL at pod level, let's add envs like in chatbot UI and follow similar notation.
https://github.com/IBM/project-ai-services/blob/main/ai-services/assets/applications/rag/podman/templates/chat-bot.yaml.tmpl#L26to%20L29
And you need to add proper model config to refer using env.

dharaneeshvrd: You can take a ref from this PR https://github.com/IBM/project-ai-services/pull/760 too

dharaneeshvrd:
@Bleama02 Please get help from @mayuka-c to add these similarity envs on chatbot to new asset structure.

Bleama02:
@dharaneeshvrd or @mayuka-c, Im trying to test to see if my most recent code works but when running `ai-services application create rag-cpu --runtime podman --template rag-cpu --skip-model-download --skip-validation spyre` the deployment now fails at the ingest-docs pod saying:

`"Deploying application 'rag-cpu'...Error: layer 3: 'ingest-docs.yaml.tmpl': Failed to deploy pod and do readiness check: containers creation check failed for pod: 'rag-cpu--ingest-docs' with error: operation timed out waiting for container creation"`

Do you know if this is because of something i changed? It wasnt doing that yesterday when i was testing it.

services/similarity/similarity_utils.py
```diff
@@ -83,18 +84,26 @@ def perform_similarity_search(
...
 
-    Returns:
+    Returns (return_timings=False):
         docs       - list of document dicts (page_content, filename, type, source, chunk_id)
         scores     - parallel list of float scores
         score_type - "cosine", "bm25", "hybrid", or "relevance" (when reranked)
+
+    Returns (return_timings=True):
+        docs, scores, score_type, perf_stat_dict
+        where perf_stat_dict contains "retrieve_time" and, when rerank=True, "rerank_time".
```
sats-23, line 102: Is there a specific purpose to these perf metrics being returned from here other than the tests?
If it is for generic storage of metrics, I suggest you implement a similar endpoint like chatbot at `/v1/perf_metrics` and keep it consistent with the same request ID (pass through from chatbot when called there), so the metrics are more traceable and usable for diagnosing.

Bleama02: The return_timings parameter is currently only used for unit tests. The similarity API endpoint doesn't expose these metrics - chatbot tracks its own API call timing via perf_stat_dict["similarity_api_time"].

For cross-service tracing with request IDs, we could add a /v1/perf_metrics endpoint to similarity service as a follow-up enhancement. Would you like me to create a separate issue for that?

sats-23: I believe yes. If an end-user is trying to debug why their similarity search queries are failing/slow and are verifying it through the chatbot calls, they may need to be able to correlate what input question/request is causing this slowness/failure to be able to debug further.

@dharaneeshvrd Please suggest

dharaneeshvrd: Agree an explicit perf_metrics endpoint would be good at similarity search service. 
Definitely not part of this PR. Let's create a story for that to handle later.
But I see this is not being sent as a response to the similarity search call, how this can be passed and used at chatbot side?

dharaneeshvrd:
@Bleama02 even though the app create is failing with that specific error, the pods would have been deployed successfully and can be used for testing. Please try to test your changes and fix the comments provided.

services/similarity/similarity_utils.py
Bleama02: You're right, currently the timing metrics aren't returned by the similarity API, so chatbot can't access them. The return_timings parameter exists but isn't used by the API endpoint. To enable this, i think we'd need to modify the similarity API response to include timing data either in response body, headers, or a separate metrics endpoint. Could also be part of the follow-up story.

Bleama02:
Okay, it took me a while to figure out how to test it but i believe I got the test working. Here is a screenshot of chatbot making the request and similarity search receiving it. Im not sure if there is a way to test it with a document while there is still that bug with the ingest pod deployment.

image

services/similarity/similarity_utils.py
dharaneeshvrd: Ideally this should have been added in this PR, But I m ok to add it in a follow up PR too.

dharaneeshvrd (approved):

services/similarity/similarity_utils.py
dharaneeshvrd: > modify the similarity API response to include timing data either in response body, headers

@sats-23 Can you please take care of this? IMO it can be part of the headers for now.

sats-23: Sure, will do

manju956: I have raised PR https://github.com/IBM/project-ai-services/pull/896 to include `timing_info` as part of similarity API response header.
CC: @dharaneeshvrd @sats-23 @Bleama02

PR #842, opened 2026-06-02
fix: Add registry authentication for catalog image pulls 

Shubhamag12:
Using secret to mount auth.json to pull images inside the container


shwetha-s-poojary (approved):
lgtm

yussufsh (approved):
lgtm
please take care of non-root user flow.

PR #795, opened 2026-05-22
Validate request body for create application

Shubhamag12:
Added logic to validate the POST request body

# Validations Added
## 1. Catalog ID Validation
- Validates that the catalog ID exists as either an architecture or a service
- Returns 404 error if catalog ID is not found

## 2. Architecture Deployment Validations
- Architecture exists in catalog
- Architecture version is valid
- All requested services exist in catalog
- All requested services have valid versions
- All requested services are compatible with the architecture
- Service-level parameters are validated against service schema
- All component providers exist in catalog
- Component versions are valid
- Component parameters are validated against component schema
- Component parameters are consistent across all services

## 3. Standalone Service Deployment Validations
- Service exists in catalog
- Service has standalone: true flag (prevents deployment of non-standalone services)
- Service version is valid
- Exactly one service is specified in the request
- Service ID in request matches the catalog ID
- Service-level parameters are validated against service schema
- All component providers exist in catalog
- Component versions are valid
- Component parameters are validated against component schema
- Component parameters are consistent

## 4. Component Consistency Validation
- Same component (type + provider) must have identical parameters across all occurrences
- Prevents configuration conflicts

## 5. Parameter Schema Validation
- Validates all provided parameters are supported (prevents unsupported parameters)
- Validates parameter values against JSON schema definitions
- Validates nested object parameters
- Provides detailed error messages with supported parameter lists

PR #844, opened 2026-06-02
Introduce secrets for WatsonX and VLLM providers

mayuka-c:
- Store WatsonX configurations as secrets
- Expose VLLM API Key within values.schema.json. Also store it as secrets


ai-services/assets/components/llm/watsonx/podman/templates/watsonx-secret.yaml.tmpl
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: "llm-secret-{{ .InstanceSlug }}"
```
mkumatag: can we rename to `watsonx-secret-{{ .InstanceSlug }}`?

mkumatag (changes requested):

ai-services/assets/components/llm/vllm-cpu/podman/templates/vllm-server.yaml.tmpl
```diff
@@ -36,10 +36,11 @@ spec:
         value: auto
       - name: VLLM_CPU_KVCACHE_SPACE
         value: "32"
-      {{- if .Values.apiKey }}
```
mkumatag, line 39: why are we removing if block? we were thinking about having an optional variable right? curious to know how this works when user doesn't supply the key?

mkumatag (changes requested):

ai-services/assets/components/llm/vllm-cpu/podman/templates/vllm-server.yaml.tmpl
mayuka-c: Yeah, Currently with secret introduction if an API key is not passed, then simply a podman secret will be generated with empty string, which I feel might not be correct.

I will make changes to have conditional deployment of secrets and I feel let us have `if` condition here because of it. Will make the updates

mkumatag (dismissed):

mayuka-c:
PR is good for review.
I did validations and found couple of issues which I have fixed. Pls take a look
- Introduce PodTemplateExecutions
- Podman kube play will fail if the rendered yaml is empty (vllm secret file will be empty if key is not passed). So I have added a condition in deployer.go to skip `podman kube play` if renderedBytes is empty.

yussufsh:
@mayuka-c  do we have a validation in place if the user provides different API keys for two VLLM components with the same backend?

mayuka-c:
> @mayuka-c do we a validation in place if the user provides different API keys for two VLLM components with the same backend?

That should be taken care in validation flow of `Create Application` flow here: https://github.com/IBM/project-ai-services/pull/795

I have already informed @Shubhamag12 to take care of params validation in his validation PR

Shubhamag12 (dismissed):
lgtm
just one question
how are we handling deletion of these secrets? as part of application delete?

mayuka-c:
> lgtm just one question how are we handling deletion of these secrets? as part of application delete?

Yes

yussufsh (approved):

PR #846, opened 2026-06-02
 docs: update non-root proposal for catalog related changes

Shubhamag12:
Updated proposal as per: #817

### DIFF
--- a/docs/proposals/non-root-user-configuration-proposal.md
+++ b/docs/proposals/non-root-user-configuration-proposal.md
@@ -38,7 +38,7 @@ This document outlines the configuration changes and system requirements impleme
    # Enable lingering to keep user services running
    sudo loginctl enable-linger <username>
    ```
-5. While running `ai-services catalog configure` cmd, make sure to add `--https-port 8443`, as 443 is a privileged port. For PowerVS environments, use `--https-port 6443` along with a custom domain configuration, as 6443 is the only available port option for non-root users on PowerVS.
+5. While running `ai-services catalog configure` cmd, make sure to add `--https-port 8443`, as 443 is a privileged port. For PowerVS environments, use `--https-port 6443` along with a custom domain configuration, as 6443 is one of the available port option for non-root users on PowerVS.
 
 **Why This Is Required:**
 - Group membership changes require a new login session
--- a/docs/proposals/non-root-user-configuration-proposal.md
+++ b/docs/proposals/non-root-user-configuration-proposal.md
@@ -79,10 +79,17 @@ drwxr-xr-x. 6 root root 360 Apr 27 11:53 ..
 - Containers could not access Spyre cards, causing crashes
 
 ### Solution Implemented
-Created custom SELinux policy module `vllm_vfio_policy` that:
+Created custom SELinux policy modules:
+
+**1. VFIO Device Access Policy (`vllm_vfio_policy`)**:
 - Allows `container_t` type to access `vfio_device_t` devices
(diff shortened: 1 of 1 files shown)


docs/proposals/non-root-user-configuration-proposal.md
```diff
@@ -242,6 +254,31 @@ SMT level affects CPU performance for Spyre card operations. Previously configur
...
+ERRO[0003] "rootlessport cannot expose privileged port 443, you can add 'net.ipv4.ip_unprivileged_port_start=443' to /etc/sysctl.conf (currently 1024), or choose a larger port number (>= 1024): listen tcp 0.0.0.0:443: bind: permission denied"
+Error: unable to start container "f72329a97540825bb753ec41806021c9fd15890ae54f6874b2f13e2c6bc8890a": starting some containers: internal libpod error
+```
+
+### Solution Implemented
+**Requirement**: Users must specify a non-privileged port when configuring the catalog service.
+
+**Usage**:
+```bash
+ai-services catalog configure --http-port 8443
```
mkumatag: if I'm not wrong, this is for the caddy server, the reason for asking to use 443 is for using in the PowerVS kind of an env where we expose only limited ports, e.g: here are the list of ports allowed for the powervs infra,

Image

mkumatag: cc @shwetha-s-poojary @yussufsh

mkumatag (dismissed):

docs/proposals/non-root-user-configuration-proposal.md
yussufsh: We can guide them to use 6443 for powervs which is only available option I guess. For non-root, on PowerVS they need to use custom domain and custom port (ie. 6443).

shwetha-s-poojary: We could also add a validation check and direct them to this doc, if we want to be defensive

mkumatag: I'm not sure if there is way to validate this! ;) it's always a challenge to check ingress port and a floating ip address.!

mkumatag: Please create a documentation task to take of those for the powervs environment somewhere..

Shubhamag12: CC: @irapandey

docs/proposals/non-root-user-configuration-proposal.md
```diff
@@ -38,6 +38,7 @@ This document outlines the configuration changes and system requirements impleme
    # Enable lingering to keep user services running
    sudo loginctl enable-linger <username>
    ```
+5. While running `ai-services catalog configure` cmd, make sure to add `--https-port 8443`, as 443 is a privileged port. For PowerVS environments, use `--https-port 6443` along with a custom domain configuration, as 6443 is the only available port option for non-root users on PowerVS.
```
yussufsh: 6443 is not the only available port.. there are a few more if I am not wrong.
You can mention 6443 is an available alternate port.

yussufsh (commented):
Please include the policy labels we are creating as part of the proposal and then we can close this PR.

docs/proposals/non-root-user-configuration-proposal.md
Shubhamag12: there are, yes
> You can mention 6443 is an available alternate port.

so above comment should be enough, right?

yussufsh (approved):

PR #850, opened 2026-06-03
[Catalog-API] Update Service and Component Endpoints in DB during App Creation flow

mayuka-c:
- Update externally registered routes using Caddy in DB
- Also update component internal endpoints in DB

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
+++ b/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
@@ -376,7 +376,10 @@ func (s *ApplicationService) insertComponentRecords(
 		instanceUUID := uuid.New()
 
 		// Filter metadata to exclude sensitive data based on schema
-		metadata := s.filterComponentMetadata(comp.ComponentType, comp.ProviderID, comp.Params)
+		metadata, err := s.filterComponentMetadata(comp.ComponentType, comp.ProviderID, comp.Params)
+		if err != nil {
+			return nil, fmt.Errorf("failed to filter component metadata for %s: %w", hash, err)
+		}
 
 		component := &models.Component{
 			ID:       instanceUUID,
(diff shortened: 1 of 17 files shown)


mayuka-c:
Updated Validation

image

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -1148,62 +1155,132 @@ func generateInstanceSlug(id string) string {
...
 		}
+
+		// Convert registered routes to endpoint format
+		for _, route := range registeredRoutes {
+			endpoint := map[string]any{
+				"endpoint": fmt.Sprintf("https://%s", route.Domain),
+				"type":     "external",
+			}
+			serviceEndpoints = append(serviceEndpoints, endpoint)
+		}
```
shwetha-s-poojary, line 1251: @mayuka-c  Have you tested with custom https-port?
You need to have that port  (which you can access via CADDY_HTTPS_PORT env var )as well here in the endpoint, we won't be able to access endpoints  if user specifies a custom one otherwise.

mayuka-c: Oh I havent tried.
Thanks let me add it and try it once

mayuka-c: Done made updates, working fine

shwetha-s-poojary (dismissed):
lgtm

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
```diff
@@ -661,4 +664,21 @@ func (s *ApplicationService) performDeletion(ctx context.Context, appID uuid.UUI
...
+func (s *ApplicationService) filterComponentMetadata(params map[string]any) map[string]any {
+	if params == nil {
+		return nil
+	}
+
+	metadata := make(map[string]any)
+
+	// Only include model-related fields
+	if model, ok := params["model"]; ok {
+		metadata["model"] = model
```
yussufsh: Instead of hardcoding filter params here can we skip only sensitive items that are stored in the secret somehow? Maybe read the schema format and skip adding to metadata.

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -38,6 +38,8 @@ import (
 	k8syaml "sigs.k8s.io/yaml"
 )
 
+const defaultHTTPSPort = "443"
```
yussufsh: Use common const.

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -1148,62 +1157,141 @@ func generateInstanceSlug(id string) string {
...
 			continue
 		}
+
+		// Convert registered routes to endpoint format
+		for _, route := range registeredRoutes {
+			url := fmt.Sprintf("https://%s", route.Domain)
+			if httpsPort != defaultHTTPSPort {
+				// if not 443 then we need to append the port too
+				url = url + ":" + httpsPort
+			}
```
yussufsh: Lets move to a common util function... to remove 443 whereever needed.

yussufsh (changes requested):

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -1148,62 +1157,141 @@ func generateInstanceSlug(id string) string {
...
+	// Add type field to indicate this is an internal endpoint
+	endpointMap["type"] = "internal"
+
+	// Update component endpoints in database
+	if err := d.componentRepo.UpdateEndpoints(context.Background(), comp.DatabaseID, endpointMap); err != nil {
+		return fmt.Errorf("failed to update component %s endpoints: %w", comp.ComponentType, err)
 	}
 
-	logger.Infof("Registered %d route(s) for application '%s'\n", len(allRoutes), plan.ApplicationName)
+	logger.Infof("Updated component %s with internal endpoint in database: %v\n", comp.ComponentType, endpointMap)
```
mkumatag: 1300+ lines in a file is insane, please consider splitting into multiple files later..

mkumatag:
> Validation
> 
> image

I'm just wondering why we are calling  that `host`in the endpoints field in the components table?

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
mayuka-c: Might have to read values.schema to determin sensitivity and based on it store it in DB. Will make the change

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
mayuka-c: Yeah we need to do bit of refactoring too as we added lots of things like Caddy registration of routes and few more

Now since the deploy flow is almost done with this PR and stabilized, we will do a bit of refactoring/cleanup next

mayuka-c:
> > Validation
> > image
> 
> I'm just wondering why we are calling that `host`in the endpoints field in the components table?

Not any particular reason. I see all of our services which talks to components uses separate host and port. So I thought if type is `internal` we will keep host and port as separate.

But open for any suggestions

mkumatag:
> Not any particular reason. I see all of our services which talks to components uses separate host and port. So I thought if type is `internal` we will keep host and port as separate.
> 
> But open for any suggestions

let us discuss on this tomorrow, I want to understand more on this.

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
mayuka-c: Done

mayuka-c:
> > > Validation
> > > image
> > 
> > 
> > I'm just wondering why we are calling that `host`in the endpoints field in the components table?
> 
> Not any particular reason. I see all of our services which talks to components uses separate host and port. So I thought if type is `internal` we will keep host and port as separate.
> 
> But open for any suggestions

@mkumatag I updated the screenshots with latest changes.
So now the endpoints column in both services and components table are having `type` and `url` (instead of endpoint key). PTAL. Also we could sync up to discuss more if required

mayuka-c:
Based on suggestions from @mkumatag I have changed to the type to be "ui" and "api" (which is basically determined based on the route type present in annotation)

Here is the updated DB with endpoints

image

mkumatag:
> Based on suggestions from @mkumatag I have changed to the type to be "ui" and "api" (which is basically determined based on the route type present in annotation)
> 
> Here is the updated DB with endpoints
> 
> image

This looks better, I'm not sure how this works for the openshift kind of environment, please that also in mind in the future.

mayuka-c:
> > Based on suggestions from @mkumatag I have changed to the type to be "ui" and "api" (which is basically determined based on the route type present in annotation)
> > Here is the updated DB with endpoints
> > image
> 
> This looks better, I'm not sure how this works for the openshift kind of environment, please that also in mind in the future.

Sure. I feel even in OpenShift the DB structure can still work (storing endpoints).
We would not have route annotation in OpenShift (as we dont have to run external Reverse proxy - Caddy). But we define `routes` resource to configure externally exposed routes and there we can have a label/annotation to determine the type

yussufsh (approved):
lgtm

PR #852, opened 2026-06-03
Escape HTML char

sats-23:
-Chatbot breaks when rendering stream response having HTML characters like angle brackets (</>)
-Refer [AISERVICES-1011](https://jsw.ibm.com/browse/AISERVICES-1011)
escape_chat_fix

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=0.0.33
+TAG?=0.0.34
 CONTAINER_BUILDER?=podman
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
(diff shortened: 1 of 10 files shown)


dharaneeshvrd (commented):
Bob reporting this, please see if it is valid.

**Potential Double-Escaping Issue**

- The escapeHtml() function is called on textChunk during streaming (line 151), which accumulates into fullText
- Then escapeHtml() is called again on fullText at lines 174 and 236

Risk: This could lead to double-escaping where < becomes &amp;lt; instead of &lt;

Recommendation: Only escape at the final rendering points (lines 174 and 236), not during streaming accumulation

sats-23:
> Bob reporting this, please see if it is valid.
> 
> **Potential Double-Escaping Issue**
> 
> * The escapeHtml() function is called on textChunk during streaming (line 151), which accumulates into fullText
> * Then escapeHtml() is called again on fullText at lines 174 and 236
> 
> Risk: This could lead to double-escaping where < becomes &lt; instead of <
> 
> Recommendation: Only escape at the final rendering points (lines 174 and 236), not during streaming accumulation

I guess not, the full text var stored is actually the concatenation of original textChunks. 
The escaping happens on the partial chunks (to display on UI) but concatenated raw to full text.
And when finally done, we are escaping on the un-escaped full text, so there should be no problem of double escaping

dharaneeshvrd (dismissed):
@pranithraoibm Can you please take a look at this PR? Its a minor one

pranithraoibm (dismissed):
LGTM

Niharika0306 (dismissed):

dharaneeshvrd (approved):

PR #853, opened 2026-06-03
Updating service-base image to use devpi for python pkgs

adarshagrawal38:
`lingua-language-detector `and `ujson` are now available in devpi index, so updating service-base to make use of it.

Verified docling in updated service base image by parsing one pdf document. Snapshot of parsed result.

image

### DIFF
--- a/images/service-base/Containerfile
+++ b/images/service-base/Containerfile
@@ -1,6 +1,4 @@
-FROM localhost/prebuilder as prebuilder
-
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.7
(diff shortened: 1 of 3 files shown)


dharaneeshvrd (commented):
Now prebuilder can be removed right?

mkumatag:
> Now prebuilder can be removed right?

may be we can keep it for now with some comment saying - not used and for reference for the future purpose.

mkumatag (dismissed):

mkumatag (approved):

mkumatag:
@adarshagrawal38 please update other images with this base image

PR #854, opened 2026-06-03
Direct and chunked summ

manalilatkar:
Features covered in this PR: Background task of async summarization handling implemented, the process decides whether summarization is going to happen in direct or chunked format and then summarization is completed according to the strategy chosen.


services/summarize/app.py
```diff
@@ -419,27 +421,341 @@ async def summarize(request: Request):
...
-    # 4. Perform summarization
-    # 5. Write result file
-    # 6. Update job status to 'completed'
-    # 7. Clean up staging directory
-    logger.info(f"[STUB] Background processing completed for job {job_id}")
+    from pathlib import Path
+    import json
+    from datetime import datetime, timezone
+    from summarize.db.repository import db_repo
+    from summarize.summ_types import JobStatus, SummarizationType
```
dharaneeshvrd: these imports are invalid 
maybe renamed

services/summarize/app.py
```diff
@@ -419,27 +421,341 @@ async def summarize(request: Request):
...
+            job_id,
+            status=JobStatus.FAILED,
+            error=str(e),
+            completed_at=datetime.now(timezone.utc)
+        )
+    
+    finally:
+        # Step 7: Clean up staging directory
+        try:
+            cleanup_staging_directory(job_id)
```
dharaneeshvrd: seems base_dir arg is missing

services/summarize/app.py
```diff
@@ -419,27 +421,341 @@ async def summarize(request: Request):
...
-    # TODO: Implement actual summarization logic:
-    # 1. Read staged file
-    # 2. Extract text (PDF or TXT)
-    # 3. Determine strategy (direct vs chunked)
-    # 4. Perform summarization
-    # 5. Write result file
-    # 6. Update job status to 'completed'
-    # 7. Clean up staging directory
-    logger.info(f"[STUB] Background processing completed for job {job_id}")
+    from pathlib import Path
```
dharaneeshvrd: Unused

services/summarize/app.py
```diff
@@ -419,27 +421,341 @@ async def summarize(request: Request):
...
+        
+        # Update job with word count and set to in_progress
+        db_repo.update_job(
+            job_id,
+            status=JobStatus.IN_PROGRESS,
+            metadata={"document_word_count": input_word_count}
+        )
+        
+        # Get job details for level
+        job = db_repo.get_job_by_id(job_id)
```
dharaneeshvrd: can we not get this from caller instead of checking in db?

services/summarize/app.py
dharaneeshvrd: setting default level is already done, can directly assign the level

services/summarize/app.py
```diff
@@ -419,27 +421,341 @@ async def summarize(request: Request):
...
+        # Step 4: Determine strategy - check if input alone fits in context window
+        max_model_len = get_llm_max_model_len()
+        prompt_tokens = settings.summarize.summarization_prompt_token_count
+        
+        # Calculate available space for output
+        available_output_tokens = max_model_len - input_tokens - prompt_tokens
+        
+        # Always compute target/max tokens (needed for both strategies)
+        target_words, min_words, max_words, max_tokens = compute_target_and_max_tokens(
+            input_tokens, max(available_output_tokens, 1000), level, None  # Use minimum 1000 if negative
```
dharaneeshvrd: I think calling compute_target_and_max_tokens can go inside else block of below if, since this would be used anyway for direct summarization.
And `max(available_output_tokens, 1000)` this logic can also be avoided

services/summarize/app.py
```diff
@@ -419,27 +421,341 @@ async def summarize(request: Request):
...
+                content_text,
+                MAX_INPUT_WORDS,
+                settings.summarize.chunk_overlap_sentences
+            )
+            
+            num_chunks = len(chunks)
+            logger.info(f"Split into {num_chunks} chunks")
+            
+            # Pre-check: Estimate if combined summaries will fit
+            estimated_chunk_summary_tokens = estimate_chunk_summary_tokens(num_chunks, max_tokens)
```
dharaneeshvrd: you can't use the file level max_tokens here
you might want to calculate the chunk level values and use it here

services/summarize/app.py
manalilatkar: @dharaneeshvrd you mean send the level along with job id to the background task?

dharaneeshvrd: yes

services/summarize/app.py
manalilatkar: Bob type checker was giving an error for just level as a parameter, but it will work.

services/summarize/app.py
manalilatkar: @dharaneeshvrd how does this sound to calculate chunk level max words:

```
# Calculate chunk-level token values
# Estimate tokens for a typical chunk (use MAX_INPUT_WORDS as reference)
chunk_input_tokens = int(MAX_INPUT_WORDS / settings.common.llm.token_to_word_ratio_en)
chunk_available_output_tokens = (
    get_llm_max_model_len()
    - chunk_input_tokens
    - settings.summarize.summarization_prompt_token_count
)

# Compute chunk-level target and max tokens
chunk_target_words, chunk_min_words, chunk_max_words, chunk_max_tokens = compute_target_and_max_tokens(
    chunk_input_tokens,
    chunk_available_output_tokens,
    level,
    None
)
```

dharaneeshvrd: > chunk_input_tokens = int(MAX_INPUT_WORDS / settings.common.llm.token_to_word_ratio_en)

Instead of this, you need to calculate based on length of words present in each chunk. For each chunk need to calculate max tokens.
Basically treat each chunk as direct summarization.

manalilatkar: @dharaneeshvrd  When we create chunks, the size of each chunk is at maximum MAX_INPUT_WORDS, still you are saying we need to count the tokens separately of each chunk (using tokenize_with_llm everytime)?

dharaneeshvrd: ok, but how much it would deviate from MAX_INPUT_WORDS?
Since we are creating chunks on the range of MAX_INPUT_WORDS not exactly with this value right?

manalilatkar: correct, not exactly this value. We can't say how much it would deviate because the logic checks for paragraph based and then sentence based splitting.

dharaneeshvrd: That's why I wanted to calculate

dharaneeshvrd:
@manalilatkar Let's update the description with more info

services/coverage.xml
dharaneeshvrd, line 1: please remove this file and update gitignore

manalilatkar: Done.

services/summarize/app.py
```diff
@@ -419,27 +421,446 @@ async def summarize(request: Request):
...
-    # 3. Determine strategy (direct vs chunked)
-    # 4. Perform summarization
-    # 5. Write result file
-    # 6. Update job status to 'completed'
-    # 7. Clean up staging directory
-    logger.info(f"[STUB] Background processing completed for job {job_id}")
+    import json
+    from datetime import datetime, timezone
+    from summarize.db.manager import db_repo
+    from summarize.models import JobStatus, SummarizationType
```
dharaneeshvrd:
```suggestion
    from summarize.models import JobStatus
```

services/summarize/app.py
```diff
@@ -419,27 +421,446 @@ async def summarize(request: Request):
...
+    from summarize.chunk_utils import (
+        split_text_into_chunks,
+        estimate_chunk_summary_tokens,
+        build_merge_messages
+    )
```
dharaneeshvrd, line 443:
```suggestion
    from summarize.chunk_utils import (
        split_text_into_chunks,
        build_merge_messages
    )
```

services/summarize/app.py
```diff
@@ -419,27 +421,446 @@ async def summarize(request: Request):
...
+        # Step 3: Compute metrics
+        input_word_count = word_count(content_text)
+        logger.info(f"Document word count: {input_word_count}")
+        
+        # Get LLM endpoint for tokenization
+        llm_endpoint = llm_model_dict['llm_endpoint']
+        llm_model = llm_model_dict['llm_model']
+        
+        # Tokenize input
+        input_tokens = await asyncio.to_thread(
```
dharaneeshvrd, line 491: Missed this last time, will it work for content larger than max model length?
What is the biggest file you have validated?
Also it would be great if you can document the results in the JIRA.

services/summarize/app.py
```diff
@@ -419,27 +421,446 @@ async def summarize(request: Request):
...
+            
+            if total_required_tokens <= max_model_len:
+                # Direct summarization strategy
+                logger.info(f"Using DIRECT strategy (fits in context window)")
+                strategy = "direct"
+            else:
+                logger.info(f"Using CHUNKED strategy (exceeds context window)")
+                strategy = "chunked"
+                # Reset file-level tokens - will be calculated later for merge step
+                target_words, min_words, max_words, max_tokens = None, None, None, 0
```
dharaneeshvrd, line 547: defining this to None is not needed IMO, since chunked flow never uses this.

dharaneeshvrd (commented):
Overall chunk summarization logic looks good to me

services/summarize/app.py
manalilatkar: MAx I have tested a 218 pages Introduction to PowerVM book, which is chunked into 6 parts. The token count for that is 126852.

manalilatkar: will test some more.

dharaneeshvrd: Let's ensure this still works for content that's larger than max model length, if not we need to look at how to calculate tokens in parts.

manalilatkar: Works for content larger than model length, for which we use chunked summarization. Upto chunk 7-8, I have checked.

services/summarize/app.py
manalilatkar: @dharaneeshvrd, it stops unbound access errors in Bob. Can we let it be? It's safe code.

dharaneeshvrd: Ok

services/summarize/app.py
dharaneeshvrd: Ok, please post the document and summary output in JIRA

services/summarize/app.py
dharaneeshvrd: Let me know once this is taken care of, we can merge it.

services/summarize/app.py
manalilatkar: Posted details of documents and their summary outputs in Jira.

services/summarize/app.py
manalilatkar: @dharaneeshvrd all issues taken care of.

dharaneeshvrd:
Let's fix the version and merge it

dharaneeshvrd (approved):

PR #857, opened 2026-06-03
AISERVICES-1139: Proposal for extending conversation support to German.

manalilatkar:


sats-23:
Do we need a guardrails flag to set the language of the session based on the first message?
Thinking of a case when user starts a chat in English and later on proceeds to ask follow up questions in German, what about those instances?

docs/proposals/conversational_rag.md
```diff
@@ -1188,3 +1188,500 @@ def test_get_conversation_context():
...
+```
+
+### 11.3 Proposed Solution
+
+**Extend the existing conversational framework to support German** by:
+
+1. Adding German-specific conversational prompts
+2. Enabling German query rephrasing with German prompt templates
+3. Refactoring payload construction to use conversational mode for both languages
+4. Maintaining language-specific token budgets
```
sats-23, line 1238: Please come up with appropriate numbers for vars like max_response_tokens, max_response_tokens_multiplier, history_token_budget, prompt_template_token_count initial_system_token_overhead, rag_system_token_overhead that are `DE` specific

docs/proposals/conversational_rag.md
```diff
@@ -1188,3 +1188,500 @@ def test_get_conversation_context():
...
+    llm_model: str,
+    config: Optional[Dict] = None,
+    api_key: str | None = None,
+    lang: Optional[str] = None
+) -> str:
+    # Use provided lang or detect if not provided
+    detected_lang = lang if lang is not None else detect_language(current_query)
+    
+    # Support both English and German
+    if detected_lang not in [lang_en, lang_de]:
```
sats-23, line 1385: Since the goal will be to support conversational mode across all other languages, I suggest we don't add conditionals like this but rather focus on removing the legacy system prompts and all code pointing to non-English languages - https://github.com/IBM/project-ai-services/blob/0720e706d48bcb6c6e7868b6bb5e826f43aeca02/services/common/llm_utils.py#L258-L295

dharaneeshvrd: yes, let's do this.

manalilatkar:
Closing this as it was implemented in another PR.

PR #860, opened 2026-06-03
Add missing similarity in rag app and introduce standalone flag

yussufsh:
1. Added similarity service which was missed before under the rag architecture.
2. Added a standalone flag in services metadata to allow installing individual services.

### DIFF
--- a/ai-services/assets/services/similarity/metadata.yaml
+++ b/ai-services/assets/services/similarity/metadata.yaml
@@ -1,6 +1,6 @@
 # Service identification
 id: similarity
-name: "Similarity Search"
+name: "Find Similar Items"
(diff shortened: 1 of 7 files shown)


yussufsh:
@Shubhamag12 FYI `standalone` is something we need to take care in the validation for services deployment. ie. deploy only if standalone=true.

mkumatag (approved):

ai-services/assets/services/similarity/metadata.yaml
```diff
@@ -17,4 +17,6 @@ dependencies:
   - id: embedding
   - id: reranker
 
+standalone: true
```
mayuka-c, line 20: @yussufsh 
This should be standalone: false and digitize should be true

PR #867, opened 2026-06-04
[Catalog] New Postgres instance for Summarize

mayuka-c:
- Fixes the Summarize failure with the Catalog flow of Application Deployment as Postgres instance is not present

Note: This ensures that a separate DB instance is used for Summarize

### DIFF
--- a/ai-services/assets/services/digitize/podman/templates/postgres.yaml.tmpl
+++ b/ai-services/assets/services/digitize/podman/templates/postgres.yaml.tmpl
@@ -67,5 +67,5 @@ spec:
   volumes:
     - name: postgres-data
       hostPath:
-        path: "{{ .BaseDir }}/applications/{{ .InstanceSlug }}/volumes/digitize/postgres"
+        path: "{{ .BaseDir }}/applications/{{ .InstanceSlug }}/volumes/postgres-digitize"
         type: DirectoryOrCreate
--- a/ai-services/assets/services/summarize/podman/values.yaml
+++ b/ai-services/assets/services/summarize/podman/values.yaml
@@ -2,3 +2,12 @@ summarize:
   port: ""
   image: icr.io/ai-services-cicd/summarize-service:v0.0.6
   log_level: "INFO"
+  database: "summarize_metadata"
+
+postgres:
+  image: icr.io/ai-services-cicd/postgres:18-2
+  username: "postgres"
+  # @generate:password
+  password: ""
+
+# Made with Bob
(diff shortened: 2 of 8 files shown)


ai-services/assets/services/digitize/podman/templates/postgres.yaml.tmpl
```diff
@@ -67,5 +67,5 @@ spec:
   volumes:
     - name: postgres-data
       hostPath:
-        path: "{{ .BaseDir }}/applications/{{ .InstanceSlug }}/volumes/postgres"
+        path: "{{ .BaseDir }}/applications/{{ .InstanceSlug }}/volumes/digitize/postgres"
```
yussufsh:
```suggestion
        path: "{{ .BaseDir }}/applications/{{ .InstanceSlug }}/volumes/postgres-digitize"
```
Let's follow the same convention as pod, secret name. Also I do not see this list for digitize growing.

ai-services/assets/services/summarize/podman/values.yaml
```diff
@@ -2,3 +2,12 @@ summarize:
   port: ""
   image: icr.io/ai-services-cicd/summarize-service:v0.0.6
   log_level: "INFO"
+  database: "summarize_metadata"
+
+postgres:
```
yussufsh, line 7: Should not this param change to `postgres-summarize`? Else it will conflict with postgres in digitize.
Let's rename it in digitize values.yaml as well IMO.

yussufsh (changes requested):

ai-services/assets/services/digitize/podman/templates/postgres.yaml.tmpl
mayuka-c: Done

ai-services/assets/services/summarize/podman/values.yaml
mayuka-c: We dont have change it as discussed. We construct the Values map for each service and need be unique across services. But will confirm once by validating!

mayuka-c: Attaching the screenshot working fine
image

yussufsh: Thanks for verifying.

yussufsh:
Wondering why Makefile and version bump jobs were skipped. @adarshagrawal38 ?

yussufsh:
> Wondering why Makefile and version bump jobs were skipped. @adarshagrawal38 ?

Working now I guess.

yussufsh (approved):

adarshagrawal38:
@yussufsh we skip make bump check when PR is in draft stage
PTAL [here](https://github.com/IBM/project-ai-services/blob/main/.github/workflows/bump-makefile-version.yml#L21)

PR #870, opened 2026-06-04
feat(catalog-ui): Services Page

Niranjan-J-Joshi:
**Changes**

- Added services cards to display available services**
Screenshot 2026-06-04 at 1 17 01 PM

- Added dynamic button text to service cards based on the selected service
image

- Added a side details panel to view service-specific information
image
image

- Added a table to display deployed services
image


ui/catalog/public/data/services.json
mkumatag, line 1: I think services api is already there, please talk to @mayuka-c and get the real dump and work on a gap..

mkumatag: cc @yussufsh

mkumatag (changes requested):

ui/catalog/public/data/services.json
mayuka-c: The API definitions can be found in swagger: https://github.com/IBM/project-ai-services/blob/main/ai-services/docs/swagger.yaml

mayuka-c: We can start integrating those

yussufsh: Yes, they will use the APIs next. 
I had informed them about integrating APIs from now on, but this one was already implemented. Currently, I am setting up a backend server for the UI team to test.

ui/catalog/src/components/DeployedServicesTable/types.ts
```diff
@@ -0,0 +1,326 @@
...
+} as const;
+
+export type AppAction =
+  | { type: typeof ACTION_TYPES.SET_SEARCH; payload: string }
+  | { type: typeof ACTION_TYPES.SET_PAGE; payload: number }
+  | { type: typeof ACTION_TYPES.SET_PAGE_SIZE; payload: number }
+  | { type: typeof ACTION_TYPES.OPEN_DELETE_DIALOG; payload: string }
+  | { type: typeof ACTION_TYPES.CLOSE_DELETE_DIALOG }
+  | { type: typeof ACTION_TYPES.SET_CONFIRMED; payload: boolean }
+  | { type: typeof ACTION_TYPES.DELETE_ROW; payload: string }
```
mayuka-c: Dont we need to use keys with specific page prefix to make it unique (for example DeployedServices_).
Since we are storing in global state, wouldnt this collide with other places if the keys are same?

ui/catalog/src/components/DeployedServicesTable/DeployedServicesTable.tsx
```diff
@@ -0,0 +1,601 @@
...
+        payload: { message: "No service selected for deletion" },
+      });
+      return;
+    }
+
+    dispatch({ type: ACTION_TYPES.SET_IS_DELETING, payload: true });
+
+    try {
+      // Attempt server-side delete; if no backend exists this may fail.
+      const res = await fetch(`/api/applications/${state.selectedRowId}`, {
```
mayuka-c: This will be /api/v1/applications

ui/catalog/src/components/DeployedServicesTable/DeployedServicesTable.tsx
```diff
@@ -0,0 +1,601 @@
...
+          ? err.message
+          : "Failed deleting service deployment";
+      const name =
+        state.rowsData.find((r) => r.id === state.selectedRowId)?.name ?? "";
+      dispatch({
+        type: ACTION_TYPES.SHOW_ERROR,
+        payload: { message: msg, rowName: name },
+      });
+    } finally {
+      dispatch({ type: ACTION_TYPES.SET_IS_DELETING, payload: false });
```
mayuka-c: Wouldnt be better if we have action deleted (instead of relying on the payload with boolean) and if dispatched with that key, it goes and deletes the row in global store?

The payload is kind of thing used for updating the state variable and I feel we dont even need payload here

ui/catalog/public/data/services.json
mkumatag: My only worry is that I see a gap between this JSON vs. the services list backend team has implemented. If we don't match, there will be a lot of gaps and it won't be possible to fix them. I expect both of you to collaborate and settle on a specification, using the same content.

ui/catalog/src/components/DeployedServicesTable/DeployedServicesTable.tsx
```diff
@@ -0,0 +1,601 @@
...
+              onRequestClose={() => {
+                dispatch({ type: ACTION_TYPES.CLOSE_DELETE_DIALOG });
+              }}
+              onRequestSubmit={handleDelete}
+            >
+              <p>
+                Deleting an service deployment permanently deletes all
+                associated components, including connected services, runtime
+                metadata, and configurations will be permanently deleted, and it
+                cannot be undone.
```
ryanmarc:
```
Deleting a service deployment permanently deletes all associated components, including connected services, runtime metadata, and configurations. This action cannot be undone.
```

ui/catalog/src/pages/Services/Services.module.scss
ryanmarc, line 1: Please evaluate the use of `!important`. Are these explicit overrides really needed?

ui/catalog/src/components/ServiceDetailPanel/ServiceDetailPanel.tsx
```diff
@@ -0,0 +1,467 @@
...
+      className={styles.sidePanel}
+    >
+      <div className={styles.content}>
+        <p className={styles.description}>{service.description}</p>
+
+        <div className={styles.tagContainer}>
+          {service.tags?.map((tag, index) => (
+            <div key={index} className={styles.tag}>
+              {tag}
+            </div>
```
ryanmarc: This should be rendering a tag component, not creating a new visual style to look like a tag.

ryanmarc:
Just one comment right now. This change: https://github.com/IBM/project-ai-services/pull/870/changes#diff-e73d02e835d9b0a8174ff28240bca4abbeab23958266f171259aa785082eebadR15-R17

was this update from localhost an intentional change that should really be pushed into main?

Aside from this, I think what's here is ok to be merged.

To merge this, @mkumatag needs to update his review status from `changes requested`. Once that's done, and the question above is answered and/or addressed, this can be merged.

ui/catalog/public/data/services.json
Niranjan-J-Joshi: API integration for Services Cards, Deployed Services table  implemented , API integration services details coming in next PR

ui/catalog/src/components/DeployedServicesTable/DeployedServicesTable.tsx
Niranjan-J-Joshi: URL changed

ryanmarc (approved):

mkumatag (approved):
I will rely on others' approval and defer the changes required to remove the mock data and align the API schema until the follow-up PR, as per our agreement in the Slack thread.

ui/catalog/src/components/ServiceCard/ServiceCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+        secondaryButtonText={
+          isPartOfDigitalAssistants ? "Part of digital assistants" : "Deploy"
+        }
+        onPrimaryButtonClick={() => onLearnMore?.(id)}
+        onClick={() => onLearnMore?.(id)}
+        onSecondaryButtonClick={() => onDeploy?.(id)}
+        clickZone="two"
+        className={`${styles.productiveCard} ${
+          isPartOfDigitalAssistants ? styles.digitalAssistantCard : ""
+        }`}
```
mayuka-c, line 39: This should be done based on standalone field as discussed

ui/catalog/src/components/DeployedServicesTable/DeployedServicesTable.tsx
```diff
@@ -0,0 +1,785 @@
...
+            case "deleted":
+              // If deleted, we'll remove the row after refresh
+              return "Deleting...";
+            case "error":
+            case "failed":
+              return "Error";
+            default:
+              return "Deleting...";
+          }
+        };
```
mayuka-c, line 228: Any reason for maintaing UI own states here?
I see deleting and pending is "Deleting..". Please recheck

ryanmarc: this is consolidating and display format of different possible states from the backend into either "Deleting..." or "Error".

For example, if the UI was to show German language, we'd take the state from the backend and map it to the UI display.

"Pending" is ambiguous.  What is pending when that's the status on something? I think consolidating that in the UI display gives a better understanding to what is happening.

PR #873, opened 2026-06-04
fix: ignore ulimit check for root users

Shubhamag12:
- Ulimit values are set from `bootstrap` part. 
   - For non-root users, a re-login is required for the values to take effect. 
   - For root users, ulimit values are passed via Podman YAML annotations and are applied at runtime, so the re-login is not applicable and hence the check for ulimit values in bootstrap is made optional for root users.
- Rest of the refactoring is done as part of lint fix

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=0.0.44
+TAG?=0.0.45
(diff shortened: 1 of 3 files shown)


yussufsh (dismissed):
lgtm

yussufsh:
@Shubhamag12 please rebase and bump up the version again.

Shubhamag12:
@yussufsh done

yussufsh (approved):

PR #874, opened 2026-06-04
Conversational Support for Other languages

sats-23:
-Code re-org to support scalability for more languages support in future.
-German conversational mode 
-German Custom System Prompt
-Supporting UTs

### DIFF
--- a/services/chatbot/app.py
+++ b/services/chatbot/app.py
@@ -340,12 +340,25 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
     if not current_query or not current_query.strip():
         APIError.raise_error(ErrorCode.EMPTY_INPUT, "Query cannot be empty")
 
-    first_user_message = next(
-        (message.content for message in req.messages if message.role == "user"),
-        current_query,
-    )
-    session_lang = detect_language(first_user_message)
-    if session_lang not in {language_codes["English"], language_codes["German"]}:
+    # Detect language from current query (stateless - detect on every message)
+    try:
+        session_lang = detect_language(current_query)
+        
+        # Fallback to English if unsupported language detected
+        if session_lang not in language_codes.values():
+            logging.debug(
+                f"Unsupported language detected ({session_lang}). "
+                "Falling back to English."
+            )
+            session_lang = language_codes["English"]
+        
+        logging.debug(f"Detected language for current message: {session_lang}")
+        
+    except Exception as e:
+        logging.warning(
+            f"Language detection failed: {e}. "
+            "Falling back to English."
+        )
         session_lang = language_codes["English"]
 
     # Ensure vectorstore is initialized on first request
--- a/services/chatbot/app.py
+++ b/services/chatbot/app.py
@@ -18,8 +18,6 @@ from lingua import Language
 from common.misc_utils import set_log_level
 from common.lang_utils import setup_language_detector, detect_language, language_codes, get_max_tokens_map
 
-setup_language_detector([Language.ENGLISH, Language.GERMAN])
-
 from chatbot.settings import settings
 from chatbot.conversation_utils import get_conversation_context, truncate_history_by_tokens
 from chatbot.query_rephrasing import rephrase_query_with_context
@@ -95,6 +93,7 @@ async def lifespan(app):
     filtered_paths = ['/health']
     configure_uvicorn_logging(settings.common.app.log_level, filtered_paths)
     initialize_models()
+    setup_language_detector([Language.ENGLISH, Language.GERMAN])
     create_llm_session(pool_maxsize=settings.common.llm.max_batch_size)
     yield
     stderr_monitor.stop()
--- a/services/chatbot/app.py
+++ b/services/chatbot/app.py
@@ -380,7 +380,7 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
         max_tokens = req.max_tokens
         # giving priority to max_tokens passed in the request, otherwise according to session language
         if not max_tokens:
-            max_tokens = max_tokens_map.get(lang, settings.llm.max_tokens)
+            max_tokens = max_tokens_map.get(lang, settings.llm.english.max_tokens)
 
         rephrased_query = current_query
         
--- a/services/chatbot/app.py
+++ b/services/chatbot/app.py
@@ -24,7 +24,7 @@ set_log_level(settings.common.app.log_level)
 
 from common.diagnostic_logger import setup_comprehensive_crash_handler
 import common.db_utils as db
-from common.lang_utils import setup_language_detector, detect_language, lang_en, lang_de, max_tokens_map
+from common.lang_utils import setup_language_detector, detect_language, language_codes, max_tokens_map
 from common.misc_utils import get_model_endpoints, set_request_id, create_llm_session, configure_uvicorn_logging
 from common.llm_utils import query_vllm_stream, query_vllm_non_stream, query_vllm_models, tokenize_with_llm
 from common.perf_utils import perf_registry
@@ -345,8 +345,8 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
         current_query,
     )
     session_lang = detect_language(first_user_message)
-    if session_lang not in {lang_en, lang_de}:
-        session_lang = lang_en
+    if session_lang not in {language_codes["English"], language_codes["German"]}:
+        session_lang = language_codes["English"]
 
     # Ensure vectorstore is initialized on first request
     if vectorstore is None:
@@ -416,7 +416,7 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
         
         if not docs:
             message = "No documents found in the knowledge base for this query."
-            if lang == lang_de:
+            if lang == language_codes["German"]:
                 message = "Für diese Anfrage wurden keine Dokumente in der Wissensdatenbank gefunden."
             if req.stream:
                 async def stream_docs_not_found():
--- a/services/chatbot/app.py
+++ b/services/chatbot/app.py
@@ -24,7 +24,7 @@ set_log_level(settings.common.app.log_level)
 
 from common.diagnostic_logger import setup_comprehensive_crash_handler
 import common.db_utils as db
-from common.lang_utils import setup_language_detector, detect_language, lang_de, max_tokens_map
+from common.lang_utils import setup_language_detector, detect_language, lang_en, lang_de, max_tokens_map
 from common.misc_utils import get_model_endpoints, set_request_id, create_llm_session, configure_uvicorn_logging
 from common.llm_utils import query_vllm_stream, query_vllm_non_stream, query_vllm_models, tokenize_with_llm
 from common.perf_utils import perf_registry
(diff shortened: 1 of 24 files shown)


services/chatbot/app.py
```diff
@@ -340,6 +346,27 @@ async def stream_auth_error():
     if not current_query or not current_query.strip():
         APIError.raise_error(ErrorCode.EMPTY_INPUT, "Query cannot be empty")
 
+    # Detect language from current query (stateless - detect on every message)
+    try:
+        session_lang = detect_language(current_query)
```
dharaneeshvrd: I think its better to call it as query_lang since there is no session exists.

services/chatbot/app.py
```diff
@@ -367,33 +394,33 @@ async def stream_query_length_error():
                 return StreamingResponse(stream_query_length_error(), media_type="text/event-stream")
             APIError.raise_error(ErrorCode.INVALID_PARAMETER, error_msg)
 
-        lang = detect_language(current_query)
-
         max_tokens = req.max_tokens
-        # giving priority to max_tokens passed in the request, otherwise according to detected language of query
+        # giving priority to max_tokens passed in the request, otherwise according to session language
```
dharaneeshvrd: Here too, earlier comment itself makes sense to me.

services/chatbot/app.py
```diff
@@ -367,33 +394,33 @@ async def stream_query_length_error():
...
         if not max_tokens:
-            max_tokens = max_tokens_map.get(lang, settings.llm.max_tokens)
+            if session_lang == language_codes["German"]:
+                max_tokens = settings.llm.german.max_tokens
```
dharaneeshvrd: I see max_tokens_map containes German too, why are we directly accessing it?

services/chatbot/prompt_validator.py
dharaneeshvrd, line 1: Did nt we discuss to add support for custom prompt for other languages later? :(
I m just worried since it would add more complication. But I see you have made the code changes anyway.
How easy is it to extend this to other languages?

services/common/settings.py
```diff
@@ -13,6 +13,42 @@
 class LLMConfig(BaseSettings):
     """LLM model configuration."""
     
+    class EnglishConfig(BaseSettings):
```
dharaneeshvrd: I think this can be removed, only chatbot's settings can be used.

services/common/llm_utils.py
```diff
@@ -182,117 +181,89 @@ def query_vllm_payload(
...
-            lang,
-            {
-                lang_en: chatbot_settings.chatbot.legacy_query_vllm_stream_en_prompt,
-                lang_de: chatbot_settings.chatbot.query_vllm_stream_de_prompt
-            }
+
+    logger.debug(f"Truncated Context: {context}")
+
+    system_prompt = (
+        chatbot_settings.chatbot.german.system_prompt
```
dharaneeshvrd: Let's use switch/case here to extend other languages neatly.

services/common/llm_utils.py
```diff
@@ -182,117 +181,89 @@ def query_vllm_payload(
...
+    llm_max_model_len = resolve_model_max_len(
+        llm_endpoint,
+        llm_model,
+        settings.llm.max_model_len,
+        api_key,
+    )
+
+    # Calculate budget for context first (prioritize context over history)
+    budget_for_context = llm_max_model_len - (
+        chatbot_settings.chatbot.initial_system_token_overhead +
```
dharaneeshvrd, line 197: Did we visit these vars when we introduced these new prompts?
If not let's check once they are within the limit used on these settings.

services/common/llm_utils.py
sats-23: Done

services/common/llm_utils.py
sats-23: I'll do one round of testing using debug logs. That should give us better idea

services/chatbot/prompt_validator.py
sats-23: Should be do-able, we just need to validate the semantic validation and prompt injection validation. 
No much difference in core logic

services/chatbot/app.py
sats-23: Mistake, fixed it

services/chatbot/app.py
sats-23: Updated it!

services/chatbot/app.py
sats-23: Handled this as well

services/common/settings.py
sats-23: Done, removed

services/chatbot/prompt_validator.py
dharaneeshvrd, line 1: Let's use switch case whereever if else is used for lang selection

services/chatbot/query_rephrasing.py
```diff
@@ -63,39 +63,61 @@ def calculate_dynamic_max_response_tokens(
...
-    Will be updated when integrating with UI
+        'User: What is Spyre?\\nAssistant: Spyre is an AI accelerator...'
+        >>> format_messages_for_rephrasing(messages, lang=language_codes["German"])
+        'Benutzer: What is Spyre?\\nAssistent: Spyre is an AI accelerator...'
     """
     if not messages:
         return ""
-    
+
+    role_labels = {
```
dharaneeshvrd: Can be integrated into the language specific settings and referred here. 
So that all these kind of configs can be accessed from a single place.

services/chatbot/query_rephrasing.py
```diff
@@ -138,7 +160,7 @@ def call_llm_for_rephrasing(
         ],
         "temperature": temperature,
         "max_tokens": max_tokens,
-        "stop": ["\n\n", "Question:", "Current Question:"],
+        "stop": ["\n\n", "Question:", "Current Question:", "Frage:", "Aktuelle Frage:"],
```
dharaneeshvrd: Stop words too

services/chatbot/app.py
```diff
@@ -49,6 +49,12 @@
 llm_model_dict = {}
 reranker_model_dict = {}
 
+# Language-specific messages
+NO_DOCUMENTS_FOUND_MESSAGES = {
+    "EN": "No documents found in the knowledge base for this query.",
+    "DE": "Für diese Anfrage wurden keine Dokumente in der Wissensdatenbank gefunden."
```
manalilatkar, line 56: If we are adding conversation support for all languages, we should include no_documents_found messages also in ITalian and French, not just English and German.

dharaneeshvrd (commented):
Let's validate [this](https://github.com/IBM/project-ai-services/pull/874/changes#r3371551916) once.
Remaining looks good to me
Can be merged post addressing small nits.

services/chatbot/prompt_validator.py
```diff
@@ -11,62 +11,149 @@
...
-    "VERDICT: SAFE\n"
-    "REASON: No injection patterns detected, prompt contains standard conversational instructions.\n"
-    "CONFIDENCE: 0.92\n\n"
-    "Example unsafe response:\n"
-    "VERDICT: UNSAFE\n"
-    "REASON: Contains role manipulation attempt with \"ignore previous instructions\" pattern.\n"
-    "CONFIDENCE: 0.95"
-)
+
+class GermanConstants:
```
manalilatkar, line 89: Do we really need separate  German and other language constants for this purpose. Here, the purpose is to analyse the content and get a specific verdict. That verdict can be in English also; the generated content doesn't need to be in German or Italian, so unless we came up with some numbers to justify a separate German prompt for this, I don't see the point for this.
@dharaneeshvrd WDYT?

services/chatbot/settings.py
```diff
@@ -214,36 +283,61 @@ class RAGConfig(BaseSettings):
...
+        2. Validate the prompt (if validation enabled)
+        3. Override the appropriate language config only if validation passes
+        4. Fallback to English on any failure
+        """
+        # Handle custom system_prompt
+        if self.system_prompt:
+            try:
+                from common.lang_utils import detect_language, language_codes
+                
+                detected_lang = detect_language(self.system_prompt, min_confidence=0.7)
```
manalilatkar, line 346: On what basis did you come up with this 0.7 number?

services/chatbot/prompt_validator.py
sats-23: A thought went around for keeping all system and query prompts in English only and to add an additional instruction to "answer in the language of user's query". 
This is a feasible solution in terms of a huger model (>50B params), but the same hasn't performed very well in terms of smaller models, especially considering our roadmap around Granite 3.3 8B, 4H Small (32B) and Mistral 24B. 
So the recommended approach was to keep everything in the native language to prevent language hallucinations, hence the strict approach.

services/chatbot/settings.py
sats-23: Extract of [existing](https://github.com/IBM/project-ai-services/blob/8b5d53b6113c5a0475633eee58769aa1c2e245bb/services/chatbot/settings.py#L311) logic

services/chatbot/app.py
sats-23: The PR was more of an approach to setup base work (make code scalable) for other languages to be added easily

services/chatbot/query_rephrasing.py
sats-23: Done

services/chatbot/query_rephrasing.py
sats-23: Done

services/chatbot/prompt_validator.py
dharaneeshvrd: We are following this language specific prompts in different parts of our services, I don't see a point in not using it here. If it produces better results, let's use it.

services/chatbot/prompt_validator.py
sats-23: I shall go with a map approach. 
Using match case doesn't optimize access time as we can't access dictionary values directly in the case statement. 
We would have to use the if Guard, which would not optimize our access time (still goes through all cases). 

```
match language:
            case _ if language == language_codes["German"]:
                keywords = GermanConstants.RESPONSE_KEYWORDS
            case _:
                keywords = EnglishConstants.RESPONSE_KEYWORDS
```

sats-23: Done using Map

sats-23:
@dharaneeshvrd PTAL at recent changes.
Both languages have been validated against cases:
-Default prompt
-Custom prompt
-Prompt Injection 
-Semantically Invalid Prompt

Supporting media is attached [here](https://jsw.ibm.com/browse/AISERVICES-1139)

services/common/llm_utils.py
dharaneeshvrd: @sats-23 what is the result of this?

services/chatbot/app.py
```diff
@@ -340,6 +347,27 @@ async def stream_auth_error():
     if not current_query or not current_query.strip():
         APIError.raise_error(ErrorCode.EMPTY_INPUT, "Query cannot be empty")
 
+    # Detect language from current query (stateless - detect on every message)
+    try:
+        query_lang = detect_language(current_query)
+        
+        # Fallback to English if unsupported language detected
+        if query_lang not in language_codes.values():
```
dharaneeshvrd, line 355: Seems query_lang returned would be in en, de, it & fr format and small letters, but language codes values returns capital codes.
Am I missing something here?

dharaneeshvrd: Never mind, it returns in capital letters only.

dharaneeshvrd (dismissed):

services/common/llm_utils.py
sats-23: Tested it out, the truncation happens fine and contextual meaning is maintained in the flow. 
Edge cases if any must be easy and minor to handle

dharaneeshvrd: Ok thanks.

dharaneeshvrd (approved):

PR #875, opened 2026-06-04
Make version as mandatory else the GET applications APIs will fail

yussufsh:
There is minor issue (since validation is not in place) if version is not provided for each service and architecture deploy params, the GET APIs are failing.
Making it required in the POST API for deployment and also in A/S/C struct while reading.


mayuka-c (dismissed):
LGTM

mayuka-c (dismissed):
LGTM

mayuka-c (approved):

PR #880, opened 2026-06-05
Fix table merging logic for docx files

Shilpa-Gokul:


manalilatkar:
LGTM but needs version bump and rebase.

Shilpa-Gokul:
Thank you @manalilatkar. I have rebased and bumped the version, could you PTAL?
@Niharika0306

Niharika0306:
changes LGTM.  
I was thinking if we should add any additional check in case of docx files , to check for matching parent headers or so ( along with the current logic of matching column headers). But given the complexity and the fact that we didn't find a failure case till now - the current logic should suffice.

Niharika0306 (dismissed):

manalilatkar (dismissed):

Niharika0306:
@Shilpa-Gokul can you update the makefile as well?

dharaneeshvrd (commented):
Please update the version in Makefile present under services/digitze

dharaneeshvrd (changes requested):
Need to bump ai-services version too
https://github.com/IBM/project-ai-services/blob/ff397d288bf0961eaee6b6252a8a115d9cd6176a/ai-services/Makefile#L3

Niharika0306 (approved):

services/digitize/doc_utils.py
```diff
@@ -610,6 +610,11 @@ def process_table(converted_doc, pdf_path, out_path, gen_model, gen_endpoint):
         out_path.write_text(json.dumps({}, indent=2), encoding="utf-8")
         return table_count, process_time
 
+    # Determine if this is a DOCX file
```
dharaneeshvrd, line 613: Ideally this should be done earlier in the pipeline and just referred here. 
@Niharika0306 please take care of this in your changes.

dharaneeshvrd (approved):

services/digitize/doc_utils.py
dharaneeshvrd: @Shilpa-Gokul can you please take this up?
Currently the methods are tightly coupled to work with pdf?
Like the `pdf_path` arg of `process_text` method.
It would be great if we clean the code also a bit by moving the methods defined for table caption extraction to a different util like docx_utils.py.

Niharika0306: Agree. I am introducing a docx_utils.py file as part of my PR. So we can then move some of these table_docx functions to there.

Shilpa-Gokul: Sure @dharaneeshvrd Will create a follow-up PR for the above

PR #882, opened 2026-06-07
Integrate Catalog API for Application PS Command

adarshagrawal38:
# Summary
This PR integrates the Catalog API to retrieve and display application process status information for the application ps command. The new functionality is hidden behind an `--experimental` flag to allow gradual rollout and testing.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/ps.go
+++ b/ai-services/cmd/ai-services/cmd/application/ps.go
@@ -3,16 +3,26 @@ package application
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/project-ai-services/ai-services/internal/pkg/application"
 	appTypes "github.com/project-ai-services/ai-services/internal/pkg/application/types"
+	catalogClient "github.com/project-ai-services/ai-services/internal/pkg/catalog/client"
+	catalogConstants "github.com/project-ai-services/ai-services/internal/pkg/catalog/constants"
+	catalogTypes "github.com/project-ai-services/ai-services/internal/pkg/catalog/types"
 	appFlags "github.com/project-ai-services/ai-services/internal/pkg/cli/constants/application"
 	"github.com/project-ai-services/ai-services/internal/pkg/cli/flagvalidator"
+	"github.com/project-ai-services/ai-services/internal/pkg/constants"
+	"github.com/project-ai-services/ai-services/internal/pkg/runtime/types"
+	"github.com/project-ai-services/ai-services/internal/pkg/utils"
 	"github.com/project-ai-services/ai-services/internal/pkg/vars"
 	"github.com/spf13/cobra"
 )
 
-var output string
+var (
+	output         string
+	experimentalPs bool
+)
 
 func isOutputWide() bool {
 	return strings.ToLower(output) == "wide"
@@ -44,6 +54,12 @@ Arguments
 
 		rt := vars.RuntimeFactory.GetRuntimeType()
 
+		// When experimentalTemplates is true and runtime is podman, use experimental catalog ps api
+		// For openshift runtime, always use the older/stable code path regardless of experimental flag
+		if experimentalPs && rt == types.RuntimeTypePodman {
+			return processApplication(applicationName)
+		}
+
 		// Create application instance using factory
 		factory := application.NewFactory(rt)
 		app, err := factory.Create(applicationName)
@@ -70,6 +86,13 @@ func init() {
 }
 
 func initPsCommonFlags() {
+	psCmd.Flags().BoolVar(
+		&experimentalPs,
+		"experimental",
+		false,
+		"Include experimental application templates",
+	)
+
 	psCmd.Flags().StringVarP(
 		&output,
 		appFlags.Ps.Output,
@@ -90,4 +113,186 @@ func buildPsFlagValidator() *flagvalidator.FlagValidator {
 		AddCommonFlag(appFlags.Ps.Output, nil)
 
 	return builder.Build()
+}
+
+func processApplication(appName string) error {
+	// Read base URL from environment variable with fallback
+	baseURL := utils.GetEnv("CATALOG_API_BASE_URL", "http://10.48.64.151:8080")
+
+	token, err := getAccessToken()
+	if err != nil {
+		return err
+	}
+	// Create application client with server URL and token
+	appClient := catalogClient.NewApplicationClient(baseURL, token)
+
+	appIds, err := getAppIds(appClient, appName, token)
+	if err != nil {
+		return err
+	}
+	if len(appIds) == 0 {
+		return fmt.Errorf("no application found with name %s", appName)
+	}
+
+	return render(appClient, appIds, token)
+}
+
+// getAccessToken retrieves the access token from the stored credentials.
+// It uses the catalog client to load credentials from the config file.
+func getAccessToken() (string, error) {
+	// Create a new client which loads credentials from config
+	client, err := catalogClient.NewWithLogin("http://10.48.64.151:8080", "admin", "Admin@123")
+	if err != nil {
+		return "", fmt.Errorf("failed to load credentials: %w", err)
+	}
+
+	// Return the access token
+	return client.AccessToken(), nil
+}
+
+// getAppIds retrieves application ID(s) from the catalog API.
+// If appName is empty, returns all application IDs.
+// If appName is provided, returns the ID of the matching application.
+// The base URL is read from the CATALOG_API_BASE_URL environment variable,
+// with a fallback to http://10.9.7.151:8080 if not set.
+func getAppIds(appClient *catalogClient.ApplicationClient, appName string, token string) ([]string, error) {
+	// List all applications
+	resp, err := appClient.ListApplications(nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch applications: %w", err)
+	}
+
+	// If appName is empty, return all IDs
+	if appName == "" {
+		ids := make([]string, 0, len(resp.Data))
+		for _, app := range resp.Data {
+			ids = append(ids, app.ID)
+		}
+
+		return ids, nil
+	}
+
+	// If appName is provided, find matching application and return its ID
+	for _, app := range resp.Data {
+		if app.Name == appName {
+			return []string{app.ID}, nil
+		}
+	}
+
+	// No matching application found
+	return nil, fmt.Errorf("application with name '%s' not found", appName)
+}
+
+// render retrieves and processes the PS information for multiple application IDs.
+// It fetches the process status for each application using the catalog API and prints the results in tabular format.
+func render(appClient *catalogClient.ApplicationClient, appIds []string, token string) error {
+	// Create table writer
+	printer := utils.NewTableWriter()
+	defer printer.CloseTableWriter()
+
+	// Set table headers based on output format
+	outputWide := isOutputWide()
+	setTableHeaders(printer, outputWide)
+
+	// Process each application ID
+	for _, appID := range appIds {
+		// Get PS information for the application
+		psResp, err := appClient.GetApplicationPS(appID)
+		if err != nil {
+			fmt.Printf("Error fetching PS for application %s: %v\n", appID, err)
+
+			continue
+		}
+
+		// Process services pods
+		for _, pod := range psResp.Services {
+			rows := buildPodRowFromAPI(psResp.Name, pod, outputWide)
+			printer.AppendRow(rows...)
+		}
+
+		// Process components pods
+		for _, pod := range psResp.Components {
+			rows := buildPodRowFromAPI(psResp.Name, pod, outputWide)
+			printer.AppendRow(rows...)
+		}
+	}
+
+	return nil
+}
+
+// setTableHeaders sets the table headers based on output format.
+func setTableHeaders(printer *utils.Printer, outputWide bool) {
+	if outputWide {
+		printer.SetHeaders("APPLICATION NAME", "POD ID", "POD NAME", "STATUS", "CREATED", "CONTAINERS")
+	} else {
+		printer.SetHeaders("APPLICATION NAME", "POD NAME", "STATUS")
+	}
+}
+
+// buildPodRowFromAPI builds a table row from API response data.
+func buildPodRowFromAPI(appName string, pod catalogTypes.Pod, wideOutput bool) []string {
+	status := getPodStatusFromAPI(pod)
+
+	// If wide option flag is not set, return appName, podName and status only
+	if !wideOutput {
+		return []string{appName, pod.PodName, status}
+	}
+
+	containerNames := getContainerNamesFromAPI(pod)
+
+	// Parse the Created string and convert to TimeAgo format
+	created := "N/A"
+	if pod.Created != "" {
+		// Try to parse the Created timestamp
+		parsedTime, err := time.Parse(catalogConstants.RFC3339WithTimezone, pod.Created)
+		if err == nil {
+			created = utils.TimeAgo(parsedTime)
+		} else {
+			// If parsing fails, use the original string
+			created = pod.Created
+		}
+	}
+
+	return []string{
+		appName,
+		pod.PodID[:12],
+		pod.PodName,
+		status,
+		created,
+		strings.Join(containerNames, ", "),
+	}
+}
+
+// getPodStatusFromAPI determines the pod status from API response.
+func getPodStatusFromAPI(pod catalogTypes.Pod) string {
+	status := string(pod.Status)
+
+	// If the pod is running, check if it's healthy
+	if strings.ToLower(status) == "running" {
+		if pod.Healthy {
+			status += fmt.Sprintf(" (%s)", constants.Ready)
+		} else {
+			status += fmt.Sprintf(" (%s)", constants.NotReady)
+		}
+	}
+
+	return status
+}
+
+// getContainerNamesFromAPI extracts container names with their status from API response.
+func getContainerNamesFromAPI(pod catalogTypes.Pod) []string {
+	if len(pod.Containers) == 0 {
+		return []string{"none"}
+	}
+
+	containerNames := make([]string, 0, len(pod.Containers))
+	for _, container := range pod.Containers {
+		health := constants.NotReady
+		if container.Healthy {
+			health = constants.Ready
+		}
+		containerNames = append(containerNames, fmt.Sprintf("%s (%s)", container.Name, health))
+	}
+
+	return containerNames
 }
--- a/ai-services/cmd/ai-services/cmd/application/ps.go
+++ b/ai-services/cmd/ai-services/cmd/application/ps.go
@@ -8,9 +8,9 @@ import (
 	appTypes "github.com/project-ai-services/ai-services/internal/pkg/application/types"
 	appFlags "github.com/project-ai-services/ai-services/internal/pkg/cli/constants/application"
 	"github.com/project-ai-services/ai-services/internal/pkg/cli/flagvalidator"
+	cliUtils "github.com/project-ai-services/ai-services/internal/pkg/cli/utils"
 	"github.com/project-ai-services/ai-services/internal/pkg/runtime/types"
 	"github.com/project-ai-services/ai-services/internal/pkg/vars"
-	cliUtils "github.com/project-ai-services/ai-services/internal/pkg/cli/utils"
 	"github.com/spf13/cobra"
 )
 
--- a/ai-services/cmd/ai-services/cmd/application/ps.go
+++ b/ai-services/cmd/ai-services/cmd/application/ps.go
@@ -119,6 +119,9 @@ func buildPsFlagValidator() *flagvalidator.FlagValidator {
 func processApplication(appName string) error {
 	// Create application client with server URL and token
 	appClient, err := catalogClient.NewApplicationClient()
+	if err != nil {
+		return fmt.Errorf("failed to create application client: %v", err)
+	}
 
 	appIDList, err := getAppIDList(appClient, appName)
 	if err != nil {
--- a/ai-services/cmd/ai-services/cmd/application/ps.go
+++ b/ai-services/cmd/ai-services/cmd/application/ps.go
@@ -3,20 +3,14 @@ package application
 import (
 	"fmt"
 	"strings"
-	"time"
 
 	"github.com/project-ai-services/ai-services/internal/pkg/application"
 	appTypes "github.com/project-ai-services/ai-services/internal/pkg/application/types"
-	catalogClient "github.com/project-ai-services/ai-services/internal/pkg/catalog/client"
-	catalogConstants "github.com/project-ai-services/ai-services/internal/pkg/catalog/constants"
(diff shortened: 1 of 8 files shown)


adarshagrawal38:
Sample Response:
Screenshot 2026-06-08 at 2 01 03 AM

ai-services/cmd/ai-services/cmd/application/ps.go
```diff
@@ -91,3 +114,179 @@ func buildPsFlagValidator() *flagvalidator.FlagValidator {
 
 	return builder.Build()
 }
+
+func processApplication(appName string) error {
+	// Read base URL from environment variable with fallback
+	baseURL := utils.GetEnv("CATALOG_API_BASE_URL", "")
```
mayuka-c: The baseURL should be present in client, err := catalogClient.New(), we need to read from there `ServerURL`.

Refer whoami command for example

ai-services/cmd/ai-services/cmd/application/ps.go
```diff
@@ -44,6 +54,12 @@ Arguments
 
 		rt := vars.RuntimeFactory.GetRuntimeType()
 
+		// When experimentalTemplates is true and runtime is podman, use experimental catalog ps api
+		// For openshift runtime, always use the older/stable code path regardless of experimental flag
```
Shubhamag12, line 60: one question here:
If the idea is to keep only openshift working same as before, shouldn't we change `/internal/pkg/application/podman/ps.go`?

@yussufsh @mayuka-c Any thoughts?

ai-services/cmd/ai-services/cmd/application/ps.go
```diff
@@ -91,3 +114,179 @@ func buildPsFlagValidator() *flagvalidator.FlagValidator {
...
+	// Set table headers based on output format
+	outputWide := isOutputWide()
+	setTableHeaders(printer, outputWide)
+
+	// Process each application ID
+	for _, appID := range appIds {
+		// Get PS information for the application
+		psResp, err := appClient.GetApplicationPS(appID)
+		if err != nil {
+			fmt.Printf("Error fetching PS for application %s: %v\n", appID, err)
```
mayuka-c: Make use of logger instead of fmt

ai-services/internal/pkg/catalog/client/application.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	var resp types.ApplicationListResponse
+	err := c.httpClient.Do(httpclient.Request{
+		Method:   http.MethodGet,
+		Endpoint: "/api/v1/applications",
+		Headers:  map[string]string{"Authorization": "Bearer " + c.token},
+		Query:    query,
+		Out:      &resp,
+	})
```
mayuka-c: Make use of resty package instead of http client

ai-services/internal/pkg/catalog/client/application.go
```diff
@@ -0,0 +1,102 @@
...
+// GetApplicationPS retrieves the process status and runtime information for an application.
+// It returns details about pods, containers, and their health status.
+func (c *ApplicationClient) GetApplicationPS(id string) (*types.ApplicationPSResponse, error) {
+	var psResp types.ApplicationPSResponse
+	err := c.httpClient.Do(httpclient.Request{
+		Method:   http.MethodGet,
+		Endpoint: fmt.Sprintf("/api/v1/applications/%s/ps", id),
+		Headers:  map[string]string{"Authorization": "Bearer " + c.token},
+		Out:      &psResp,
+	})
```
mayuka-c: Also here, please use resty

ai-services/cmd/ai-services/cmd/application/ps.go
mayuka-c: Yes we would need to update it there

ai-services/cmd/ai-services/cmd/application/ps.go
mayuka-c: Reference: https://github.com/IBM/project-ai-services/blob/main/ai-services/internal/pkg/catalog/client/client.go#L48

ai-services/cmd/ai-services/cmd/application/ps.go
adarshagrawal38: are we not going to remove `application` pkg in long term?

ai-services/cmd/ai-services/cmd/application/ps.go
adarshagrawal38: updated

ai-services/cmd/ai-services/cmd/application/ps.go
Shubhamag12: why do we want to remove it though? 🤔

ai-services/internal/pkg/catalog/client/application.go
```diff
@@ -0,0 +1,100 @@
...
+
+// NewApplicationClient creates a new ApplicationClient with the given server URL and token.
+func NewApplicationClient(serverURL string) *ApplicationClient {
+	return &ApplicationClient{
+		serverURL:  serverURL,
+		httpClient: httpclient.New(serverURL),
+	}
+}
+
+// ListApplicationsParams holds optional query parameters for listing applications.
```
Shubhamag12: can we separate out list application to a different file? because as we migrate more cmds, this file would keep on insreasing

ai-services/internal/pkg/catalog/client/application.go
mkumatag: @mayuka-c or do you think writing some client package will be helpful for any consumers? may be part of follow up PR or later..

mayuka-c: Correct true.
We can have all application Catalog CRUD calls in some client package and utilize it.

ai-services/cmd/ai-services/cmd/application/ps.go
```diff
@@ -91,3 +115,165 @@ func buildPsFlagValidator() *flagvalidator.FlagValidator {
...
+	appIds, err := getAppIds(appClient, appName, client.AccessToken())
+	if err != nil {
+		return err
+	}
+	if len(appIds) == 0 {
+		return fmt.Errorf("no application found with name %s", appName)
+	}
+
+	return render(appClient, appIds, client.AccessToken())
+}
```
mayuka-c: Please move this to application utils.
I see all the CLI commands can make use of it

ai-services/cmd/ai-services/cmd/application/ps.go
```diff
@@ -91,3 +115,165 @@ func buildPsFlagValidator() *flagvalidator.FlagValidator {
...
+	// If appName is provided, find matching application and return its ID
+	for _, app := range resp.Data {
+		if app.Name == appName {
+			return []string{app.ID}, nil
+		}
+	}
+
+	// No matching application found
+	return nil, fmt.Errorf("application with name '%s' not found", appName)
+}
```
mayuka-c: This too will be utilized by other CLI commands, we can reuse

ai-services/cmd/ai-services/cmd/application/ps.go
adarshagrawal38: done

ai-services/internal/pkg/catalog/client/application.go
adarshagrawal38: using resty now

ai-services/cmd/ai-services/cmd/application/ps.go
adarshagrawal38: I was thinking as we will integrate `API` calls for `cli` cmd, so we then stop using application.

ai-services/internal/pkg/catalog/client/application.go
```diff
@@ -0,0 +1,101 @@
...
+	if resp.IsError() {
+		return nil, fmt.Errorf("list applications: server returned HTTP %d", resp.StatusCode())
+	}
+
+	return &result, nil
+}
+
+// GetApplicationPS retrieves the process status and runtime information for an application.
+// It returns details about pods, containers, and their health status.
+func (c *ApplicationClient) GetApplicationPS(id string, token string) (*types.ApplicationPSResponse, error) {
```
mkumatag: ideally token should be set at the package object level while creating it or may be via some function, sending token in every call doesn't make sense..

adarshagrawal38: okay will set token at object level

mayuka-c: @adarshagrawal38 You can actually set the catalogClient object itself too within this Application Client struct.
catalogClient has accessToken and baseServerURL

ai-services/cmd/ai-services/cmd/application/ps.go
mayuka-c: application layer would still be there :)
So flow will be cmd -> internal/pkg/application (runtime based calls respective runtimeType) 
Acutally we need to have all business logic there inside application

ai-services/internal/pkg/catalog/client/application.go
adarshagrawal38: are you suggesting to have one file for each API call?

ai-services/cmd/ai-services/cmd/application/ps.go
adarshagrawal38: where is the application utils?
Are you talking about [application/common](https://github.com/IBM/project-ai-services/tree/main/ai-services/internal/pkg/application/common) or [utils.go](https://github.com/IBM/project-ai-services/blob/main/ai-services/internal/pkg/utils/util.go)

ai-services/internal/pkg/catalog/client/application.go
Shubhamag12: yes, because delete, create, and others may/may not add contents to this files based on implementation. 
wydt?

ai-services/cmd/ai-services/cmd/application/ps.go
Shubhamag12: as discussed, we are not moving until we have migrated everything.
Closing this comment @adarshagrawal38

ai-services/internal/pkg/catalog/client/application.go
adarshagrawal38: here we will only sending API call and parsing the response,
So I think let's keep all functions one file, as I don't see file growing large.

Are you saying we may not have delete/create api call in this client?

ai-services/cmd/ai-services/cmd/application/ps.go
mayuka-c: I feel let us have CLI utils here
The other utils `application_utils` is for the Catalog Backend server

ai-services/cmd/ai-services/cmd/application/ps.go
```diff
@@ -91,3 +115,162 @@ func buildPsFlagValidator() *flagvalidator.FlagValidator {
...
+
+	return render(appClient, appIDList)
+}
+
+// getAppIDList retrieves application ID(s) from the catalog API.
+// If appName is empty, returns all application IDs.
+// If appName is provided, returns the ID of the matching application.
+func getAppIDList(appClient *catalogClient.ApplicationClient, appName string) ([]string, error) {
+	// List all applications
+	resp, err := appClient.ListApplications(nil)
```
yussufsh: IMO we should have two funcs

getAllApps() ([]Application, error)- when appName is empty. Calls /applications [get]
getAppByName() (Application, error) - List and search the list for app.Name.
@Shubhamag12 all commands can use the same getAppByName() func.

ai-services/internal/pkg/catalog/client/application.go
Shubhamag12: > Are you saying we may not have delete/create api call in this client? 

no, I am saying create will definitely add contents to this file. Others, may or may not add to this file (depends on implementation).

We can take it up when we implement other migrations. No issues

ai-services/cmd/ai-services/cmd/application/ps.go
adarshagrawal38: Introduced this methods in utils.

ai-services/internal/pkg/catalog/client/application.go
adarshagrawal38: done

ai-services/cmd/ai-services/cmd/application/ps.go
adarshagrawal38: moved to utils

ai-services/cmd/ai-services/cmd/application/ps.go
adarshagrawal38: moved to utils

ai-services/internal/pkg/cli/utils/common.go
```diff
@@ -0,0 +1,163 @@
...
+	setTableHeaders(printer, outputWide)
+
+	// Process each application ID
+	for _, app := range applicationList {
+		// Get PS information for the application
+		psResp, err := appClient.GetApplicationPS(app.ID)
+		if err != nil {
+			logger.Errorf("Error fetching PS for application %s: %v\n", app.ID, err)
+
+			continue
```
yussufsh: Why are not throwing error if API fails?

ai-services/internal/pkg/catalog/client/application.go
```diff
@@ -0,0 +1,106 @@
...
+type ApplicationClient struct {
+	httpClient *resty.Client
+	client     *Client
+}
+
+// NewApplicationClient creates a new ApplicationClient with the given server URL and token.
+func NewApplicationClient() (*ApplicationClient, error) {
+	client, err := New()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load credentials: %v", err)
```
yussufsh:
```suggestion
		return nil, fmt.Errorf("failed to initialize client: %w", err)
```

ai-services/internal/pkg/catalog/client/application.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+
+	resp, err := req.Get("/api/v1/applications")
+	if err != nil {
+		return nil, fmt.Errorf("list applications: %w", err)
+	}
+
+	if resp.IsError() {
+		return nil, fmt.Errorf("list applications: server returned HTTP %d", resp.StatusCode())
```
yussufsh: With `statusCode` we should also read the error response body. This is more of a generic failure message.

ai-services/internal/pkg/catalog/client/application.go
```diff
@@ -0,0 +1,106 @@
...
+	resp, err := c.httpClient.R().
+		SetHeader("Authorization", "Bearer "+c.client.AccessToken()).
+		SetResult(&result).
+		Get(fmt.Sprintf("/api/v1/applications/%s/ps", id))
+	if err != nil {
+		return nil, fmt.Errorf("get application ps: %w", err)
+	}
+
+	if resp.IsError() {
+		return nil, fmt.Errorf("get application ps: server returned HTTP %d", resp.StatusCode())
```
yussufsh: Same here.

ai-services/internal/pkg/cli/utils/application_utils.go
```diff
@@ -0,0 +1,32 @@
...
+	listResponse, err := appClient.ListApplications(nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch applications: %w", err)
+	}
+
+	return listResponse.Data, nil
+}
+
+func GetAppByName(appClient *catalogClient.ApplicationClient, appName string) (*types.Application, error) {
+	listResponse, err := appClient.ListApplications(nil)
```
yussufsh, line 21: @mayuka-c in the future, let's have a filter based on name for the List API.
Eg:
```
listResponse, err := appClient.ListApplications(&catalogClient.ListApplicationsParams{
        Name: appName,  // ← Filter on server side
    })
```

ai-services/internal/pkg/cli/utils/common.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch application '%s': %w", appName, err)
+	}
+
+	return []catalogTypes.Application{*application}, nil
+}
+
+// render retrieves and processes the PS information for multiple application IDs.
+// It fetches the process status for each application using the catalog API and prints the results in tabular format.
+func render(appClient *catalogClient.ApplicationClient, outputWide bool, applicationList []catalogTypes.Application) error {
```
yussufsh:
```suggestion
func renderApplicationPS(appClient *catalogClient.ApplicationClient, outputWide bool, applicationList []catalogTypes.Application) error {
```
`render` name is too generic.

ai-services/internal/pkg/cli/utils/common.go
```diff
@@ -0,0 +1,163 @@
...
+			rows := buildPodRowFromAPI(psResp.Name, pod, outputWide)
+			printer.AppendRow(rows...)
+		}
+	}
+
+	return nil
+}
+
+// setTableHeaders sets the table headers based on output format.
+func setTableHeaders(printer *utils.Printer, outputWide bool) {
```
yussufsh:
```suggestion
func setApplicationPSTableHeaders(printer *utils.Printer, outputWide bool) {
```

ai-services/internal/pkg/cli/utils/common.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create application client: %w", err)
+	}
+
+	applicationList, err := fetchApplications(appClient, opts.ApplicationName)
+	if err != nil {
+		return err
+	}
+
+	return render(appClient, opts.OutputWide, applicationList)
```
yussufsh: Please check for empty applicationList.

yussufsh (changes requested):

ai-services/internal/pkg/cli/utils/common.go
adarshagrawal38: added empty check

ai-services/internal/pkg/cli/utils/common.go
adarshagrawal38: Updated

ai-services/internal/pkg/cli/utils/common.go
adarshagrawal38: updated

ai-services/internal/pkg/cli/utils/common.go
adarshagrawal38: If it fails for one app, then I was attempting to process next app in this list.
Which will give partial result.

adarshagrawal38: if we don't want partial result, will throw the error.

ai-services/internal/pkg/cli/utils/common.go
adarshagrawal38: PS cmd Response when no app is running

```
➜  ai-services git:(application-ps-cli) ✗ ./bin/ai-services-darwin-arm64 application ps --experimental --runtime podman
Error: no application found
```

ai-services/internal/pkg/cli/utils/common.go
yussufsh: Yes we don't need error messages during rendering.

ai-services/internal/pkg/catalog/client/application.go
```diff
@@ -0,0 +1,123 @@
...
+	if resp.IsError() {
+		return nil, fmt.Errorf("get application ps: server returned HTTP %d: %s", resp.StatusCode(), parseErrorResponse(resp))
+	}
+
+	return &result, nil
+}
+
+// parseErrorResponse attempts to parse the error response from the API.
+// It returns the error message if successfully parsed, otherwise returns the raw response body.
+func parseErrorResponse(resp *resty.Response) string {
```
yussufsh: Lets move to utils.

adarshagrawal38: moved method to ./internal/pkg/utils/util.go:

Avoided `application_util.go`, as it was creating circular dependency issue.

ai-services/internal/pkg/cli/utils/common.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+
+		// Process components pods
+		for _, pod := range psResp.Components {
+			rows := buildPodRowFromAPI(psResp.Name, pod, outputWide)
+			printer.AppendRow(rows...)
+		}
+	}
+
+	return nil
```
mayuka-c: Shouldnt this be in ps.go itself? I dont see how this can be reused.
This looks only for ps.go

ai-services/internal/pkg/cli/utils/common.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// setApplicationPSTableHeaders sets the table headers based on output format.
+func setApplicationPSTableHeaders(printer *utils.Printer, outputWide bool) {
+	if outputWide {
+		printer.SetHeaders("APPLICATION NAME", "POD ID", "POD NAME", "STATUS", "CREATED", "CONTAINERS")
+	} else {
+		printer.SetHeaders("APPLICATION NAME", "POD NAME", "STATUS")
+	}
+}
```
mayuka-c: Also this

ai-services/internal/pkg/cli/utils/application_utils.go
mayuka-c: Yeah we can introduce a new query for name

ai-services/internal/pkg/catalog/client/application.go
```diff
@@ -0,0 +1,107 @@
...
+type ListApplicationsParams struct {
+	// Page is the page number (1-indexed). Default: 1
+	Page int
+	// PageSize is the number of items per page (max: 100). Default: 20
+	PageSize int
+	// DeploymentType filters by deployment type: 'architectures' or 'services'
+	DeploymentType string
+	// CatalogID filters by catalog ID (e.g., 'rag', 'chat', 'digitize', 'summarize')
+	CatalogID string
+}
```
mayuka-c: Please create a models.go here in this pkg and move the struct there

mayuka-c (commented):
Rest LGTM

ai-services/internal/pkg/catalog/client/application.go
adarshagrawal38: done

ai-services/internal/pkg/cli/utils/common.go
adarshagrawal38: moved to ps.go

ai-services/internal/pkg/cli/utils/common.go
adarshagrawal38: okay, moved to ps.go

Shubhamag12 (approved):
lgtm

adarshagrawal38:
@mayuka-c PTAL

yussufsh (approved):

ai-services/internal/pkg/catalog/client/application.go
adarshagrawal38: parsing error response now.

PR #885, opened 2026-06-08
feat: support generic public/private key validation

shwetha-s-poojary:
Replace hardcoded RSA/ECDSA type checking with generic crypto.Signer interface for key pair validation. This enables support for all current and future key types including Ed25519.

Changes:
- Use crypto.Signer interface for generic key validation
- Add fallback parsing for SEC1 (EC) and PKCS1 (RSA) formats
- Remove specific crypto/rsa and crypto/ecdsa imports
- Leverage built-in Equal() method for public key comparison

Supported key types:
- RSA (2048, 3072, 4096-bit)
- ECDSA (P-256, P-384, P-521)
- Ed25519

Tested with OpenSSL-generated certificates in all formats.


ai-services/internal/pkg/utils/cert.go
```diff
@@ -259,44 +258,60 @@ func verifyKeyPairMatch(cert *x509.Certificate, keyBytes []byte) error {
...
+		return fmt.Errorf("private key does not implement crypto.Signer interface")
+	}
+
+	// Get the public key from the private key
+	signerPubKey := signer.Public()
+
+	// Define interface for types that support Equal method
+	type equalable interface {
+		Equal(crypto.PublicKey) bool
+	}
```
mayuka-c, line 303: Any reason for defining new type interface `equalable`?

ai-services/internal/pkg/utils/cert.go
```diff
@@ -259,44 +258,60 @@ func verifyKeyPairMatch(cert *x509.Certificate, keyBytes []byte) error {
...
 
-	return privateKey, nil
+	// Try PKCS1 format for RSA keys (legacy format)
+	rsaKey, rsaErr := x509.ParsePKCS1PrivateKey(keyData)
+	if rsaErr == nil {
+		return rsaKey, nil
+	}
+
+	// Return detailed error if all formats fail
+	return nil, fmt.Errorf("failed to parse private key: tried PKCS8 (%v), SEC1/EC (%v), PKCS1/RSA (%v)", err, ecErr, rsaErr)
```
mayuka-c: I dont think we need to mention on what we tried.
Maybe better to say in the error what are all supported

mayuka-c (commented):
Rest LGTM

ai-services/internal/pkg/utils/cert.go
shwetha-s-poojary: Without the equalable interface, you cannot call `Equal()` on `crypto.PublicKey` - it will error out at compile-time with `type crypto.PublicKey has no field or method Equal`

Go's empty marker interface doesn't declare the method.

ai-services/internal/pkg/utils/cert.go
shwetha-s-poojary: Good point - will modify.

yussufsh (approved):

PR #886, opened 2026-06-08
migrate pod logs to use new catalog structure api way

Shubhamag12:
This PR implements new way of consuming application logs. 
`--container` flag is not required anymore for podman, since each component is deployed as a standalone pod.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/logs.go
+++ b/ai-services/cmd/ai-services/cmd/application/logs.go
@@ -12,7 +12,10 @@ import (
 )
 
 var (
-	podName           string
+	// common flags.
+	podName string
+
+	// openshift flags.
 	containerNameOrID string
 )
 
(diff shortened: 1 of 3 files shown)


ai-services/cmd/ai-services/cmd/application/logs.go
```diff
@@ -62,14 +65,18 @@ Arguments
...
 
 func initLogsCommonFlags() {
 	logsCmd.Flags().StringVar(&podName, appFlags.Logs.Pod, "", "Pod name to show logs from (required)")
-	logsCmd.Flags().StringVar(&containerNameOrID, appFlags.Logs.Container, "", "Container logs to show logs from (Optional)")
 	_ = logsCmd.MarkFlagRequired(appFlags.Logs.Pod)
 }
 
+func initLogsOpenshiftFlags() {
+	logsCmd.Flags().StringVar(&containerNameOrID, appFlags.Logs.Container, "", "Container logs to show logs from (Optional)")
+}
```
mayuka-c, line 78: Any specific reason for removing container flag in podman?
I still feel we can keep it

mayuka-c (commented):
Also you would have to do this as done by @adarshagrawal38 here: https://github.com/IBM/project-ai-services/pull/882/changes#diff-3e137d1e9cfebe3df9234e991d05272a6db814124d93168843156fcd468160d5R119.

We would need to fetch the application ID and validate for a given appName

ai-services/internal/pkg/application/podman/logs.go
```diff
@@ -10,28 +10,10 @@ import (
...
-	}
-	if !exists {
-		return fmt.Errorf("container %s doesn't exists", opts.ContainerNameOrID)
-	}
-
-	logger.Infof("Fetching logs for container: %s", opts.ContainerNameOrID)
-	if err := p.runtime.ContainerLogs(opts.ContainerNameOrID); err != nil {
-		return fmt.Errorf("failed to fetch container: %s logs; err: %w", opts.ContainerNameOrID, err)
+	logger.Infof("Fetching logs for pod: %s", opts.PodName)
+	if err := p.runtime.PodLogs(opts.PodName); err != nil {
```
yussufsh, line 15: Lets use app id from DB.

PR #887, opened 2026-06-08
fix: query Caddy API for actual catalog endpoints instead of hardcoding domains

shwetha-s-poojary:
Replaces hardcoded domain suffix logic with Caddy admin API query to retrieve actual registered routes, fixing 'catalog info' showing wrong endpoints.

catalog info now:

```
-------
✔ Catalog service deployed successfully
-------
Next Steps:
-------
- Access the Catalog UI at https://catalog-ui.example.com:8446

- Access the Catalog Backend at https://catalog-api.example.com:8446


[root@onprem133726180 ~]# ai-services catalog info --runtime podman
Catalog Service Name: ai-services
Catalog Template: catalog
Version: 0.0.1
Info:
-------
Day N:

- Catalog UI is available at https://catalog-ui.example.com:8446

- Catalog Backend API is available at https://catalog-api.example.com:8446
```

### DIFF
--- a/ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
+++ b/ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
@@ -35,6 +35,7 @@ const (
 	kindSecret            = "Secret"
 	caddyCertsDirName     = "certs"
 	caddyContainerDataDir = "/data/caddy"
+	routeFormatParts      = 3 // Expected number of parts in route format: "port:subdomain:type"
 )
 
 // catalogDeploymentContext holds cached values during catalog deployment to avoid redundant lookups.
@@ -750,6 +751,55 @@ func getCaddyHTTPSPort(rt *podman.PodmanClient, caddyPodName string) (string, er
 	return "", fmt.Errorf("HTTPS port mapping not found in pod ports")
 }
 
+// parseRouteEntry parses a single route entry and returns the subdomain.
+// Route format: "port:subdomain:type"
+// Returns empty string if the entry is invalid.
+func parseRouteEntry(routeEntry, podName string) string {
+	routeEntry = strings.TrimSpace(routeEntry)
+	if routeEntry == "" {
+		return ""
+	}
+
+	// Split by colon to get subdomain (second part)
+	parts := strings.Split(routeEntry, ":")
+	if len(parts) != routeFormatParts {
+		logger.Warningf("Invalid route format '%s' in pod %s, expected 'port:subdomain:type'", routeEntry, podName)
+
+		return ""
+	}
+
+	subdomain := strings.TrimSpace(parts[1])
+
+	return subdomain
+}
+
+// processRouteInfo processes route information and populates the routeDomains map.
+func processRouteInfo(info TemplateRouteInfo, proxyManager proxy.ProxyManager, routeDomains map[string]string) {
+	// Parse routes annotation directly to extract subdomains
+	// Format: "port:subdomain:type, port:subdomain:type, ..."
+	// Example: "8081:catalog-ui:ui, 8080:catalog-api:api"
+	for _, routeEntry := range strings.Split(info.RoutesAnnotation, ",") {
+		subdomain := parseRouteEntry(routeEntry, info.PodName)
+		if subdomain == "" {
+			continue
+		}
+
+		// Query Caddy for this route (route ID is just the subdomain)
+		actualRoute, err := proxyManager.GetRouteByID(subdomain)
+		if err != nil {
+			// Log warning but continue - route might not exist yet
+			logger.Warningf("Failed to query route %s from Caddy: %v", subdomain, err)
+
+			continue
+		}
+
+		// Create variable name from subdomain
+		sanitizedSubdomain := strings.ReplaceAll(subdomain, "-", "_")
+		varName := strings.ToUpper(fmt.Sprintf("%s_DOMAIN", sanitizedSubdomain))
+		routeDomains[varName] = actualRoute.Domain
+	}
+}
+
 // GetCatalogRouteInfo retrieves route domains and HTTPS port for the catalog service.
 func GetCatalogRouteInfo(rt *podman.PodmanClient, tp templates.Template, appTemplateName string, argParams map[string]string) (map[string]string, string, error) {
 	// Find Caddy pod from templates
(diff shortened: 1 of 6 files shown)


ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -131,6 +131,80 @@ func (c *caddyManager) createRoute(routeConfig map[string]interface{}) error {
...
+	}
+
+	return domain, true
+}
+
+// GetRegisteredRoutes retrieves all routes currently registered with Caddy.
+// Only extracts route ID and domain for efficiency.
+func (c *caddyManager) GetRegisteredRoutes() ([]Route, error) {
+	// Build URL to get routes from Caddy config
+	routesURL, err := url.JoinPath(c.adminURL, "config", "apps", "http", "servers", c.serverName, "routes")
```
mayuka-c: Can we not do GetRoutesByID directly instead of doing ListRoutes?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -752,42 +752,33 @@ func getCaddyHTTPSPort(rt *podman.PodmanClient, caddyPodName string) (string, er
 
 // GetCatalogRouteInfo retrieves route domains and HTTPS port for the catalog service.
 func GetCatalogRouteInfo(rt *podman.PodmanClient, tp templates.Template, appTemplateName string, argParams map[string]string) (map[string]string, string, error) {
```
mayuka-c, line 796: Can we move this to some utility if its also used by info command

ai-services/internal/pkg/proxy/caddy.go
mayuka-c: I guess this should be fine too since you might end up making two different calls to fetch both catalog-ui and backend

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -752,42 +752,33 @@ func getCaddyHTTPSPort(rt *podman.PodmanClient, caddyPodName string) (string, er
...
-
-	// Build routes from annotations to get domains
-	for _, info := range routeInfos {
-		routes, err := proxy.BuildRoutesFromAnnotation(info.RoutesAnnotation, domainSuffix, info.PodName)
-		if err != nil {
-			continue // Skip if routes can't be built
-		}
-
-		for _, route := range routes {
+	catalogRoutePrefix := catalogconstants.CatalogAppName + "--"
```
mayuka-c: Here no need to append --, just checking for catalogAppName prefix is fine right?

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -131,6 +131,80 @@ func (c *caddyManager) createRoute(routeConfig map[string]interface{}) error {
 	return nil
 }
 
+// extractDomainFromRoute extracts the domain from a Caddy route configuration.
+// Returns the domain and a boolean indicating success.
+func extractDomainFromRoute(rawRoute map[string]interface{}) (string, bool) {
```
Shubhamag12: nit: should we use any instead of interface{}?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
shwetha-s-poojary: Yes should be , doesn't make a difference.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
shwetha-s-poojary: It's actually only used by info subcommand.So will move it to `ai-services/internal/pkg/catalog/cli/info/podman/info.go`

shwetha-s-poojary: All the helper functions used here are currently in this file. Maybe we can handle this refactoring later, once we move those helpers to the right utility files, and keep it like this for now?

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: Cool - I can do that.

mayuka-c (dismissed):
LGTM

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -752,42 +752,32 @@ func getCaddyHTTPSPort(rt *podman.PodmanClient, caddyPodName string) (string, er
...
-		routes, err := proxy.BuildRoutesFromAnnotation(info.RoutesAnnotation, domainSuffix, info.PodName)
-		if err != nil {
-			continue // Skip if routes can't be built
-		}
-
-		for _, route := range routes {
+	for _, route := range registeredRoutes {
+		// Filter catalog routes by checking if route ID starts with catalog app name
+		// Route IDs follow pattern: "appName--podName--subdomain" (e.g., "ai-services--catalog--catalog-ui")
+		if strings.HasPrefix(route.ID, catalogconstants.CatalogAppName) {
```
yussufsh: This route id can conflict if any application is named `ai-services`?

shwetha-s-poojary: I've addressed this by simplifying route-ids naming to subdomain/route in template, which we fetch and GetRouteByID on the fetched routes to get the domains and use it further

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -131,6 +131,70 @@ func (c *caddyManager) createRoute(routeConfig map[string]interface{}) error {
 	return nil
 }
 
+// extractDomainFromRoute extracts the domain from a Caddy route configuration.
+// Returns the domain and a boolean indicating success.
+func extractDomainFromRoute(rawRoute map[string]any) (string, bool) {
```
yussufsh: Return error instead of boolean for better error messages

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -752,49 +752,64 @@ func getCaddyHTTPSPort(rt *podman.PodmanClient, caddyPodName string) (string, er
...
 
-		for _, route := range routes {
-			parts := strings.Split(route.Domain, ".")
-			if len(parts) > 0 {
-				subdomain := parts[0]
-				sanitizedSubdomain := strings.ReplaceAll(subdomain, "-", "_")
-				varName := strings.ToUpper(fmt.Sprintf("%s_DOMAIN", sanitizedSubdomain))
-				routeDomains[varName] = route.Domain
+			// Split by colon to get subdomain (second part)
+			parts := strings.Split(routeEntry, ":")
```
yussufsh: This split logic still exist in `BuildRoutesFromAnnotation` which is also used in `caddy.go`.
Let's have a single place for such logic from maintainability perspective.

yussufsh (changes requested):

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
shwetha-s-poojary: Addressed.

mayuka-c (dismissed):
LGTM

yussufsh (approved):

PR #891, opened 2026-06-08
Add component id in the get application by id response

prachikurade1:
- Update the applications by id API to return component id
- This is required to obtain the template id for the components required to identify the opensearch pod for the restore cli command


prachikurade1:
application_by_id_resp

yussufsh (approved):

PR #901, opened 2026-06-09
[CATALOG-CLI] Making catalog password idempotent and password reset support

adarshagrawal38:
# Summary

Improved user experience for the catalog configure command by skipping password prompt when the catalog secret already exists.

### What Changed
**Before:**

- The catalog configure command always prompted users for an admin password, even when the catalog-secret already existed in the runtime environment

After:
- Added secret existence check before prompting for password
- If catalog-secret exists, the password prompt is skipped entirely
- Password is only requested when the secret is not present

### DIFF
--- a/ai-services/cmd/ai-services/cmd/catalog/configure.go
+++ b/ai-services/cmd/ai-services/cmd/catalog/configure.go
@@ -1,15 +1,17 @@
 package catalog
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
-	"syscall"
+	"strconv"
 
 	"github.com/spf13/cobra"
-	"golang.org/x/term"
 
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog/cli/configure"
+	catalogPodman "github.com/project-ai-services/ai-services/internal/pkg/catalog/cli/configure/podman"
+	catalogConstant "github.com/project-ai-services/ai-services/internal/pkg/catalog/constants"
 	"github.com/project-ai-services/ai-services/internal/pkg/constants"
 	"github.com/project-ai-services/ai-services/internal/pkg/logger"
 	"github.com/project-ai-services/ai-services/internal/pkg/runtime"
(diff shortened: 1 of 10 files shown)


ai-services/cmd/ai-services/cmd/catalog/configure.go
```diff
@@ -73,9 +74,22 @@ Examples:
...
+		return fmt.Errorf("failed to create runtime: %w", err)
+	}
+
+	existingSecret, err := helpers.ExistingSecrets(rt, []string{"catalog-secret"})
+	if err != nil {
+		return fmt.Errorf("failed to check existing secrets: %w", err)
+	}
+
+	var adminPassword string
+	if len(existingSecret) == 0 {
```
yussufsh: If secret exist should we ask if user want to overwrite?
cc @mayuka-c 
This will help with password reset requirement.

mayuka-c: I was thinking maybe to give some flag `--reset-admin` which would then prompt the user to enter new password and we override?

This is so that user might run `configure` for different purpose like cert renewal, pods not deployed properly, routes not registered in Caddy etc and we should not be asking him the same prompt again

mayuka-c: Also if we are updating the existing secret, we also need to ensure that we restart the pods (consuming these secret) @adarshagrawal38

adarshagrawal38: If we want to allow reset of password, then having a `--reset-admin` flag is a good idea
@yussufsh WDYT?

yussufsh: I'm okay with it, but please change the name for the parameter `reset-admin` seems something else.

ai-services/cmd/ai-services/cmd/catalog/configure.go
```diff
@@ -66,16 +69,21 @@ Examples:
...
+	configureBoolFlags(cmd)
 
 	return cmd
 }
 
 // runConfigure executes the catalog configuration process.
 func runConfigure(argParams map[string]string) error {
-	// Prompt for admin password
-	adminPassword, err := promptForPassword()
+	rt, err := vars.RuntimeFactory.Create("")
```
yussufsh, line 126: Let's have a different func for just resetting the password when the parameter is defined. Also lets restart the containers instead of delete.

PR #903, opened 2026-06-09
docx utilities to preserve hierarchical metadata

Niharika0306:
This PR aims at adding utilities and support to extract toc and headings from docx files, and maintain the hierarchical  metadata.

1. toc extraction first happens while checking for a table_paragraph and list_paragraph. 
2. And then falls back to check via headings in word documents
3. there is also page_mapping done via an estimation logic that helps for local search - ( can remove if not needed)


Niharika0306:
@dharaneeshvrd the changes are ready. ptal.

services/digitize/doc_utils.py
```diff
@@ -351,17 +357,6 @@ def process_text(converted_doc, pdf_path, out_path):
         if label == "section_header":
             prov_list = text_obj.prov
 
-            # Handle empty prov list (e.g., for DOCX files)
```
dharaneeshvrd, line 354: Any reason to remove this in pdf method?

services/digitize/doc_utils.py
```diff
@@ -316,15 +317,20 @@ def recover_table_caption_from_body_context(converted_doc, table_ix: int, search
     'page_header', 'page_footer', 'caption', 'reference', 'footnote'
 }
 
-def process_text(converted_doc, pdf_path, out_path):
```
dharaneeshvrd, line 319: I prefer not to edit this method, since we dont know what would break.
Maybe we can alter if needed in next release.

services/digitize/doc_utils.py
Niharika0306: so this block was added as part of the initial PR for docx enablement. Now that I have a separate function for docx, no prov check is needed there because we are not using it at all in docx.

dharaneeshvrd: ok got it

services/digitize/doc_utils.py
```diff
@@ -397,23 +478,16 @@ def process_text(converted_doc, pdf_path, out_path):
...
-                            })
-                    else:
-                        # No pdf_pages available (DOCX), just add without font size
-                        structured_output.append({
-                            "label": label,
-                            "text": text_obj.text,
-                            "page": page_no,
-                            "font_size": None
-                        })
+                    structured_output.append({
```
dharaneeshvrd: this block needs to be inside `if len(matches):`

Niharika0306: done.

services/digitize/doc_utils.py
Niharika0306: handled.

dharaneeshvrd (approved):

PR #904, opened 2026-06-09
Refactor images part in deployer code

yussufsh:
This refactor will help in writing new code for image list/pull cli.

1. `CollectImagesFromTemplates` will be reused and does append images to the imageSet by using the given templates.
2. Also collecting images specified in `podSpec.Spec.InitContainers`

### DIFF
--- a/ai-services/internal/pkg/catalog/template_extractor.go
+++ b/ai-services/internal/pkg/catalog/template_extractor.go
@@ -2,6 +2,8 @@ package catalog
 
 import (
 	"bytes"
+	"errors"
+	"fmt"
 	texttemplate "text/template"
 
 	k8syaml "sigs.k8s.io/yaml"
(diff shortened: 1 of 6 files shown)


ai-services/internal/pkg/catalog/template_extractor.go
```diff
@@ -0,0 +1,64 @@
...
+			"Values":       values,
+			"env":          map[string]map[string]string{},
+		}
+
+		// Render the template
+		var rendered bytes.Buffer
+		if err := tmpl.Execute(&rendered, initialParams); err != nil {
+			logger.Errorf("Failed to render template %s for image extraction: %v", templateName, err)
+			continue
+		}
```
mayuka-c, line 53: Dont we need to capture the errors in array like `errorResponses`.
And then fail if any template it failed and return error?

yussufsh: Yes, let me add error handling there.
I see we are doing the same in `planner.go` as well. Will try to fix it.

ai-services/internal/pkg/catalog/template_extractor.go
```diff
@@ -0,0 +1,170 @@
...
+	return totalSpyreCards, nil
+}
+
+// fetchSpyreCardsFromPodAnnotations extracts Spyre card requirements from pod annotations.
+// Returns the total number of Spyre cards and a map of container names to their card requirements.
+func fetchSpyreCardsFromPodAnnotations(annotations map[string]string) (int, map[string]int, error) {
+	var spyreCards int
+	spyreCardContainerMap := map[string]int{}
+
+	spyreCardAnnotationRegex := regexp.MustCompile(`^ai-services\.io\/([A-Za-z0-9][-A-Za-z0-9_.]*)--spyre-cards$`)
```
mayuka-c: There's already a regex for this in vars package

mayuka-c (commented):
Rest LGTM

mayuka-c (dismissed):

ai-services/internal/pkg/catalog/template_extractor.go
yussufsh: Updated in template_extractor and also in deployer.go

mayuka-c (approved):

PR #905, opened 2026-06-09
migrating cli start cmd to use catalog structure 

Shubhamag12:
1. Added experimental mode for starting application pods via CLI.
2. If experimental mode is enabled, we do not filter pods via label, but fetch all pods using application ps api

### DIFF
--- a/ai-services/internal/pkg/cli/utils/common.go
+++ b/ai-services/internal/pkg/cli/utils/common.go
@@ -9,6 +9,7 @@ import (
 	catalogConstants "github.com/project-ai-services/ai-services/internal/pkg/catalog/constants"
 	catalogTypes "github.com/project-ai-services/ai-services/internal/pkg/catalog/types"
 	"github.com/project-ai-services/ai-services/internal/pkg/constants"
+	"github.com/project-ai-services/ai-services/internal/pkg/runtime/types"
 	"github.com/project-ai-services/ai-services/internal/pkg/utils"
 )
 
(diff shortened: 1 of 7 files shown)


ai-services/internal/pkg/cli/utils/common.go
```diff
@@ -98,17 +99,40 @@ func getContainerNamesFromAPI(pod catalogTypes.Pod) []string {
 	return containerNames
 }
 
-func ValidateApplicationName(appName string) error {
+func GetPodsFromApplicationsPS(appName string) ([]types.Pod, error) {
```
mayuka-c, line 102: I feel let us have both util method,
Validate Application and then this GetPods too which takes app as input

So basically Validate Application will validate and return the app if valid

mayuka-c (commented):
Rest LGTM

ai-services/internal/pkg/cli/utils/common.go
Shubhamag12: I removed it because its already there in application_utils.go

mayuka-c: Ah ok, also any reason why again we need to do one more API call, I feel the Validate call simply return the app and we can reuse here

mayuka-c: Also please move that here in cli/utils, ValidateApp is required for CLI

Shubhamag12: > also any reason why again we need to do one more API call.

because we need to fetch all pods of that application

Shubhamag12: > Also please move that here in cli/utils, ValidateApp is required for CLI

its already in cli/utils/application_utils.go

yussufsh (approved):

PR #909, opened 2026-06-09
Support image pull and list with new catalog architecture

yussufsh:
- Created image_extractor.go with CollectImagesFromTemplates() function to extract container images from templates by parsing YAML and collecting image references from pod specs.
- Integrated image extraction into catalog CLI commands (image pull and image list) to enable pulling and listing container images required by catalog services.
- Added tools and catalog infrastructure images (Caddy, Catalog, Postgres, etc.) to support the catalog deployment.

Note for reviewers: Please look at each commit, it will be easy to review. 😄

### DIFF
--- a/ai-services/internal/pkg/catalog/image_extractor.go
+++ b/ai-services/internal/pkg/catalog/image_extractor.go
@@ -1,0 +1,144 @@
+package catalog
+
+import (
+	"fmt"
+
+	"github.com/project-ai-services/ai-services/internal/pkg/catalog/types"
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+	"github.com/project-ai-services/ai-services/internal/pkg/utils"
+)
+
+// GetCatalogImages collects all unique images from a service or architecture template.
+// This is the main entry point for catalog-based image collection from CLI or API.
+func (p *CatalogProvider) GetCatalogImages(templateID string) ([]string, error) {
+	allImages := make(map[string]bool)
+
+	// Try to load as architecture first
+	arch, err := p.LoadArchitecture(templateID)
+	if err == nil {
+		if err := p.collectArchitectureImages(arch.Services, allImages); err != nil {
+			return nil, err
+		}
+	} else {
+		// Try to load as service
+		service, err := p.LoadService(templateID)
+		if err == nil {
+			if err := p.collectServiceWithDependencies(templateID, service.Dependencies, allImages); err != nil {
+				return nil, err
+			}
+		} else {
+			return nil, fmt.Errorf("template '%s' not found as service or architecture", templateID)
+		}
+	}
+
+	return utils.ExtractMapKeys(allImages), nil
+}
+
+// collectArchitectureImages collects all images for all services in an architecture.
+func (p *CatalogProvider) collectArchitectureImages(services []types.ServiceReference, allImages map[string]bool) error {
+	for _, svcRef := range services {
+		service, err := p.LoadService(svcRef.ID)
+		if err != nil {
+			continue
+		}
+
+		// Reuse collectServiceWithDependencies to collect both service and component images
+		if err := p.collectServiceWithDependencies(svcRef.ID, service.Dependencies, allImages); err != nil {
+			logger.Errorf("Failed to collect images for service %s: %v", svcRef.ID, err)
+		}
+	}
+
+	return nil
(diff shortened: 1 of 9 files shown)


ai-services/internal/pkg/catalog/image_extractor.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/project-ai-services/ai-services/assets"
+	"github.com/project-ai-services/ai-services/internal/pkg/catalog/constants"
+	"github.com/project-ai-services/ai-services/internal/pkg/catalog/types"
+	"github.com/project-ai-services/ai-services/internal/pkg/cli/templates"
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+	"github.com/project-ai-services/ai-services/internal/pkg/utils"
+	"github.com/project-ai-services/ai-services/internal/pkg/vars"
+)
+
+// GetCatalogImages collects all unique images from a service or architecture template.
```
mayuka-c, line 15: Do we want to even support for a service?
I feel lets support for architecture itself (Both listing and pull)

ai-services/internal/pkg/catalog/image_extractor.go
```diff
@@ -0,0 +1,179 @@
...
+	return nil
+}
+
+// collectArchitectureImages collects all images for all services in an architecture.
+func (p *CatalogProvider) collectArchitectureImages(services []types.ServiceReference, allImages map[string]bool) error {
+	for _, svcRef := range services {
+		service, err := p.LoadService(svcRef.ID)
+		if err != nil {
+			continue
+		}
```
mayuka-c, line 83: Do we want to collect array of errorResponses and fail?

ai-services/internal/pkg/catalog/image_extractor.go
```diff
@@ -0,0 +1,179 @@
...
+
+		// Reuse collectServiceWithDependencies to collect both service and component images
+		if err := p.collectServiceWithDependencies(svcRef.ID, service.Dependencies, allImages); err != nil {
+			logger.Errorf("Failed to collect images for service %s: %v", svcRef.ID, err)
+		}
```
mayuka-c, line 88: Dont we need to return error here?

ai-services/internal/pkg/catalog/image_extractor.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	return nil
+}
+
+// collectServiceWithDependencies collects images for a specific service and its dependencies.
+func (p *CatalogProvider) collectServiceWithDependencies(serviceID string, dependencies []types.DependencyReference, allImages map[string]bool) error {
```
mayuka-c, line 95: Here is 1 way I'm thinking, to simplify process

For a given architecture lets collect all services and components in a set (this avoid duplicate).
Then loop over each services -> collect all images
Loop over each components -> collect all images

ai-services/internal/pkg/catalog/image_extractor.go
yussufsh: A template would be an architecture or service. Also, if I want to deploy, say summarize on an air-gapped machine, I do not need everything.

mayuka-c: But dont we need to pull all the images/models before hand?
Otherwise once they install catalog, it will list everything there (Diigital Assistant and services) and then it will fail if user tries to provision other.

So I feel we need to pull everything

ai-services/internal/pkg/catalog/image_extractor.go
yussufsh: I am reusing the `collectServiceWithDependencies` method, so I don't need to store it in a set. All that it is doing is getting all services in an architecture and call the same method using `service.Dependencies`.

ai-services/internal/pkg/catalog/image_extractor.go
yussufsh: > But dont we need to pull all the images/models before hand?

Yes, but this can be done anytime on an internet-connected machine and loaded into podman.
While deploying, they need to ensure they have all the prerequisites needed for it to work.

Also IMO we should keep all templates since we are showing them in the application templates subcommand.

mayuka-c (approved):
LGTM

PR #910, opened 2026-06-09
migrating cli stop cmd to use catalog structure

Shubhamag12:
Added experimental mode for stopping application pods via CLI.
If experimental mode is enabled, we do not filter pods via label, but fetch all pods using application ps api


ai-services/internal/pkg/application/podman/stop.go
```diff
@@ -5,18 +5,28 @@ import (
...
 // Stop stops a running application.
 func (p *PodmanApplication) Stop(opts appTypes.StopOptions) error {
-	pods, err := p.runtime.ListPods(map[string][]string{
-		"label": {fmt.Sprintf("ai-services.io/application=%s", opts.Name)},
-	})
-	if err != nil {
-		return fmt.Errorf("failed to list pods: %w", err)
+	var pods []types.Pod
+	var err error
+	if !opts.Experimental {
```
yussufsh: For easier deprecation and removal, have another func that is experimental and do a check before this func (parent func).

yussufsh: I think you are trying to reuse the existing code.. ignore my previous comment.

Shubhamag12: yes, in future when we migrate fully, we will have to remove just one part of code

yussufsh (approved):

PR #912, opened 2026-06-10
Add jq to tools image

prachikurade1:
Added the jq package to the tools image since it is needed for restore opensearch functionality


ai-services/internal/pkg/vars/var.go
```diff
@@ -15,7 +15,7 @@ var (
 var (
 	// SpyreCardAnnotationRegex -> ai-services.io/<containerName>--spyre-cards.
 	SpyreCardAnnotationRegex = regexp.MustCompile(`^ai-services\.io\/([A-Za-z0-9][-A-Za-z0-9_.]*)--spyre-cards$`)
-	ToolImage                = "icr.io/ai-services/tools:0.7"
+	ToolImage                = "icr.io/ai-services-private/tools:0.9"
```
mkumatag: can you point to the cicd repo for now, we will find a way to override the tools image later to the public image.

mkumatag (dismissed):

ai-services/internal/pkg/vars/var.go
mkumatag: @adarshagrawal38 we need to find a way to use the public image later during the release cut binaries.

adarshagrawal38: How about adding a `make target`  updates the image paths to point to the public repository?

prachikurade1: updated to point to cicd repo

adarshagrawal38 (dismissed):
LGTM

images/tools/requirements.txt
```diff
@@ -1,2 +1,24 @@
 hf_xet==1.2.0
```
yussufsh:
```suggestion
hf_xet==1.5.0
```

images/tools/requirements.txt
```diff
@@ -1,2 +1,24 @@
...
 huggingface_hub[cli]==1.0.1
```
yussufsh: Let's update and use latest versions of hf and all dependencies which are available on devpi.

yussufsh (changes requested):

images/tools/requirements.txt
prachikurade1: done

images/tools/requirements.txt
prachikurade1: Done

yussufsh (dismissed):

yussufsh (approved):

ai-services/internal/pkg/vars/var.go
mkumatag: > How about adding a `make target` updates the image paths to point to the public repository?

yes, we might need to add one more target called `release` and override the registry only via `-ldflags "-X ..."` flag and ship it.

PR #913, opened 2026-06-10
fix: use values instead of param due to load failure in opensearch template

yussufsh:
After adding error handling to planner code while loading templates for calculating spyre card numbers we are facing following issue:
```
{
  "error": "Failed to create application: failed to create deployment plan: failed to allocate Spyre cards: failed to get Spyre card requirements for component vector_store: failed to collect Spyre cards from templates: failed to process templates for Spyre card calculation: failed to process 1 template(s): error unmarshaling JSON: while decoding JSON: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'"
}
```

This is due to an empty memoryLimit in opensearch template, even with values.yaml, which contains the default value.

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=0.0.63
+TAG?=0.0.64
 CONTAINER_BUILDER?=podman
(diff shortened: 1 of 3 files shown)


Shubhamag12 (dismissed):
lgtm

Shubhamag12 (approved):
lgtm

PR #916, opened 2026-06-10
Support model list and download cli with the new template

yussufsh:
1. Support model list/download with new catalog structure.
2. Implement model_extractor.go for collecting component models for a given template.
3. Skip watsonx model during download.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/model/download.go
+++ b/ai-services/cmd/ai-services/cmd/application/model/download.go
@@ -36,6 +36,10 @@ func init() {
 }
 
 func download(cmd *cobra.Command) error {
+	if experimentalModels && vars.RuntimeFactory.GetRuntimeType() == types.RuntimeTypePodman {
+		return downloadCatalogModels(templateName)
+	}
+
 	if vars.RuntimeFactory.GetRuntimeType() == types.RuntimeTypeOpenShift {
 		// Since we do not have tmpl files in OpenShift marking it as unsupported for now
 		logger.Warningln("Not supported for openshift runtime")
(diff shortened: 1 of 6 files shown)


ai-services/cmd/ai-services/cmd/application/model/download.go
```diff
@@ -57,3 +61,32 @@ func download(cmd *cobra.Command) error {
 
 	return nil
 }
+
+// downloadCatalogModels downloads models for services or architectures from the catalog.
+func downloadCatalogModels(templateID string) error {
+	models, err := getCatalogModels(templateID, "watsonx")
```
mayuka-c, line 67: Isnt watson component type provider?
I see in the function parameter it says excludeComponents which might sound like excluding embedding, llm, vector_Store right?

mayuka-c (commented):
Minor comment, Rest LGTM

ai-services/internal/pkg/catalog/model_extractor.go
```diff
@@ -0,0 +1,173 @@
...
+					modelSet[constVal] = true
+				}
+			}
+		}
+	}
+
+	// Also check for default value
+	if defaultVal, ok := propMap["default"].(string); ok && defaultVal != "" {
+		modelSet[defaultVal] = true
+	}
```
mayuka-c: We need not have to check for default, just checking for models params should be enough `oneOf` property

ai-services/cmd/ai-services/cmd/application/model/download.go
yussufsh: Changed the param name to `excludeComponentProviders`.

ai-services/internal/pkg/catalog/model_extractor.go
yussufsh: Removed default.

mayuka-c (approved):
LGTM

PR #922, opened 2026-06-11
refactor code to use the helper function GetCaddyProxyManager()

shwetha-s-poojary:
Update RegisterRoutesForAppAndReturn to accept ProxyManager.Simplified function signature by passing ProxyManager instead of adminURL and serverName parameters

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
+++ b/ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
@@ -1121,7 +1121,7 @@ func (d *PodmanDeployer) getEnvParamsForComponent(podSpec *podmodels.PodSpec, pl
 func (d *PodmanDeployer) registerApplicationRoutes(ctx context.Context, plan *DeploymentPlan) error {
 	logger.Infof("Registering routes for application '%s'\n", plan.ApplicationName)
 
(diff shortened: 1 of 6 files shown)


yussufsh (approved):

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -1121,7 +1121,7 @@ func (d *PodmanDeployer) getEnvParamsForComponent(podSpec *podmodels.PodSpec, pl
 func (d *PodmanDeployer) registerApplicationRoutes(ctx context.Context, plan *DeploymentPlan) error {
 	logger.Infof("Registering routes for application '%s'\n", plan.ApplicationName)
 
-	adminURL, domainSuffix, httpsPort, err := d.getCaddyConfiguration()
+	domainSuffix, httpsPort, proxyManager, err := d.getCaddyConfiguration()
 	if err != nil {
```
shwetha-s-poojary, line 1125: @yussufsh answering your question earlier about why we are calling `getCaddyConfiguration` in `registerApplicationRoutes` and not in `registerServiceRoutes` is because don't want to call this function every time for each service (we are looping over services of the application below)Instead we just calculate it at application level once and use it for every service)

PR #926, opened 2026-06-11
Move catalog db data to volume

yussufsh:
In manifests:
- Migrated from hostPath to persistentVolumeClaim.
- Added volume tracking and skip-cleanup labels

In uninstall:
- Implemented dynamic volume extraction from pod labels with volume-skip-cleanup support

### DIFF
--- a/ai-services/assets/catalog/podman/templates/catalog-db.yaml.tmpl
+++ b/ai-services/assets/catalog/podman/templates/catalog-db.yaml.tmpl
@@ -8,21 +8,11 @@ metadata:
     ai-services.io/version: "{{ .Version }}"
     ai-services.io/secret: "catalog-db-secret"
     ai-services.io/secret-skip-cleanup: "true"
+    ai-services.io/volume: "postgres-catalog"
+    ai-services.io/volume-skip-cleanup: "true"
+  annotations:
+    io.podman.annotations.pids-limit/postgresql: "4096"
 spec:
-  initContainers:
-  - name: fix-permissions
-    image: "{{ .Values.db.image }}"
-    command:
-    - /bin/sh
-    - -c
-    - |
-      chown -R 26:0 /var/lib/pgsql
-      chmod -R g=u /var/lib/pgsql
-    volumeMounts:
-    - name: postgres-data
-      mountPath: /var/lib/pgsql
-    securityContext:
-      runAsUser: 0
   containers:
   - name: postgresql
     image: "{{ .Values.db.image }}"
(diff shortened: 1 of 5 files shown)


ai-services/assets/catalog/podman/templates/catalog-db.yaml.tmpl
```diff
@@ -8,21 +8,11 @@ metadata:
     ai-services.io/version: "{{ .Version }}"
     ai-services.io/secret: "catalog-db-secret"
     ai-services.io/secret-skip-cleanup: "true"
+    ai-services.io/volume: "postgres-catalog"
+    ai-services.io/volume-skip-cleanup: "true"
```
mayuka-c, line 12: Any reason of adding volume-skip-cleanup?
By default volume will exist if user dont use `skip-cleanup`

ai-services/internal/pkg/catalog/cli/uninstall/podman/uninstall.go
```diff
@@ -221,4 +268,42 @@ func dataDeletion(dataPath string) error {
 	return nil
 }
 
+// deleteVolumes removes the specified volumes.
+func deleteVolumes(rt *podman.PodmanClient, volumeNames []string) error {
+	if len(volumeNames) == 0 {
+		logger.Infoln("No volumes to delete")
```
mayuka-c: I guess we dont have to even log this

ai-services/assets/catalog/podman/templates/catalog-db.yaml.tmpl
yussufsh: I just used the secret cleanup logic.
VolumeDelete logic was missing for the catalog; I have now added it to delete volumes for both persistent data and normal volumes, e.g. `podman-auth-secret`.

ai-services/internal/pkg/catalog/cli/uninstall/podman/uninstall.go
yussufsh: ok will skip printing.

ai-services/assets/catalog/podman/templates/catalog-db.yaml.tmpl
mayuka-c: I feel by default for volumes, we can follow the logic without this label right?
Preserve all the volumes if `--skip-cleanup` is used as part of `uninstall`?

yussufsh: Now I see what you are saying...
But my counter is keeping volume like `podman-auth-secret` does not make sense in catalog pods.

When skip-cleanup=false:
Delete `podman-auth-secret`
Delete `postgres-catalog`

When skip-cleanup=true:
Delete `podman-auth-secret`
Keep `postgres-catalog`

Shubhamag12 (approved):
lgtm

mayuka-c (commented):
LGTM

PR #928, opened 2026-06-11
migrate create CLI cmd to use catalog API

Shubhamag12:
Migrated ai-services application create command to use the Catalog API when `--experimental` flag is enabled for podman runtime.

1. Defaults to vllm-spyre for LLM/reranker (Spyre acceleration)
2. Switches to CPU when user provides provider-specific params: llm.vllm-cpu.model=...

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/create.go
+++ b/ai-services/cmd/ai-services/cmd/application/create.go
@@ -260,6 +260,12 @@ func buildFlagValidator() *flagvalidator.FlagValidator {
 
 // validateTemplateFlag validates the template flag.
 func validateTemplateFlag(cmd *cobra.Command) error {
+	// Skip template validation in experimental mode for podman runtime
+	// In experimental mode, templates are validated against catalog API
+	if experimentalCreate && vars.RuntimeFactory.GetRuntimeType() == types.RuntimeTypePodman {
+		return nil
+	}
+
 	tp := templates.NewEmbedTemplateProvider(&assets.ApplicationFS)
 	if err := tp.AppTemplateExist(templateName); err != nil {
 		return err
@@ -395,7 +401,7 @@ func createApp(appName string) error {
 	if err != nil {
 		logger.Warningf("Failed to marshal payload for debugging: %v\n", err)
 	} else {
-		logger.Infof("=== CLI REQUEST PAYLOAD ===\n%s\n", string(payloadJSON))
+		logger.Infof("CLI REQUEST PAYLOAD:\n%s\n", string(payloadJSON))
 	}
 
 	// 7. Create application via catalog API
@@ -694,7 +700,7 @@ func selectProviderFromDeployOptions(compDeployOpt catalogTypes.DeployOptionsCom
 
 	// Special logic for LLM and reranker component types - prefer Spyre acceleration
 	if compDeployOpt.Type == "llm" || compDeployOpt.Type == "reranker" {
-		// Default to vllm-spyre for LLM/reranker if available (Spyre acceleration)
+		// Default to vllm-spyre for LLM/reranker if available
 		for _, p := range compDeployOpt.Providers {
 			if p.ID == "vllm-spyre" {
 				return p.ID, make(map[string]string)
--- a/ai-services/cmd/ai-services/cmd/application/create.go
+++ b/ai-services/cmd/ai-services/cmd/application/create.go
@@ -2,7 +2,10 @@ package application
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"maps"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


Shubhamag12:
@mayuka-c @yussufsh PR is ready to review

ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -324,4 +362,490 @@ func validateSkipChecksFlag(cmd *cobra.Command) error {
...
+	}
+
+	// 3. Build the catalog API payload
+	payload, err := buildCatalogPayload(appName)
+	if err != nil {
+		return err
+	}
+
+	// 4. Debug: Print the payload being sent
+	logPayload(payload)
```
yussufsh: No need to log, might contain sensitive information.

ai-services/cmd/ai-services/cmd/application/create.go
yussufsh: Please add some examples for 5 cards, reranker-cpu, everything on cpu, etc.

ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -324,4 +361,477 @@ func validateSkipChecksFlag(cmd *cobra.Command) error {
...
+func findSpyreProvider(compDeployOpt catalogTypes.DeployOptionsComponent) string {
+	if compDeployOpt.Type != "llm" && compDeployOpt.Type != "reranker" {
+		return ""
+	}
+
+	for _, p := range compDeployOpt.Providers {
+		if p.ID == "vllm-spyre" {
+			return p.ID
+		}
+	}
```
mayuka-c: This needs to be generic, if tomo if we add embedding with spyre then again we need to rework on code.

Can we not determine this based on LoadingComponents and checking if it has spyre provider or not?

ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -324,4 +361,477 @@ func validateSkipChecksFlag(cmd *cobra.Command) error {
...
+	logger.Infof("Creating application '%s' using template '%s'...\n", appName, templateName)
+	resp, err := appClient.CreateApplication(payload)
+	if err != nil {
+		return fmt.Errorf("failed to create application: %w", err)
+	}
+
+	logger.Infof("Application creation initiated (ID: %s)\n", resp.ID)
+
+	// 5. Poll for application status
+	return pollApplicationStatus(appClient, appName)
```
mayuka-c: We would need to print Next steps after polling if succeeded

mayuka-c (commented):
Rest LGTM, approving if we want to handle the comments in next PR

mayuka-c (dismissed):

ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -324,4 +361,477 @@ func validateSkipChecksFlag(cmd *cobra.Command) error {
...
+			}
+		}
+
+		// If provider not found in deploy options, return error
+		if !providerFound {
+			return apiModels.Service{}, fmt.Errorf("provider '%s' not found in deploy options for component type '%s'", providerID, compDeployOpt.Type)
+		}
+
+		// Fetch schema and apply defaults, merging with user params
+		componentParamsAny, err := applySchemaDefaults(appClient, compDeployOpt.Type, providerID, userParams)
```
mayuka-c, line 617: Also from the schema we need to evaluate other things like min-max length, required and other properties present

ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -324,4 +361,477 @@ func validateSkipChecksFlag(cmd *cobra.Command) error {
...
+
+	return &apiModels.CreateApplicationRequest{
+		CatalogID: serviceID,
+		Name:      appName,
+		Services:  []apiModels.Service{svc},
+		Version:   svc.Version,
+	}, nil
+}
+
+// buildServiceEntryWithDeployOptions builds a single service entry with its components using deploy options.
```
mayuka-c, line 585: I do not see service params read for example `system-prompt` for chat. Are we doing it somewhere?

ai-services/cmd/ai-services/cmd/application/create.go
Shubhamag12: I will remove check on line 765. 
Then the function simply checks if vllm-spyre is available in the component's provider list, regardless of what component type is.

ai-services/cmd/ai-services/cmd/application/create.go
Shubhamag12: isnt that handled from the api implementation side?

ai-services/cmd/ai-services/cmd/application/create.go
Shubhamag12: yes, we are doing it below in https://github.com/IBM/project-ai-services/pull/928/changes#diff-6722bcebe66658a9d8011492f9b3ecc5ae37a4ebecb0b4728fdc353709ca754cR622

ai-services/cmd/ai-services/cmd/application/create.go
yussufsh: > Can we not determine this based on LoadingComponents and checking if it has spyre provider or not?

Should there be validation for this? If handled in API then just go ahead and allow this.

ai-services/cmd/ai-services/cmd/application/create.go
mayuka-c: Not sure, please check if the validations are handling it. I think since you had done validations you might be knowing it.

ai-services/cmd/ai-services/cmd/application/create.go
mayuka-c: Yes that should be good and better option here. Anyway API would return 400, we can just error out the message in terminal

ai-services/cmd/ai-services/cmd/application/create.go
Shubhamag12: anyway, we are going to raise a PR for param validation (watsonx keys), will take in that if not done already

ai-services/cmd/ai-services/cmd/application/create.go
Shubhamag12: taking it up as part of next PR

ai-services/cmd/ai-services/cmd/application/create.go
Shubhamag12: done

ai-services/cmd/ai-services/cmd/application/create.go
Shubhamag12: will handle it as part of subsequent PRs

yussufsh (approved):
lgtm

mayuka-c (approved):
LGTM

PR #932, opened 2026-06-11
fix: clean up auth-secret volume during uninstall 

yussufsh:
After uninstall `podman-auth-secret` was left out.
Added the missing label to cleanup volumes properly.


mayuka-c (approved):
LGTM

Shubhamag12 (approved):

PR #936, opened 2026-06-12
[Catalog] Backend Asset updates

mayuka-c:
- Update descriptions for Digital Assistant and services
- Minor updates on About for reordering
- Make Watsonx params as required and also WatsonX naming to all lower

### DIFF
--- a/ai-services/assets/architectures/rag/metadata.yaml
+++ b/ai-services/assets/architectures/rag/metadata.yaml
@@ -34,9 +34,9 @@ services:
 about:
   - title: "Services"
     values:
-      - "Question and Answer"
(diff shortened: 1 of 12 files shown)


ai-services/assets/architectures/rag/metadata.yaml
yussufsh, line 20: We will need order for UI rendering.

mayuka-c: We are not maintaing order numbers anymore as we are handling it differently here: https://github.com/IBM/project-ai-services/pull/941

yussufsh (approved):

PR #941, opened 2026-06-12
read about as yaml for maintaining the order

mkumatag:
Instead of using the standard gin JSON marshaller, use a custom JSON, yaml marshaller and unmarshaller to preserve order.


mayuka-c:
Added UnmarshalYAML as with default unmarshaller used by GIN there was an issue.
Rest all changes that you see are refactoring due to linters

yussufsh (approved):

PR #943, opened 2026-06-12
Get Reference Doc is broken if Similarity Search Pod restarts

sats-23:
[AISERVICES-1251](https://jsw.ibm.com/browse/AISERVICES-1251)

-When similarity search pod restarts, nginx stores a cached IP for the resolved hostname which is invalid when the pod restarts. 
-Passing PODMAN_DNS as env which is used to resolve to an IP which is valid for just 5 seconds (to resolve on every request - only for similarity search)
-These changes should not affect the OpenShift env as /etc/resolv.conf points to the Cluster DNS IP and does the same (just for ss call)


sats-23:
Refer: https://jsw.ibm.com/browse/AISERVICES-1251

PR #946, opened 2026-06-12
feat: implement idempotent catalog reconfigure with validation

shwetha-s-poojary:
- Add domain immutability validation during reconfigure
- Add HTTPS port immutability validation during reconfigure
- Block certificate content changes (require reset-cert)
- Skip certificate reload when identical
- Simplify route registration to check existence only
- Remove verbose logging for cleaner output


shwetha-s-poojary:
closing this as we have a different PR to take care of this.

PR #948, opened 2026-06-14
[Catalog] Implement Pod-DB Periodic Status Synchronization

mayuka-c:
- Spawn a goroutine during container start up which periodically syncs pod and db status periodically for all applications (every 30 sec).
- The status is only synced for those Applications which are currently either in `Running` or `Error` state. If the application is in any other state that means its still in deploying or deletion, so dont want to do syncing for those.
- Uses a mutex to avoid multiple sync process to run at the same time (to avoid overhead).

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.90
+TAG?=v0.0.91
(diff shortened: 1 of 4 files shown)


mayuka-c:
Validation is been added in other PR here: https://github.com/IBM/project-ai-services/pull/949#issuecomment-4701730914

yussufsh (approved):

PR #952, opened 2026-06-14
[Catalog-API]: Bug - Fixing application ps endpoint

adarshagrawal38:
Fixes the `/v1/application/{id}/ps` response so it returns all pods associated with the specified application ID, instead of omitting some pods due to incomplete handler behavior.


adarshagrawal38:
Response of application ps cmd
image

yussufsh (approved):

mayuka-c (approved):
LGTM

PR #955, opened 2026-06-15
Fix truncated responses issue

sats-23:
-[AISERVICES-1164](https://jsw.ibm.com/browse/AISERVICES-1164)
-Fix bug with max_tokens not being picked up language specifically (always defaulted to English)
-Get token count of system prompt and query prompt
-UI nit fix to not append ...

### DIFF
--- a/services/common/llm_utils.py
+++ b/services/common/llm_utils.py
@@ -203,6 +203,7 @@ def query_vllm_payload(
     query_system_prompt_sample = query_system_prompt.format(
         context="",
         rephrased_query="",
+        max_tokens=max_new_tokens,
     )
     rag_system_token_overhead = len(tokenize_with_llm(query_system_prompt_sample, llm_endpoint))
 
@@ -262,6 +263,7 @@ def query_vllm_payload(
     final_system_content = query_system_prompt.format(
         context=context,
         rephrased_query=rephrased_query or question,
+        max_tokens=max_new_tokens,
     )
     message_array.append({
         "role": "system",
--- a/services/common/llm_utils.py
+++ b/services/common/llm_utils.py
@@ -15,6 +15,38 @@ logger = get_logger("LLM")
 
 is_debug = logger.isEnabledFor(logging.DEBUG)
 
+def apply_token_buffer(max_tokens: int, token_buffer_ratio: float | None = None, context: str = "LLM") -> int:
+    """
+    Apply token buffer to give LLM breathing room to respect prompt word limits.
+    
+    The prompt instructs the LLM to limit to max_tokens, but we reduce the API limit
+    by token_buffer_ratio to ensure the LLM can naturally complete before hitting the hard limit.
+    
+    Args:
+        max_tokens: The maximum tokens specified in the prompt instruction
+        token_buffer_ratio: Buffer ratio to apply (0.0-0.5). If None, uses settings.llm.token_buffer_ratio
+        context: Context string for logging (e.g., "chatbot", "summarization")
+    
+    Returns:
+        Effective max tokens with buffer applied
+    
+    Example:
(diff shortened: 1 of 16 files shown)


ui/chatbot/src/components/customSendMessage.jsx
```diff
@@ -61,14 +61,7 @@ async function customSendMessage(
...
   function finalizeResponse(fullText) {
-    let trimmed = fullText.trim(); // to remove trailing white-space
-    // Define acceptable sentence-ending punctuation (both Hindi + English)
-    const validEndings = ['।', '.', '?', '!', '…']; // also includes ellipsis itself
-    const lastChar = trimmed.charAt(trimmed.length - 1);
-    if (!validEndings.includes(lastChar)) {
-      trimmed += ' ...';
-    }
-    return trimmed;
+    return fullText.trim();
```
sats-23, line 64: @Niharika0306 Are there any side effects to this change?

I plan on removing this as most answers in different languages have ... appended in the end even though they are grammatically stopped gracefully and within the token budget of the LLM

Niharika0306: Is the false ellipsis being added in case of english language as well ?

sats-23: Nope, happens only in other languages.

dharaneeshvrd: What about adding an instruction to end the answer with a full stop? Will it be too much?

Niharika0306: Okay then. I think this function was written keeping in mind only english and hindi language initially, but not sure how much work would it be extending to the other languages we support. 
For time being, if we are seeing a lot of false cases, I think we can get rid of the utility.

dharaneeshvrd: Yeah I think we can get rid of it, it's not really helping.

sats-23:
@dharaneeshvrd @Niharika0306 PTAL
The combination of removal of UI append combined with the token restriction in the query prompt seems to be stable.
I have tested this against all languages and found them to be performing with graceful stop (grammatically) in the end of every response.

services/common/llm_utils.py
```diff
@@ -194,8 +249,8 @@ def query_vllm_payload(
 
     # Calculate budget for context first (prioritize context over history)
     budget_for_context = llm_max_model_len - (
-        chatbot_settings.chatbot.initial_system_token_overhead +
-        chatbot_settings.chatbot.rag_system_token_overhead +
```
Niharika0306, line 198: so instead of using static values, we are now dynamically calculating it and checking?

Niharika0306 (commented):
Just asking to confirm - why was the language specific max_tokens not picked up till now? can you point in the code changes specifically?

services/common/llm_utils.py
sats-23: Yes, this will be more useful in custom-system prompt scenario where user can put a much larger prompt and can be accurately accounted for.

sats-23:
> Just asking to confirm - why was the language specific max_tokens not picked up till now? can you point in the code changes specifically?

Yes, the first [commit](https://github.com/IBM/project-ai-services/pull/955/changes/f55dedfd90fce8145d61a59db1691e82b8f5e875) in this PR

The max_tokens param had the English max_tokens passed as default, so within the app, the max_tokens was never null, therefore there was no language specific override happening. This was fixed

Niharika0306:
LGTM

dharaneeshvrd (dismissed):

dharaneeshvrd (approved):

PR #956, opened 2026-06-15
return proper errors codes from catalog API calls

Shubhamag12:
Refactored error handling to use ValidationError with proper HTTP status codes across API calls

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
+++ b/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
@@ -227,13 +227,13 @@ func (s *ApplicationService) UpdateApplication(ctx context.Context, id uuid.UUID
 	if app == nil {
 		return nil, &ValidationError{
 			Code:    http.StatusNotFound,
-			Message: "application does not exist",
+			Message: ErrMsgApplicationNotFound,
 		}
 	}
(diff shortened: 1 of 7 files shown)


yussufsh (dismissed):

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
```diff
@@ -491,7 +494,10 @@ func (s *ApplicationService) GetApplicationByID(ctx context.Context, id uuid.UUI
 		return nil, fmt.Errorf("failed to get application: %w", err)
 	}
 	if app == nil {
-		return nil, ErrApplicationNotFound
+		return nil, &ValidationError{
+			Code:    http.StatusNotFound,
+			Message: "application does not exist",
```
mayuka-c: Can we have move this to a constant?. Similarly for other messages we can have it in some file so we can reuse it

mayuka-c (commented):
Rest LGTM

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
Shubhamag12: done, ptal

yussufsh (dismissed):

yussufsh (approved):

PR #914, opened 2026-06-10
feat(catalog-ui): Digital Assistants API Integrations

pranithraoibm:
- Digital Assistants deploy flow UI
image
image
image

- Image links replaced with stored images
- API Integration for list, deploy and delete digital assistants and also the user profile API

PR #963, opened 2026-06-15
Design team feedback changes

pranithraoibm:
- Changes requested by the design team for the Digital Assistants page and the use cases page.
- Added validation for the watsonX fields
- Making all the params calls and storing it persistently in the store than in cache as they don't change and helps in better UX.
- Removed the hardcoded heading and description of Digital assistants and replaced them with API response
- Addressed both the review comments from [here](https://github.com/IBM/project-ai-services/pull/914/changes#pullrequestreview-4495297114) as well

New changes
- Removed images and its references completely except architecture diagram (Still a TODO as its WIP)
- Added Q&A service custom prompt functionality
- Fixed deletion API
- Changed name from `IBM AI Foundation for Power` to `AI Services` everywhere
- Added the `Why this model is good?` dropdown for each service
- Fixed the vLLM API key not getting passed in the deployment payload

Latest changes
- Added 15mins cache for deploy-options related API calls and 5mins for the resources API


ui/catalog/src/hooks/useProviderParams.ts
ryanmarc, line 1: This change completely removes the ability to invalidate the local cache, there needs to be a mechanism that allows for invalidating this data to pull fresh data from an API call if something does actually change in the api response.

ui/catalog/src/utils/inferenceComponentHelper.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    return false;
+  } catch {
+    // Schema might be a URL or invalid JSON, not a JSON schema
+    return false;
+  }
+}
+
+// Determines if component is an inference component (multiple providers with model input)
+export function isInferenceComponent(component: Component): boolean {
```
ryanmarc, line 36: `isInferenceComponent()` logic returns false when a component has `<= 1` provider, and it also returns `false` if provider schema is absent, invalid JSON, or a URL instead of inline JSON. The helper explicitly catches parse failures and treats them as non-inference.

That is dangerous because `buildDeploymentComponent()` now only uses the selected `inferenceBackend` and only merges service-level params when `isInferenceComp` is `true`. Before this PR, deployment transform treated `llm` and `reranker` as inference components directly by component type, so they always got the inference backend/service params path.

An LLM/reranker component with a single provider, a schema URL, or a schema shape that does not contain inline `model` can silently build the wrong deployment payload: selected inference backend ignored and WatsonX/service-level params omitted.

ui/catalog/src/hooks/useProviderParams.ts
pranithraoibm: The reason I moved away from cache was these API responses (like the deploy-options and params and schema calls) aren't something that change frequently, so invalidating them and pulling them again isn't really needed. Also I had set the cache time as 5mins, and if the user is on the deploy flow screen for more than 5mins the dropdown options go off (as the data is invalidated) and the user needs to close and open the deploy flow screen again to make the API calls. 

If cache approach is still needed then we might have to increase the cache time to 15mins or something for better UX. Let me know what should I do.

ui/catalog/src/utils/inferenceComponentHelper.ts
pranithraoibm:
```
export function isInferenceComponent(component: Component): boolean {
  // Primary: Generic detection based on schema (any provider with model input)
  const hasModelInput = component.providers.some(providerExpectsModelInput);
  if (hasModelInput) {
    return true;
  }

  // Fallback: Known inference component types for backward compatibility
  // This ensures LLM/reranker always work even with:
  // - Single provider
  // - URL schemas
  // - Schema parse failures
  if (component.type === "llm" || component.type === "reranker") {
    return true;
  }

  return false;
}
```

Planning to update it this way

pranithraoibm:
@ryanmarc I have updated the description with all the new changes added in the latest 2 commits. PTAL

ui/catalog/src/utils/deploymentTransform.ts
```diff
@@ -156,6 +158,153 @@ function buildDeploymentComponent(
...
+    "project_id",
+    "region",
+    "watsonx", // watsonx-specific params
+  ];
+
+  for (const [key, value] of Object.entries(allParams)) {
+    const lowerKey = key.toLowerCase();
+    const isInferenceBackendParam = inferenceBackendParamPatterns.some(
+      (pattern) => lowerKey.includes(pattern),
+    );
```
ryanmarc: This classification by lowerKey using only name doesn't take the actual provider schemas into account. It's possible for params to be classified incorrectly.

ui/catalog/src/utils/deploymentTransform.ts
```diff
@@ -156,6 +158,153 @@ function buildDeploymentComponent(
...
+    // Store params by provider + model combination
+    if (Object.keys(inferenceBackendParams).length > 0 && model) {
+      const key = `${serviceConfig.inferenceBackend}:${model}`;
+      // Merge params, later services override earlier ones
+      backendParamsMap[key] = {
+        ...(backendParamsMap[key] || {}),
+        ...inferenceBackendParams,
+      };
+    }
+  }
```
ryanmarc: Is it really intended for one service to override the params of another service simply because they are using the same model? What about a scenario where one service may use a model in vllm and another uses that model in a different provider?

ui/catalog/src/components/DeployFlow/components/ServiceConfigCard.tsx
```diff
@@ -514,6 +776,123 @@ export const ServiceConfigCard: React.FC<ServiceConfigCardProps> = ({
...
+                    if (hasValidationError) {
+                      setHasValidationError(false);
+                    }
+                    onUpdateConfig({
+                      params: {
+                        ...currentConfig?.params,
+                        ...params,
+                      },
+                    });
+                  }}
```
ryanmarc, line 799: This update looks like it spreads the current config back in. If things are removed it looks like they'd be re-added to the params.

ui/catalog/src/utils/deploymentTransform.ts
mayuka-c: Not sure, I didnt go through deep into implementation.
But heres what it should be

If you have multiple services like Chat, Digitize and Summarize using llm. If the models selected with given provider are same, then we will only deploy 1 instance of llm pod in backend. So we need to ensure that params (For Eg: VLLM API Key) remain the same as long as models are same for a given provider.

But if models are different or a provider is different across services then there is no such requirement as it will go deploy 2 instances of llm pods each could have its own params.

So models with provider selection determines whether single deployment or multi deployment

mayuka-c: So the logic should be if users have selected same model with same provider across services then ensure that params provided by user are same across both!

pranithraoibm: Sure - the previous implementation was silently overriding parameters. If a user entered different vLLM API keys for the same provider+model combination, the code would merge them and use the latest entered one.

I've updated the implementation to remove this merging behavior. Now the frontend passes the payload as-is with each service's configured parameters. The backend will validate that services sharing the same provider+model have identical parameters and throw an error if they don't. This way, users are explicitly aware of the conflict and can make the necessary corrections in the UI.

mayuka-c: Yes, this makes complete sense to do

pranithraoibm: Thank you @mayuka-c

pranithraoibm:
@mayuka-c could you please verify the changes made as per the review comments?

ui/catalog/src/components/DeployFlow/components/ServiceConfigCard.tsx
```diff
@@ -56,6 +159,53 @@ export const ServiceConfigCard: React.FC<ServiceConfigCardProps> = ({
...
   };
 
+  // Validate inference backend parameters
+  const validateInferenceBackendParams = (): boolean => {
+    if (!currentConfig?.inferenceBackend || !currentConfig?.params) {
+      return true; // No params to validate
+    }
+
+    const componentType = llmComponent ? "llm" : "reranker";
+    const paramsMap = providerParamsByType[componentType]?.paramsMap || {};
```
mayuka-c, line 170: Do we need to hardcode llm and reranker?
Maybe add a TODO and keep this check in a single file and reuse. We may need to revisit in next release I feel.

I feel better would be to keep llm and reranker as constants in a file and then reference everywhere. In this way it  will be easier to know

ui/catalog/src/components/DeployFlow/components/ServiceConfigCard.tsx
```diff
@@ -305,6 +467,161 @@ export const ServiceConfigCard: React.FC<ServiceConfigCardProps> = ({
...
+                    )}
+                  </span>
+                </div>
+              );
+            })}
+
+          {/* Model Description Accordion - Visible in both view and edit modes */}
+          {(llmComponent || rerankerComponent) &&
+            (() => {
+              const componentType = llmComponent ? "llm" : "reranker";
```
mayuka-c, line 547: Same here too

ui/catalog/src/components/DeployFlow/components/ServiceConfigCard.tsx
```diff
@@ -449,11 +770,126 @@ export const ServiceConfigCard: React.FC<ServiceConfigCardProps> = ({
...
+                  }}
+                  hasValidationError={hasValidationError}
+                />
+              </div>
+            )}
+
+            {/* Model Description Accordion - In edit mode */}
+            {(llmComponent || rerankerComponent) &&
+              (() => {
+                const componentType = llmComponent ? "llm" : "reranker";
```
mayuka-c, line 813: Here as well

mayuka-c: Isnt this the same as above, why do we have same code here too (either in edit or apply it should be same right)?

ui/catalog/src/utils/deploymentTransform.ts
```diff
@@ -161,6 +242,32 @@ export function transformToDeploymentPayload(
...
     }
 
+    // Find the component type that uses the inference backend (llm or reranker)
+    let componentType = "llm";
+    const hasReranker = serviceDefinition.components.some(
+      (c) => c.type === "reranker",
+    );
+    if (hasReranker && serviceConfig.components?.reranker) {
+      componentType = "reranker";
+    }
```
mayuka-c, line 253: Again here, I think I had maybe added a comment somewhere on how to get this without even having to know type.

But anyway as mentioned above we need to add TODOs otherwise we will miss

mayuka-c (dismissed):
Rest LGTM, please add TODOs whereever we see `llm` and `reranker` referenced
Approving this

ui/catalog/src/utils/inferenceComponentHelper.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  // Fallback: Known inference component types for backward compatibility
+  // This ensures LLM/reranker always work even with:
+  // - Single provider
+  // - URL schemas
+  // - Schema parse failures
+  if (component.type === "llm" || component.type === "reranker") {
+    return true;
+  }
```
mayuka-c, line 51: Here as well add a TODO and let us use constants for llm and reranker and use it across.

It will be easier in next release to remove and also know where it is used

pranithraoibm:
@ryanmarc everything is addressed, can we merge this?

ryanmarc (approved):

ui/catalog/src/hooks/useProviderParams.ts
ryanmarc: Without a way to ever invalidate the cache, these live forever. If new things are added, they'll never be fetched from the API. How will the cache be invalidated when it actually needs to be?

ui/catalog/src/utils/inferenceComponentHelper.ts
ryanmarc: This was marked as resolved, but no change was actually made.

ryanmarc (dismissed):

ui/catalog/src/hooks/useProviderParams.ts
pranithraoibm: When the user logs out the stored is cleared, if this isn't ok then I'll add a cache of 15mins?

ui/catalog/src/utils/inferenceComponentHelper.ts
pranithraoibm: Changes are present here 
https://github.com/IBM/project-ai-services/pull/963/changes#diff-7064249aad45d30cbf983a55ab5a05a715bc28dc4374f55c008e75105124c5d1R36

ui/catalog/src/hooks/useProviderParams.ts
pranithraoibm: Added 15mins cache

mkumatag (approved):

PR #966, opened 2026-06-15
feat (catalog UI) : Services Deploy Flow

Niranjan-J-Joshi:
- Services Deploy Flow
image

image

image

- Integrate Service Details side panel with API's 

image


ui/catalog/src/components/ServicesDeployFlow/ServicesDeployFlow.tsx
```diff
@@ -0,0 +1,528 @@
...
+
+    const serviceConfig = state.formData.services[state.selectedServiceId];
+    const llmComponent = serviceConfig?.components?.llm;
+
+    if (!llmComponent?.providerId) {
+      return true; // If no LLM provider selected, allow proceeding
+    }
+
+    // Get the provider schema for the selected LLM provider
+    const schemaKey = `${state.selectedServiceId}-llm-${llmComponent.providerId}`;
```
ryanmarc: This builds the keys with `-` but the provider schemas are cached and read everywhere else with `:` separated keys such as serviceId:componentType:providerId. The store’s key helper joins parts with ":", and StepOne reads provider schemas using the colon format. Because of that mismatch, providerSchemas[schemaKey] will usually be undefined; the validation then returns true when no schema is found, which means the final Deploy action can be enabled even when required LLM credentials are missing.

ui/catalog/src/components/ServicesDeployFlow/steps/StepOne.tsx
```diff
@@ -0,0 +1,287 @@
...
+    // Get all component types that exist in the service
+    const serviceComponentTypes = Object.keys(serviceConfig.components);
+
+    // Find matching components from deployOptions to get provider info and display names
+    // Filter out llm and reranker - they belong in Step 2 (advanced configuration)
+    return deployOptions.components
+      ?.filter(
+        (component) =>
+          serviceComponentTypes.includes(component.type) &&
+          ["version", "vector_store", "embedding"].includes(component.type),
```
ryanmarc: This filter contradicts the API-driven behavior described above. We initialize all components returned by deployOptions, and the comment says Step 1 should show everything except llm/reranker, but this allowlist only renders version/vector_store/embedding. That can hide future required service components while still submitting them with defaults. Please invert this filter to exclude only llm/reranker.

`serviceComponentTypes.includes(component.type) &&
!["llm", "reranker"].includes(component.type)`

ui/catalog/src/components/ServicesDeployFlow/steps/StepOne.tsx
```diff
@@ -0,0 +1,287 @@
...
+              params: {
+                ...currentComponent.params,
+                model, // Set the model parameter
+              },
+            },
+          },
+        },
+      },
+    });
+  };
```
ryanmarc, line 170: This loses the provider/model relationship. For model-backed components, the UI only updates params.model and keeps the existing providerId, while fetchComponentModelsWithSchemas also dedupes model options by model id and can discard other provider mappings. That can submit a provider_id that does not actually support the selected model. Please preserve model/provider pairs and update providerId + params.model together, or require provider + model selection as a paired choice.

ryanmarc (changes requested):

ui/catalog/src/utils/serviceDeploymentTransform.ts
```diff
@@ -0,0 +1,225 @@
...
+    if (value !== undefined && value !== null && value !== "") {
+      merged[key] = value;
+    }
+  }
+
+  // CRITICAL FIX: vector_store components should never have a model parameter
+  // Remove it even if it comes from schema defaults or user values
+  if (componentType === "vector_store" && "model" in merged) {
+    delete merged.model;
+  }
```
mayuka-c: Do we need this?
Whatever you get from deploy-options thats final

mayuka-c (dismissed):
Couldnt go through completely due to shortage of time but Overall LGTM
Pls address all comments

ui/catalog/src/components/ServicesDeployFlow/steps/StepOne.tsx
```diff
@@ -0,0 +1,294 @@
...
+  Column,
+  Toggletip,
+  ToggletipButton,
+  ToggletipContent,
+} from "@carbon/react";
+import { Information } from "@carbon/icons-react";
+import styles from "../ServicesDeployFlow.module.scss";
+import type { StepProps } from "../types";
+import { useMemo } from "react";
+import { useServiceDeployStore } from "@/store/serviceDeploy.store";
```
mayuka-c, line 14: Are we planning to combine all these steps later?. Because all these are also present in Digital Asistants flow and almost all remains the same

Niranjan-J-Joshi: Yes , In next releases we will be working on creating one reusable component and use that in both digital assistants and services.

ui/catalog/src/utils/serviceDeploymentTransform.ts
Niranjan-J-Joshi: Yes, removed this condition and made change in deployment.api.ts file

ui/catalog/src/components/ServicesDeployFlow/steps/StepOne.tsx
Niranjan-J-Joshi: changed filter to  !["llm", "reranker"]

ui/catalog/src/components/ServicesDeployFlow/ServicesDeployFlow.tsx
Niranjan-J-Joshi: Updated the schema key to use : separators, matching the provider schema cache

ui/catalog/src/components/ServicesDeployFlow/steps/StepOne.tsx
ryanmarc: I don't see a change for this - was this addressed?

ryanmarc (dismissed):

ui/catalog/src/components/ServicesDeployFlow/steps/StepOne.tsx
Niranjan-J-Joshi: Addressed this request in below commit
[Implement review comments](https://github.com/IBM/project-ai-services/pull/966/changes/54c615172a4851dfcc5dba36a75fc238e9fb8f90)

mkumatag (approved):
Merging the code as is for now and deferring the reviews later

PR #976, opened 2026-06-16
fix: move sentient group and ulimit values to configure

Shubhamag12:
LLM, reranker, and embedding pods were failing to start with errors:
```
crun: controller 'cpu' is not available (cgroup error)
crun: setrlimit 'RLIMIT_NOFILE': Operation not permitted (ulimit error)
```

**Root Cause:** The sentient group and ulimit configurations were only created during Spyre-specific setups causing failures in non-Spyre setups.

**Solution:** Moved sentient group creation and ulimit configurations from spyre.gpo code to the podman configure, ensuring they're configured for both spyre and non-spyre environments.

### DIFF
--- a/ai-services/internal/pkg/validators/podman/ulimits/ulimits.go
+++ b/ai-services/internal/pkg/validators/podman/ulimits/ulimits.go
@@ -26,7 +26,7 @@ func (r *UlimitsRule) Description() string {
 }
 
 func (r *UlimitsRule) Verify() error {
-	logger.Infoln("Validating ulimit configurations", logger.VerbosityLevelDebug)
+	logger.Debugln("Validating ulimit configurations")
 
 	// Validate memlock configuration
 	if err := r.validateMemlockConf(); err != nil {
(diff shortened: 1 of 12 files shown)


ai-services/internal/pkg/validators/podman/ulimits/ulimits.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return fmt.Errorf("memlock configuration not found or invalid in %s", memlockConfFile)
+}
+
+func (r *UlimitsRule) validateNofileConf() error {
+	const (
+		nofileConfFile   = "/etc/security/limits.conf"
+		minNofileLimit   = 134217728
+		nofileFieldCount = 4
+	)
```
mayuka-c: I see all these are again being defined and used in configureNoLimit method too. Can we have the constant defined at global

mayuka-c (commented):
@Shubhamag12 Are most of these methods refactoring/cleanups?
If we have retained the existing functionality (which we have already reviewed), then its fine (as I'm not going through the method implementation again)

ai-services/internal/pkg/validators/podman/ulimits/ulimits.go
Shubhamag12: done

Shubhamag12:
> Are most of these methods refactoring/cleanups?

yes
We were doing mixed operations in spyre/repair.go
I have moved non-spyre related to configure.go and created validation of each configuration in `validators`.

mayuka-c (approved):
LGTM

PR #977, opened 2026-06-16
[Catalog] Use common DB helper methods

mayuka-c:
- Move DB helper methods to application_utils.go.
- Use it across the entire Application Flow where DB insert/update is done.
- Handle errors in deploy.go wherever not handled


yussufsh (approved):
lgtm

PR #983, opened 2026-06-17
remove experimental flags and use new way of deploying application

Shubhamag12:
This PR removes all experimental flags from application commands and migrates catalog API logic to the implementation layer, making it the default behavior for Podman runtime.


Shubhamag12:
Closing this

PR #985, opened 2026-06-17
Enable deploy button in DA deploy flow

pranithraoibm:
- Enabled the deploy button in the digital assistants even when the resources are insufficient and let backend manage the error handling.
- Added `Insufficient resources available` to the yellow warning symbol in the resources cards.

image

### DIFF
--- a/ai-services/assets/catalog/podman/values.yaml
+++ b/ai-services/assets/catalog/podman/values.yaml
@@ -1,6 +1,6 @@
 ui:
   port: ""
-  image: icr.io/ai-services-cicd/catalog-ui:v0.0.16
+  image: icr.io/ai-services-cicd/catalog-ui:v0.0.17
 
 backend:
   port: ""
--- a/ui/catalog/Makefile
+++ b/ui/catalog/Makefile
@@ -1,6 +1,6 @@
 REGISTRY ?= icr.io/ai-services-private
 IMAGE ?= catalog-ui
-TAG ?= v0.0.16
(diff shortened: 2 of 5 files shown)


mayuka-c (commented):
Just a query,
Shouldnt this be only for cores?
Do we want to not restrict for memory and spyre?

pranithraoibm:
In the spyre cards case the backend doesn't allow the deployment to happen when there is insufficient cards, memory I'm not sure. What should be done for that?

mayuka-c:
> In the spyre cards case the backend doesn't allow the deployment to happen when there is insufficient cards, memory I'm not sure. What should be done for that?

I feel for spyre we can block it in the UI itself right, instead of call even coming to backend and failing?

pranithraoibm:
@dharaneeshvrd Any suggestion what should be done wrt memory? I can relax the check on cores and still disable the deploy button when there is insufficient resources for memory and spyre but will need hardcoding like `resource.type !== "cpu"`

mkumatag:
> @dharaneeshvrd Any suggestion what should be done wrt memory? I can relax the check on cores and still disable the deploy button when there is insufficient resources for memory and spyre but will need hardcoding like `resource.type !== "cpu"`

let us relax only for the cpu and other resources we still can be strict for now, we will revisit this entire flow before the GA.

mkumatag (approved):
Merging the code as is for now and revisit this before the GA to decide what should be the soft vs hard limitation for the deployment.

PR #987, opened 2026-06-17
Make catalog API implementation default, move old implementation behind --legacy flag

Shubhamag12:
Replaced `--experimental` flag with --legacy flag across all application commands (create, delete, info, logs, ps, start, stop). The catalog API implementation is now the default for Podman runtime, with the previous implementation available via `--legacy` flag for backward compatibility.


mayuka-c (dismissed):
LGTM

prachikurade1 (approved):
Approved

PR #989, opened 2026-06-17
Fix digitize UI duration and timestamps

Niharika0306:
The database schema used TIMESTAMP (without WITH TIME ZONE) for all timestamp columns. When timezone-aware datetime objects were stored in PostgreSQL:

1. PostgreSQL stripped the timezone information and stored timestamps as local time
2. Upon retrieval, SQLAlchemy returned timezone-naive datetime objects
3. When serialized to ISO 8601 format, these timestamps lacked the Z (UTC) suffix
4. JavaScript in the frontend interpreted these timestamps as local timezone instead of UTC
5. This caused incorrect duration calculations, off by the timezone offset (e.g., 5.5 hours for IST)

The full issue is briefed here: https://jsw.ibm.com/browse/AISERVICES-1249

Screenshot 2026-06-17 at 5 24 16 PM

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.113
+TAG?=v0.0.114
 CONTAINER_BUILDER?=podman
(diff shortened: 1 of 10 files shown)


mkumatag:
Please bump the image version and also mention the jira item somewhere for the reference, otherwise good to merge.

mkumatag (approved):
LGTM

PR #994, opened 2026-06-18
Update service-base image to v0.0.33 and increment dependent service versions

mkumatag:
## Summary
This PR updates all services to use the latest service-base image (v0.0.33) and increments the version numbers of all dependent services accordingly.

## Changes
- Updated all service Containerfiles to use `service-base:v0.0.33` (from v0.0.32)
- Incremented service versions in Makefiles:
  - chatbot-service: v0.0.18 → v0.0.19
  - digitize-service: v0.0.25 → v0.0.26
  - similarity-service: v0.0.11 → v0.0.12
  - summarize-service: v0.0.12 → v0.0.13
- Updated all asset values.yaml files with new service versions across:
  - Individual service assets
  - RAG application assets (podman and openshift variants)

## Files Modified
- Service Containerfiles (4 files)
- Service Makefiles (4 files)
- Asset configuration files (9 files)

## Testing - expect to be covered by the test team during the test
- [ ] Build and test all services with updated base image
- [ ] Verify asset deployments with new versions

PR #995, opened 2026-06-18
Bump up service-base image to v0.0.33

adarshagrawal38:


mkumatag:
Already created in #994

PR #999, opened 2026-06-18
UI Enhancements and Default Initialization Updates

Niranjan-J-Joshi:
This PR contains updates to the Details page, default value initialization, and validation enhancements.

- Updated the tag display to dynamically show the appropriate type:
- Digital Assistant for assistant resources
- Service Name for service resources
- Display the IBM Certified badge only for IBM-certified services.
- Added Accelerators information to the Allocated Resources card.
- Adjusted the width of the Cores and Memory fields for improved layout consistency.

Screenshot 2026-06-18 at 10 58 45 PM
image

- Dynamically initialize the LLM Model and Inference Backend based on the configured default flag.
- Added validation to display error messages when required fields are left empty.

image

### DIFF
--- a/ui/catalog/src/components/DeployedServicesTable/CellRenderers.tsx
+++ b/ui/catalog/src/components/DeployedServicesTable/CellRenderers.tsx
@@ -1,4 +1,4 @@
-import { Tag, OverflowMenu, OverflowMenuItem } from "@carbon/react";
+import { Tag, OverflowMenu, OverflowMenuItem, Link } from "@carbon/react";
 import {
   Delete,
   CheckmarkFilled,
@@ -88,7 +88,7 @@ export const ActionCell = ({ rowId, dispatch, rowData }: CellRendererProps) => {
 };
 
 export const NameCell = ({ value }: CellRendererProps) => (
-  <span style={{ fontWeight: 500 }}>{String(value)}</span>
+  <Link style={{ fontWeight: 500 }}>{String(value)}</Link>
 );
 export const StatusCell = ({ value }: CellRendererProps) => {
   const status = String(value);
--- a/ui/catalog/src/components/DeployedServicesTable/CellRenderers.tsx
+++ b/ui/catalog/src/components/DeployedServicesTable/CellRenderers.tsx
@@ -10,6 +10,7 @@ import type { Dispatch } from "react";
 import type { AppAction } from "./types";
 import { ACTION_TYPES } from "./types";
 import styles from "./DeployedServices.module.scss";
+import type { DeploymentDetails } from "@/types/digitalAssistants";
 
 // Status configuration
 const STATUS_CONFIG = {
@@ -57,7 +58,8 @@ interface CellRendererProps {
   value: unknown;
   rowId: string;
   dispatch: Dispatch<AppAction>;
-  rowData?: { status: string };
+  rowData?: { status: string; name?: string; type?: string };
+  onRowClick?: (deployment: DeploymentDetails) => void;
 }
 
 export const ActionCell = ({ rowId, dispatch, rowData }: CellRendererProps) => {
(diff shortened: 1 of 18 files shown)


ui/catalog/src/services/auth.ts
```diff
@@ -48,6 +49,9 @@ export const logout = async () => {
 
   // Clear all deploy store data
   useDeployStore.getState().clearAll();
+
+  // Clear all service deploy store data
+  useServiceDeployStore.getState().clearAllCache();
```
ryanmarc: Local cleanup should not depend on the remote logout request succeeding. If /logout fails or times out, this code skips clearTokens, clearUserInfo, clearAll, and clearAllCache, leaving stale local state behind. Please move local clearing into a finally block.

ui/catalog/src/components/DeploymentDetails/DeploymentDetails.tsx
```diff
@@ -166,16 +166,32 @@ const DeploymentDetails = ({
...
+              deploymentServices.every(
+                (service) =>
+                  serviceMetadataById[service.catalog_id]?.certifiedBy ===
+                  "IBM",
+              )
+            : deploymentServices.some(
+                (service) =>
+                  serviceMetadataById[service.catalog_id]?.certifiedBy ===
+                  "IBM",
+              );
```
ryanmarc: is it intentional to show as IBM certified when only "some" items of a service other than "Digital Assistant" have certified by IBM?

ui/catalog/src/components/DeployedServicesTable/CellRenderers.tsx
```diff
@@ -88,7 +88,7 @@ export const ActionCell = ({ rowId, dispatch, rowData }: CellRendererProps) => {
 };
 
 export const NameCell = ({ value }: CellRendererProps) => (
-  <span style={{ fontWeight: 500 }}>{String(value)}</span>
+  <Link style={{ fontWeight: 500 }}>{String(value)}</Link>
```
ryanmarc: This changes the service name from static text to a link-looking element without adding any navigation target or click handler. That creates misleading UI and likely an accessibility issue. Please either wire this an action or make it not a link.

ui/catalog/src/components/ServicesDeployFlow/ServicesDeployFlow.tsx
```diff
@@ -236,14 +236,18 @@ export const ServicesDeployFlow = ({
       // Get component models from store
       const componentKey = `${state.selectedServiceId}:${component.type}`;
       const models = componentModels[componentKey] || [];
+      const selectedProviderId = componentConfig.providerId;
+      const matchingModelForProvider =
+        models.find((model) => model.providerId === selectedProviderId) ||
+        models[0];
```
ryanmarc: This fallback can create an invalid provider/model combination. If no model matches the selected/default provider, falling back to models[0] may select a model owned by another provider while keeping the current providerId. Please only set params.model when the model belongs to the selected provider, or switch the provider to match the fallback model’s provider. Otherwise the generated deployment payload can be internally inconsistent.

ryanmarc (changes requested):

ui/catalog/src/services/auth.ts
Niranjan-J-Joshi: I've moved the local cleanup (clearTokens, clearUserInfo, clearAll, and clearAllCache) into a finally block so it runs regardless of whether the /logout request succeeds, fails, or times out.

ui/catalog/src/components/DeploymentDetails/DeploymentDetails.tsx
Niranjan-J-Joshi: Yes. For deployment types other than Digital Assistant, there will be  only a single service in deploymentServices.  I replaced the .some() check with a direct lookup on the first service

ui/catalog/src/components/DeployedServicesTable/CellRenderers.tsx
Niranjan-J-Joshi: I added row click event for this link, making the behavior consistent with the Digital Assistants table

ui/catalog/src/components/ServicesDeployFlow/ServicesDeployFlow.tsx
Niranjan-J-Joshi: I've removed the fallback to models[0]. We now only set params.model when a valid model exists for the selected provider

ryanmarc (dismissed):

ryanmarc (approved):

PR #1000, opened 2026-06-18
Deploy flow options bug fix

pranithraoibm:
- Preserved the cache only for deploy-options API and removed them for others as it was resulting in race condition and the dropdown values were getting messed up. 
- Fixed the old values get preserved on the view mode of deploy options even after removal by the user.

### DIFF
--- a/ui/catalog/src/components/DeployFlow/components/ServiceConfigCard.tsx
+++ b/ui/catalog/src/components/DeployFlow/components/ServiceConfigCard.tsx
@@ -696,8 +696,12 @@ export const ServiceConfigCard: React.FC<ServiceConfigCardProps> = ({
                           if (hasValidationError) {
                             setHasValidationError(false);
                           }
+                          // Merge provider params with existing params to preserve service-level params
                           onUpdateConfig({
-                            params,
+                            params: {
+                              ...currentConfig?.params,
+                              ...params,
+                            },
                           });
                         }}
                         providerParamsMap={
@@ -726,8 +730,12 @@ export const ServiceConfigCard: React.FC<ServiceConfigCardProps> = ({
                     if (hasValidationError) {
                       setHasValidationError(false);
                     }
+                    // Merge service-level params with existing params to preserve provider params
                     onUpdateConfig({
-                      params,
+                      params: {
+                        ...currentConfig?.params,
+                        ...params,
+                      },
                     });
                   }}
                   providerParamsMap={{
--- a/ui/catalog/src/components/DeployFlow/components/ServiceConfigCard.tsx
+++ b/ui/catalog/src/components/DeployFlow/components/ServiceConfigCard.tsx
@@ -696,12 +696,34 @@ export const ServiceConfigCard: React.FC<ServiceConfigCardProps> = ({
                           if (hasValidationError) {
                             setHasValidationError(false);
                           }
-                          // Merge provider params with existing params to preserve service-level params
-                          onUpdateConfig({
-                            params: {
-                              ...currentConfig?.params,
-                              ...params,
+                          // Get provider schema keys to know which params belong to provider
+                          const providerSchema =
+                            providerParamsByType[componentType]?.paramsMap?.[
+                              fieldValue || ""
+                            ];
+                          const providerKeys = new Set(
+                            providerSchema?.properties
+                              ? Object.keys(providerSchema.properties)
+                              : [],
+                          );
+
+                          // Preserve service-level params, update only provider params
+                          const mergedParams: Record<string, unknown> = {};
+                          Object.entries(currentConfig?.params || {}).forEach(
+                            ([key, value]) => {
+                              // Keep service-level params (not in provider schema)
+                              if (!providerKeys.has(key)) {
+                                mergedParams[key] = value;
+                              }
                             },
+                          );
+                          // Add/update provider params from DynamicSchemaFields
+                          Object.entries(params).forEach(([key, value]) => {
+                            mergedParams[key] = value;
+                          });
+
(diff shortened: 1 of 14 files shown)


ui/catalog/src/store/deploy.store.ts
ryanmarc, line 1: This makes catalog metadata effectively permanent in persisted deploy storage. `architectures`, `serviceSummaries`, `architectureDetails`, `providerParams`, and `serviceParams` are persisted, but the staleness checks were removed, so users can keep stale schemas/dropdown data across backend catalog changes or UI releases until localStorage is cleared. Can we keep TTL/versioned invalidation for catalog metadata and fix the race by de-duping in-flight fetches or scoping cache keys instead of removing invalidation globally?

ui/catalog/src/hooks/useServiceParams.ts
```diff
@@ -10,28 +10,25 @@ interface UseServiceParamsResult {
...
   useEffect(() => {
     // Don't fetch if no serviceId
     if (!serviceId) {
       return;
     }
 
-    // Fetch if we don't have data or if cache is stale, and we haven't already started fetching
-    if ((params && !isStale) || hasFetched.current) {
+    // Fetch only if we don't have data and haven't already started fetching
+    if (params || hasFetched.current) {
```
ryanmarc: `hasFetched` is global to this hook instance, not to the provider/service key. After one fetch completes, switching to another provider or service with no cached params can leave `params` null while `hasFetched.current === true`, causing the effect to return without fetching the new key. Can we key this ref by `componentType:providerId` / `componentType:serviceId`, reset it when the key changes, or replace it with an in-flight request map?

ui/catalog/src/components/DeployFlow/components/ServiceConfigCard.tsx
```diff
@@ -677,10 +697,7 @@ export const ServiceConfigCard: React.FC<ServiceConfigCardProps> = ({
                             setHasValidationError(false);
                           }
                           onUpdateConfig({
-                            params: {
-                              ...currentConfig?.params,
-                              ...params,
-                            },
+                            params,
```
ryanmarc: This replacement is risky because inference-backend params and service-level params both live in `currentConfig.params`. The backend dropdown change above preserves service-level keys when switching backend, but the next provider-field edit calls `onUpdateConfig({ params })` and can drop those preserved service-level keys unless `DynamicSchemaFields` always emits the complete merged params object. Can we preserve service-level keys when provider params change, and preserve provider keys when service params change, instead of replacing the whole params object?

ryanmarc (changes requested):

ui/catalog/src/hooks/useDeployOptions.ts
```diff
@@ -35,17 +37,15 @@ export const useDeployOptions = () => {
...
-      hasFetched.current = true;
+    // Fetch if we don't have data or if cache is stale
+    // dedupe() handles preventing duplicate in-flight requests
+    if ((!deployOptions || isStale) && !deployOptionsLoading) {
       setDeployOptionsLoading(true);
       setDeployOptionsError(null);
 
-      fetchDeployOptions(selectedArchitectureId)
+      // Use dedupe to prevent simultaneous requests
+      const requestKey = `deployOptions:${selectedArchitectureId}`;
```
ryanmarc, line 55: `useDeployOptions` keys the in-flight request by `selectedArchitectureId`, but the persisted cache is still a single global `deployOptions` value and timestamp. After deploy options are fetched for architecture A, switching to architecture B leaves `deployOptions` truthy and `isDeployOptionsStale()` false, so this guard will skip fetching B’s options until the TTL expires. Can we key `deployOptions` / `deployOptionsFetchedAt` by architecture ID, or store the architecture ID with the cached options and treat a mismatch as stale?

ui/catalog/src/components/DeployFlow/components/ServiceConfigCard.tsx
```diff
@@ -696,8 +696,12 @@ export const ServiceConfigCard: React.FC<ServiceConfigCardProps> = ({
                           if (hasValidationError) {
                             setHasValidationError(false);
                           }
+                          // Merge provider params with existing params to preserve service-level params
                           onUpdateConfig({
-                            params,
+                            params: {
+                              ...currentConfig?.params,
+                              ...params,
+                            },
```
ryanmarc: I commented on this in a previous PR and this change reintroduces the problem.

This merge can reapply params that `DynamicSchemaFields` intentionally deleted. For example, when a UI-only controlling checkbox is unchecked, the controlled field is removed from the object passed to `onChange`; spreading `...currentConfig?.params` before `...params` brings that deleted key back because omission cannot express deletion. Can we either have `DynamicSchemaFields` send a patch with explicit deleted keys, or merge only within the active schema key set and remove keys that are absent from the emitted params?

ui/catalog/src/components/DeployFlow/components/ServiceConfigCard.tsx
```diff
@@ -726,8 +730,12 @@ export const ServiceConfigCard: React.FC<ServiceConfigCardProps> = ({
                     if (hasValidationError) {
                       setHasValidationError(false);
                     }
+                    // Merge service-level params with existing params to preserve provider params
                     onUpdateConfig({
-                      params,
+                      params: {
+                        ...currentConfig?.params,
+                        ...params,
+                      },
```
ryanmarc: Same thing here.

ryanmarc (changes requested):

ryanmarc (approved):

PR #1001, opened 2026-06-19
Make target to update the image tag in values.yaml files

adarshagrawal38:
### Summary
Automated solution for updating container image references in `values.yaml` files and updating Makefile TAGs, when files are modified for the services, which is eliminating manual updates across 12 components and 44+ files.

### DIFF
--- a/hack/update_image_tags.py
+++ b/hack/update_image_tags.py
@@ -1,0 +1,301 @@
+#!/usr/bin/env python3
+"""
+Updates image tags in values.yaml files when a Makefile is modified.
+
+This script reads the IMAGE and TAG values from a specified Makefile and updates
+all corresponding values.yaml files based on the COMPONENTS mapping from the
+check_image_names.py script.
+
+Usage:
+    python hack/update_image_tags.py --makefile services/chatbot/Makefile --registry icr.io/ai-services-cicd
+    python hack/update_image_tags.py -m services/digitize/Makefile -r icr.io/ai-services-cicd
+"""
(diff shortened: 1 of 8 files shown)


adarshagrawal38:
Example: 
Makefile updated for chatbot-service

image



`Values updated in yaml file for chanbot-service.`
Screenshot 2026-06-19 at 5 51 59 PM

hack/update_image_tags.py
```diff
@@ -0,0 +1,538 @@
...
+import subprocess
+import sys
+from pathlib import Path
+from typing import Dict, List, Tuple, Optional
+
+# Map of: makefile_path -> list of (values_yaml_path, values_key)
+# This should match the COMPONENTS mapping in .github/scripts/check_image_names.py
+COMPONENTS = {
+    "services/chatbot/Makefile": [
+        ("ai-services/assets/applications/rag/podman/values.yaml", "backend"),
```
yussufsh: This will be hard to maintain.
You are using such logic in `.github/scripts/check_image_names.py`

adarshagrawal38:
Bob response of version bump:
image

hack/update_image_tags.py
adarshagrawal38: I have extracted the components in common.py, and keeping the logic of updating values.yaml file in different file.

adarshagrawal38:
Closing of PR as updating tags and values.yaml file is getting complicated via make target.

PR #1002, opened 2026-06-19
fix(catalog-ui): improve authentication routing and error handling

pranithraoibm:
- Renamed ProtectedRoute → AuthRoute with requireAuth prop
- /login redirects authenticated users to /digital-assistants
- Logout always clears tokens (even if API fails)
- Login shows different errors: inline for credentials, toast for network
- Logout shows toast if API fails
- AppHeader just navigates to /logout (no redundant logout call)


ui/catalog/src/pages/Login/Login.tsx
```diff
@@ -33,15 +37,31 @@ const LoginPage = () => {
...
       navigate(ROUTES.DIGITAL_ASSISTANTS);
-    } catch {
-      setError(true);
+    } catch (error) {
+      // Check if it's a 401 Unauthorized (wrong credentials)
+      if (axios.isAxiosError(error) && error.response?.status === 401) {
+        setCredentialError(true);
+      } else {
+        // Network error or other server errors
+        setNetworkError(true);
```
ryanmarc: This only treats 401 as invalid credentials. If the backend can return 400, 403, or another validation/auth status for bad credentials, users will see a misleading “Network error.” Perhaps we should classify all expected auth-failure statuses as credential errors and reserve the toast for connectivity/5xx failures.

ryanmarc (dismissed):

Niranjan-J-Joshi (approved):

PR #1006, opened 2026-06-22
[DB-Sync] Fix missing multi pods and add syncing other resources like secrets, volumes, pods missing

mayuka-c:
Fixes: https://jsw.ibm.com/browse/AISERVICES-1270 and also fixes: https://jsw.ibm.com/browse/AISERVICES-1289

- For a given service/component, if there are multiple pods, earlier syncing was done to only 1 pod. Now all the pods are synced and even if any pod is in error, then overall service/component/application status will be error.
- Also added syncing for missing pods, secrets and volumes and update the status appropriately.

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/services/sync/sync_service.go
+++ b/ai-services/internal/pkg/catalog/apiserver/services/sync/sync_service.go
@@ -5,14 +5,17 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	texttemplate "text/template"
 	"time"
 
 	"github.com/google/uuid"
+	catalogpkg "github.com/project-ai-services/ai-services/internal/pkg/catalog"
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog/db/models"
 	dbrepo "github.com/project-ai-services/ai-services/internal/pkg/catalog/db/repository"
 	catalogutils "github.com/project-ai-services/ai-services/internal/pkg/catalog/utils"
 	"github.com/project-ai-services/ai-services/internal/pkg/constants"
 	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+	modelpkg "github.com/project-ai-services/ai-services/internal/pkg/models"
 	"github.com/project-ai-services/ai-services/internal/pkg/runtime"
 	"github.com/project-ai-services/ai-services/internal/pkg/runtime/common"
 	"github.com/project-ai-services/ai-services/internal/pkg/vars"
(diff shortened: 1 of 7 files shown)


ai-services/internal/pkg/catalog/apiserver/services/sync/sync_service.go
```diff
@@ -47,13 +67,20 @@ func NewSyncService(
 		syncInterval = DefaultSyncInterval
 	}
 
+	catalogProvider, err := catalogpkg.NewCatalogProvider()
+	if err != nil {
+		logger.Errorf("Failed to create catalog provider for sync service: %v", err)
```
Shubhamag12: are we not returning error here?

ai-services/internal/pkg/catalog/apiserver/services/sync/sync_service.go
```diff
@@ -291,44 +358,90 @@ func (s *SyncService) syncComponentPod(ctx context.Context, rt runtime.Runtime,
...
 
-	// Determine component status based on pod health
-	newStatus, message := s.determineComponentStatus(pod)
+	return newStatus, "", nil
+}
+
+// handleComponentPodFetchError handles the case when pods cannot be fetched for a component.
+func (s *SyncService) handleComponentPodFetchError(ctx context.Context, component *models.Component, componentID uuid.UUID, fetchErr error) (models.ComponentStatus, string, error) {
+	newStatus := models.ComponentStatusError
+	message := fmt.Sprintf("Pod not found or error: %v", fetchErr)
```
Shubhamag12: can we store this in a constant?
I see it being used here as well: https://github.com/IBM/project-ai-services/pull/1006/changes#diff-6783f0a6e60373a48b28bbfbf5498bc5b5991c3f5554f2b66f6bab446551fb03R299

Shubhamag12 (commented):
minor comments

ai-services/internal/pkg/catalog/apiserver/services/sync/sync_service.go
mayuka-c: If we return error then we need to handle in caller which is during pod startup.
I'm thinking maybe better to return error and there we can panic as sync is really important. I will make the update

ai-services/internal/pkg/catalog/apiserver/services/sync/sync_service.go
mayuka-c: Done

Shubhamag12 (approved):
lgtm

PR #1007, opened 2026-06-22
fix: do not allow domain change during --reset-certificate

shwetha-s-poojary:
Added domain validation to `--reset-certificate` to prevent certificate resets when the domain has changed. The operation now fails with a clear error if the domain extracted from new SSL certificates doesn't match the deployed configuration, preventing certificate-domain mismatches.


mayuka-c (dismissed):
LGTM

mayuka-c (approved):
LGTM

PR #1011, opened 2026-06-22
Misc fixes

mayuka-c:
- Add `--insecure` for `ai-services login` in dev guide
- Fix message in List Application API for services. Also populate Component Status and message.


Shubhamag12 (approved):
lgtm

PR #1012, opened 2026-06-22
Session timeout implementation

pranithraoibm:
- Warning before logout: A popup warns users 2 minutes before logout with a countdown timer

image

- Auto-logout after inactivity: Users get logged out automatically after 15 minutes with no activity
image

- Keeps you logged in: Automatically refreshes your session in the background so you don't get logged out unexpectedly

New changes:
- Removed the logout button from the warning modal to prevent 401 authentication errors that occurred when users clicked logout after being inactive, as the access token would expire during inactivity and fail server validation. Users who want to logout can use the header dropdown menu where the token remains fresh due to active user interaction.
- Improved the error notification placement in the logout page as it was using the same value as login but logout page doesn't have the page header so positioning wasn't good, now it's fixed on top right.


ui/catalog/src/constants/session.constants.ts
```diff
@@ -0,0 +1,6 @@
+export const SESSION_CONFIG = {
+  INACTIVITY_TIMEOUT: 30 * 1000,
+  WARNING_TIME: 20 * 1000,
+  TOKEN_REFRESH_BUFFER: 2 * 60 * 1000,
```
ryanmarc: This would warn someone in 20 seconds and log them out in 30 seconds. Are these really the timings that we want the application to have?

ryanmarc (changes requested):

ui/catalog/src/constants/session.constants.ts
pranithraoibm: Sorry, those were test values. In production, users will see a warning modal after 13 minutes of inactivity and will be automatically logged out after 15 minutes of inactivity.

ryanmarc (approved):

PR #1013, opened 2026-06-22
AISERVICES-1277 : Input validation for additional fields 

Niranjan-J-Joshi:
Changes:

- Added schema-based validation for additional provider fields.

Added validation checks for:
- Required fields
- Pattern/regex constraints
- Minimum and maximum length constraints

Prevents users from applying changes when validation errors are present.
Displays field-level validation messages for invalid inputs.


ui/catalog/src/components/ServicesDeployFlow/steps/StepTwo.tsx
```diff
@@ -174,22 +182,37 @@ export const StepTwo: React.FC<StepProps> = ({
...
-      );
+    // Parse schema to get all fields with their validation rules
+    const fields = parseSchema(
+      providerSchema as import("@/utils/schemaParser").JSONSchema,
+    );
+
+    // Validate each field
+    fields.forEach((field) => {
+      const value = llmParams[field.key];
+      const error = validateField(value, field);
```
ryanmarc, line 205: This changes required-field behavior: whitespace-only strings now pass validation because validateField only checks value === "" and does not trim. The previous helper rejected whitespace-only values with String(value).trim() !== "". Please either trim string values before calling validateField or update validateField to treat whitespace-only strings as missing, otherwise required credential fields can be applied as spaces.

ryanmarc (changes requested):

ryanmarc (approved):

PR #1018, opened 2026-06-23
[CVE-FIX] AISERVICES-1294: Bumping by caddy to fix cves

adarshagrawal38:
PR is for fixing `8 High cve` out of `10 High cve` present in caddy web server


adarshagrawal38:
After bumping up caddy, we still have 2 HIGH cves coming from stdlib pkg, upstream repo of caddy is not yet moved to latest go version that why cves are present.

Any suggestion on how, to handle this cves?

I am thinking of adding a patch file to bump up go version.

cc: @mkumatag @yussufsh @shwetha-s-poojary

adarshagrawal38:
@tsambari2 suggested we can live with this cve, once upstream caddy fix the issue, we can bump up caddy.

mkumatag (approved):

mkumatag (changes requested):
@adarshagrawal38 I see there are golang issues in the caddy image, wondering if we can build instead of pulling built binary?

mkumatag (approved):

PR #1020, opened 2026-06-23
fix: fetch and delete dir from catalog pod env

Shubhamag12:
Problem: As part of catalog uninstall, if a custom dir path is specified, it fails to cleanup as it is always removing from the default path

Solution: We should inspect the catalog pod, read the custom dir set and use that path to remove all the time.

### DIFF
--- a/ai-services/internal/pkg/catalog/cli/uninstall/podman/uninstall.go
+++ b/ai-services/internal/pkg/catalog/cli/uninstall/podman/uninstall.go
@@ -126,6 +126,12 @@ func performCleanup(rt *podman.PodmanClient, pods []types.Pod, skipCleanup bool)
 		return err
 	}
 
+	// Delete models data
+	modelsDataPath := filepath.Join(baseDir, "models")
+	if err := dataDeletion(modelsDataPath); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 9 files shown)


ai-services/internal/pkg/catalog/cli/uninstall/podman/uninstall.go
```diff
@@ -112,7 +121,7 @@ func performCleanup(rt *podman.PodmanClient, pods []types.Pod, skipCleanup bool)
 	}
 
 	// Delete caddy data
-	caddyDataPath := getDataPath(baseDir, "common")
+	caddyDataPath := filepath.Join(baseDir, "common")
 	if err := dataDeletion(caddyDataPath); err != nil {
 		return err
 	}
```
mayuka-c, line 127: Also add models deletion

Shubhamag12: done

mayuka-c (dismissed):
LGTM, pls bump up the image

Shubhamag12:
> LGTM, pls bump up the image

done, ptal

mayuka-c (approved):

PR #1024, opened 2026-06-23
feat: enhance schema validation with pattern enforcement and security improvements

mkumatag:
## Overview
This PR improves the schema validation system with better security, validation patterns, and comprehensive test coverage.

### 1. Schema Validation ()
- **Fix duplicate error messages** by collecting only leaf errors from the validation tree
- **Add ** to redact sensitive values from error logs
- Maintain field names and validation rules while protecting sensitive data
- Pattern: `at '/fieldName': '[REDACTED]'` instead of exposing actual values

### 2. Watsonx Schema ()
- **Add pattern validation** for `watsonxApiKey` (44-char alphanumeric with `-` and `_`)
- **Add minLength/maxLength constraints** for API key (44 characters)
- **Add UUID pattern validation** for `watsonxProjectId`
- **Add HTTPS URL pattern validation** for `watsonxUrl`
- Ensures client-side validation matches server-side requirements

### 3. Tests
- Added unit tests for schema.go file

## Benefits
- ✅ **Enhanced security** by redacting sensitive values in logs
- ✅ **Better user experience** with pattern validation
- ✅ **Improved test maintainability** and coverage
- ✅ **Consistent validation** between client and server
- ✅ **No breaking changes** - fully backward compatible

## Testing
```bash
cd ai-services && go test ./internal/pkg/catalog/validators/...
# PASS
# ok  	github.com/project-ai-services/ai-services/internal/pkg/catalog/validators	1.377s
```

### DIFF
--- a/ai-services/internal/pkg/catalog/validators/schema.go
+++ b/ai-services/internal/pkg/catalog/validators/schema.go
@@ -119,6 +119,7 @@ func sanitizeErrorMessage(msg string) string {
 	// We want to keep '/fieldName' but redact 'actual-value'
 	// This pattern finds the first quoted value after ": " which is typically the actual value being validated
 	re := regexp.MustCompile(`(at '[^']+': )'[^']+'`)
+
 	return re.ReplaceAllString(msg, "${1}'[REDACTED]'")
 }
 
--- a/ai-services/internal/pkg/catalog/validators/schema.go
+++ b/ai-services/internal/pkg/catalog/validators/schema.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"maps"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/santhosh-tekuri/jsonschema/v6"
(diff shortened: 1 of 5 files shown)


ai-services/internal/pkg/catalog/validators/schema.go
```diff
@@ -96,17 +97,30 @@ func validateAgainstSchema(compiledSchema *jsonschema.Schema, params map[string]
...
+// sanitizeErrorMessage removes sensitive values from error messages while keeping field names and validation details.
+// It only redacts the actual value being validated (the first quoted string after a colon in the message).
+func sanitizeErrorMessage(msg string) string {
+	// Pattern to match: at '/fieldName': 'actual-value' does not match...
+	// We want to keep '/fieldName' but redact 'actual-value'
+	// This pattern finds the first quoted value after ": " which is typically the actual value being validated
+	re := regexp.MustCompile(`(at '[^']+': )'[^']+'`)
+
+	return re.ReplaceAllString(msg, "${1}'[REDACTED]'")
+}
```
mayuka-c, line 124: Would this redact all the values?
I was wondering if we can redact only those values which are having format: password

mkumatag: Yes, but it's very difficult to differentiate between passwords and normal ones. And I don't think it's worth displaying the values as-is to the client, since we're already showing the variable name.

mkumatag: I did try exploring adding this code but adding unnecessary complicated code, I fee we need to revisit all these errors later I believe. For now may be we can push this code and enhance it later..

Shubhamag12 (dismissed):
lgtm

ai-services/internal/pkg/catalog/validators/schema.go
mayuka-c: I feel we may also be printing the Values map in logs, maybe I will handle it separately.
Maybe better to just print the keys there too in logs

mayuka-c (dismissed):
LGTM

mayuka-c:
Pls bump up the image

ai-services/internal/pkg/catalog/validators/schema.go
Shubhamag12: @mayuka-c but that too will print the apikey, right? because same values map logs was added earlier and I had removed it because it was showing sensitive info.

Shubhamag12 (approved):
lgtm

PR #1027, opened 2026-06-23
Invalid args

shwetha-s-poojary:
Added `Args: cobra.NoArgs` validation to the catalog configure command to reject unexpected positional arguments. This prevents silent failures when users accidentally use em-dashes (`—`) instead of double hyphens (`--`) for flags, which are often auto-inserted by word processors and cause flags to be ignored without error messages.


mayuka-c (approved):
LGTM

PR #1033, opened 2026-06-24
fix: refresh token if expired while polling application

Shubhamag12:
- Problem: 
Create cmd was throwing below error while deploying application.
```
Deploying, Message: Deploying application
Error: failed to get application status: HTTP 401: invalid token 
```

- Solution: refresh token if it expires

### DIFF
--- a/ai-services/internal/pkg/catalog/client/application.go
+++ b/ai-services/internal/pkg/catalog/client/application.go
@@ -242,4 +242,10 @@ func (c *ApplicationClient) GetComponentProviderParams(componentType, providerID
 	return result, nil
 }
 
+// Client returns the underlying authenticated HTTP client.
+// This allows access to client-level operations like token refresh.
+func (c *ApplicationClient) Client() *Client {
+	return c.client
+}
+
 // Made with Bob
--- a/ai-services/internal/pkg/catalog/client/application.go
+++ b/ai-services/internal/pkg/catalog/client/application.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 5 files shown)


ai-services/internal/pkg/cli/utils/application_utils.go
```diff
@@ -20,8 +22,23 @@ func GetAllApps(appClient *catalogClient.ApplicationClient) ([]types.Application
...
+		if !isUnauthorizedError(err) {
+			return nil, err
+		}
+
+		// Refresh token and retry once
+		if refreshErr := appClient.Client().RefreshToken(); refreshErr != nil {
+			return nil, fmt.Errorf("failed to refresh token: %w", refreshErr)
+		}
+
+		// Retry the request with refreshed token
```
adarshagrawal38: How about moving the retry logic to `appClient.ListApplications(nil)`
So that it is applicable for `GetAllApps` method as well

WDYT?

Shubhamag12: using getAppbyID now

ai-services/internal/pkg/catalog/client/application.go
```diff
@@ -159,10 +161,19 @@ func (c *ApplicationClient) GetApplication(id string) (*types.Application, error
...
+		httpErr := &HTTPError{
 			StatusCode: resp.StatusCode(),
 			Message:    utils.ParseErrorResponse(resp),
 		}
+		if httpErr.StatusCode == http.StatusUnauthorized {
+			if refreshErr := c.client.RefreshToken(); refreshErr != nil {
+				return nil, httpErr
+			}
+
+			return c.GetApplication(id)
```
yussufsh, line 184: This will lead to infinite recursion on repeated 401.

yussufsh (changes requested):

ai-services/internal/pkg/catalog/client/application.go
Shubhamag12: add a new func

yussufsh (approved):

PR #1035, opened 2026-06-24
[CVE-FIX] Bump up pip for tools images

adarshagrawal38:


images/tools/Containerfile
```diff
@@ -10,6 +10,8 @@ COPY requirements.txt .
 RUN dnf update ${DNF_FLAGS} -y \
   && dnf install ${DNF_FLAGS} -y gcc \
   python3.12-pip python3.12 python3.12-devel python3-pexpect python3-ptyprocess python3-pyyaml python3-file-magic \
+  && python3.12 -m pip install --upgrade pip \
```
manju956: We have go implementation of Servicereport tool which does repair and validation of spyre cards. Why do we still need to maintain tools image here? 
CC: @mkumatag

mkumatag: we use other tools like hf for downloading the models!

manju956: Ok, got it.

images/tools/Containerfile
```diff
@@ -10,6 +10,8 @@ COPY requirements.txt .
...
+  && rm -rf /usr/lib/python3.12/site-packages/pip-23.2.1.dist-info \
```
manju956: can we avoid hardcoding of specific pip directory like this? maybe use wildcard or something?

mkumatag: yes, this is problem.. we need to handle this better, deleting the file with version is not a good thigh

images/tools/Containerfile
```diff
@@ -10,6 +10,8 @@ COPY requirements.txt .
 RUN dnf update ${DNF_FLAGS} -y \
   && dnf install ${DNF_FLAGS} -y gcc \
   python3.12-pip python3.12 python3.12-devel python3-pexpect python3-ptyprocess python3-pyyaml python3-file-magic \
+  && python3.12 -m pip install --upgrade pip \
+  && rm -rf /usr/lib/python3.12/site-packages/pip-*.dist-info \
```
mkumatag: but this will delete all the pip package..

images/tools/Containerfile
manju956: right

adarshagrawal38: Bob suggested: We can install PIP via get-pip.py

`curl -sSL https://bootstrap.pypa.io/get-pip.py | python3.12 -`

images/tools/Containerfile
adarshagrawal38: instead of rm we can do `microdnf remove python3.12-pip`
Which will remove `pip-23.2.1` from the machine.

We will only have latest version of `pip-26`

adarshagrawal38:
I feel using `get-pip.py` is better way to install pip.

Another option is:
```
microdnf install -y python3.12-devel python3.12-pip
python3.12 -m pip install --upgrade pip
microdnf remove python3.12-pip
```

cc: @mkumatag WDYT?

mkumatag:
> I feel using `get-pip.py` is better way to install pip.
> 
> Another option is:
> 
> ```
> microdnf install -y python3.12-devel python3.12-pip
> python3.12 -m pip install --upgrade pip
> microdnf remove python3.12-pip
> ```
> 
> cc: @mkumatag WDYT?

I'm not an expert here but try something works for long run and the standard way, if get-pip.py then lets do it.

images/service-base/Containerfile
```diff
@@ -1,25 +1,22 @@
...
 
-RUN microdnf install -y python3.12-devel python3.12-pip \
+RUN microdnf install -y python3.12-devel \
         lcms2-devel openblas-devel freetype libicu libjpeg-turbo && \
     curl -o /tmp/epel-release.rpm https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm && \
     rpm -ivh /tmp/epel-release.rpm && \
     rm -f /tmp/epel-release.rpm && \
     microdnf install -y spatialindex-devel && \
     microdnf clean all && \
+    curl -sSL https://bootstrap.pypa.io/get-pip.py | python3.12 - && \
```
manju956, line 10: Is it safe to remove get-pip.py script once run?

adarshagrawal38: we have multi stage build, so file won't be present in final stage

manju956: Ok

manju956 (approved):
LGTM

PR #1036, opened 2026-06-24
Missing langs for not showing Get Ref Docs button

sats-23:
Minor nit

### DIFF
--- a/ui/chatbot/src/components/customSendMessage.jsx
+++ b/ui/chatbot/src/components/customSendMessage.jsx
@@ -246,6 +246,8 @@ async function customSendMessage(
     const noDocsMessages = [
       'No documents found in the knowledge base for this query.',
       'Für diese Anfrage wurden keine Dokumente in der Wissensdatenbank gefunden.',
+      'Nessun documento trovato nella base di conoscenza per questa richiesta.',
+      'Aucun document trouvé dans la base de connaissances pour cette requête.',
     ];
     const hasNoDocsMessage = noDocsMessages.some((msg) =>
       fullText.includes(msg),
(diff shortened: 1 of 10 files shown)


ui/chatbot/src/components/customSendMessage.jsx
```diff
@@ -246,6 +246,8 @@ async function customSendMessage(
     const noDocsMessages = [
       'No documents found in the knowledge base for this query.',
       'Für diese Anfrage wurden keine Dokumente in der Wissensdatenbank gefunden.',
+      'Nessun documento trovato nella base di conoscenza per questa richiesta.',
+      'Aucun document trouvé dans la base de connaissances pour cette requête.',
```
dharaneeshvrd, line 250: Wondering how it would show respective language message for the user query?

sats-23: User sends query, we do lang detection on that query. 
Based on that the No docs message is chosen.

sats-23: No docs message comes from backend, but here we are just doing another layer of handling just to not show the Get Reference Docs button

Niharika0306: so on the UI, it checks if the array has a message that matches what comes from backend ( as a validation), and then displays that. 
but no real detection happening on the UI side right.

sats-23: Yes, that's right

Niharika0306: okay, lgtm.

Maybe in future, we can try to relay the message directly based on backend response structure.

Niharika0306 (approved):

mkumatag (approved):

PR #1037, opened 2026-06-24
fix: add retries while creating and deleting applications

Shubhamag12:
Added logic to retry while creating or deleting applications.


mayuka-c (approved):
LGTM

PR #1041, opened 2026-06-24
doc: update non root proposal doc

Shubhamag12:
Adding `su - <username>` which was missing from the non-root doc.


manju956 (approved):
lgtm

PR #1042, opened 2026-06-24
Add delay to GET applications call post delete

pranithraoibm:
- Added a 5secs delay to the GET applications API call post delete as it was throwing 500 error immediately. Adding a retry immediately also didn't help as it was throwing the same 500 error. 
- Fixed the icon displayed in the messages column as per the application status 
- Added padding to the services table message column for better UX

JIRA: https://jsw.ibm.com/browse/AISERVICES-1295

### DIFF
--- a/ui/catalog/src/pages/DigitalAssistants/DigitalAssistants.tsx
+++ b/ui/catalog/src/pages/DigitalAssistants/DigitalAssistants.tsx
@@ -93,6 +93,8 @@ const renderCell = ({
   );
 };
 
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
 const DigitalAssistantsPage = () => {
   const [state, dispatch] = useReducer(appReducer, INITIAL_STATE);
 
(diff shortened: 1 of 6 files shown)


ui/catalog/src/pages/DigitalAssistants/DigitalAssistants.tsx
```diff
@@ -198,7 +200,11 @@ const DigitalAssistantsPage = () => {
     try {
       await deleteApplication(state.selectedRowId);
       dispatch({ type: ACTION_TYPES.CLOSE_DELETE_DIALOG });
-      await loadApplications();
+
+      void (async () => {
+        await sleep(5000);
+        await loadApplications();
+      })();
```
ryanmarc: This should not run the delayed refresh as fire-and-forget. Since `loadApplications()` runs outside the surrounding `try/catch`, any failure is unhandled and the UI can show stale deleted data with no loading/error state. Can we await the delayed reload or move the refresh/error handling into the same async flow?

ryanmarc (changes requested):

ui/catalog/src/pages/DigitalAssistants/DigitalAssistants.tsx
pranithraoibm: Fixed by awaiting the delayed reload directly within the try/catch block.

ryanmarc (approved):

PR #1045, opened 2026-06-24
[AISERVICES-1310] fix: clear provider-specific params on inference backend change

Niranjan-J-Joshi:
This change clears provider-specific parameters when switching inference backends to prevent stale configuration values from being carried over to the newly selected backend.

Problem:

- Previously, when a user changed the inference backend, all existing parameters were retained, including inference backend-specific fields such as credentials, endpoints, and other optional configuration values.

Changes

- Reset provider-specific parameters when the inference backend changes.
- Prevent stale inference backend-specific values from being reused across different providers.
- Added validation for service name


ui/catalog/src/components/ServicesDeployFlow/steps/StepTwo.tsx
```diff
@@ -605,8 +605,14 @@ export const StepTwo: React.FC<StepProps> = ({
...
-                                  params:
-                                    currentConfig?.components.llm?.params || {},
+                                  params: currentConfig?.components.llm?.params
+                                    ?.model
+                                    ? {
+                                        model:
+                                          currentConfig.components.llm.params
+                                            .model,
+                                      }
+                                    : {},
```
ryanmarc, line 614: When changing the inference backend, we should clear or revalidate `params.model`. Keeping the previous model can create an invalid provider/model combination and submit a config the backend cannot resolve.

ryanmarc (changes requested):

ui/catalog/src/components/ServicesDeployFlow/steps/StepTwo.tsx
Niranjan-J-Joshi: I updated this to clear params.model when the inference backend changes, so we no longer persist a potentially invalid provider/model combination in the config.

We still need to retain the selected model temporarily for UI filtering. The Inference backend dropdown filtering logic depends on knowing which model is currently selected; if the model is completely removed, the dropdown falls back to showing all providers, including incompatible ones.

To address this, I keep the model only in local component state for filtering purposes. It is also cleared whenever the user selects a different LLM model, so filtering always follows the current model selection and does not retain stale model values.

ui/catalog/src/components/ServicesDeployFlow/steps/StepTwo.tsx
```diff
@@ -594,32 +607,34 @@ export const StepTwo: React.FC<StepProps> = ({
...
                         itemToString={(item) => (item ? item.text : "")}
                         selectedItem={selectedItem}
                         onChange={({ selectedItem }) => {
                           if (field.isInferenceMethod) {
+                            // Store current model for filtering before clearing params
+                            const currentModel = currentConfig?.components.llm
+                              ?.params?.model as string | undefined;
+                            if (currentModel) {
+                              setModelForFiltering(currentModel);
+                            }
```
ryanmarc: This fixes the invalid provider/model carry-over, but clearing params.model unconditionally can leave the UI showing a fallback model while the config has params: {}. Can we either keep the current model when it is valid for the newly selected provider, or show the model dropdown as empty/required until the user explicitly selects one?

ryanmarc (changes requested):

ui/catalog/src/components/ServicesDeployFlow/steps/StepTwo.tsx
ryanmarc: Based on our conversation - lets switch back to this code.

Niranjan-J-Joshi: I have reverted to the this implementation and pushed the changes for review.

pranithraoibm (approved):
LGTM

ryanmarc (approved):

PR #935, opened 2026-06-12
add --insecure flag to catalog login cmd

Shubhamag12:
1. This PR adds --insecure flag to the catalog login command to skip TLS certificate verification. 
2. logs insecure message saying "Not recommended for production" if the user passes the flag as part of `ai-services login` command.

PR #1047, opened 2026-06-25
Fix issues with backup and restore

prachikurade1:
Issue 1: The service type earlier was "digitize" which was used to fetch the Digitize API url. However now it is changed to "Digitize Documents". Hence updated the code to use catalog id field instead of the type field.

Issue 2: The calls to the Digitize API were not respecting the --insecure flag passed during catalog login. Hence updated the backup and restore clients to use that.


ai-services/internal/pkg/application/common/restore/digitize.go
```diff
@@ -147,9 +149,19 @@ type DigitizeRestoreClient struct {
...
 	client := resty.New().SetBaseURL(serviceURL)
 
+	// Load catalog credentials to check if insecure mode is enabled
+	creds, err := config.Load()
+	if err == nil && creds.Insecure {
+		logger.Infoln("Using insecure TLS mode for digitize import API (certificate verification disabled)")
+		client.SetTLSClientConfig(&tls.Config{
+			InsecureSkipVerify: true,
+		})
+	}
```
mayuka-c: It should be done within the resty client, any reason for doing this again?
@Shubhamag12 Are we not adding insecure within the client?

mayuka-c: If not, we should be adding it there and making it more generic without caller needing to specify this

prachikurade1: It is implemented here - ai-services/internal/pkg/catalog/client/client.go - however this is specific for the catalog client. So for Digitize API calls it still fails.
@Shubhamag12 - can you please check ?

Shubhamag12: I think we will have to use different client for B/R altogether since we use the same client for openshift as well :(
and it will not work there.

maybe revert the httpclient file: https://github.com/IBM/project-ai-services/pull/935/changes#diff-42065640454a0a7abc4eb180dd44100c9e78aef1e40e805f18dff68ba50a1a08

prachikurade1: As discussed, tls verification will be skipped for openshift always and for podman if login was done using insecure flag.

mayuka-c (approved):
LGTM

yussufsh (approved):

PR #1048, opened 2026-06-25
[CATALOG-API] Updating app name lookup from DB to be case-insensitive and checking if appName present before updating name

adarshagrawal38:
- Updated `GetAppByName()` to return app from database, by perform case-insensitive matching for `appName` field.
- Added validation to check if App is already present in update flow.

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
+++ b/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
@@ -219,8 +219,16 @@ func ValidatePaginationParams(page, pageSize int) (int, int, error) {
 }
 
 func (s *ApplicationService) UpdateApplication(ctx context.Context, id uuid.UUID, userID, newName string) (*types.Application, error) {
-	if err := s.validateApplicationName(ctx, newName); err != nil {
-		return nil, err
+	existingApp, err := s.appRepo.GetByName(ctx, newName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check for existing application: %w", err)
+	}
+	if existingApp != nil {
+		// Application with this name already exists - return conflict error
+		return nil, &ValidationError{
+			Code:    http.StatusConflict,
+			Message: fmt.Sprintf(ErrMsgApplicationNameExists, newName),
+		}
 	}
 
 	app, err := s.appRepo.GetByID(ctx, id)
@@ -268,8 +276,16 @@ func (s *ApplicationService) UpdateApplication(ctx context.Context, id uuid.UUID
 // for deployment execution, returning 202 Accepted immediately.
 func (s *ApplicationService) CreateApplication(ctx context.Context, req apimodels.CreateApplicationRequest) (*apimodels.CreateApplicationResponse, error) {
 	// Phase 1: Validate request and check for duplicate application name
-	if err := s.validateApplicationName(ctx, req.Name); err != nil {
-		return nil, err
+	existingApp, err := s.appRepo.GetByName(ctx, req.Name)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check for existing application: %w", err)
+	}
+	if existingApp != nil {
+		// Application with this name already exists - return conflict error
+		return nil, &ValidationError{
+			Code:    http.StatusConflict,
+			Message: fmt.Sprintf(ErrMsgApplicationNameExists, req.Name),
+		}
 	}
 
 	// Phase 2: Validate request payload
@@ -1195,27 +1211,6 @@ func loadApplicationPods(rt runtime.Runtime, appID string) ([]types.Pod, error) 
 
 	// Build pod response with metadata and container details
 	return appPodList, nil
-}
-
-// validateApplicationName validates the application name and checks if it already exists in DB.
-func (s *ApplicationService) validateApplicationName(ctx context.Context, appName string) error {
-	if err := s.validator.ValidateAppName(appName); err != nil {
-		return err
-	}
-
-	existingApp, err := s.appRepo.GetByName(ctx, appName)
-	if err != nil {
-		return fmt.Errorf("failed to check for existing application: %w", err)
-	}
-	if existingApp != nil {
-		// Application with this name already exists - return conflict error
-		return &ValidationError{
-			Code:    http.StatusConflict,
-			Message: fmt.Sprintf(ErrMsgApplicationNameExists, appName),
-		}
-	}
-
-	return nil
 }
 
 // Made with Bob
--- a/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
+++ b/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
@@ -223,6 +223,18 @@ func (s *ApplicationService) UpdateApplication(ctx context.Context, id uuid.UUID
 		return nil, err
 	}
 
+	existingApp, err := s.appRepo.GetByName(ctx, newName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check for existing application: %w", err)
+	}
+	if existingApp != nil {
+		// Application with this name already exists - return conflict error
+		return nil, &ValidationError{
+			Code:    http.StatusConflict,
+			Message: fmt.Sprintf(ErrMsgApplicationNameExists, newName),
+		}
+	}
+
 	app, err := s.appRepo.GetByID(ctx, id)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get application: %w", err)
--- a/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
+++ b/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
@@ -275,6 +275,9 @@ func (s *ApplicationService) CreateApplication(ctx context.Context, req apimodel
 		}
 	}
 
+	// Update app name by trimming prefix and suffix spaces
+	req.Name = strings.TrimSpace(req.Name)
+
 	// Phase 2: Validate request payload
 	if err := s.validator.ValidateDeploymentRequest(ctx, req); err != nil {
 		return nil, err
--- a/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
+++ b/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
@@ -219,20 +219,8 @@ func ValidatePaginationParams(page, pageSize int) (int, int, error) {
 }
 
 func (s *ApplicationService) UpdateApplication(ctx context.Context, id uuid.UUID, userID, newName string) (*types.Application, error) {
-	if err := s.validator.ValidateAppName(newName); err != nil {
+	if err := s.validateApplicationName(ctx, newName); err != nil {
 		return nil, err
-	}
-
-	existingApp, err := s.appRepo.GetByName(ctx, newName)
-	if err != nil {
-		return nil, fmt.Errorf("failed to check for existing application: %w", err)
-	}
-	if existingApp != nil {
-		// Application with this name already exists - return conflict error
-		return nil, &ValidationError{
-			Code:    http.StatusConflict,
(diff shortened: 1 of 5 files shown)


ai-services/internal/pkg/catalog/validators/validation.go
```diff
@@ -48,6 +55,44 @@ func (v *ApplicationValidator) ValidateDeploymentRequest(ctx context.Context, re
...
+		}
+	}
+
+	return nil
+}
+
+// isAlphanumeric checks if a character is alphanumeric (a-z, A-Z, 0-9).
+func isAlphanumeric(char rune) bool {
+	return (char >= 'a' && char <= 'z') || (char >= 'A' && char <= 'Z') || (char >= '0' && char <= '9')
+}
```
mayuka-c: Use regex

ai-services/internal/pkg/catalog/validators/validation.go
```diff
@@ -48,6 +55,44 @@ func (v *ApplicationValidator) ValidateDeploymentRequest(ctx context.Context, re
...
 }
 
+func (v *ApplicationValidator) ValidateAppName(appName string) error {
+	// Check if app name is empty
+	if appName == "" {
+		return fmt.Errorf("application name cannot be empty")
+	}
+
+	// Restrict max character length to 64
+	if len(appName) > 64 {
```
mayuka-c: Also add min length

adarshagrawal38: add min check as well

ai-services/internal/pkg/catalog/validators/validation.go
adarshagrawal38: using regex now

yussufsh:
@adarshagrawal38 I think UI has this check `Name must be at least 3 characters`... and maximum 100.
cc @pranithraoibm

pranithraoibm:
image

No @yussufsh only empty check is present, no restrictions added on the length

yussufsh:
@pranithraoibm I saw it in update deployment page.. could you please confirm?
image
image

pranithraoibm:
Yes @yussufsh confirmed, in the deployment details rename place we have this check

yussufsh:
> Yes @yussufsh confirmed, in the deployment details rename place we have this check

Are we planning to add it in deploy page as well or remove check from update page? let us know.

yussufsh:
@adarshagrawal38 take care of the length once @pranithraoibm confirms the final way UI is integrating.
Also, please add case-insensitive checks for existing app names in both the create and update flows.

pranithraoibm:
So this check happens when we call the deploy API? for deploy flows

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
```diff
@@ -276,6 +281,10 @@ func (s *ApplicationService) CreateApplication(ctx context.Context, req apimodel
 	}
 
 	// Phase 2: Validate request payload
+	if err := s.validator.ValidateAppName(req.Name); err != nil {
+		return nil, err
+	}
```
yussufsh: This should be right on top in Phase 1. Only valid names need to be checked against db.

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
```diff
@@ -235,6 +235,11 @@ func (s *ApplicationService) UpdateApplication(ctx context.Context, id uuid.UUID
 			Message: ErrMsgUserNotOwner,
 		}
 	}
+
+	if err := s.validator.ValidateAppName(newName); err != nil {
+		return nil, err
+	}
+
```
yussufsh: Are we not missing a check for existing apps with the same name here?

yussufsh (changes requested):

yussufsh:
> So this check happens when we call the deploy API? for deploy flows

App name check will happen for both deploy and update flow.
Now what I see is that UI check is not consistent among these flows. This PR needs to be aligned what the UI and design team had decided for app name.

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
```diff
@@ -235,6 +235,11 @@ func (s *ApplicationService) UpdateApplication(ctx context.Context, id uuid.UUID
...
 	err = s.appRepo.UpdateDeploymentName(ctx, id, newName)
 	if err != nil {
 		return nil, fmt.Errorf("failed to update application name: %w", err)
```
yussufsh:
```suggestion
		return nil, fmt.Errorf("failed to update name: %w", err)
```

ai-services/internal/pkg/catalog/validators/validation.go
```diff
@@ -50,6 +50,21 @@ func (v *ApplicationValidator) ValidateDeploymentRequest(ctx context.Context, re
...
 
+// ValidateAppName validates the application name.
+func (v *ApplicationValidator) ValidateAppName(appName string) error {
+	// Restrict character length between 4 and 64
+	const minAppNameLength = 4
+	const maxAppNameLength = 64
+	if len(appName) < minAppNameLength || len(appName) > maxAppNameLength {
+		return &ValidationError{
+			Code:    http.StatusBadRequest,
+			Message: fmt.Sprintf("application name must be between %d and %d characters", minAppNameLength, maxAppNameLength),
```
yussufsh:
```suggestion
			Message: fmt.Sprintf("name must be between %d and %d characters", minAppNameLength, maxAppNameLength),
```

yussufsh (changes requested):

ai-services/internal/pkg/catalog/validators/validation.go
adarshagrawal38: removed applicatiob.

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
adarshagrawal38: removed

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
adarshagrawal38: We already have that check 
PTAL [here](https://github.com/adarshagrawal38/project-ai-services/blob/62b9e80f786cca922883a7a6318b5e674d1ec485/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go#L275)

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
adarshagrawal38: Moved to phase 1 now.

ai-services/internal/pkg/catalog/validators/validation.go
```diff
@@ -50,6 +50,21 @@ func (v *ApplicationValidator) ValidateDeploymentRequest(ctx context.Context, re
 	}
 }
 
+// ValidateAppName validates the application name.
+func (v *ApplicationValidator) ValidateAppName(appName string) error {
+	// Restrict character length between 4 and 64
+	const minAppNameLength = 4
```
mayuka-c: Any reason for min being 4?
Maybe 2 or 3 should be fine I think

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
```diff
@@ -263,6 +268,10 @@ func (s *ApplicationService) UpdateApplication(ctx context.Context, id uuid.UUID
 // for deployment execution, returning 202 Accepted immediately.
 func (s *ApplicationService) CreateApplication(ctx context.Context, req apimodels.CreateApplicationRequest) (*apimodels.CreateApplicationResponse, error) {
```
mayuka-c, line 277: Can we trim spaces in prefix and suffix for appName in both Create and Update

adarshagrawal38: @yussufsh suggested not to do that now, to keep parity with UI

ai-services/internal/pkg/catalog/validators/validation.go
adarshagrawal38: We discussed min length right, If you remember.

yussufsh: Let's do 3-100?

adarshagrawal38: Wondering, can APP name length can go upto 100.

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
yussufsh: > We already have that check PTAL [here](https://github.com/adarshagrawal38/project-ai-services/blob/62b9e80f786cca922883a7a6318b5e674d1ec485/ai-services/internal/pkg/catalog/apiserver/repository/application_service.go#L275)

This is for create application. We are talking about update func.

adarshagrawal38: okay I will Add.

ai-services/internal/pkg/catalog/validators/validation.go
yussufsh: https://github.com/IBM/project-ai-services/blob/main/ai-services/internal/pkg/catalog/db/migrations/assets/20260430094502_create_applications_table.sql#L12

adarshagrawal38: Update range to 3-100 chars PTAL

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
adarshagrawal38: Added check for existing app

ai-services/internal/pkg/catalog/validators/validation.go
```diff
@@ -50,6 +50,21 @@ func (v *ApplicationValidator) ValidateDeploymentRequest(ctx context.Context, re
 	}
 }
 
+// ValidateAppName validates the application name.
+func (v *ApplicationValidator) ValidateAppName(appName string) error {
+	// Restrict character length between 3 and 100
+	const minAppNameLength = 3
+	const maxAppNameLength = 100
+	if len(appName) < minAppNameLength || len(appName) > maxAppNameLength {
```
mkumatag: why ware we doing it here if you are already controlling here 

https://github.com/IBM/project-ai-services/blob/efa0ac6359d51e6d631aa7cf07c28dce5287ad51/ai-services/internal/pkg/catalog/apiserver/models/create_application.go#L5

mkumatag (changes requested):

ai-services/internal/pkg/catalog/validators/validation.go
adarshagrawal38: Initially there were multiple checks on `appName`, but after addressing review comments, only one remains. I've kept this method in anticipation of additional checks being added in the future.

mkumatag: What I'm trying to say is that this check is redundant and duplicated, and may become stale if handled in two different places, which is unnecessary. Furthermore, your call will never actually hit this code in a real-world length match scenario because the Gin layer itself will throw an error message and return.

adarshagrawal38: right removed len check

ai-services/internal/pkg/catalog/apiserver/repository/application_service.go
yussufsh: Multiple spaces in between chars is what we talked about, but trimming spaces at start and end makes sense. This doesn't allow app names with just spaces. Such checks and manipulation should be client-side, in my view.

yussufsh: That said it seems not urgent for this release.

yussufsh (approved):
lgtm

mkumatag (approved):

PR #1049, opened 2026-06-25
Fix caching during sync

mayuka-c:
- Since the local cache was using catalogID, if there were multiple entries in cache with different instances, then it used to get override causing issues with slug for secret and volumes.
- The above was causing DB sync to update to error as it was saying secret and volumes are missing.


adarshagrawal38 (approved):
LGTM

Shubhamag12 (approved):

PR #1050, opened 2026-06-25
summarize bugfix: store chunking strategy and metadata in db

manalilatkar:
Earlier there were 2 bugs in the summarize async jobs flow:
1. When strategy is chunked, it was not updated in db, so every job was direct strategy in db.
2. In list jobs, metadata was being sent as empty dict, now we have removed the key.

Now both these bugs are fixed.


services/bugfix.patch
dharaneeshvrd, line 1: What is this patch for?

dharaneeshvrd (commented):
Can you please add more info on the bug and the JIRA story in description?

services/bugfix.patch
manalilatkar: Removed the patch.

services/summarize/app.py
```diff
@@ -1070,6 +1078,21 @@ async def list_jobs(
             # Convert status string to JobStatus enum if needed
             job_status = job.status if isinstance(job.status, JobStatus) else JobStatus(job.status)
             
+            # Create JobMetadata from job metadata if available
```
dharaneeshvrd: I think metadata should be part of individual get right instead of the list API?

dharaneeshvrd (approved):

PR #1054, opened 2026-06-25
[DB-Sync] Reducing Service/Component unwanted syncs

mayuka-c:
- This is to avoid unwanted service/component syncs for those which are not in Running/Error state.
- Also skips syncing services/components which are in Initializing state.

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/services/sync/sync_service.go
+++ b/ai-services/internal/pkg/catalog/apiserver/services/sync/sync_service.go
@@ -201,7 +201,7 @@ func (s *SyncService) syncApplication(ctx context.Context, app *models.Applicati
 	// Step 3: If any service or component is still initialising, skip updating the
 	// application status this cycle — we don't have a complete picture yet.
 	if componentsPending || servicesPending {
-		logger.InfofCtx(ctx, "Skipping application %s status update: some services/components are still initialising", app.Name)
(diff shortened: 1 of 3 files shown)


ai-services/internal/pkg/catalog/apiserver/services/sync/sync_service.go
```diff
@@ -185,20 +185,28 @@ func (s *SyncService) syncApplication(ctx context.Context, app *models.Applicati
...
 	if len(serviceErrors) > 0 {
 		errorMessages = append(errorMessages, serviceErrors...)
 		allHealthy = false
 	}
 
-	// Step 3: Update application status based on collected errors
+	// Step 3: If any service or component is still initialising, skip updating the
+	// application status this cycle — we don't have a complete picture yet.
+	if componentsPending || servicesPending {
+		logger.InfofCtx(ctx, "Skipping application %s status update: some services/components are still initialising", app.Name)
```
yussufsh:
```suggestion
		logger.InfofCtx(ctx, "Skipping application %s status update: some services/components are still in Initializing state", app.Name)
```

yussufsh (commented):
minor nit.

ai-services/internal/pkg/catalog/apiserver/services/sync/sync_service.go
mayuka-c: Done

yussufsh (approved):

PR #1056, opened 2026-06-25
Set empty System prompt in values.yaml

mayuka-c:
- This is required to ensure that if no prompt is set within the params from client (UI/CLI) then need to use "" (empty) by default.
- If not used currently `"CHATBOT_SYSTEM_PROMPT=\u003cno value\u003e",` will be set

### DIFF
--- a/ai-services/assets/services/chat/podman/templates/chat-bot.yaml.tmpl
+++ b/ai-services/assets/services/chat/podman/templates/chat-bot.yaml.tmpl
@@ -104,7 +104,11 @@ spec:
         - name: CHATBOT_SIMILARITY_SERVICE_URL
           value: "http://similarity-api-{{ .InstanceSlug }}:7000"
         - name: CHATBOT_SYSTEM_PROMPT
+          {{- if .Values.backend.systemPrompt }}
           value: "{{ .Values.backend.systemPrompt }}"
+          {{- else }}
+          value: ""
+          {{ end }}
       ports:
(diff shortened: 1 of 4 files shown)


ai-services/assets/services/chat/podman/templates/chat-bot.yaml.tmpl
```diff
@@ -104,7 +104,11 @@ spec:
         - name: CHATBOT_SIMILARITY_SERVICE_URL
           value: "http://similarity-api-{{ .InstanceSlug }}:7000"
         - name: CHATBOT_SYSTEM_PROMPT
+          {{- if .Values.backend.systemPrompt }}
```
yussufsh, line 107: I am wondering since we are setting default to "" do we really need this if check?

mayuka-c: Yes even if we have "Values.backend.systemPrompt" to "", the podman sets as env being: `\u003cno value\u003e"`

So this if condition is explicitly saying keep it empty

mayuka-c: I was also not aware but found this out while I was validating :). This might be the issue with how go templating works where it adds those string if we want to pass empty string

yussufsh (dismissed):

yussufsh (approved):

PR #1058, opened 2026-06-25
Both tables UI fix

pranithraoibm:
https://jsw.ibm.com/browse/AISERVICES-1275 - Grey background fix in both the tables, this was because of the sticky behavior we added to the name and the actions column, removed them for now. (Verified both on Safari and Chrome)
https://jsw.ibm.com/browse/AISERVICES-1307 - Fixed this column visibility issue in DA page. Works as expected now
https://jsw.ibm.com/browse/AISERVICES-1276 - 'New' word was missing added that


ui/catalog/src/pages/DigitalAssistants/DigitalAssistants.tsx
```diff
@@ -560,8 +560,10 @@ const DigitalAssistantsPage = () => {
                                                     dispatch={dispatch}
                                                   />
                                                 </TableCell>
-                                                <TableCell />
-                                                <TableCell />
+                                                {state.visibleColumns
+                                                  .uptime && <TableCell />}
+                                                {state.visibleColumns
+                                                  .messages && <TableCell />}
```
ryanmarc, line 569: Please double check that this renders correctly. The parent table headers are driven by state.visibleColumns, but expanded child rows always render the Status cell and only conditionally render uptime / messages. If the user hides the Status column, child rows still include that extra cell, shifting later cells under the wrong headers/actions column.

ryanmarc (dismissed):

ui/catalog/src/pages/DigitalAssistants/DigitalAssistants.tsx
pranithraoibm: Fixed by wrapping the Status cell in child rows with {state.visibleColumns.status && (...)} so it only renders when the Status column is visible, matching the parent row behavior. This ensures child rows always have the same number of cells as parent rows, preventing column misalignment.

pranithraoibm:
https://github.com/user-attachments/assets/579a356e-28a1-4c20-a17c-48231a65edc7

Niranjan-J-Joshi (approved):

PR #1059, opened 2026-06-25
[AISERVICES-1305] Refresh services/applications after rename on back navigation and fix margin issue

Niranjan-J-Joshi:
This PR includes the following fixes:

1. Service list refresh after rename

- After renaming a service, navigating back to the services/ digital assistants page now refreshes the services and application table.
- This ensures the updated service name is immediately visible without requiring a manual refresh.

2. Responsive margin fix

- Fixed a margin/layout issue in details page header


ui/catalog/src/components/DeploymentDetails/DeploymentDetails.tsx
```diff
@@ -393,6 +395,10 @@ const DeploymentDetails = ({
                   onClick={(e) => {
                     e.preventDefault();
                     onBack();
+                    // Call refresh after navigation so table is mounted
+                    setTimeout(() => {
+                      onRefresh?.();
+                    }, 0);
```
ryanmarc: This feels a bit fragile because the refresh is scheduled with `setTimeout` after `onBack()`, so it is not tied to React’s render/commit lifecycle and could run after the details view unmounts or with stale state. Could we move the refresh responsibility up to the owning page so `DeploymentDetails` only handles navigation and the parent decides what needs to be refreshed?

ryanmarc (changes requested):

ui/catalog/src/components/DeploymentDetails/DeploymentDetails.tsx
Niranjan-J-Joshi: Yes, I agree

Changes made:
- Removed onRefresh prop
- Removed setTimeout pattern
- Component now only calls onBack()

ryanmarc (dismissed):

pranithraoibm (approved):

PR #1060, opened 2026-06-25
Increase llm-cpu memory limits to 150G

mayuka-c:
- The vllm cpu pod for llm crashes intermittently due to OOM.
- Based on the suggestion received from the folks, with the configurations we have it requires around 120G, but stable would be 150G.
- Tried with 120G, but still intermittently saw the issue. After moving to 150G didnt see the issue (ran 4-5 times)

### DIFF
--- a/ai-services/assets/components/llm/vllm-cpu/podman/metadata.yaml
+++ b/ai-services/assets/components/llm/vllm-cpu/podman/metadata.yaml
@@ -8,5 +8,5 @@ podTemplateExecutions:
 # Resource requirements
 resources:
   cpu: 12
-  memory: 161061273600 # 150GB in bytes
+  memory: 161061273600 # 150Gi in bytes
   storage: 53687091200 # 50Gi for model storage in bytes
--- a/ai-services/assets/components/llm/vllm-cpu/podman/metadata.yaml
+++ b/ai-services/assets/components/llm/vllm-cpu/podman/metadata.yaml
@@ -8,5 +8,5 @@ podTemplateExecutions:
 # Resource requirements
 resources:
   cpu: 12
-  memory: 107374182400 # 100Gi in bytes
+  memory: 161061273600 # 150GB in bytes
   storage: 53687091200 # 50Gi for model storage in bytes
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.141
+TAG?=v0.0.142
 CONTAINER_BUILDER?=podman
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
--- a/ai-services/assets/catalog/podman/values.yaml
+++ b/ai-services/assets/catalog/podman/values.yaml
@@ -4,7 +4,7 @@ ui:
 
 backend:
   port: ""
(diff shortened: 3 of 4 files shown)


yussufsh:
@mayuka-c I see hardcoded values at https://github.com/IBM/project-ai-services/blob/main/ai-services/assets/components/llm/vllm-cpu/podman/templates/vllm-server.yaml.tmpl#L27-L28
Could you also fix them here? please validate if we have not missed anything else.

mayuka-c:
> @mayuka-c I see hardcoded values at https://github.com/IBM/project-ai-services/blob/main/ai-services/assets/components/llm/vllm-cpu/podman/templates/vllm-server.yaml.tmpl#L27-L28 Could you also fix them here? please validate if we have not missed anything else.

oh do you mean referencing from values.yaml file (max_model_len and batch_size). Can I take it up separately if fine?

ai-services/assets/components/llm/vllm-cpu/podman/metadata.yaml
```diff
@@ -8,5 +8,5 @@ podTemplateExecutions:
 # Resource requirements
 resources:
   cpu: 12
-  memory: 107374182400 # 100Gi in bytes
+  memory: 161061273600 # 150GB in bytes
```
mkumatag:
```suggestion
  memory: 161061273600 # 150Gi in bytes
```

yussufsh:
I don't mind a different PR but I see some changes which are still not in sync, particularly from https://github.com/IBM/project-ai-services/commit/42ba7514fe546cf325029dcab84ca9498ba326c5#diff-233e1312d64e828dca70a13c62af3f1f7b3facfe863e34a2fda3aa5a8babac74

mkumatag:
> I don't mind a different PR but I see some changes which are still not in sync, particularly from [42ba751#diff-233e1312d64e828dca70a13c62af3f1f7b3facfe863e34a2fda3aa5a8babac74](https://github.com/IBM/project-ai-services/commit/42ba7514fe546cf325029dcab84ca9498ba326c5#diff-233e1312d64e828dca70a13c62af3f1f7b3facfe863e34a2fda3aa5a8babac74)

but we shouldn't depend how applications use them IMO, we need to have standard default settings for the models we server and allow users configure(advanced options may be in future)

yussufsh:
> but we shouldn't depend how applications use them IMO, we need to have standard default settings for the models we server and allow users configure(advanced options may be in future)

Yes, but I am just making sure we are not diverging from applications manifests which were extensively tested. We can take this up later as well but want to ensure we are not missing anything.

mayuka-c:
> I don't mind a different PR but I see some changes which are still not in sync, particularly from [42ba751#diff-233e1312d64e828dca70a13c62af3f1f7b3facfe863e34a2fda3aa5a8babac74](https://github.com/IBM/project-ai-services/commit/42ba7514fe546cf325029dcab84ca9498ba326c5#diff-233e1312d64e828dca70a13c62af3f1f7b3facfe863e34a2fda3aa5a8babac74)

Arent those all in sync. It is just that its hardocded in template and not getting from values.yaml which should be fine right?

There in that PR they are referencing it everywhere since it was a single values.yaml. Now for each component we have a different values.yml

Do you see any divergence in values?

ai-services/assets/components/llm/vllm-cpu/podman/metadata.yaml
mayuka-c: Done

yussufsh (approved):
Approving based on https://github.com/IBM/project-ai-services/pull/1064
Thanks @mayuka-c for looking at the differences.

PR #1062, opened 2026-06-25
fix: add actions to udev rules

Shubhamag12:
certain actions are required to take rules into effect.
udevadm trigger with the default change action does not re-apply group ownership to virtual VFIO device nodes that already exist on disk. So changed udevadm trigger to use --action=add|change to force re-evaluation of GROUP assignment on existing virtual device nodes.


yussufsh (approved):

mkumatag:
Please ensure your PRs include clear and detailed descriptions. Providing additional context around the changes helps the team review them more effectively and keeps everyone on the same page.

PR #1064, opened 2026-06-26
Ensure service/components assets are accurate

mayuka-c:
- Jira issue: [AISERVICES-1320](https://jsw.ibm.com/browse/AISERVICES-1320)
- Compared services/components assets with application assets to ensure we are not missing anything.
- Remove unnecesary fields from values.yaml and use references of it in templates wherever applicable.

- The only major discrepancy found was for summarize, below envs were not set under services assets which is been added
```
        - name: LLM_MAX_MODEL_LEN
          value: "{{ .Values.llm.maxModelLen }}"
        - name: LLM_MAX_BATCH_SIZE
          value: "{{ .Values.llm.maxBatchSize }}"
        - name: LLM_API_KEY
          value: "{{ .Values.llm.apiKey }}"
```
- Also watsonx provider for llm had `restartPolicy` missing which needs to be set to `always`

### DIFF
--- a/ai-services/assets/components/llm/watsonx/podman/values.yaml
+++ b/ai-services/assets/components/llm/watsonx/podman/values.yaml
@@ -3,6 +3,5 @@ model: "ibm/granite-4-h-small"
 watsonxApiKey: ""
 watsonxProjectId: ""
 watsonxUrl: ""
-maxNumBatchedTokens: 26208
 maxModelLen: 26208
 maxBatchSize: 32
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.144
+TAG?=v0.0.145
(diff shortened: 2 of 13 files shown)


ai-services/assets/components/llm/watsonx/podman/values.yaml
```diff
@@ -3,6 +3,5 @@ model: "ibm/granite-4-h-small"
 watsonxApiKey: ""
 watsonxProjectId: ""
 watsonxUrl: ""
-maxNumBatchedTokens: 26208
 maxModelLen: 26208
 maxBatchSize: 32
```
yussufsh, line 7: We are not using these as well right?

mayuka-c: Its being used in chat, digitize template
https://github.com/IBM/project-ai-services/blob/main/ai-services/assets/services/chat/podman/templates/chat-bot.yaml.tmpl#L75

yussufsh (dismissed):
lgtm

mkumatag (approved):

PR #1066, opened 2026-06-26
Fix job deletion affecting document due to cascade delete

dharaneeshvrd:
- Currently when we delete a job, underlying docs also getting cleaned up in db due to the cascade delete.
- Changing it to `set null` when the parent job is deleted.

Fixes https://jsw.ibm.com/browse/AISERVICES-1318


manju956 (approved):
lgtm

PR #675, opened 2026-04-25
must-gather tool

mkumatag:

PR #1067, opened 2026-06-26
[Catalog] Implement Log Sanitizer

mayuka-c:
- Implement a log sanitizer so that sensitive information are not logged (redacted)
- Sanitize at logger pkg so that its extensible and avoids printing/adding any sensitive info as part of args in logs.

Ref for sanitizing patterns: https://github.com/IBM/project-ai-services/pull/675

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.149
+TAG?=v0.0.150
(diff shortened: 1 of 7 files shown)


yussufsh:
Do we also want to sanitize API responses?

mayuka-c:
> Do we also want to sanitize API responses?

I thought of doing it, but couldnt find proper way for handling sanitization for API responses. Lets see if I can find a way for API response. Maybe will it take it separately or I'm thinking not to even send err object for this release in API response

yussufsh (approved):

PR #1068, opened 2026-06-26
[Catalog] Set Gin mode to release

mayuka-c:
- Add env within catalog template and set GIN_MODE to `release`.
- If not set uses `dev` mode.


yussufsh (dismissed):

mkumatag (commented):
may be you need to find a way to override this..

mayuka-c:
> may be you need to find a way to override this..

We could as part of catalog configure if we want to override gin-mode but maybe we can do it later is what I feel

mkumatag (approved):

PR #1070, opened 2026-06-26
[Podman] Do secret mounting instead of using secretRef

mayuka-c:
- Jira ID: [AISERVICES-1323](https://jsw.ibm.com/browse/AISERVICES-1323)
- Currently with secretRef, when we do podman inspect -> it displays the apiKey/password referenced from the secrets which can be a security concern.
- So migrating to use mounting as it just displays the mountPath without exposing secrets.

### DIFF
--- a/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
+++ b/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
@@ -7,7 +7,7 @@ metadata:
     ai-services.io/template: "{{ .AppTemplateName }}"
     ai-services.io/version: "{{ .Version }}"
     ai-services.io/secret: "catalog-secret"
-    ai-services.io/volume: "podman-auth-secret"
+    ai-services.io/volume: "podman-auth-secret,catalog-secret"
   annotations:
     run.oci.keep_original_groups: "1"
     ai-services.io/routes: "8081:catalog-ui:ui, 8080:catalog-api:api"
--- a/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
+++ b/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
@@ -20,9 +20,8 @@ spec:
       command: ["/bin/sh", "-c"]
       args:
         - |
+          set -e
           export DB_PASSWORD=$(cat /etc/secret/catalog-db-secret/db-password)
-
-          set -e  # Exit immediately if any command fails
 
           max_retries=5
           retry_delay=5
--- a/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
+++ b/ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
@@ -20,6 +20,8 @@ spec:
       command: ["/bin/sh", "-c"]
       args:
         - |
+          export DB_PASSWORD=$(cat /etc/secret/catalog-db-secret/db-password)
+
           set -e  # Exit immediately if any command fails
 
           max_retries=5
@@ -49,17 +51,15 @@ spec:
               exit 1
             fi
           done
-      env:
-        - name: DB_PASSWORD
-          valueFrom:
-            secretKeyRef:
-              name: catalog-db-secret
-              key: db-password
       resources:
         requests:
           memory: "512Mi"
         limits:
           memory: "512Mi"
+      volumeMounts:
+        - name: catalog-db-secret
+          mountPath: /etc/secret/catalog-db-secret
+          readOnly: true
   containers:
     - name: ui
       image: "{{ .Values.ui.image }}"
@@ -95,7 +95,10 @@ spec:
           {{- end }}
       command: ["/bin/sh", "-c"]
       args:
-        - "/usr/bin/ai-services catalog apiserver --port=8080 --admin-username=admin --admin-password-hash=${ADMIN_PASSWORD} --runtime={{ .Values.backend.runtime }}"
+        - |
+          export ADMIN_PASSWORD=$(cat /etc/secret/catalog-secret/admin-password)
+          export DB_PASSWORD=$(cat /etc/secret/catalog-db-secret/db-password)
+          exec /usr/bin/ai-services catalog apiserver --port=8080 --admin-username=admin --admin-password-hash=${ADMIN_PASSWORD} --runtime={{ .Values.backend.runtime }}
       env:
         - name: GIN_MODE
           value: "release"
@@ -107,22 +110,12 @@ spec:
           value: "unix://{{ .Values.backend.podman.uri }}"
         - name: REGISTRY_AUTH_FILE
           value: "/run/containers/auth.json"
-        - name: ADMIN_PASSWORD
-          valueFrom:
-            secretKeyRef:
-              name: catalog-secret
-              key: admin-password
         - name: DB_HOST
           value: {{ .AppName }}--db
         - name: DB_PORT
           value: "5432"
         - name: DB_USER
           value: "{{ .Values.db.user }}"
-        - name: DB_PASSWORD
-          valueFrom:
-            secretKeyRef:
-              name: catalog-db-secret
-              key: db-password
         - name: AI_SERVICES_BASE_DIR
           value: "{{ .BaseDir }}"
         - name: CADDY_ADMIN_URL
@@ -154,6 +147,12 @@ spec:
           readOnly: true
         - name: ai-services-data
           mountPath: {{ .BaseDir }}
+        - name: catalog-secret
+          mountPath: /etc/secret/catalog-secret
+          readOnly: true
+        - name: catalog-db-secret
+          mountPath: /etc/secret/catalog-db-secret
+          readOnly: true
       resources:
         requests:
           podman.io/device=/dev/vfio: 4
@@ -178,3 +177,9 @@ spec:
       hostPath:
         path: {{ .BaseDir }}
         type: DirectoryOrCreate
+    - name: catalog-secret
+      secret:
+        secretName: catalog-secret
+    - name: catalog-db-secret
+      secret:
+        secretName: catalog-db-secret
--- a/ai-services/assets/components/llm/vllm-cpu/podman/templates/vllm-server.yaml.tmpl
+++ b/ai-services/assets/components/llm/vllm-cpu/podman/templates/vllm-server.yaml.tmpl
@@ -32,8 +32,7 @@ spec:
           {{- if .Values.apiKey }}
           export VLLM_API_KEY=$(cat /etc/secret/vllm-secret/apiKey)
           {{- end }}
-          exec /usr/local/bin/python -m vllm.entrypoints.openai.api_server \
-            --model /models/{{ .Values.model }} \
+          exec vllm serve /models/{{ .Values.model }} \
             --served-model-name {{ .Values.model }} \
             --max-num-batched-tokens={{ .Values.maxNumBatchedTokens }} \
             --max-model-len={{ .Values.maxModelLen }} \
--- a/ai-services/assets/components/llm/vllm-cpu/podman/templates/vllm-server.yaml.tmpl
+++ b/ai-services/assets/components/llm/vllm-cpu/podman/templates/vllm-server.yaml.tmpl
@@ -6,6 +6,7 @@ metadata:
     ai-services.io/template: "{{ .TemplateID }}"
     {{- if .Values.apiKey }}
     ai-services.io/secret: "vllm-secret-{{ .InstanceSlug }}"
+    ai-services.io/volume: "vllm-secret-{{ .InstanceSlug }}"
     {{- end }}
   annotations:
     io.podman.annotations.ulimit: "nofile=134217728:134217728,memlock=-1:-1"
--- a/ai-services/assets/components/llm/vllm-cpu/podman/templates/vllm-server.yaml.tmpl
+++ b/ai-services/assets/components/llm/vllm-cpu/podman/templates/vllm-server.yaml.tmpl
@@ -16,17 +16,27 @@ spec:
       hostPath:
         path: "{{ .BaseDir }}/models"
         type: Directory
+    {{- if .Values.apiKey }}
+    - name: vllm-secret
+      secret:
+        secretName: "vllm-secret-{{ .InstanceSlug }}"
+    {{- end }}
   containers:
     - name: llm
       image: "{{ .Values.image }}"
+      command: ["/bin/sh"]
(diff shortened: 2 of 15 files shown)


mkumatag:
looks promising, @dharaneeshvrd you also review the changes.

mayuka-c:
I'm yet to validate. Will make it `Ready` once done.
PR can be reviewed meanwhile

mayuka-c:
PR is ready for review

Have validated all scenarios, working fine
- Spyre
- CPU
- WatsonX
- Setting VLLM Api Key
- Performing ingestion/summarize/chatbot

cc. @mkumatag @yussufsh @dharaneeshvrd

ai-services/internal/pkg/catalog/apiserver/services/sync/sync_service.go
```diff
@@ -696,11 +696,19 @@ func (s *SyncService) extractResourceLabelsFromPodSpec(podSpec *modelpkg.PodSpec
 	}
 
 	if secretLabel, ok := podSpec.Labels["ai-services.io/secret"]; ok && secretLabel != "" {
-		counts.SecretNames = append(counts.SecretNames, secretLabel)
+		for name := range strings.SplitSeq(secretLabel, ",") {
+			if name = strings.TrimSpace(name); name != "" {
+				counts.SecretNames = append(counts.SecretNames, name)
+			}
+		}
 	}
```
mayuka-c, line 704: This was needed as currently we have secret,volume labels with comma separated and during sync we need to ensure all secrets/volumes are present.

mkumatag (dismissed):
please get an ack from other 2 folks as well :)

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
```diff
@@ -7,7 +7,7 @@ metadata:
     ai-services.io/template: "{{ .AppTemplateName }}"
     ai-services.io/version: "{{ .Version }}"
     ai-services.io/secret: "catalog-secret"
-    ai-services.io/volume: "podman-auth-secret"
+    ai-services.io/volume: "podman-auth-secret,catalog-secret"
```
yussufsh, line 10: Missing `catalog-db-secret` here?

ai-services/assets/components/llm/vllm-cpu/podman/templates/vllm-server.yaml.tmpl
```diff
@@ -6,6 +6,7 @@ metadata:
     ai-services.io/template: "{{ .TemplateID }}"
     {{- if .Values.apiKey }}
     ai-services.io/secret: "vllm-secret-{{ .InstanceSlug }}"
+    ai-services.io/volume: "vllm-secret-{{ .InstanceSlug }}"
```
yussufsh, line 9: But the volume name is `vllm-secret` at the bottom of this file right?
Same with others. Did we check if they are getting deleted properly?

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
```diff
@@ -20,6 +20,8 @@ spec:
       command: ["/bin/sh", "-c"]
       args:
         - |
+          export DB_PASSWORD=$(cat /etc/secret/catalog-db-secret/db-password)
```
yussufsh, line 24: After reading into env should we rm the files? I think we might need it in readiness checks since it is different shell session but if not used in readiness is it a good idea to remove?
We can think about this later as well if needed.

yussufsh:
@mayuka-c Please also verify `--reset-*` flow during catalog configure once to make sure it works fine with these changes.

mkumatag (commented):
## Code Review Summary(Automated Bob's review comments)

Thanks for this PR — the security motivation is solid and the implementation is consistent across all templates. Posting inline comments on specific issues below. Overall the approach is correct: mounting secrets as files and reading them in a shell wrapper prevents values from leaking via `podman inspect`. A few things need attention before merging.

@yussufsh @mayuka-c see if these are valid review comments from Bob?

ai-services/assets/services/digitize/podman/templates/digitize.yaml.tmpl
```diff
@@ -4,8 +4,6 @@ metadata:
   name: "digitize-{{ .InstanceSlug }}"
   labels:
     ai-services.io/template: "{{ .TemplateID }}"
```
mkumatag, line 6: **🔴 High — Secret lifecycle risk: `ai-services.io/secret` removed from the consuming pod**

The `ai-services.io/secret: "digitize-db-secret-{{ .InstanceSlug }}"` and `ai-services.io/secret-skip-cleanup: "true"` labels have been removed from the `digitize` pod and moved to `postgres.yaml.tmpl`. However, the digitize pod *also* mounts and reads `digitize-db-secret` in both its init container and backend-server.

If the sync service uses these labels to track secret lifecycle (cleanup), and the postgres pod is torn down (or its labels are processed) before the digitize pod is done, the secret could be garbage-collected while still in active use by the digitize containers.

**Suggestion:** Keep `ai-services.io/secret: "digitize-db-secret-{{ .InstanceSlug }}"` and `ai-services.io/secret-skip-cleanup: "true"` on the digitize pod since it is also a consumer of that secret. The label should live on the pod that *owns* the secret's lifecycle — and in this case both pods depend on it.

ai-services/assets/services/digitize/podman/templates/digitize.yaml.tmpl
```diff
@@ -4,8 +4,6 @@ metadata:
...
-    ai-services.io/secret: "digitize-db-secret-{{ .InstanceSlug }}"
-    ai-services.io/secret-skip-cleanup: "true"
     ai-services.io/volume: "digitize-cache-{{ .InstanceSlug }}"
```
mkumatag, line 7: **🟠 Medium — `opensearch-secret` and `vllm-secret` volumes not tracked in `ai-services.io/volume` label**

The opensearch secret volume (and the conditional `vllm-secret`) are added to the pod spec but are not listed in the `ai-services.io/volume` label. This means `extractResourceLabelsFromPodSpec` will not include them in volume cleanup tracking.

Compare to how `opensearch.yaml.tmpl` correctly handles this:
```yaml
ai-services.io/volume: "opensearch-{{ .InstanceSlug }},opensearch-secret-{{ .InstanceSlug }}"
```

**Suggestion:** Update the label to include all tracked volumes:
```yaml
ai-services.io/volume: "digitize-cache-{{ .InstanceSlug }},opensearch-secret-{{ .Values.vector_store.instanceSlug }}"
```
And conditionally append `vllm-secret-{{ .Values.llm.instanceSlug }}` when `.Values.llm.apiKey` is set (if the label supports templating).

ai-services/assets/services/chat/podman/templates/chat-bot.yaml.tmpl
```diff
@@ -9,6 +9,10 @@ metadata:
     # ai-services.io/ports: "{{ .Values.ui.port }}:3000,{{ .Values.backend.port }}:5000"  # Uncomment to expose ports directly (bypasses Caddy)
 spec:
   restartPolicy: always
+  volumes:
```
mkumatag, line 12: **🟠 Medium — `opensearch-secret` volume not tracked in any label**

The opensearch secret volume is added to the pod spec here, but this pod has no `ai-services.io/volume` label at all. The volume will not be tracked for cleanup by `extractResourceLabelsFromPodSpec`.

**Suggestion:** Add a volume tracking label to the pod metadata:
```yaml
ai-services.io/volume: "opensearch-secret-{{ .Values.vector_store.instanceSlug }}"
```

ai-services/assets/services/similarity/podman/templates/similarity-api.yaml.tmpl
```diff
@@ -9,22 +9,25 @@ metadata:
     # ai-services.io/ports: "{{ or .Values.similarity.port }}:7000"  # Uncomment to expose ports directly (bypasses Caddy)
 spec:
   restartPolicy: always
+  volumes:
```
mkumatag, line 12: **🟠 Medium — `opensearch-secret` volume not tracked in any label**

Same issue as `chat-bot.yaml.tmpl`: the opensearch secret volume is mounted but this pod has no `ai-services.io/volume` label, so the volume won't be tracked for cleanup.

**Suggestion:** Add to the pod metadata labels:
```yaml
ai-services.io/volume: "opensearch-secret-{{ .Values.vector_store.instanceSlug }}"
```

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
mkumatag, line 10: **🟡 Low — `export` runs before `set -e`, inconsistent with other templates**

The `export DB_PASSWORD=$(cat ...)` is inserted *before* `set -e`. If the `cat` fails (e.g., volume not mounted, wrong path), the error will be silently swallowed and the script will proceed with an empty `DB_PASSWORD`, leading to a confusing downstream failure rather than a fast, obvious abort.

All other templates in this PR do not use `set -e` at all, making this inconsistent.

**Suggestion:** Either move the `export` after `set -e`:
```sh
set -e
export DB_PASSWORD=$(cat /etc/secret/catalog-db-secret/db-password)
```
Or remove `set -e` here and align with the rest of the templates (the retry loop already handles failures explicitly).

ai-services/internal/pkg/catalog/apiserver/services/sync/sync_service.go
```diff
@@ -696,11 +696,19 @@ func (s *SyncService) extractResourceLabelsFromPodSpec(podSpec *modelpkg.PodSpec
...
 
 	if secretLabel, ok := podSpec.Labels["ai-services.io/secret"]; ok && secretLabel != "" {
-		counts.SecretNames = append(counts.SecretNames, secretLabel)
+		for name := range strings.SplitSeq(secretLabel, ",") {
+			if name = strings.TrimSpace(name); name != "" {
+				counts.SecretNames = append(counts.SecretNames, name)
```
mkumatag, line 701: **✅ Correct fix — but note: `strings.SplitSeq` requires Go 1.24+**

The comma-split loop is the right approach to support the new multi-value labels like `"postgres-catalog,catalog-db-secret"`. `strings.SplitSeq` (iterator form) was introduced in Go 1.24 — confirmed compatible since the project is on Go 1.26.

The `strings.TrimSpace` guard against accidental whitespace is a nice touch. ✅

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
mayuka-c: catalog-db-secret is present within the catalog-db pod as its responsibility of catalog-db pod to manage the secret/volume.

Here: https://github.com/IBM/project-ai-services/pull/1070/changes#diff-bcb58cae41bdb6eb5647a2acbc127fee5b8819337bbe0363187a43a5c26b2ce1R11

ai-services/assets/components/llm/vllm-cpu/podman/templates/vllm-server.yaml.tmpl
mayuka-c: Ah, the volume present here with name is internal volumes ref: https://github.com/IBM/project-ai-services/pull/1070/changes#diff-0bd225c6776738d2f3d71e0f7003db9b72e83299e1adc5a981b2609117f2d816R28

The global volume created will be with slug and its coming from secretName from above. This is similar to persistentVolumeClaim

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
mayuka-c: This is read only mount, not sure if we can delete it.
But we can see it later

ai-services/assets/services/digitize/podman/templates/digitize.yaml.tmpl
mayuka-c: This will not be required because opensearch-secret (along with volume) is owned by opensearch component lifecycle.
So opensearch should have the resposibility to manage it.

If we include it here in digitize, if digitize pod gets deleted it will delete opensearch secret too (But this is wrong as component can be reused by multiple services). Same thing for vllm-key too.

ai-services/assets/services/similarity/podman/templates/similarity-api.yaml.tmpl
mayuka-c: Same as above

ai-services/assets/catalog/podman/templates/catalog.yaml.tmpl
mayuka-c: Done

ai-services/assets/services/digitize/podman/templates/digitize.yaml.tmpl
mayuka-c: I feel it should be part of producer (digitize-db) owning it similar to component.
There might be a situation where digitize-db could be used by multiple services. So if services are deleted first, then we should not be cleaning the digitize-db-secret volume.

PLMK your thoughts

mayuka-c:
> @mayuka-c Please also verify `--reset-*` flow during catalog configure once to make sure it works fine with these changes.

Yes will do a quick validation.

Edit: Validated reset password and auth, working fine.

ai-services/assets/services/chat/podman/templates/chat-bot.yaml.tmpl
mayuka-c: Same as above

mayuka-c:
> ## Code Review Summary(Automated Bob's review comments)
> Thanks for this PR — the security motivation is solid and the implementation is consistent across all templates. Posting inline comments on specific issues below. Overall the approach is correct: mounting secrets as files and reading them in a shell wrapper prevents values from leaking via `podman inspect`. A few things need attention before merging.
> 
> @yussufsh @mayuka-c see if these are valid review comments from Bob?

I have addressed and made few updates.
For few of them, I feel its not quite right. Have replied comments for respective suggestions.

ai-services/assets/services/digitize/podman/templates/digitize.yaml.tmpl
mkumatag: should be the responsibility of the service who creates this cred, if `digitize-db` then yes

mkumatag (approved):

dharaneeshvrd:
Changes lgtm, sry to look at late.

PR #1075, opened 2026-06-26
build: add multi-arch support and release targets

mkumatag:
## Summary

Two build-system improvements to `ai-services/Containerfile` and `ai-services/Makefile`.

Main purpose of this PR is for controlling tools reference via makefile during the release after the promotion

---

### Containerfile

**Multi-arch builder stage**
- Added `TARGETOS` / `TARGETARCH` ARGs (defaulting to `linux/ppc64le`) so the Go download URL is constructed from the platform args rather than being hard-coded.
- Enables `--platform`-based cross-builds (e.g. `podman build --platform linux/amd64`) without editing the file.

**Locked DNF repos in the runtime stage**
- Added a `DNF_FLAGS` ARG that restricts `microdnf install` to `ubi-9-baseos-rpms` and `ubi-9-appstream-rpms`.
- Prevents accidental package resolution from unintended repos that may be registered on the build host.

---

### Makefile

**`RELEASE_REGISTRY` variable**
- Introduces a separate `RELEASE_REGISTRY` (`icr.io/ai-services`) distinct from the default CICD registry (`icr.io/ai-services-private`).

**Configurable `TOOLS_IMAGE`**
- Adds `TOOLS_IMAGE_NAME`, `TOOLS_IMAGE_TAG`, and `TOOLS_IMAGE` variables so the tools image reference is overridable at call time without editing the file.

**`LDFLAGS` extracted as a top-level variable**
- Removes the duplicated inline `-ldflags` blocks in `BUILD_CMD` and the `binaries` target (DRY).
- Adds `-X vars.ToolImage=$(TOOLS_IMAGE)` to embed the tools image ref directly into built binaries.

**`release` targets**
- Adds `release`, `release-binaries`, `release-build`, and `release-push` targets.
- All release artifacts (binaries + container image) are built with `RELEASE_TOOLS_IMAGE` derived from `RELEASE_REGISTRY`, keeping the embedded ref and the pushed image in sync.
- The full ref can still be overridden independently if needed:
  ```
  make release RELEASE_REGISTRY=icr.io/my-org
  make release RELEASE_REGISTRY=icr.io/my-org TOOLS_IMAGE_TAG=0.12
  ```

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,4 +1,5 @@
 REGISTRY?=icr.io/ai-services-private
+RELEASE_REGISTRY?=icr.io/ai-services
 IMAGE=ai-services
 TAG?=v0.0.152
 CONTAINER_BUILDER?=podman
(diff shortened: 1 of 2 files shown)


adarshagrawal38:
Can we add a comment [here](https://github.com/IBM/project-ai-services/blob/main/ai-services/internal/pkg/vars/var.go#L18)

Stating that values will be overridden by ldflags, during the Go binary build via the make target.

adarshagrawal38:
I think we also need to update release workflow [here](https://github.com/mkumatag/project-ai-services/blob/3b51383c14cb51253e6a414bca8b5aca375a6b0b/.github/workflows/release.yml#L29)

Suggestion
```
- name: Build Go binaries for all platforms (cross-compilation)
   working-directory: ai-services
   run: |
      make release-binaries
```

ai-services/Makefile
```diff
@@ -106,6 +114,29 @@ push:
...
+release-binaries:
+	$(MAKE) binaries TOOLS_IMAGE=$(RELEASE_TOOLS_IMAGE)
+.PHONY: release-binaries
+
+release-build:
+	$(MAKE) build REGISTRY=$(RELEASE_REGISTRY) TOOLS_IMAGE=$(RELEASE_TOOLS_IMAGE)
+.PHONY: release-build
+
+release-push:
+	$(MAKE) push REGISTRY=$(RELEASE_REGISTRY)
```
adarshagrawal38, line 137: If we directly push image to release registry, then code signature pipelines won't be able to sign the image, with current implementation.

adarshagrawal38:
[here](https://github.com/mkumatag/project-ai-services/blob/3b51383c14cb51253e6a414bca8b5aca375a6b0b/hack/cicd/code-signing-service#L241) as well Make target would change from binaries to release-binaries

mkumatag:
@adarshagrawal38 well, this is getting complicated, we can revisit later.. let us go ahead with promoting the tool image alone and update the ref in the code directly.

mkumatag:
We don't need this for now, will revisit if needed later

PR #1077, opened 2026-06-29
[Podman] Add SupplementGroup during reboot

mayuka-c:
- During system reboot, the podman-restart systemd takes care of restarting all the pods. There is an issue where the pods requiring `vfio` devices fails with permission denied as sentient group is not supplied by podman restart systemd due to which the `catalog, llm, reranker (requiring spyre)` fails with permission denied.

Fix: The fix is to ensure that we add the override.conf with `SupplementaryGroups: sentient` while configuring podman-restart systemd during `ai-services bootstrap` so that necessary groups are supplied when it restarts.

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.154
+TAG?=v0.0.155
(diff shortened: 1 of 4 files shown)


yussufsh (dismissed):

mkumatag:
@mayuka-c please rebase and bump the version again.

mkumatag (approved):

PR #1078, opened 2026-06-29
Pagination related fixes

pranithraoibm:
Before

- Tables only ever showed 10 rows, even when there were more
- No way to see rows beyond the first 10 — pagination bar never appeared
- Deleting the last item on a page showed a blank table
- The "Filter by service" filter only searched through rows visible on screen, not all data
- Exporting downloaded only what was on the current page (max 20 rows)

After

- Tables now show up to 20 rows by default and fetch the correct page from the server
- Pagination bar appears when there are more than 20 total items, allowing navigation through all data
- Deleting the last item on a page automatically takes you back to the previous page
- "Filter by service" now filters on the server — works across all data, not just what's on screen. Changed the filter menu option from checkbox to radio button as backend supports only one filter type at a time (Type string as seen [here](https://github.com/IBM/project-ai-services/blob/efa0ac6359d51e6d631aa7cf07c28dce5287ad51/ai-services/docs/swagger.yaml#L556))
- Export now fetches all data (up to 100 rows as limit fixed by the GET API call), applies any active search or filter, and respects hidden columns

Known Limitation:
Search and sort by status column value works only for the current page data

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.156
+TAG?=v0.0.157
 CONTAINER_BUILDER?=podman
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.155
+TAG?=v0.0.156
 CONTAINER_BUILDER?=podman
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.157
+TAG?=v0.0.156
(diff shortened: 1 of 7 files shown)


ui/catalog/src/components/DeployedServicesTable/DeployedServicesTable.tsx
ryanmarc, line 1: Please don’t feed paginated results into the existing full-list cache. setDeployedServices(rawData) now stores only the current page, but the cache path later treats that array as the full/page-1 result and sets totalItems to cachedServices.length. Either bypass this cache for paginated reads or key it by page/filter and store pagination metadata with it.

ryanmarc (changes requested):
This only exports the first page with page_size=100, so tenants with more than 100 rows get a successful but incomplete CSV. Please fetch all pages until pagination.total_pages / total_items is exhausted, or explicitly block/label the export limit. This needs to be fixed in both DeployedServicesTable and DigitalAssistants.

ui/catalog/src/components/DeployedServicesTable/DeployedServicesTable.tsx
pranithraoibm: Fixed by bypassing the cache for paginated reads

pranithraoibm:
Export now fetches all pages sequentially (has_next loop with page_size=100) instead of a single capped request. We have has_next in the GET API response. The modal button shows "Exporting..." and is disabled while pages are being fetched, closing only after all data is collected and the CSV is ready.

ryanmarc (dismissed):

ai-services/Makefile
```diff
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.155
+TAG?=v0.0.157
```
mkumatag: why not .156?

prachikurade1: sorry, my bad. The UI code was not updated in this PR, so had to increment again for UI image, and incremented for ai-services as well. Fixed it.

mkumatag (approved):

PR #1087, opened 2026-07-01
Optimizations for Chatbot

sats-23:
-To help keep code more maintainable and scalable 

Changes:
-Predefine config maps for languages to avoid map creation on every call
-Reuse connection for all similarity search calls made via chatbot
-New helper function _stream_error_response for all StreamingResponse exceptions

### DIFF
--- a/services/chatbot/backend_utils.py
+++ b/services/chatbot/backend_utils.py
@@ -1,10 +1,21 @@
 import time
 import requests
+from requests.adapters import HTTPAdapter
 from common.misc_utils import get_logger
 from common.validation_utils import validate_query_length as _validate_query_length
 from chatbot.settings import settings
 
 logger = get_logger("backend_utils")
+
+_similarity_session = None
+
+def get_similarity_session():
+    global _similarity_session
+    if _similarity_session is None:
+        _similarity_session = requests.Session()
+        adapter = HTTPAdapter(pool_connections=5, pool_maxsize=10)
+        _similarity_session.mount("http://", adapter)
+    return _similarity_session
 
 
 def validate_query_length(query, emb_endpoint):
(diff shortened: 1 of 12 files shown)


services/chatbot/backend_utils.py
```diff
@@ -1,11 +1,22 @@
...
 logger = get_logger("backend_utils")
 
+_similarity_session = None
+
+def get_similarity_session():
+    global _similarity_session
+    if _similarity_session is None:
+        _similarity_session = requests.Session()
+        adapter = HTTPAdapter(pool_connections=5, pool_maxsize=10)
+        _similarity_session.mount("http://", adapter)
```
dharaneeshvrd, line 17:
```suggestion
        _similarity_session.mount("http://", adapter)
        _similarity_session.mount("https://", adapter)
```
Let's add https as well

services/chatbot/backend_utils.py
```diff
@@ -1,11 +1,22 @@
...
 
 logger = get_logger("backend_utils")
 
+_similarity_session = None
+
+def get_similarity_session():
+    global _similarity_session
+    if _similarity_session is None:
+        _similarity_session = requests.Session()
+        adapter = HTTPAdapter(pool_connections=5, pool_maxsize=10)
```
dharaneeshvrd: Not sure we really need 5 connections, let's keep it to 3 to even handle scaled opensearch
For max size, we need to use `settings.common.llm.max_batch_size` to handle the num of concurrent requests.

dharaneeshvrd (commented):
Other changes lgtm

services/chatbot/backend_utils.py
sats-23: Done

services/chatbot/backend_utils.py
sats-23: Done

dharaneeshvrd (approved):

PR #1089, opened 2026-07-01
[CVE-FIX] Bump up python pkg to fix cves

adarshagrawal38:
Bump up ujson and  pydantic-settings python pkg to fix moderate cves.


yussufsh (approved):
lgtm

PR #1090, opened 2026-07-01
Update Initial Catalog API & DB design proposals

mayuka-c:
- This is to ensure that Catalog API design proposed initially is updated with latest implementation.
- This ensures that the design doc is in sync with latest implementation done.
- Also ensuring the DB design proposal is in sync with actual DB schema implementation.


yussufsh (approved):
lgtm

PR #1094, opened 2026-07-02
[Catalog-API] Populate catalog_id and version in List Applications

mayuka-c:
- Currently the catalog_id and version key is present in List Applications but is empty (Since we are using the same underlying struct for both List Applications and GetAppByID).
- Instead of it being empty, let the values be populated.


yussufsh (approved):
lgtm

PR #1095, opened 2026-07-02
docs: add model management service proposal

mkumatag:
## Summary

Adds a full implementation plan for a **dynamic Model Management Service** that replaces the current static vLLM pod deployment (models hard-wired into the application pod at creation time) with an on-demand model lifecycle system backed by a universal LiteLLM gateway.

## Motivation

Currently, model deployments are static — they are bundled into the application pod template and deployed as a single unit with the application. There is no way to:
- Deploy or swap models independently from the application lifecycle
- Use different model backends (vLLM, WatsonX, OpenAI-compatible) through a unified interface
- Gate deployments on available hardware resources (CPU, memory, Spyre accelerator cards)
- Manage models via a REST API or CLI at runtime

## Proposal

The plan introduces a `ModelManager` interface with runtime-aware implementations (Podman, OpenShift, Docker Compose) and promotes **LiteLLM as the universal model gateway** — all consumer services talk to LiteLLM, which dynamically routes to the correct backend.

### Architecture diagrams included

- Component interaction overview
- Model deployment flow (step-by-step including resource pre-flight)
- Resource gating decision tree (CPU / Memory / Spyre accelerator cards)
- LiteLLM as universal gateway with dynamic route table
- Weight source strategies by runtime
- Sub-task dependency graph

### 10 sub-tasks planned

| # | Sub-Task |
|---|----------|
| 1 | `ModelManager` interface and core types |
| 2 | Resource pre-flight check package (CPU, memory, Spyre cards) |
| 3 | LiteLLM gateway promotion and dynamic route management |
| 4 | Podman `ModelManager` implementation |
| 5 | OpenShift `ModelManager` implementation |
| 6 | Docker Compose `ModelManager` implementation |
| 7 | CLI commands: `ai-services model deploy/undeploy/list/status` |
| 8 | REST API: `/api/v1/models` endpoints on Catalog API Server |
| 9 | Application template updates to route through LiteLLM |
| 10 | Documentation and developer guide |

### Weight source strategies per runtime

- **Podman**: local host volume, Hugging Face pull-at-start
- **OpenShift**: modelcar init-container, PersistentVolumeClaim, ICR image
- **Docker Compose**: bind-mount volume, Hugging Face pull-at-start

## Files changed

- `docs/proposals/model-management-service-plan.md` — new proposal document

## No code changes

This PR contains only a proposal document. Implementation will follow as separate PRs per sub-task once the proposal is reviewed and approved.

### DIFF
--- a/docs/proposals/model-management-service-plan.md
+++ b/docs/proposals/model-management-service-plan.md
@@ -1,0 +1,579 @@
+# Model Management Service — Implementation Plan
+
+## Overview
+
+### Goal
+Replace the current static, hard-wired model deployment (models bundled inside the application pod at creation time) with a **dynamic Model Management Service** that can deploy, undeploy, and list model inference backends on demand, regardless of runtime. All consumer services talk exclusively through a **LiteLLM Gateway** — they never reach vLLM, WatsonX, or any other backend directly.
+
+### Scope
+- A new `model-manager` internal package with a runtime-aware `ModelManager` interface.
+- A **resource pre-flight check** that queries available system resources (CPU, memory, accelerator cards) before any deployment to fail-fast with a clear error when capacity is insufficient.
+- New `ai-services model deploy / undeploy / list / status` CLI commands.
+- New `/api/v1/models` REST endpoints added to the existing Catalog API Server (Gin).
+- LiteLLM Gateway promoted from a static WatsonX-only component to the **universal model gateway** — reconfigured dynamically via its Admin API when models are added or removed.
+- Runtime-specific model weight source strategies:
+  - **Podman**: local host volume (`~/.local/share/ai-services/models`), Hugging Face pull-at-start.
+  - **OpenShift**: modelcar init-container, PersistentVolumeClaim, IBM Container Registry (ICR) image.
+  - **Docker Compose**: bind-mount volume, Hugging Face pull-at-start.
+- No changes to existing consumer services (`chatbot`, `digitize`, `similarity`, `summarize`) — they continue to read `LLM_ENDPOINT`, `EMB_ENDPOINT`, `RERANKER_ENDPOINT` from environment variables, but those will now always point to the LiteLLM Gateway.
+
+### Non-Goals
+- UI/dashboard for model management.
+- Multi-tenant model isolation.
+- GPU auto-scaling / auto-provisioning.
+- Changes to Python service code.
+
+---
+
+## Architecture Overview
+
+### Component Interaction
+
+```mermaid
+flowchart TD
+    DevCI["Developer / CI"]
+
+    subgraph Orch["Orchestration Layer"]
+        CLI["ai-services model deploy/undeploy/list"]
+        API["POST /api/v1/models\nCatalog API Server"]
+    end
+
+    DevCI --> CLI
+    DevCI --> API
+
+    subgraph MMLayer["ModelManager interface - runtime-aware factory"]
+        Podman["Podman impl"]
+        OCP["OpenShift impl"]
+        DC["Docker Compose impl"]
+    end
+
+    CLI --> MMLayer
+    API --> MMLayer
+
+    Preflight["Pre-flight resource check\nCPU / Memory / Accelerator cards"]
+
+    Podman --> Preflight
+    OCP --> Preflight
+    DC --> Preflight
+
+    subgraph Backends["Backends"]
+        vLLM["vLLM pod\ninstruct / emb / reranker role"]
+        LiteLLM["LiteLLM Gateway pod\nalways-on port 4000\nAdmin API manages route table"]
+        WatsonX["WatsonX.ai API\nno local pod needed"]
+    end
+
+    Preflight --> vLLM
+    Preflight --> LiteLLM
+    Preflight --> WatsonX
+
+    vLLM -- "registered via LiteLLM Admin API" --> LiteLLM
+
+    subgraph Consumers["Consumer Services"]
+        Chatbot["chatbot\nLLM_ENDPOINT → litellm:4000/v1"]
+        Digitize["digitize\nEMB_ENDPOINT → litellm:4000/v1"]
+        Similarity["similarity\nLLM_ENDPOINT → litellm:4000/v1"]
+    end
+
+    LiteLLM --> Chatbot
+    LiteLLM --> Digitize
+    LiteLLM --> Similarity
+```
+
+### Model Deployment Flow (per runtime)
+
+```mermaid
+flowchart TD
+    Start(["ai-services model deploy\n--provider vllm-spyre\n--model ibm-granite/granite-3.3-8b-instruct\n--role llm --source local --app myapp"])
+
+    Step1["1. Load component metadata\nassets/components/llm/vllm-spyre/podman/metadata.yaml\nresources: cpu=8, memory=150Gi, accelerators={ibm.com/spyre_pf: 4}"]
+
+    Step2["2. Resource Pre-flight Check\nruntime.GetSystemInfo()\nCPU available? Memory available? Spyre cards free?"]
+
+    Fail(["Abort with PreflightError\nlist all violated constraints"])
+
+    Step3["3. Render pod template\nvllm-server.yaml.tmpl\nwith InstanceSlug, BaseDir, Values"]
+
+    Step4["4. runtime.CreatePod\nPod: llm-instanceSlug\nLabels: ai-services.io/role=model\nai-services.io/model-role=llm"]
+
+    Step5["5. Poll runtime.InspectPod\nwait for liveness probe to pass"]
+
+    Step6["6. Register route in LiteLLM Gateway\nPOST /model/new\nmodel_name=llm, api_base=llm-instanceSlug:8000/v1"]
+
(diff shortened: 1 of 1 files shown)


docs/proposals/model-management-service-plan.md
```diff
@@ -0,0 +1,579 @@
...
+
+    Preflight["Pre-flight resource check\nCPU / Memory / Accelerator cards"]
+
+    Podman --> Preflight
+    OCP --> Preflight
+    DC --> Preflight
+
+    subgraph Backends["Backends"]
+        vLLM["vLLM pod\ninstruct / emb / reranker role"]
+        LiteLLM["LiteLLM Gateway pod\nalways-on port 4000\nAdmin API manages route table"]
```
mayuka-c, line 61: Will we have 1 instance of LiteLLM pod already deployed even before model deploy is done?

I guess this LiteLLM pod takes care of routing the traffic to VLLM and watsonx respectively?

docs/proposals/model-management-service-plan.md
```diff
@@ -0,0 +1,579 @@
...
+    Preflight --> vLLM
+    Preflight --> LiteLLM
+    Preflight --> WatsonX
+
+    vLLM -- "registered via LiteLLM Admin API" --> LiteLLM
+
+    subgraph Consumers["Consumer Services"]
+        Chatbot["chatbot\nLLM_ENDPOINT → litellm:4000/v1"]
+        Digitize["digitize\nEMB_ENDPOINT → litellm:4000/v1"]
+        Similarity["similarity\nLLM_ENDPOINT → litellm:4000/v1"]
```
mayuka-c, line 74: How does the endpoint work (I see in the diagram its same across)?, how does it route appropriately based on EMB, LLM, RERANKER Endpoint?

docs/proposals/model-management-service-plan.md
```diff
@@ -0,0 +1,579 @@
...
+    LiteLLM --> Chatbot
+    LiteLLM --> Digitize
+    LiteLLM --> Similarity
+```
+
+### Model Deployment Flow (per runtime)
+
+```mermaid
+flowchart TD
+    Start(["ai-services model deploy\n--provider vllm-spyre\n--model ibm-granite/granite-3.3-8b-instruct\n--role llm --source local --app myapp"])
```
mayuka-c, line 86: I guess role here is `component_Type`?
Also why do we need --source and also what is `--app` (Insnt model deploy agnostic of application)?

mayuka-c, line 86: Also when we say `model deploy`, it will only deploy 1 instance/pod of a given model?.
Can we have multiple instances of same model deployed?. If yes then that is where `--app` being used?

docs/proposals/model-management-service-plan.md
```diff
@@ -0,0 +1,579 @@
...
+
+    Fail(["Abort with PreflightError\nlist all violated constraints"])
+
+    Step3["3. Render pod template\nvllm-server.yaml.tmpl\nwith InstanceSlug, BaseDir, Values"]
+
+    Step4["4. runtime.CreatePod\nPod: llm-instanceSlug\nLabels: ai-services.io/role=model\nai-services.io/model-role=llm"]
+
+    Step5["5. Poll runtime.InspectPod\nwait for liveness probe to pass"]
+
+    Step6["6. Register route in LiteLLM Gateway\nPOST /model/new\nmodel_name=llm, api_base=llm-instanceSlug:8000/v1"]
```
mayuka-c, line 100: For watsonx, I guess we will skip above steps and just do the register route in LiteLLM?

docs/proposals/model-management-service-plan.md
```diff
@@ -0,0 +1,579 @@
...
+    Step2 -- "any check fails" --> Fail
+    Step2 -- "all pass" --> Step3 --> Step4 --> Step5 --> Step6 --> Step7 --> Done
+```
+
+### Resource Gating Decision Tree
+
+```mermaid
+flowchart TD
+    A(["Deploy request received"])
+    B["Load required resources\nfrom component metadata\ncpu, memory, storage, accelerators"]
```
mayuka-c, line 116: `Deploy request` meaning when we are actually deploying the application?
So during that time, we were currently selecting the model but I guess now user can only select the deployed model?

docs/proposals/model-management-service-plan.md
```diff
@@ -0,0 +1,579 @@
...
+    CpuOk{"CPU available\n>= required?"}
+    CpuErr(["error: insufficient CPU\nneed X, have Y available"])
+
+    MemOk{"Memory available\n>= required?"}
+    MemErr(["error: insufficient memory\nneed XGi, have YGi available"])
+
+    SpyreCheck{"Provider ==\nvllm-spyre?"}
+    FindCards["spyre.FindFreeCards"]
+    CardsOk{"free_count >=\nrequired spyre_pf?"}
+    CardsErr(["error: insufficient Spyre cards\nneed X, Y currently free"])
```
mayuka-c, line 128: If my above comment is right, then any reason we need to do spyre checks here during deploy flow as we would be reusing the deployed model?

docs/proposals/model-management-service-plan.md
```diff
@@ -0,0 +1,579 @@
...
+
+**Intent**
+Promotes LiteLLM from a static WatsonX-only component to the **universal model gateway** for all providers. Adds a Go client that wraps the LiteLLM Admin API so any runtime implementation can register and deregister model routes at deploy/undeploy time.
+
+**Expected Outcomes**
+- New package `ai-services/internal/pkg/model_manager/litellm/client.go` with `AddRoute`, `RemoveRoute`, `ListRoutes` wrapping LiteLLM Admin API (`POST /model/new`, `DELETE /model/delete`, `GET /model/info`).
+- `RouteConfig` struct captures provider-specific fields: for vLLM — `api_base`, optional `api_key`; for WatsonX — `watsonx_project_id`, `watsonx_url`, `api_key`; for OpenAI-compatible — `api_base`.
+- Updated `images/litellm/config/config.yaml` starts with an **empty `model_list`** — all routes are added dynamically at runtime. The `LITELLM_MASTER_KEY` env var gates Admin API access.
+- New standalone `litellm-server` pod template: `ai-services/assets/components/llm-gateway/litellm/podman/templates/litellm-server.yaml.tmpl` (and OpenShift Deployment equivalent) — always deployed as phase-1 infra alongside postgres/opensearch, not co-located with any vLLM container.
```
mayuka-c, line 295: Instead of being a component, maybe we can have this included as part of catalog assets where we currently deploy catalog ui, backend, caddy and also the litellm pod

docs/proposals/model-management-service-plan.md
```diff
@@ -0,0 +1,579 @@
...
+**Relevant Context**
+- Existing OpenShift application package: `ai-services/internal/pkg/application/openshift/`.
+- Helm package: `ai-services/internal/pkg/helm/`.
+- Runtime interface: `ListRoutes`, `DeletePVCs`, `InspectPod` already present.
+
+**Status** — `[ ] pending`
+
+---
+
+### Sub-Task 6 — Docker Compose `ModelManager` Implementation
```
mayuka-c, line 389: Maybe Docker Compose we can remove?

mayuka-c (commented):
These are some of my intial Queries, PTAL.
Didnt quite completely go through implementation, might require some time :)

docs/proposals/model-management-service-plan.md
mkumatag: you can ignore for now, this is picked because I have docker compose file in my local repo(under construction)

docs/proposals/model-management-service-plan.md
mkumatag: may be based on the model name we register inside the litellm, maybe some litellm experts can comment @dharaneeshvrd @hasan4791 @sats-23

docs/proposals/model-management-service-plan.md
mkumatag: yes, one single router takes care of all the routing.

docs/proposals/model-management-service-plan.md
mkumatag: some bob generated stuff, we can remove what is not required and readjust ;)

mkumatag: pod(gateway) is single entity and can server multiple entities(models - different or even same models)

may be we need double click here..

docs/proposals/model-management-service-plan.md
dharaneeshvrd: Yes https://github.com/IBM/project-ai-services/blob/511f1fda30dd42c8fab47a7c803dca92bd0abe4f/images/litellm/config/config.yaml#L2
When we have different models, based on the model name, mapped litellm config would be utilised to route the request.

dharaneeshvrd (commented):
Overall lgtm 
Only thing is to note is that if we want to onboard a provider other than vllm and WatsonX, it requires significant code changes to add the pass through endpoints like how we did it for WatsonX.
Other option is to have a local tokenizer which brings challenges like downloading model files onto the service pods.
If we don't have plans on supporting new provider in near term, we are good with this design for now.

PR #1099, opened 2026-07-02
ci: add concurrency controls to all GitHub Actions workflows

yussufsh:
## Summary

Adds workflow-level concurrency controls to all 22 GitHub Actions workflows, following [GitHub Actions concurrency docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency).

## Changes

1. Added to every workflow file:

```yaml
concurrency:
  group: "${{ github.workflow }}-${{ github.ref }}"
  cancel-in-progress: true
```

2. Deleted stale `.github/workflows/python-base-image.yml` file.
This was failing because the dir path did not exist. On further investigation, it felt as not used at all, hence removed.

## Behaviour

Trigger | github.ref value | Effect
-- | -- | --
pull_request | refs/pull/<PR_NUMBER>/merge | Redundant pushes to a PR branch cancel the previous run for that PR only
push to main | refs/heads/main | Rapid successive merges to main are serialised; older run is cancelled
push tag (v*) | refs/tags/v1.2.3 | Each tag ref is unique; no unintended cancellation


dharaneeshvrd (approved):

mkumatag:
> Rapid successive merges to main are serialised; older run is cancelled

One concern I have is that if we cancel an older build that is still running, it will skip building that specific version and only push the newer one. For example, the image for ai-service:v0.0.159 might never get pushed if the build is cancelled by merging changes for the next version (v0.0.160).

yussufsh:
> One concern I have is that if we cancel an older build that is still running, it will skip building that specific version and only push the newer one. For example, the image for ai-service:v0.0.159 might never get pushed if the build is cancelled by merging changes for the next version (v0.0.160).

Excellent catch.
Push to main is a real problem.

The fix is to only cancel in-progress runs for pull_request events (where we just want the latest commit to be tested), and never cancel runs triggered by a push to main.

Will raise a PR for the fix.

PR #1100, opened 2026-07-03
fix: AISERVICES-1355: handle create parameters for components correctly

yussufsh:
- Provider selection with =false now opts out: previously llm.vllm-cpu=false would still select vllm-cpu. Now =false is treated as "don't use this provider."
- Two providers active at once now errors: previously if two providers were both active for the same component type, one was silently picked at random due to Go map iteration. Now it fails with a clear message.
- Any value other than true/false on a bare provider key logs a warning and defaults to opt-out: previously it would silently select the provider.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/create.go
+++ b/ai-services/cmd/ai-services/cmd/application/create.go
@@ -737,7 +737,11 @@ func buildServiceEntryWithDeployOptions(appClient *catalogClient.ApplicationClie
 		providerParams := extractComponentParamsForService(serviceID, compDeployOpt.Type, argParams)
 
 		// Determine provider ID and get its params
-		providerID, userParams := selectProviderFromDeployOptions(compDeployOpt, providerParams)
+		providerID, userParams, err := selectProviderFromDeployOptions(compDeployOpt, providerParams)
+		if err != nil {
+			return apiModels.Service{}, err
+		}
+
 		if providerID == "" {
 			return apiModels.Service{}, fmt.Errorf("no provider found for component type '%s'", compDeployOpt.Type)
 		}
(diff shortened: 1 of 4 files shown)


ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -870,6 +874,10 @@ func extractComponentParamsForService(serviceID string, componentType string, al
 }
 
 // extractProviderParams extracts provider parameters from allParams with the given prefix.
+// For bare provider keys (e.g., llm.vllm-cpu=true/false):
+//   - "true" selects the provider.
+//   - "false" opts out of the provider (skipped).
+//   - Any other value logs a warning and is treated as "false".
```
mayuka-c, line 889: Shouldnt we be rejecting with error for unknown value set?

ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -902,42 +921,62 @@ func extractProviderParams(prefix string, allParams map[string]string, providerP
...
+	if providerID != "" {
+		return providerID, params, nil
 	}
 
 	// Special logic for LLM and reranker component types - prefer Spyre acceleration.
-	if providerID := findSpyreProvider(compDeployOpt); providerID != "" {
-		return providerID, make(map[string]string)
+	if spyreID := findSpyreProvider(compDeployOpt); spyreID != "" {
+		return spyreID, make(map[string]string), nil
 	}
```
mayuka-c, line 950: Can we make use of default set within the components metadata?
In this way it will be generic and it will be config driven tomo if we want to change: https://github.com/IBM/project-ai-services/blob/main/ai-services/assets/components/reranker/vllm-spyre/metadata.yaml#L10

mayuka-c: By default llm and reranker will have default: true for `vllm-spyre` provider

yussufsh: Good catch. Let us fix it in a different PR.

mayuka-c (dismissed):
LGTM

mayuka-c (approved):
LGTM

PR #1103, opened 2026-07-06
AISERVICES-1232: Supporting caddy for non-root user

adarshagrawal38:
Refactors the Caddy Dockerfile to use a multi-stage build pattern, separating the Go build environment from the final runtime image
Switching to caddy user via the `USER` directive to ensure the process does not run as root.


adarshagrawal38:
will post a new PR

PR #1105, opened 2026-07-07
Fetch max length from litellm & lang multipliers for token budget

sats-23:
-Fetch correct max_length in following failover order - litellm (/model/info) > vLLM (/v1/models) > Fallback param passed
-Ratio multiplier on history and query rephrasing budget

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.172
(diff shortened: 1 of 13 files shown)


mkumatag:
@sats-23 can you update the description with the detailed information like issue we are trying to solve and the proposal?

dharaneeshvrd (approved):
Let's rename the PR title and description to reflect the current changes we are making

sats-23:
> Let's rename the PR title and description to reflect the current changes we are making

Done

PR #1106, opened 2026-07-07
fix : error out on non-running pod during catalog configure re-run

shwetha-s-poojary:
#### Issue:

If ai-services catalog configure --runtime podman is run without logging into the registry, the catalog pod fails to pull the image and remains in the Created state.
After logging in and re-running catalog configure, the CLI reports:
`✔ Catalog service already deployed`
even though the catalog pod is not running

#### Expected behavior:
It should verify that the catalog pod is Running, not just that it exists.
If the pod is in Created (or any non-running) state, treat the deployment as incomplete and redeploy.

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.164
+TAG?=v0.0.165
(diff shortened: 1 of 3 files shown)


yussufsh (approved):
lgtm
@mayuka-c ptal

mayuka-c (approved):
LGTM

PR #1107, opened 2026-07-07
handle failed chunk sumamrize process in async summarization for large files

manalilatkar:
When you stop the LLM server mid-summarizing for a large file, we see a correct tally of chunks that completed summarizing and chunks that failed with the correct error status.[
Screenshot 2026-07-07 at 5 08 38 PM
](url)

### DIFF
--- a/services/summarize/app.py
+++ b/services/summarize/app.py
@@ -726,40 +726,72 @@ async def process_summarization_job(job_id: str, level):
                         True
                     )
                     
-                    async with concurrency_limiter:  # Global vLLM limit
-                        chunk_result, chunk_in_tokens, chunk_out_tokens = await asyncio.to_thread(
-                            query_vllm_summarize,
-                            llm_endpoint=llm_endpoint,
-                            messages=chunk_messages,
-                            model=llm_model,
-                            max_tokens=chunk_info['max_tokens'],
-                            temperature=settings.summarize.summarization_temperature,
-                        )
-                    
-                    if isinstance(chunk_result, dict) and "error" in chunk_result:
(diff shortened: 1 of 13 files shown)


services/summarize/app.py
```diff
@@ -726,40 +726,72 @@ async def process_chunk(chunk_text: str, chunk_index: int):
...
+                        logger.error(
+                            f"Chunk {chunk_index + 1}/{num_chunks} failed for job {job_id}: {chunk_exc}",
+                            exc_info=True
+                        )
+                        # Record the failure immediately in DB metadata before re-raising
+                        async with metadata_lock:
+                            job_record = db_repo.get_job_by_id(job_id)
+                            if job_record:
+                                current_metadata = job_record.job_metadata or {}
+                                current_metadata["failed_chunks"] = current_metadata.get("failed_chunks", 0) + 1
```
dharaneeshvrd: Just wondering, do we need a field called failed_chunks? since we are stopping the summarization as soon as a chunk failed and not proceeding the summarization for remaining ones. So it would be just one, it won't increase at all.
I think just marking the whole job as failure would be sufficient 
Wdyt?

manalilatkar: Fixed as discussed.

services/summarize/models.py
dharaneeshvrd: I see this still exists here.

dharaneeshvrd: This should use `JobMetadata` defined down below.

manalilatkar: removed.

manalilatkar: fixed

dharaneeshvrd (approved):

PR #1108, opened 2026-07-08
Propagate error status code in non-stream

sats-23:
-Earlier, in case of 4xx/5xx exceptions, chatbot would return the exception wrapped in a 200 (non-stream only)
-Enhanced to raise HTTPError as is for non-stream cases
-Fixed utility to use right logger and redirect to sys out (earlier used root logger)

Screenshot 2026-07-08 at 08 57 23

image

### DIFF
--- a/services/chatbot/app.py
+++ b/services/chatbot/app.py
@@ -636,11 +636,7 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
                 return response_data
 
             APIError.raise_error(ErrorCode.LLM_ERROR, "Unexpected response format from LLM")
-        except HTTPException:
-            # Re-raise HTTPException to preserve status codes
-            raise
         except requests.exceptions.HTTPError as e:
-            # Propagate upstream 4xx/5xx status codes (vLLM/LiteLLM) back to the caller
             status_code = e.response.status_code if e.response is not None else 502
             logger.error(f"Error in non-streaming response: {e}", exc_info=True)
             raise HTTPException(
@@ -654,11 +650,16 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
                 },
             )
         except Exception as e:
-            # For non-streaming requests, return error in chat response format
-            error_message = f"Error: {str(e)}"
-            logger.error(f"Error in non-streaming response: {error_message}", exc_info=True)
-            return ChatCompletionResponse(
-                choices=[ChatChoice(message=ChatMessage(content=error_message))]
+            logger.error(f"Error in non-streaming response: {e}", exc_info=True)
+            raise HTTPException(
+                status_code=500,
+                detail={
+                    "error": {
+                        "code": ErrorCode.LLM_ERROR.value,
+                        "message": str(e),
+                        "status": 500,
+                    }
+                },
             )
         finally:
             # Release semaphore for non-streaming requests
--- a/services/chatbot/app.py
+++ b/services/chatbot/app.py
@@ -8,6 +8,7 @@ from fastapi.security import HTTPBearer, HTTPAuthorizationCredentials
 from fastapi.openapi.docs import get_swagger_ui_html
 from fastapi.responses import StreamingResponse, Response
 import json
+import requests
 from contextlib import asynccontextmanager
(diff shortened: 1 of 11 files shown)


services/chatbot/app.py
```diff
@@ -8,21 +8,23 @@
...
-from common.misc_utils import set_log_level
+from common.misc_utils import set_log_level, get_logger
 from common.lang_utils import detect_language, LanguageCodes, get_max_tokens_map
 
 from chatbot.settings import settings
 from chatbot.conversation_utils import get_conversation_context, truncate_history_by_tokens
 from chatbot.query_rephrasing import rephrase_query_with_context
 
 set_log_level(settings.common.app.log_level)
+logger = get_logger("chatbot")
```
dharaneeshvrd, line 27: What? we never had a logger for chatbot :(

services/chatbot/app.py
```diff
@@ -637,10 +639,24 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
...
+                        "status": status_code,
+                    }
+                },
+            )
         except Exception as e:
             # For non-streaming requests, return error in chat response format
             error_message = f"Error: {str(e)}"
-            logging.error(f"Error in non-streaming response: {error_message}", exc_info=True)
+            logger.error(f"Error in non-streaming response: {error_message}", exc_info=True)
             return ChatCompletionResponse(
```
dharaneeshvrd: Why are we not raising exception here too?

sats-23: This is the final except case after handling all HTTPError, so this is just for a random except which we may have no encountered or from an intermediate component. This is rare to it, but we can make it return same status code if needed

services/chatbot/app.py
sats-23: It always went to the root logger. 
I noticed that when the error exceptions logged with "ERROR: root"

We had the setup_comprehensive_crash_handler which used the logger (named chatbot) but it too never redirected to sys out. 

Fixed both 
-Redirect to sout - `console_handler = logging.StreamHandler(sys.stdout)`
-Made sure all occurrences used this logger

services/chatbot/app.py
```diff
@@ -637,10 +639,24 @@ async def chat_completion(req: ChatCompletionRequest, credentials: Optional[HTTP
         except HTTPException:
             # Re-raise HTTPException to preserve status codes
```
dharaneeshvrd: What is the difference between this and below exception?

services/common/misc_utils.py
```diff
@@ -137,7 +138,7 @@ def get_logger(name):
         # Add the filter to inject request_id
         logger.addFilter(RequestIDFilter())
 
-        console_handler = logging.StreamHandler()
+        console_handler = logging.StreamHandler(sys.stdout)
```
dharaneeshvrd, line 141: Why this change is required? dont it need to handle stderr too?

services/chatbot/app.py
dharaneeshvrd: Yes, it should return 5xx here right?
I don't know why we decided to return in chat response format like the comment says, lets correct it now.

services/common/misc_utils.py
sats-23: Without specifying, it by default used to go to stderr. Since these are part of the main application logs, we need to specify param to redirect it to stdout

services/chatbot/app.py
sats-23: Usually HTTPException is thrown and caught by FastAPI (for inbound requests)
HTTPError is for requests library (outbound requests).
This case where we catch exceptions coz of LiteLLM failures falls under the second one, hence the new block

sats-23: I have removed this first exception

services/chatbot/app.py
sats-23: Sure, done

dharaneeshvrd (commented):
Let's update the legacy templates as well
And we can merge post that

sats-23:
> Let's update the legacy templates as well And we can merge post that

Done, PTAL

dharaneeshvrd (approved):

PR #1109, opened 2026-07-08
fix: cancel in-progress runs for pull_request events

yussufsh:
## Summary

Added checks for cancelling in-progress runs only for pull requests and not for push to branches or tags.

## Behaviour

Trigger | github.ref value | cancel-in-progress | Effect
-- | -- | -- | --
pull_request | refs/pull/<PR_NUMBER>/merge | true | Redundant pushes to a PR branch cancel the previous run for that PR only
push to main | refs/heads/main | false  | Every run completes, every version's image gets pushed
push tag (v*) | refs/tags/v1.2.3 | false | Release build always completes
schedule | (image scanner) | false | Scheduled scans never cancelled

### DIFF
--- a/.github/workflows/ai-services-image.yml
+++ b/.github/workflows/ai-services-image.yml
@@ -14,7 +14,7 @@ on:
 
 concurrency:
   group: "${{ github.workflow }}-${{ github.ref }}"
-  cancel-in-progress: true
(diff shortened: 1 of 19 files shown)


mayuka-c (approved):
LGTM

adarshagrawal38 (approved):
LGTM

PR #1111, opened 2026-07-08
Chatbot Optimizations

sats-23:
-Moved _SPECIAL_TOKENS, _SUPPORTED to a frozenset
-Removed dependency on OrderedDict
-Normalise messages[-1] separately instead of building a full normalised list then immediately re-slicing it
-Remove unused params in search_only

### DIFF
--- a/services/common/perf_utils.py
+++ b/services/common/perf_utils.py
@@ -1,12 +1,11 @@
 import threading
 import time
 from datetime import datetime
-from collections import OrderedDict
 from common.misc_utils import get_request_id
 
 class PerfMetricsRegistry:
     def __init__(self, max_size=1000):
-        self._metrics = OrderedDict()
+        self._metrics = {}
         self._max_size = max_size
         self._lock = threading.Lock()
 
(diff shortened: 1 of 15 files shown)


services/common/perf_utils.py
```diff
@@ -1,12 +1,11 @@
 import threading
 import time
 from datetime import datetime
-from collections import OrderedDict
 from common.misc_utils import get_request_id
 
 class PerfMetricsRegistry:
     def __init__(self, max_size=1000):
-        self._metrics = OrderedDict()
+        self._metrics = {}
```
dharaneeshvrd, line 8: Why this change required?

dharaneeshvrd (commented):
Remaining changes lgtm

services/common/perf_utils.py
sats-23: OrderedDict was the way to go before v3.7, but after that normal dictionaries preserve insertion order. 
Just replace as standard dict is much more optimized, uses lesser memory and is faster.

dharaneeshvrd (approved):

PR #1113, opened 2026-07-08
fix: use caddy subroutes to avoid hot reloads

shwetha-s-poojary:
#### Summary
Optimizes proxy routing performance by eliminating Caddy hot-reloads during route changes and migrating certificate provisioning to a native, on-the-fly approach.

#### Changes
Zero-Reload Subroutes (Commit 1): Moved route operations to the subroute handler's inner array (`/id/app-routes-handler/routes`). This changes route updates into in-memory mutations, eliminating the 1-2s connection drops and intermittent 500 errors during app deletion.
On-Demand TLS (Commit 2): Solves the `ERR_SSL_PROTOCOL_ERROR` introduced by the subroute change. Because in-memory mutations bypass the engine reload, Caddy stopped eagerly generating certificates for new domains. Moving to native `on_demand `TLS inside the Caddyfile ensures certificates are generated instantly (under 10ms) on the first client request.

#### Impact
Zero downtime during application creation or deletion.

#### Testing
Able to access endpoints

Tested: remove the retries workaround and test if we still see any transient 500 errors during route registration/unregistration, there should not be any caddy downtime.
image

### DIFF
--- a/ai-services/assets/catalog/podman/Caddyfile.tmpl
+++ b/ai-services/assets/catalog/podman/Caddyfile.tmpl
@@ -1,19 +1,26 @@
 {
 	admin 0.0.0.0:2019
 
-	servers :443 {
-		name {{ .CaddyServerName }}
-	}
+    servers :443 {
+        name {{ .CaddyServerName }}
+    }
+
+    # 1. Enable On-Demand TLS globally and point it to your local health check
+    on_demand_tls {
+        ask http://127.0.0.1:8080/health
+    }
 }
 
 :443 {
-	# 1. This initializes the routes array natively so it's never empty
-	handle /internal/health {
+    handle /internal/health {
         # @id internal-health
         respond "OK" 200
     }
 
-	tls internal
+    # 2. Tell the internal CA to generate certs when requested
+    tls internal {
+        on_demand
+    }
 }
 
 # Internal health check endpoint (Bypasses automatic HTTPS redirection)
--- a/ai-services/assets/catalog/podman/Caddyfile.tmpl
+++ b/ai-services/assets/catalog/podman/Caddyfile.tmpl
@@ -1,13 +1,16 @@
 {
-	admin 0.0.0.0:2019
+    admin 0.0.0.0:2019
 
     servers :443 {
         name {{ .CaddyServerName }}
     }
 
-    # 1. Enable On-Demand TLS globally and point it to your local health check
+    # Enable On-Demand TLS globally.
+    # Before issuing a certificate for any hostname Caddy calls the ask endpoint;
+    # 2xx = allow, non-2xx = deny. /tls/ask validates the requested domain query
+    # parameter against registered routes so certs are only issued for known hosts.
     on_demand_tls {
-        ask http://127.0.0.1:8080/health
+        ask {{ .CatalogBackendURL }}/tls/ask
     }
 }
 
@@ -17,13 +20,13 @@
         respond "OK" 200
     }
 
-    # 2. Tell the internal CA to generate certs when requested
+    # Tell the internal CA to generate certs on demand
     tls internal {
         on_demand
     }
 }
 
-# Internal health check endpoint (Bypasses automatic HTTPS redirection)
+# Internal endpoints (bypass automatic HTTPS redirection)
 :8080 {
     respond /health "OK" 200
 }
--- a/ai-services/internal/pkg/catalog/apiserver/router.go
+++ b/ai-services/internal/pkg/catalog/apiserver/router.go
@@ -10,12 +10,13 @@ import (
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog/apiserver/middleware"
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog/apiserver/repository"
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog/apiserver/services/auth"
+	"github.com/project-ai-services/ai-services/internal/pkg/proxy"
 	swaggerFiles "github.com/swaggo/files"
 	ginSwagger "github.com/swaggo/gin-swagger"
 )
 
 // CreateRouter sets up the Gin router with the necessary routes and authentication middleware for the API server.
-func CreateRouter(authSvc auth.Service, tokenMgr *auth.TokenManager, blacklist repository.TokenBlacklist, appService *repository.ApplicationService) *gin.Engine {
+func CreateRouter(authSvc auth.Service, tokenMgr *auth.TokenManager, blacklist repository.TokenBlacklist, appService *repository.ApplicationService, proxyManager proxy.ProxyManager) *gin.Engine {
 	if mode := os.Getenv("GIN_MODE"); mode != "" {
 		gin.SetMode(mode)
 	}
@@ -32,6 +33,35 @@ func CreateRouter(authSvc auth.Service, tokenMgr *auth.TokenManager, blacklist r
 	// Expose /health for liveness probes
 	router.GET("/health", func(c *gin.Context) {
 		c.JSON(http.StatusOK, gin.H{"message": "ok"})
+	})
+
+	// TLS ask endpoint — called by Caddy's on_demand_tls before issuing a certificate.
+	// Validates the requested domain against registered routes; returns 200 to allow or
+	// 403 to deny. If no proxy manager is configured, fails open (200) to avoid blocking
+	// legitimate cert issuance during startup.
+	router.GET("/tls/ask", func(c *gin.Context) {
+		domain := c.Query("domain")
+		if domain == "" || proxyManager == nil {
+			c.Status(http.StatusOK)
+			return
+		}
+
+		// Route IDs are the subdomain portion (the part before the first dot).
+		// GetRouteByID returns an error if the route is not registered in Caddy.
+		routeID := domain
+		for i, ch := range domain {
+			if ch == '.' {
+				routeID = domain[:i]
+				break
+			}
+		}
+
+		if _, err := proxyManager.GetRouteByID(routeID); err != nil {
+			c.Status(http.StatusForbidden)
+			return
+		}
+
+		c.Status(http.StatusOK)
 	})
 
 	// Swagger documentation endpoint
--- a/ai-services/internal/pkg/proxy/caddy.go
+++ b/ai-services/internal/pkg/proxy/caddy.go
@@ -113,9 +113,68 @@ func (c *caddyManager) RegisterRoute(ctx context.Context, route Route) error {
 	return c.createRoute(routeConfig)
 }
 
-// Helper to append a new route to the server's route array.
+// ensureSubroute guarantees the persistent subroute container exists in Caddy's live config.
+// It uses GET /id/app-routes-handler to check; if absent it creates the outer catch-all route
+// via a single POST to the top-level routes array (one reload, tolerated at first-use time).
+// On every subsequent call the GET returns 200 and the function is a no-op.
+func (c *caddyManager) ensureSubroute() error {
+	checkURL, err := url.JoinPath(c.adminURL, "id", constants.CaddySubrouteHandlerID)
+	if err != nil {
+		return err
+	}
+
+	resp, err := c.httpClient.R().Get(checkURL)
+	if err != nil {
+		return fmt.Errorf("failed to check subroute existence: %w", err)
+	}
+
+	if resp.StatusCode() == http.StatusOK {
+		// Already present — nothing to do.
+		return nil
+	}
+
+	// Subroute is absent; seed it via a single top-level POST.
+	// @id tags are JSON-only: Caddyfile has no syntax for them, so we create
+	// the subroute container here through the Admin API instead of the Caddyfile.
+	outerRoute := map[string]any{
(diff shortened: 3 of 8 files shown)


shwetha-s-poojary:
Config now with subroutes handlers:

```
{
  "admin": {
    "listen": "0.0.0.0:2019"
  },
  "apps": {
    "http": {
      "servers": {
        "ai_services": {
          "listen": [
            ":443"
          ],
          "routes": [
            {
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "handle": [
                        {
                          "body": "OK",
                          "handler": "static_response",
                          "status_code": 200
                        }
                      ]
                    }
                  ]
                }
              ],
              "match": [
                {
                  "path": [
                    "/internal/health"
                  ]
                }
              ]
            },
            {
              "@id": "app-routes",
              "handle": [
                {
                  "@id": "app-routes-handler",
                  "handler": "subroute",
                  "routes": [
                    {
                      "@id": "catalog-ui",
                      "handle": [
                        {
                          "handler": "reverse_proxy",
                          "upstreams": [
                            {
                              "dial": "ai-services--catalog:8081"
                            }
                          ]
                        }
                      ],
                      "match": [
                        {
                          "host": [
                            "catalog-ui.10.48.64.150.nip.io"
                          ]
                        }
                      ],
                      "terminal": true
                    },
                    {
                      "@id": "catalog-api",
                      "handle": [
                        {
                          "handler": "reverse_proxy",
                          "upstreams": [
                            {
                              "dial": "ai-services--catalog:8080"
                            }
                          ]
                        }
                      ],
                      "match": [
                        {
                          "host": [
                            "catalog-api.10.48.64.150.nip.io"
                          ]
                        }
                      ],
                      "terminal": true
                    }
                  ]
                }
              ],
              "terminal": false
            }
          ]
        },
        "srv1": {
          "listen": [
            ":8080"
          ],
          "routes": [
            {
              "handle": [
                {
                  "body": "OK",
                  "handler": "static_response",
                  "status_code": 200
                }
              ],
              "match": [
                {
                  "path": [
                    "/health"
                  ]
                }
              ]
            }
          ]
        }
      }
    },
    "tls": {
      "automation": {
        "on_demand": {
          "permission": {
            "endpoint": "http://127.0.0.1:8080/health",
            "module": "http"
          }
        },
        "policies": [
          {
            "issuers": [
              {
                "module": "internal"
              }
            ],
            "on_demand": true
          }
        ]
      }
    }
  }
}
```

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -113,9 +113,65 @@ func (c *caddyManager) RegisterRoute(ctx context.Context, route Route) error {
 	return c.createRoute(routeConfig)
 }
 
-// Helper to append a new route to the server's route array.
+// ensureSubroute guarantees the persistent subroute container exists in Caddy's live config.
+// It checks via GET /id/app-routes-handler; if absent it seeds the outer catch-all route
+// containing the subroute handler via a single top-level POST (one reload, first-use only).
+// On every subsequent call the GET returns 200 and the function is a no-op.
+func (c *caddyManager) ensureSubroute() error {
```
yussufsh: Can we wrap the seed path in a sync.Once?
`ensureSubroute` would be a no-op after the very first request. sync.Once costs nothing and eliminates the race condition entirely.

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -257,7 +314,14 @@ func RegisterRoutesForAppAndReturn(
 		return nil, fmt.Errorf("failed to build routes: %w", err)
 	}
 
-	// Step 3: Register each route with Caddy
+	// Step 3: Ensure the subroute container exists exactly once before registering any routes.
+	if cm, ok := proxyManager.(*caddyManager); ok {
+		if err := cm.ensureSubroute(); err != nil {
```
yussufsh: Bob suggested this as major change:
> RegisterRoute can be called without ensureSubroute — will silently fail ([caddy.go:77](vscode-webview://1oib9tqi8enmnn6jcmf486g15431kt819glrjb0m77i9s6s82c5o/ai-services/internal/pkg/proxy/caddy.go))
RegisterRoute is a ProxyManager interface method callable by any consumer (tests, future callers). It calls createRoute, which now POSTs to /id/app-routes-handler/routes — but never ensures the subroute container exists first. If called outside RegisterRoutesForAppAndReturn, Caddy returns a 404/error and the error message (caddy returned status … on creation) gives no hint at the real cause.
> Suggestion: Either move the ensureSubroute call into RegisterRoute

ai-services/assets/catalog/podman/Caddyfile.tmpl
```diff
@@ -1,19 +1,26 @@
...
 
-	servers :443 {
-		name {{ .CaddyServerName }}
-	}
+    servers :443 {
+        name {{ .CaddyServerName }}
+    }
+
+    # 1. Enable On-Demand TLS globally and point it to your local health check
+    on_demand_tls {
```
yussufsh, line 12: Bob:
> Caddy uses the ask endpoint to decide whether to allow certificate issuance — 2xx = allow, non-2xx = deny. Because /health always returns 200, Caddy will issue an internal cert for any hostname that triggers a TLS handshake, including typos or probes. Low risk with an internal CA, but worth documenting. Ideally a dedicated endpoint should validate the requested domain query parameter against registered routes.

ai-services/assets/catalog/podman/Caddyfile.tmpl
yussufsh, line 1: This file contains tabs and now spaces as well with the new code.
It's worth normalising for consistency; please use 4-space indentation.

yussufsh (changes requested):

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: moved the ensureSubroute inside RegisterRoute

ai-services/assets/catalog/podman/Caddyfile.tmpl
shwetha-s-poojary: That's a good catch - I will raise a fix soon.

shwetha-s-poojary: Added `/tls/ask `to the catalog backend as suggested , it checks the domain against Caddy's route table and returns 200/403 accordingly.One thing to be aware of: the ask endpoint is in the catalog backend pod, not in the Caddy pod, so Caddy makes a cross-pod call for every cert request. It works fine since both pods share the same network.If the catalog backend is down or not yet ready, /tls/ask returns 200 by default so Caddy doesn't accidentally block certs for real routes during startup.

Bob suggested that in  long term a sidecar in the Caddy pod would be cleaner - no cross-pod dependency.

@yussufsh PTAL and let me know your thoughts on this.

ai-services/internal/pkg/catalog/apiserver/router.go
```diff
@@ -34,6 +35,35 @@ func CreateRouter(authSvc auth.Service, tokenMgr *auth.TokenManager, blacklist r
 		c.JSON(http.StatusOK, gin.H{"message": "ok"})
 	})
 
+	// TLS ask endpoint — called by Caddy's on_demand_tls before issuing a certificate.
+	// Validates the requested domain against registered routes; returns 200 to allow or
+	// 403 to deny. If no proxy manager is configured, fails open (200) to avoid blocking
+	// legitimate cert issuance during startup.
+	router.GET("/tls/ask", func(c *gin.Context) {
+		domain := c.Query("domain")
+		if domain == "" || proxyManager == nil {
```
mayuka-c, line 44: WHy do we need to expose endpoint now?

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -76,11 +79,21 @@ func (c *caddyManager) HealthCheck() error {
...
+// going through RegisterRoutesForAppAndReturn.
 func (c *caddyManager) RegisterRoute(ctx context.Context, route Route) error {
 	if route.ID == "" {
 		return fmt.Errorf("cannot register route: route ID is empty")
 	}
 
+	// Ensure the subroute container exists before the first route is added.
+	// sync.Once guarantees exactly one attempt regardless of concurrent callers;
+	// the outcome (including any error) is cached for all subsequent calls.
+	if err := c.ensureSubroute(); err != nil {
```
mayuka-c: Why not create subroute container outside this? and then we start registering routes. In this way we dont need to have sync.Once and makes it clean.

After deploying the caddy, create the subroute container

ai-services/internal/pkg/catalog/apiserver/router.go
shwetha-s-poojary: To address this https://github.com/IBM/project-ai-services/pull/1113#discussion_r3556796206

mayuka-c: Any reason of why this is required now?

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: Done - moved subroute creation to startup in runAPIServer right after GetCaddyProxyManager succeeds, before the server begins accepting requests. EnsureSubroute is now in ProxyManager interface and sync.Once is gone.

ai-services/internal/pkg/catalog/apiserver/router.go
shwetha-s-poojary: Ok so the issue was -  on demand tls creates certs even when the domain is not valid/registered as  /health always returns 200.So we needed an endpoint to validate the legit domains.
Why are we using on demand TLS? - you can check the PR description.

Do you have anything better in mind?Happy to explore :)

mayuka-c: So this is the requirement for subroute?
What exactly is domain and who passes this?
Maybe can we have some call to discuss on this?

shwetha-s-poojary: > So this is the requirement for subroute?

So /tls/ask is just a gatekeeper who asks "do we know about this domain?It wouldn't be right to say it's a subroute requirement.

>What exactly is domain and who passes this?

Caddy passes it automatically as a query parameter e.g. `GET /tls/ask?domain=catalog-ui.example.com`  whenever it's about to issue an on-demand cert. We didn't add that call, it's built into Caddy's on_demand_tls behaviour.

> Maybe can we have some call to discuss on this?

Sure

shwetha-s-poojary:
Bob helped me summarise my experiments  😄 
@mayuka-c @yussufsh @mkumatag  need your directions here


#### The Core Problem

* **The Change:** We moved dynamic route updates to Caddy's **Zero-Reload Subroutes** (`POST /id/app-routes-handler/routes`) to perform in-memory mutations. This successfully eliminated the 1–2s connection drops and `500` errors caused by full engine reloads.
* **The Bug:** Because in-memory mutations bypass Caddy's full engine reload, Caddy's TLS provisioning step is skipped.
* **The Impact:** Caddy does not automatically generate certificates for newly registered domains. The first client connection to any new route fails with `ERR_SSL_PROTOCOL_ERROR`.

---

### Approach 1: On-Demand TLS (`ask` Endpoint) — *Implemented in this PR*

**How it solves the problem:**
Instead of trying to pre-issue certificates, we configure Caddy to generate them reactively. When the first client connection to a new domain arrives, Caddy generates the certificate on the fly.

* **Workflow:** Client TLS handshake $\rightarrow$ Caddy has no cert $\rightarrow$ Caddy queries `GET /tls/ask?domain=...` on our catalog backend $\rightarrow$ Backend returns `200` $\rightarrow$ Caddy instantly generates and caches the cert (<10ms).
* **Caveats:**
* **Requires Authorization Endpoint:** We must maintain the backend `ask` endpoint to validate domains, preventing attackers from exhausting resources by requesting certificates for fake domains.
* **First-Request Overhead:** The very first connection to a new domain carries a tiny latency penalty for the on-the-fly generation.

Note: Bob suggested that in long term a sidecar in the Caddy pod would be cleaner(no cross-pod dependency) rather than exposing the `/tls/ask` endpoint in catalog

---

### Approach 2: Programmatic Automation (`automateInternalCert`) — *Alternative Experimented*

**How it solves the problem:**
This is a proactive workaround. Right after the application registers a new route, it makes a follow-up API call to manually inject the new domain into Caddy's active certificate automation queue.

* **Workflow:** App creates route $\rightarrow$ App immediately calls Caddy's Admin API to append the hostname to `/config/apps/tls/certificates/automate` $\rightarrow$ Caddy's background loop immediately provisions and caches the certificate before any user connects.
* **Caveats:**
* **Complex Seeding Logic:** If the `tls` app block is missing in Caddy's live config, a straight POST fails. The application must perform a `GET` check, then run a `PUT` to seed the minimal configuration structure before appending.
* **Tight Coupling:** The application code becomes heavily dependent on Caddy's internal API schema, requiring custom logic to handle API failures and skip processing if wildcard certificates exist.

shwetha-s-poojary:
Closing this pull request for now as RCA didn't confirm Caddy was causing 500 transient errors.

PR #1117, opened 2026-07-09
Fix Table merging issue

manju956:
- Language-Agnostic Table Merging: Replaced hardcoded English continuation checks (like continued) with rapidfuzz. The logic now uses structural and fuzzy header matching, seamlessly supporting multi-page tables in German, French, and other languages.
- Fixed Empty Captions & Parser Glitches: Added `clean_markdown_table_and_caption`. This detects edge cases where the PDF parser mistakenly jams the caption into the table's header row. It now extracts the hidden caption properly and promotes the true column names back to the top.
- Updated `merge_markdown_tables` to dynamically detect and strip out duplicate header rows when joining tables together

### DIFF
--- a/services/digitize/processing/tables.py
+++ b/services/digitize/processing/tables.py
@@ -12,7 +12,7 @@ Responsibilities:
 
 import json
 import time
-import difflib
+from rapidfuzz import fuzz
 from pathlib import Path
 
 from common.lang_utils import LanguageCodes, get_prompt_for_language
@@ -91,10 +91,10 @@ def headers_match(headers1: list[str], headers2: list[str]) -> bool:
     return normalized1 == normalized2
 
 
-def is_table_continuation(headers1: list[str], headers2: list[str], caption1: str, caption2: str, fuzzy_threshold: float = 0.85) -> bool:
+def is_table_continuation(headers1: list[str], headers2: list[str], caption1: str, caption2: str, fuzzy_threshold: float = 85.0) -> bool:
     """
     Determines if a table is a continuation across pages.
-    Uses structural column matching and fuzzy caption matching to remain language-agnostic.
+    Uses structural column matching and RapidFuzz caption matching.
     """
     # 1. STRUCTURAL CHECK: Column headers MUST match perfectly
     if not headers_match(headers1, headers2):
@@ -123,9 +123,9 @@ def is_table_continuation(headers1: list[str], headers2: list[str], caption1: st
         if cap2.startswith(cap1) or cap1.startswith(cap2):
             return True
 
-        # 4. FUZZY MATCHING (Robust Path)
-        # Handles OCR errors or cases where prefixing is slightly broken
-        similarity = difflib.SequenceMatcher(None, cap1, cap2).ratio()
+        # 4. FUZZY MATCHING (RapidFuzz Path)
+        # fuzz.ratio calculates the normalized Indel distance (0 to 100)
+        similarity = fuzz.ratio(cap1, cap2)
         if similarity >= fuzzy_threshold:
             return True
 
--- a/services/digitize/processing/tables.py
+++ b/services/digitize/processing/tables.py
@@ -3,17 +3,18 @@ Table processing: extraction, merging, and summarisation.
 
 Responsibilities:
 - process_table  — extract/summarise tables from a converted document
-- extract_table_headers — parse markdown table header row
-- headers_match — compare two header lists
-- is_table_continuation — [NEW] verify continuation via structure & fuzzy caption matching
-- merge_markdown_tables — combine two markdown tables by dropping the second header
+- clean_markdown_table_and_caption — fix parser glitches where caption becomes table header
+- extract_table_headers — parse markdown table header row smartly
+- is_table_continuation — fuzzy matching of headers across pages
+- merge_markdown_tables — combine tables by dropping duplicate setup rows
 - merge_consecutive_tables — merge tables spanning consecutive pages
 """
 
 import json
 import time
-from rapidfuzz import fuzz
+import re
 from pathlib import Path
+from rapidfuzz import fuzz
 
 from common.lang_utils import LanguageCodes, get_prompt_for_language
 from common.llm_utils import summarize_and_classify_tables, tqdm_wrapper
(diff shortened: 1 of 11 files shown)


services/digitize/processing/tables.py
```diff
@@ -89,19 +91,47 @@ def headers_match(headers1: list[str], headers2: list[str]) -> bool:
...
+
+    # 3. PREFIX MATCHING (Fast Path)
+    # Checks if one is a direct prefix of the other (e.g. "Tabelle 19" vs "Tabelle 19 (Forts.)")
+    if cap1 and cap2:
+        if cap2.startswith(cap1) or cap1.startswith(cap2):
+            return True
+
+        # 4. FUZZY MATCHING (Robust Path)
+        # Handles OCR errors or cases where prefixing is slightly broken
+        similarity = difflib.SequenceMatcher(None, cap1, cap2).ratio()
```
dharaneeshvrd: Any specific reason to use this library?
FYI We already use fuzzy matching with `from rapidfuzz import fuzz` in other places.

services/digitize/settings.py
```diff
@@ -195,26 +195,25 @@ class GermanConfig(BaseSettings):
         )
         
         prompt: str = Field(
-        default="""Sie sind ein intelligenter Assistent, der Tabellen aus Dokumenten analysiert.
+        default="""Sie sind ein hochpräziser Datenextraktor. Ihre Aufgabe ist es, die folgende Markdown-Tabelle zu lesen und ihren Inhalt exakt und unverändert als saubere Markdown-Tabelle wiederzugeben.
```
dharaneeshvrd: What changes are you making in German prompt?

dharaneeshvrd (commented):
Can we please validate this logic works for other languages too?

services/digitize/settings.py
manju956: the earlier prompt was instructed to write bullet points and paragraphs, which was harming the merged marked down table to rewrite as bullets and form of an essay. new prompt forces the llm to passthrough markdown table data.

services/digitize/processing/tables.py
manju956: I wasn't aware of rapidfuzz usage, now used it for ufzzy matching.

services/digitize/settings.py
```diff
@@ -228,23 +227,28 @@ class GermanConfig(BaseSettings):
                 | Power10   | 16    | 8 TB     |
 
                 Ausgabe:
-                Summary: Die Tabelle präsentiert technische Spezifikationen für den Power10-Prozessor. Die Prozessorkonfiguration umfasst 16 Kerne für parallele Verarbeitungsfähigkeiten. Die Speicherkapazität unterstützt bis zu 8 TB (Terabyte) RAM und bietet erhebliche Speicherressourcen für Unternehmensworkloads und datenintensive Anwendungen.
+                Summary:
+                | Prozessor | Kerne | Speicher |
```
manalilatkar: This part should contain the summary which has been deleted, not the table again.
Same with the negative example below.

manalilatkar (commented):
just one comment about the prompt, rest LGTM.

services/digitize/settings.py
dharaneeshvrd: > new prompt forces the llm to passthrough markdown table data.

can you please elaborate a bit?

English prompt also uses `instructed to write bullet points and paragraphs` rule, what issues are we facing with this rule? do we need to change English prompt too?

services/digitize/settings.py
```diff
@@ -195,26 +195,25 @@ class GermanConfig(BaseSettings):
...
 
                 Ihre Aufgaben:
 
-                1. Extrahieren und dokumentieren Sie JEDE Information aus der Tabelle in umfassenden Details:
-                - Listen Sie ALLE Abschnitte, Unterabschnitte und deren Referenznummern auf, falls vorhanden
-                - Fügen Sie JEDE Spezifikation, Funktion, Nummer, Code, Bedingung und Anforderung hinzu
-                - Erwähnen Sie ALLE Elemente, auch wenn sie nebensächlich erscheinen - nichts sollte ausgelassen werden
-                - Verwenden Sie ein strukturiertes Format mit klarer Organisation (nummerierte Listen, Aufzählungspunkte oder detaillierte Absätze)
-                - Seien Sie äußerst gründlich und umfassend - streben Sie maximale Detailtiefe an
-                - Wenn die Tabelle mehrere Zeilen/Spalten hat, beschreiben Sie jede einzelne
```
Niharika0306, line 208: I agree. Seems like we are forcing it to output the same markdown text in the summary section , which we probably doesn't bore well for the summary expected and RAG context. 
Is there a sample output of a table summarised using this new prompt?

services/digitize/processing/tables.py
```diff
@@ -155,36 +185,41 @@ def merge_consecutive_tables(table_dict: dict) -> dict:
...
+            next_caption = next_table.get('caption', '')
             next_headers = extract_table_headers(next_markdown)
-            # Check if tables are on consecutive pages and have matching headers
+
+            # Check if tables are on consecutive pages, have matching headers AND pass the robust continuation check
             if (next_page is not None and
                     last_merged_page is not None and
                     next_page == last_merged_page + 1 and
-                    headers_match(current_headers, next_headers)):
+                    is_table_continuation(current_headers, next_headers, current_caption, next_caption)):
```
Niharika0306: Another small thing I might have to check - looks like we are now checking caption matching as well before merging tables. Its a good robust check, but in case of docx files, we don't have captions provided by default from docling , and are currently extracting captions based on custom written heuristics.
So i'm not sure if this always works in case of docx tables. There could be a few misses.

dharaneeshvrd: Do you think we need to handle this differently to cover .docx file formats?

services/digitize/settings.py
manju956: Yes
image

services/digitize/settings.py
manju956: I mean, the llm to pass the extracted table data(reformatted as bullets) to be interpreted as is, without summarization, pre-trained knowledge. 
Yes, I think we need to update english prompt too to have the table data

manju956: > I mean, the llm to pass the extracted table data(reformatted as bullets) to be interpreted as is, without summarization, pre-trained knowledge. Yes, I think we need to update english prompt too to have the table data. @Niharika0306 pls share your thought

services/digitize/settings.py
Niharika0306: @manju956 , what is the question asked here? I am more interested to see the reference chunks and how is the table info summarised/displayed

manju956: `IBM Cognos-URI-Elemente und -Beschreibungen` the table name from the pdf page no. 186

services/digitize/settings.py
manju956: per discussion, I reverted the prompt changes as we need summarization of table data.

services/digitize/processing/tables.py
manju956: reverted caption matching logic, so we can continue to use existing logic to support docx files too.

services/digitize/settings.py
manju956: reverted

services/digitize/settings.py
manju956: retained the existing behavior of LLM's table data summarization.

services/digitize/processing/tables.py
dharaneeshvrd: I see caption matching is still done using fuzzy match, am I missing something here?

dharaneeshvrd:
Overall changes looks good to me, any other open comments that still needs to be resolved @manju956  ?
But anyway I ll let @Niharika0306 to approve and merge it when she is back.

manju956:
> Overall changes looks good to me, any other open comments that still needs to be resolved @manju956 ? But anyway I ll let @Niharika0306 to approve and merge it when she is back.

I addressed all the open comments, showed the table.json structure and response to @Niharika0306 as well. 
Sure, let @Niharika0306 take a final look and approve

Niharika0306:
LGTM

Niharika0306 (dismissed):

Niharika0306 (approved):

PR #1119, opened 2026-07-09
AISERVICES-1351 proposal for entity extraction service

manalilatkar:


docs/proposals/entity_extraction_service_proposal.md
```diff
@@ -0,0 +1,1199 @@
...
+- PostgreSQL for durable metadata, initialized by an idempotent init container.
+- `/var/cache`-backed staging and result files on a persistent volume.
+- Boot-time recovery scan for zombie jobs.
+
+Two execution paths are provided:
+
+| Path                    | Endpoint                | Use case                                                                                                                                            |
+|:------------------------|:------------------------|:----------------------------------------------------------------------------------------------------------------------------------------------------|
+| **Synchronous**         | `POST /v1/extract`      | Raw text submitted inline. Blocking call, immediate JSON result.                                                                                    |
+| **Asynchronous (jobs)** | `POST /v1/extract/jobs` | File uploads (`.pdf`, `.txt`). PDFs are digitized internally via the Digitize Documents REST API before extraction. Returns a `job_id` for polling. |
```
dharaneeshvrd: No need to process .txt files via digitize service.
.docx file can be added here.

docs/proposals/entity_extraction_service_proposal.md
```diff
@@ -0,0 +1,1199 @@
...
+| Input: text (e.g., digitized invoice from Digitize service)   | Sync `text` field; async worker fetches digitized content via `GET /v1/documents/{id}/content` on the digitize service                                        |
+| Input: JSON schema with `"required": true` properties         | Registered schema in the Schema Registry (`json_schema` column, JSON Schema draft 2020-12)                                                                    |
+| Input (optional): single-shot examples (text–JSON pairs)      | `examples` array stored per schema in the registry                                                                                                            |
+| Config: Version                                               | Not managed by the service — schemas are immutable and unversioned (see Non-Goals). Users encode versions in the schema `name` if needed (e.g., `invoice-v2`) |
+| Config: LLM (granite-3.3-8b-instruct / mistral-small-3.1-24b) | `MODEL_NAME` env var (service default)                                                                                                                        |
```
dharaneeshvrd, line 33: For now we can avoid this config, since we would be tying a model when we deploy a service in current release.
It brings an overall perspective change that any service can talk to any model and its configurable at request level. It's a good point for future discussions. I will keep in mind. Thank you!

docs/proposals/entity_extraction_service_proposal.md
dharaneeshvrd: We have to design this flexibly that if the digitize service is not deployed along with this service, since this can be independent as well, we should allow plain text sync extraction.

docs/proposals/entity_extraction_service_proposal.md
```diff
@@ -0,0 +1,1199 @@
...
+| Config (optional): custom prompt                              | `custom_prompt` column per schema, appended to the system prompt                                                                                              |
+| Output: JSON populated according to input schema              | `data.extraction` in the response, validated server-side against the registered schema                                                                        |
+| Output: pointer to input text                                 | `data.source` block (digitize `doc_id` for async, input word/token stats for sync)                                                                            |
+| External dependency: OpenAI-compatible inferencing endpoint   | Existing vLLM endpoint (`OPENAI_BASE_URL`), shared semaphore                                                                                                  |
```
dharaneeshvrd, line 37: This is also not required, same as other comment regd model name.

docs/proposals/entity_extraction_service_proposal.md
```diff
@@ -0,0 +1,1199 @@
...
+
+**Request body:**
+
+| Field | Type | Required | Description |
+|:---|:---|:---|:---|
+| `name` | string | Yes | Unique, human-readable identifier (e.g., `invoice-extraction`). 1–200 chars, `[a-zA-Z0-9._-]`. |
+| `description` | string | No | Free-text description of what the schema extracts. |
+| `json_schema` | object | Yes | JSON Schema (draft 2020-12). Root must be `type: object`. Properties may be tagged `"required": true` (see 5.1.1). |
+| `examples` | array | No | Few-shot examples: `[{"text": "...", "output": {...}}]`. Max 5. Each `output` must validate against `json_schema`. |
+| `custom_prompt` | string | No | Extra instructions appended to the system prompt (e.g., domain conventions, date formats). Max 2,000 chars. |
```
dharaneeshvrd: Need to validate this prompt similar to the prompt validation happening for chatbot before actually using.

docs/proposals/entity_extraction_service_proposal.md
```diff
@@ -0,0 +1,1199 @@
...
+
+This keeps the user-facing convention from the concept while ensuring the stored schema works with standard validators (`jsonschema` library) and vLLM guided decoding.
+
+**Validation rules:**
+
+1. `name` is unique — conflict returns `409`.
+2. `json_schema` must parse as a valid draft 2020-12 schema (checked with `jsonschema.validators.Draft202012Validator.check_schema`) with root `type: object`.
+3. Every `examples[i].output` must validate against the normalized schema — otherwise `400` identifying the failing example index.
+
+#### 5.1.2 Schema size validation:
```
dharaneeshvrd, line 208: Also need to validate this prompt for maliciousness similar to the prompt validation happening for chatbot before actually using.

docs/proposals/entity_extraction_service_proposal.md
```diff
@@ -0,0 +1,1199 @@
...
+| 409 Conflict | One or more jobs reference this schema. Response lists up to 10 referencing `job_id`s. |
+| 500 Internal Server Error | Database failure. |
+
+---
+### 5.5 DELETE /v1/schemas (Bulk Delete)
+
+**Query parameters:** `confirm=true` (required).
+
+1. Validate `confirm=true` (else `400`).
+2. Reject with `409` if any schema is used in an extract_job.
```
dharaneeshvrd, line 355: Hope this is for completed jobs as well similar to the behaviour of individual schema

docs/proposals/entity_extraction_service_proposal.md
```diff
@@ -0,0 +1,1199 @@
...
+{input_text}
+
+JSON:
+```
+
+The few-shot block renders each registered example as an `Example text:` / `Example JSON:` pair. Examples are the primary lever for teaching field-mapping conventions (e.g., which of several dates on an invoice is the "invoice date").
+
+### 8.2 Guided decoding
+
+Where the vLLM deployment supports it, the request includes structured-output enforcement so the model *cannot* emit non-conformant JSON:
```
dharaneeshvrd, line 851: We support WatsonX too, but as long as this is a openai standard it should work, but let's cover here.

docs/proposals/entity_extraction_service_proposal.md
```diff
@@ -0,0 +1,1199 @@
...
+2. `json_schema` must parse as a valid draft 2020-12 schema (checked with `jsonschema.validators.Draft202012Validator.check_schema`) with root `type: object`.
+3. Every `examples[i].output` must validate against the normalized schema — otherwise `400` identifying the failing example index.
+
+#### 5.1.2 Schema size validation:
+
+REGISTRATION_BUDGET_FRACTION is the maximum share of the model's context window (MAX_MODEL_LEN) that a schema's fixed prompt overhead — the schema itself, its few-shot examples, and the prompt template — is allowed to consume. It's checked once, at POST /v1/schemas time.
+The reasoning: every extraction prompt is composed of two parts — a fixed part that's identical for every request against that schema (schema JSON + examples + system prompt + reserved output tokens) and a variable part (the user's input text). The fraction draws the line between them. With REGISTRATION_BUDGET_FRACTION = 0.5 and MAX_MODEL_LEN = 32768:
+
+```commandline
+schema_tokens + examples_tokens + PROMPT_OVERHEAD_TOKENS + MIN_OUTPUT_TOKENS
```
dharaneeshvrd: Do we need to include MIN_OUTPUT_TOKENS here?
Since reserved output is calculated and compared against the total max model length below.
Also we have put a clamp to get the max output tokens based on schema tokens, do you think we need to add some limitation for schema tokens as well?

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: @dharaneeshvrd look at env var DIGITIZE_BASE_URL, and if this endpoint is not there, async endpoint will not service pdfs and docx. We can control this in individual service deployment and RAG app deployment.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: This variable still needs to be there (along with its respective MAX_MODEL_LEN to make some claculations) because we need to send LLM name in responses.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: This is also needed, to make the vLLM calls of entity extraction.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: yes, same is mentioned in 5.12.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: I'll add a note mentioning this.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: Sure, I can remove the MIN_OUTPUT_TOKENS from here. 
Re: 

> Also we have put a clamp to get the max output tokens based on schema tokens, do you think we need to add some limitation for schema tokens as well?

The equation on line 213 itself will act as a limitation on schema tokens, I don't think we need one more separate limitation.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: @dharaneeshvrd can we abstract out the prompt validation into common so the same thing can be used here?

docs/proposals/entity_extraction_service_proposal.md
dharaneeshvrd: Are these configs documented here are part of an individual request or the config of the deployment?
I thought these individual request params, hence mentioned as not needed.

docs/proposals/entity_extraction_service_proposal.md
dharaneeshvrd: Let's mention this part of async API handling.

docs/proposals/entity_extraction_service_proposal.md
dharaneeshvrd: Can be, please see what is the feasibility of doing this.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: @dharaneeshvrd a mix of both. A list of all configs. For request-specific configs, check 5.1 and 5.7.

docs/proposals/entity_extraction_service_proposal.md
dharaneeshvrd: Let me know once the note is added

docs/proposals/entity_extraction_service_proposal.md
dharaneeshvrd: One more reason I m saying this here is that, we would limit the output token based on schema tokens to 4096, if schema tokens itself is more than that, then there output won't generate properly because we are limiting it to 4096.

dharaneeshvrd: Either we need to alter that logic to include more room or add a validation here for schema tokens to be within 2048.

docs/proposals/entity_extraction_service_proposal.md
```diff
@@ -0,0 +1,1199 @@
...
+
+**Validation rules:**
+
+1. `name` is unique — conflict returns `409`.
+2. `json_schema` must parse as a valid draft 2020-12 schema (checked with `jsonschema.validators.Draft202012Validator.check_schema`) with root `type: object`.
+3. Every `examples[i].output` must validate against the normalized schema — otherwise `400` identifying the failing example index.
+
+#### 5.1.2 Schema size validation:
+
+REGISTRATION_BUDGET_FRACTION is the maximum share of the model's context window (MAX_MODEL_LEN) that a schema's fixed prompt overhead — the schema itself, its few-shot examples, and the prompt template — is allowed to consume. It's checked once, at POST /v1/schemas time.
```
Niharika0306: can we use a different name here for 'REGISTRATION_BUDGET_FRACTION' since its difficult to tie it to the context. May be something using the existing keywords?

docs/proposals/entity_extraction_service_proposal.md
Niharika0306: Here, in this context - is MIN_OUTPUT_TOKENS considered the same as reserved_output_tokens ?

Niharika0306: Also, I feel the calculation could change a bit if there are examples included - ideally the input_tokens could be just as much as the example tokens.  We have seen it in the shipping orders case.

docs/proposals/entity_extraction_service_proposal.md
```diff
@@ -0,0 +1,1199 @@
...
+**Response codes:**
+
+| Status | Description |
+|:---|:---|
+| 202 Accepted | Job created. |
+| 400 Bad Request | Missing file, multiple files, or missing `schema_id`. |
+| 404 Not Found | Unknown `schema_id`. |
+| 415 Unsupported Media Type | Not a valid `.txt` or `.pdf`. |
+| 429 Too Many Requests | Job concurrency at capacity. |
+| 500 Internal Server Error | Unexpected failure. |
```
Niharika0306: Don't we have something similar to '422 Unprocessable Entity' as clocked during the synchronous req execution here as well?

docs/proposals/entity_extraction_service_proposal.md
```diff
@@ -0,0 +1,1199 @@
...
+max_tokens (LLM call) = reserved_output
+```
+
+**Output reservation rationale:** unlike summarization (output proportional to input), extraction output size is bounded by the **schema**, not the input — a 30-page contract still yields one JSON object shaped by the schema. `OUTPUT_TOKEN_FACTOR` defaults to `2.0` (the populated JSON is roughly the schema's keys plus values), clamped between `MIN_OUTPUT_TOKENS=512` and `MAX_OUTPUT_TOKENS=4096`. Schemas with unbounded arrays (e.g., `line_items`) are the reason for the generous factor and the 4,096 ceiling; all three knobs are env-configurable.
+
+Rationale for choosing values in above formula:
+1. OUTPUT_TOKEN_FACTOR: refer table below to figure out what number supports what kind of output. 5 is what is recommended to cover unbounded attributes in schemas. We could also obtain it from the user while creating schema in future enhancements.
+2. MAX_OUTPUT_TOKENS = 4096 = 12.5% of 32768. 12.5% feels like the right ceiling because the service's stated purpose — invoices, emails, contracts — is input-heavy and output-light: you feed pages of text and get back a compact JSON record.
```
Niharika0306, line 752: The  MAX_OUTPUT_TOKENS and  MIN_OUTPUT_TOKENS  are currently just constants. I see the reasoning mentioned, but I feel that they should still be dependent on the schema tokens in some way.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: @Niharika0306 how about INPUT_BUDGET_FRACTION

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: As discussed, I will add a note that these values are only for now and will be finalized after experimentation.

docs/proposals/entity_extraction_service_proposal.md
Niharika0306: sure. Or a SCHEMA_REGISTRATION_BUDGET_FRACTION  also makes sense for me.

manalilatkar: @dharaneeshvrd WDYT?

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: @dharaneeshvrd 
Agreed — a schema whose implied output exceeds MAX_OUTPUT_TOKENS is pre-doomed and should fail at registration. Rather than a hardcoded 2048, I'll add the derived check 
`schema_tokens × OUTPUT_TOKEN_FACTOR ≤ MAX_OUTPUT_TOKENS `
(which equals 4096 at current defaults) to the 5.1.2 admission gate, returning 400 with a token breakdown. Note this covers the deterministic case only — array-heavy outputs can still hit the ceiling input-dependently, which stays handled by the OUTPUT_BUDGET_EXCEEDED fast-fail.

manalilatkar: @Niharika0306 we'll be testing this for 2 seperate use cases - shipping orders and PII. Let's see how the numbers fare while experimentation.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: @dharaneeshvrd discussed with @sats-23 and its doable, mostly prompts we would need to relocate to common and rewrite the execution logic. I'll create a task to track this.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: mentioned in line no 20 and 21.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: added

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: Note added on line no 760

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: Note added on line no 856.

docs/proposals/entity_extraction_service_proposal.md
dharaneeshvrd: How about `MAX_SCHEMA_SHARE`?
I feel its more simpler and to the point.

docs/proposals/entity_extraction_service_proposal.md
dharaneeshvrd: I feel we can get started with this, like @manalilatkar mentioned based on the results, we can tweak this if needed.

docs/proposals/entity_extraction_service_proposal.md
dharaneeshvrd: > Note this covers the deterministic case only — array-heavy outputs can still hit the ceiling input-dependently, which stays handled by the OUTPUT_BUDGET_EXCEEDED fast-fail.

can you please elaborate the use case here?

manalilatkar: my bad, I meant CONTEXT_LIMIT_EXCEEDED, not OUTPUT_BUDGET_EXCEEDED.
This covers the use case of,  say, a shipping order email that contains 10 different items to be shipped. Now they will create 10 entries of arrays under the items field. This is a array-heavy output. This type of output can still hit the MAX_OUTPUT_TOKENS ceiling, in which case we can retry with a higher max_tokens count if the budget still exists otherwise fail with CONTEXT_LIMIT_EXCEEDED.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: Actually that makes it sound like we are talking about a max share of the schema whereas we are talking about the max share of the context window.

dharaneeshvrd: CONTEXT_SCHEMA_SHARE sounds good as per discussion

manalilatkar: changed everywhere.

docs/proposals/entity_extraction_service_proposal.md
manalilatkar: As discussed, added this step in execution logic on line 507 and line 386.

Niharika0306 (approved):

dharaneeshvrd (approved):

PR #1120, opened 2026-07-09
AISERVICES-1353 Proposal Doc for deploy flow refactor

pranithraoibm:
Proposal doc for https://jsw.ibm.com/browse/AISERVICES-1353


docs/proposals/catalog/deployflow-ui-refactor-proposal.md
```diff
@@ -0,0 +1,276 @@
...
+
+Type names are unified during the move:
+- `Component` / `DeployOptionsComponent` — same shape, unified as `Component`
+- `Service` / `DeployOptionsService` — same shape, unified as `DeployOptionsService`
+- `ResourcesResponse` — confirmed identical in both files (same endpoint), one copy kept
+- `ResourcesApiResponse` (in `digitalAssistants.ts`, uses `used_cpu`/`used_bytes`) — different endpoint, renamed to `UsedResourcesResponse` for clarity
+
+**API functions:** `src/api/digitalAssistants.ts` and `src/services/deployment.api.ts` are also parallel API clients hitting the same backend. Both export `fetchResources`, `fetchProviderParams`/`fetchProviderSchema` (same endpoint, different names), `deployApplication`, `calculateUptime`, and `transformApplicationToRow` — all duplicated. They were built separately because each flow used its own API file. This PR merges all function exports into a single `src/api/applications.api.ts`, deletes `src/api/digitalAssistants.ts`, and updates `deployment.api.ts` to re-export from `applications.api.ts` for any consumers that can't be updated atomically. All import sites are updated to `@/api/applications.api`.
```
ryanmarc: This statement, along with line 251, contradict each other, and the actual PR 0 implementation (#1123) went with full deletion.

docs/proposals/catalog/deployflow-ui-refactor-proposal.md
```diff
@@ -0,0 +1,276 @@
...
+`useServices`, `useSessionTimeout`, and `useUseCases` stay in `src/hooks/` — they are either used outside the deploy flows or unrelated entirely.
+
+**Dead code deletions:**
+
+- `src/components/DynamicField/` — `DynamicField.tsx` renders a single Carbon input from a `ParsedField` with the same `switch(type)` logic as `DynamicSchemaFields`, but is not imported anywhere in the codebase. The shared `DynamicSchemaFields` created in PR 4 supersedes it.
+- `src/hooks/useResources.ts` — fetches system resources and caches them in `useDeployStore`, but neither deploy flow imports it. Both flows call `fetchResources` directly from `@/api/applications.api` instead. Dead code with no consumers.
+
+### PR 2 — Extract shared TypeScript types
+
+Move the identical interfaces — `ComponentConfig`, `DeployFormData`, `ServiceConfig`, `ResourceItem`, base `StepProps`, and the shared action types — into `deployFlow/shared/types.ts`. Each flow's own types file shrinks to just its unique additions.
```
ryanmarc: See the review on #1123 - There's an issue with the interfaces not being identical.

docs/proposals/catalog/deployflow-ui-refactor-proposal.md
```diff
@@ -0,0 +1,276 @@
...
+- **`useDeployFlowReducer.ts`** — a React hook that owns the shared state management: the reducer, the action types, and the four callbacks (`handleFormDataChange`, `handleEditingChange`, `handleResourceStatusChange`, `handleBack`) that are identical in both root components. Each flow calls this hook and gets the shared state and callbacks back, then adds only its own flow-specific actions on top — `digitalAssistant` adds `SET_IS_LOADING` and `SET_ERROR` (driven by `useDeployOptions`), and `services` adds `SET_SELECTED_SERVICE` (for tracking which standalone service the user picked in StepZero).
+- **`DeployTearsheetShell.tsx`** — a wrapper component that renders the `Tearsheet`, the vertical `ProgressIndicator` in the influencer panel(the optional side panel), and the deployment error toast notification. Each flow passes in its own step labels and the content for the current step as children. The shell handles all the surrounding chrome.
```
ryanmarc: This and line 163 disagree on whether the deploy-toast actions are flow-specific. If the toast lives in the shared DeployTearsheetShell (PR 7 says the shell renders "the deployment error toast notification"), its SHOW/HIDE actions arguably belong in the shared reducer, right now the doc has the shell owning the UI while one flow owns its state. Please reconcile the two lists, and while there, specify the composition mechanism (e.g. shared reducer returns S | undefined, flow reducer handles the rest), same pattern question as the tables proposal (#1121), and it'd be good if both refactors answer it the same way.

docs/proposals/catalog/deployflow-ui-refactor-proposal.md
```diff
@@ -0,0 +1,276 @@
...
+
+This PR moves the shared reducer cases and callbacks into `useDeployFlowReducer`, a hook in `deployFlow/shared/hooks/`. Each flow calls the hook to get the shared state and callbacks, then handles only its own unique actions on top — `digitalAssistant` adds `SET_IS_LOADING`, `SET_ERROR`, `SHOW_DEPLOY_TOAST`, and `HIDE_DEPLOY_TOAST` (driven by `useDeployOptions` and the deploy toast lifecycle), and `services` adds `SET_SELECTED_SERVICE` (for tracking which standalone service the user picked in StepZero).
+
+### PR 7 — Extract shared DeployTearsheetShell wrapper
+
+The outermost JSX — `Tearsheet` with the vertical progress indicator influencer and the deployment error notification — is structurally the same in both root components. Wrap it in a single shell component that accepts the step definitions and page content as props, and replace the duplicated Tearsheet boilerplate in both flows.
+
+### PR 8 — Unify StepOne into a shared component
```
ryanmarc: PR 8 as scoped is (1) a shared StepOne component, (2) a user-visible change to what the dropdown means (provider-disguised-as-model → true model selection), and (3) a rearchitecture of the DA flow's data-fetching layer including deleting two hooks and extending useDeployOptions. Any one of those is a full review on its own - together they can't plausibly land under 500 lines, and a regression in one will block reverting the others. Strongly suggest splitting: 8a = eager-fetch migration in digitalAssistant (behavior: fewer mid-flow spinners, no UI change); 8b = shared StepOne (pure structure); 8c = model-first selection (the actual UX change, testable in isolation). The merge-order section already serializes 7 → 8 → 9, so this costs nothing but PR count.

docs/proposals/catalog/deployflow-ui-refactor-proposal.md
```diff
@@ -0,0 +1,276 @@
...
+- `ResourceRequirements.tsx` — fetches available system resources and calculates totals accounting for global components plus every enabled service, then passes the results to the shared `ResourceRequirementsPanel` to display
+
+`deployFlow/services/` keeps:
+
+- Its own API calls (`deployApplication` via `applications.api`) and store (`useServiceDeployStore`)
+- `StepZero` — the service selection screen where the user picks which standalone service to deploy
+- `ResourceRequirements.tsx` — fetches available system resources and calculates totals for a single service only, then passes the results to the shared `ResourceRequirementsPanel` to display
+- `formDataInitializer.ts` — initialises the form data structure when a service is selected
+
+The work is split into 10 PRs, each under 500 lines of diff so they stay easy to review. PR 0 consolidates duplicated API infrastructure. PRs 1–7 are pure refactors with no behaviour changes. PRs 8 and 9 involve behaviour alignment and structural unification.
```
ryanmarc: Both "fixes" in PR 4 are behavior changes: adding x-ui-only support to the services flow changes what renders for any future schema that uses it, and moving DeployFlow from render-time validation to validate-on-submit changes when users see errors (they lose as-you-type feedback). The second one in particular deserves UX sign-off, not just code review.

docs/proposals/catalog/deployflow-ui-refactor-proposal.md
```diff
@@ -0,0 +1,276 @@
...
...igitalAssistant` fetches provider schemas lazily — when the user selects a provider from a dropdown, `useProviderParams` fires an API call on demand for that specific `componentType + providerId` pair. This causes a loading spinner mid-flow every time the user makes a selection. The `services` flow does the opposite — `useServiceDeployOptions` fetches all provider schemas eagerly on mount in two stages (Step 1 component models first for immediate display, LLM models in the background), so by the time the user reaches any step everything is already in the store and `useProviderSchema` is just a synchronous store read. This PR migrates `digitalAssistant` to the same eager-fetch pattern: `useDeployOptions` is extended to fetch and cache all component provider schemas upfront when it loads the architecture deploy options. The per-provider lazy hooks (`useProviderParams`, `useServiceParams`) are replaced by a single `useProviderSchema`-style store selector. This eliminates the mid-flow loading states in `digitalAssistant` and makes the hook layer between the two flows structurally identical — both read from the store synchronously in components, both populate the store eagerly on mount.
```
ryanmarc: Two things worth a sentence in the doc: (1) scale: eager fetch is one GET .../params per (component × provider) across global components and every service; for the RAG architecture today that's fine, but note the assumption so someone re-checks it when a 10-service architecture appears (firing them via the existing dedupe/request-manager util with Promise.allSettled, not sequentially, should also be stated); (2) failure: lazy fetch fails at the moment of selection with a local error; eager fetch can partially fail at mount.

Define it: does the flow open with a warning, retry per-schema on demand, or block? "Structurally identical hook layers" is the right goal, just don't let the DA flow inherit a worse failure mode than the one it has.

ryanmarc (changes requested):

docs/proposals/catalog/deployflow-ui-refactor-proposal.md
pranithraoibm: Updated to match the table and #1123: both files fully deleted in PR 0, with all import sites updated.

docs/proposals/catalog/deployflow-ui-refactor-proposal.md
pranithraoibm: Fixed by keeping the shared type to have the common fields and each flow extends it with what's unique to it.

docs/proposals/catalog/deployflow-ui-refactor-proposal.md
pranithraoibm: Moved SHOW/HIDE_DEPLOY_TOAST to shared since the shell owns the toast for both flows. Also noticed isLoading and error shouldn't be in the reducer. Services already handles them as plain hook values, PR 8 will align DA to do the same.

docs/proposals/catalog/deployflow-ui-refactor-proposal.md
pranithraoibm: Flagged x-ui-only as a behaviour change in the proposal. 
For the validation one, checked the code and both flows already gate errors on hasValidationError so timing is the same for users. Updated it as a structural only change, no UX sign-off needed.

docs/proposals/catalog/deployflow-ui-refactor-proposal.md
pranithraoibm: Added both. Scale assumption documented as one GET per component/provider pair, fired in parallel via Promise.allSettled through the request-manager dedup util. Also upgrading services from Promise.all to Promise.allSettled in the same PR so both flows are consistent. For the failure mode, failed schemas are never written to the store so close/reopen triggers a re-fetch automatically. If any fail at mount, an ActionableNotification shows the affected components with a message to close and reopen.

ryanmarc (approved):

PR #1121, opened 2026-07-09
AISERVICES-1357: Applications table refactor proposal document

Niranjan-J-Joshi:
Proposal Document for [AISERVICES-1357](https://jsw.ibm.com/browse/AISERVICES-1357)


docs/proposals/catalog/applications-table-refactor-proposal.md
```diff
@@ -0,0 +1,237 @@
...
+
+* Search
+* Filtering
+* Column management
+* Export
+* Confirmation dialogs
+* Data fetching and refresh
+
+Because these features were implemented separately, both tables contain significant duplicated logic and UI code.
+This duplication increases maintenance effort, requires bug fixes to be applied in multiple places,
```
ryanmarc: this just ends mid sentence.

docs/proposals/catalog/applications-table-refactor-proposal.md
```diff
@@ -0,0 +1,237 @@
...
+
+Because these features were implemented separately, both tables contain significant duplicated logic and UI code.
+This duplication increases maintenance effort, requires bug fixes to be applied in multiple places, 
+
+---
+
+## Proposed Solution
+
+Create a shared table framework that contains the common functionality used by both Digital Assistants and Deployed Services tables.
+Common functionality will be centralized into reusable components, hooks, and utilities.This approach reduces duplication
```
ryanmarc: This seems to end mid sentence as well.

docs/proposals/catalog/applications-table-refactor-proposal.md
```diff
@@ -0,0 +1,237 @@
...
+        useAutoRefresh.ts
+        useCSVExport.ts
+        useExportToastAutoDismiss.ts
+      components/
+        CellRenderers.tsx
+        DeleteModal.tsx
+        ExportModal.tsx
+        TableToasts.tsx
+        TableEmptyStates.tsx
+        TableToolbarActions.tsx
```
ryanmarc: src/components/table/ contains nothing but shared/ — either flatten to src/components/table/ directly or note what non-shared siblings you expect to live there later.

docs/proposals/catalog/applications-table-refactor-proposal.md
```diff
@@ -0,0 +1,237 @@
...
+- Extend `DeployedServices.module.scss` to compose from `table.shared.module.scss`; remove the now-duplicated rules
+
+PR 5 depends on PRs 1, 2, and 3.
+
+---
+
+
+## File Changes Summary
+
+A complete map of every file created, modified, or deleted across all 6 PRs.
```
ryanmarc: The summary says "all 6 PRs" but the plan defines 5.

docs/proposals/catalog/applications-table-refactor-proposal.md
```diff
@@ -0,0 +1,237 @@
...
+
+- **`useAutoRefresh.ts`** — A hook that encapsulates the shared refresh lifecycle. Currently both tables set up the same `setInterval` effect independently.
+
+- **`useCSVExport.ts`** — A hook that implements the complete CSV export flow: filename validation, the multi-page fetch loop , visible-column filtering, `downloadCSVWithChildren` call, and success/error toast state.
+
+- **`useExportToastAutoDismiss.ts`** — A hook for the 5-second auto-dismiss of successful export toasts.Currently both files have an identical `useEffect` that sets a `setTimeout` when `exportToastOpen && exportToastKind === "success"`.
+
+- **`CellRenderers.tsx`** — Exports shared table cell renderers (`StatusCell`, `MessageCell`, `ActionCell`, and `NameCell`) along with a common `STATUS_CONFIG` map covering all statuses used by both tables (`Initializing`, `Downloading`, `Deploying`, `Running`, `Deleting`, `Stopped`, `Error`).
```
ryanmarc: The proposal's clean-name STATUS_CONFIG will silently miss "Deleting..." rows (they'd fall through to DEFAULT_STATUS_CONFIG, rendering gray/paused instead of blue/in-progress). Before merging the maps, the plan needs a normalization step.

ryanmarc: These two ActionCells don't behave identically today, one is case-sensitive, the other isn't. A shared ActionCell has to pick one, and that's a (small) behavior change for whichever table loses. Standardize on the normalized RowStatus from the above comment so the comparison is exact by construction. Document the choice in the proposal so PR 3 review isn't where this gets discovered.

docs/proposals/catalog/applications-table-refactor-proposal.md
```diff
@@ -0,0 +1,237 @@
...
+
+All shared table infrastructure lives in one place with a clear label, and both consumers remain as independent files that only import what they need from it.
+
+The `shared/` directory contains everything both tables import from:
+
+- **`types.ts`** — Provides the base TypeScript contracts used by both tables, including shared row models, table state, and column definitions.
+
+- **`table.shared.module.scss`** — Contains styles that are duplicated across both table implementations, including toast notifications, status tags, menu styling, and shared table presentation. Table-specific styles remain within their respective modules.
+
+- **`reducerUtils.ts`** — Helper functions for the state update cases that both reducers implement identically: search updates with pagination reset, pagination updates, export state updates, column visibility updates, and column reset.
```
ryanmarc: "Helper functions the reducers call" vs "a shared reducer the tables compose" are quite different in TypeScript ergonomics once each table extends BaseTableState and adds its own action union. Committing to the shape in PR 1, e.g. `handleSharedTableAction<S extends BaseTableState>(state: S, action: SharedTableAction): S | undefined` with each table's reducer doing `return handleSharedTableAction(state, action) ?? ownCases(state, action)`. Cheap to decide now; painful to retrofit in two parallel migration PRs. (The deploy-flow proposal in #1120 has the same problem in useDeployFlowReducer — worth using the same pattern in both.)

docs/proposals/catalog/applications-table-refactor-proposal.md
```diff
@@ -0,0 +1,237 @@
...
+
+- **`tableUtils.ts`** — Three utility functions: `filterRowsBySearch(rows, search, fields)` (both tables filter rows by joining field values and lowercasing — the only difference is which fields are included, which becomes a parameter), `getVisibleHeaders(headers, visibleColumns)` (identical in both files ), and `getToggleableHeaders(headers)`.
+
+- **`useAutoRefresh.ts`** — A hook that encapsulates the shared refresh lifecycle. Currently both tables set up the same `setInterval` effect independently.
```
ryanmarc: Extracting the setInterval is the easy part, the spec should also pin down the edge behavior both tables currently handle implicitly (or don't): (1) skip a tick while a fetch is already in flight so slow responses don't stack and apply out of order; (2) pause while the delete modal is open, so a refresh doesn't replace rows mid-confirmation or clobber the optimistic "Deleting…" state.

ryanmarc (changes requested):

docs/proposals/catalog/applications-table-refactor-proposal.md
Niranjan-J-Joshi: I've added both cases to the useAutoRefresh hook : skipping refreshes while a request is in flight and supporting a paused state during delete confirmation flows.

docs/proposals/catalog/applications-table-refactor-proposal.md
Niranjan-J-Joshi: fixed — removed the shared/ wrapper. All files now sit directly under src/components/table/.
We are not planning any sibling directories alongside it.

docs/proposals/catalog/applications-table-refactor-proposal.md
Niranjan-J-Joshi: Fixed

docs/proposals/catalog/applications-table-refactor-proposal.md
Niranjan-J-Joshi: Fixed

docs/proposals/catalog/applications-table-refactor-proposal.md
Niranjan-J-Joshi: Fixed — updated to "all 5 PRs".

docs/proposals/catalog/applications-table-refactor-proposal.md
Niranjan-J-Joshi: Updated, the shared ActionCell standardizes on exact matching (rowData?.status === "Running"), replacing the case-insensitive comparison in DeployedServicesTable.

docs/proposals/catalog/applications-table-refactor-proposal.md
Niranjan-J-Joshi: reducerUtils.ts exposes handleSharedTableAction, which has all shared table actions, each table will continue to handle its own table-specific actions. If an action isn't handled by handleSharedTableAction, it falls back to the table's reducer.

ryanmarc (dismissed):

pranithraoibm (approved):

PR #1122, opened 2026-07-10
[CI-CD] jenkins: Enhancement in code signature pipeline.

adarshagrawal38:
# Changes
Selective image skipping: added a `SKIP_IMAGE_LIST` parameter that allows operators to bypass specific build stages without modifying the pipeline. Parameter is use full when user don't wan't to build all the images.

ICR credential override: adds an `ICR_REGISTRY_PASSWORD` password parameter that accepts an IAM API key at runtime. When provided, it overrides the stored icr-iam-creds credential, allowing operators to build and publish images directly to icr.io/ai-services without updating the Jenkins credential store.


manju956 (approved):
lgtm

PR #1123, opened 2026-07-13
refactor: consolidate duplicated API infrastructure into api.types.ts and applications.api.ts

pranithraoibm:
Both deploy flows independently built parallel API layers hitting the same backend, resulting in duplicate type definitions and duplicate function implementations.

- Create `src/types/api.types.ts` — single home for all API contract types; delete `src/types/digitalAssistants.ts`
- Create `src/api/applications.api.ts` — single API client merging all functions; delete `src/api/digitalAssistants.ts` and `src/services/deployment.api.ts`
- `Component` (in `digitalAssistants.ts`) and `DeployOptionsComponent` (in `deployment.api.ts`) are type definitions with identical fields (`type`, `name`, `providers`) merged into one: `DeployOptionsComponent`; `ResourcesApiResponse` renamed to `UsedResourcesResponse` to distinguish it from `ResourcesResponse` (different endpoint, different fields)
- Update all files that imported from the deleted files to point to the new locations


ui/catalog/src/types/api.types.ts
```diff
@@ -73,30 +73,30 @@ export interface Provider {
   };
 }
 
-export interface Component {
+// Unified: was Component in digitalAssistants.ts and DeployOptionsComponent in deployment.api.ts
```
ryanmarc: These comments are unnecessary. Please remove these and check the rest of the code for comments that don't offer any useful information.

ui/catalog/src/types/api.types.ts
```diff
@@ -73,30 +73,30 @@ export interface Provider {
...
 
-export interface Service {
+export interface DeployOptionsService {
   id: string;
   name: string;
   version: string;
   schema?: string;
-  components: Component[];
-  resources?: Resources;
+  components: DeployOptionsComponent[];
```
ryanmarc, line 83: The old Service interface in digitalAssistants.ts had resources?: Resources, but the unified DeployOptionsService drops it — and StepTwo.tsx:166 reads service.resources for the service-level resource totals, so this no longer type-checks.

ui/catalog/src/api/applications.api.ts
```diff
@@ -0,0 +1,444 @@
...
+// Deploys a new application with the provided configuration payload
+export async function deployApplication(
+  payload: DeploymentPayload,
+): Promise<DeployApplicationResponse | void> {
+  const response = await api.post<DeployApplicationResponse>(
+    APPLICATION_ENDPOINTS.GET_APPLICATIONS,
+    payload,
+  );
+  return response.data;
+}
```
ryanmarc, line 327: The two deleted files had different deployApplication contracts: the DA flow posts a payload without deployment_type, the services flow posts one with deployment_type: "service". Merging them under the services-flow DeploymentPayload breaks the DA call at DeployFlow.tsx:263. Suggest modelling both contracts explicitly, e.g. ArchitectureDeploymentPayload | ServiceDeploymentPayload as the parameter type (move the DA payload interface into api.types.ts too, since it is an API contract), rather than pretending there's one payload shape.
Also: the return type Promise<DeployApplicationResponse | void> should just be Promise<DeployApplicationResponse>, the function unconditionally returns response.data. The | void only forces pointless narrowing on callers.

ui/catalog/src/api/applications.api.ts
```diff
@@ -0,0 +1,444 @@
...
+      oneOf?: Array<{ const: string; title: string; description?: string }>;
+      [key: string]: unknown;
+    }
+  >;
+}> {
+  const response = await api.get(
+    SERVICE_ENDPOINTS.GET_COMPONENT_PROVIDER_PARAMS(componentType, providerId),
+  );
+  return response.data;
+}
```
ryanmarc: The refactor proposal explicitly calls out fetchProviderParams/fetchProviderSchema as "same endpoint, different names" to be merged, but both survive here — now side by side in one file, one via the endpoint constant and one via a hardcoded template string.

ui/catalog/src/api/applications.api.ts
```diff
@@ -0,0 +1,444 @@
...
+// Fetches provider schema parameters (services flow)
+export async function fetchProviderSchema(
+  componentType: string,
+  providerId: string,
+): Promise<ProviderSchema> {
+  const response = await api.get<ProviderSchema>(
+    `/components/${componentType}/providers/${providerId}/params`,
+  );
+  return response.data;
+}
```
ryanmarc, line 124: See comment from line 98.

ui/catalog/src/api/applications.api.ts
```diff
@@ -0,0 +1,444 @@
...
+): Promise<ArchitectureDetailsResponse> {
+  const response = await api.get<ArchitectureDetailsResponse>(
+    DIGITAL_ASSISTANTS_ENDPOINTS.ARCHITECTURE_DETAILS(architectureId),
+  );
+  return response.data;
+}
+
+// Fetches details about a specific service
+export async function fetchServiceDetails(serviceId: string): Promise<Service> {
+  const response = await api.get<Service>(`/services/${serviceId}`);
```
ryanmarc: This hardcodes the path while the rest of the file uses the constants module. Use a constant here also.

ui/catalog/src/api/applications.api.ts
```diff
@@ -0,0 +1,444 @@
...
+  );
+  return response.data;
+}
+
+// Fetches deploy options for a specific service
+export async function fetchServiceDeployOptions(
+  serviceId: string,
+): Promise<ServiceDeployOptions> {
+  const response = await api.get<ServiceDeployOptions>(
+    `/services/${serviceId}/deploy-options`,
```
ryanmarc: This hardcodes the path while the rest of the file uses the constants module. Use a constant here also.

ryanmarc (changes requested):

ui/catalog/src/types/api.types.ts
pranithraoibm: Added resources? back to DeployOptionsService

ui/catalog/src/api/applications.api.ts
pranithraoibm: Fixed. Split into ArchitectureDeploymentPayload and ServiceDeploymentPayload in api.types.ts. DeploymentComponent and DeploymentService are also extracted into api.types.ts so both transforms share the same definitions. deployment_type: "service" moved into the transform itself so the caller no longer needs to spread and patch it. Also fixed the return type to `Promise <DeployApplicationResponse>`.

ui/catalog/src/api/applications.api.ts
pranithraoibm: Fixed. fetchProviderParams is deleted. fetchProviderSchema is now the single function for that endpoint. All consumers (useProviderParams, serviceDeploymentTransform) updated to use it.

ryanmarc:
This PR is still failing the typecheck after the latest changes. There are 3 places. The current `npm run typecheck` is actually broken - it doesn't actually check anything. Please update the typecheck command in the package.json file to `tsc -b  --noEmit` and also remove the `"ignoreDeprecations": "6.0",` line from `tsconfig.app.json`.

After doing that, if you run `npm run typecheck` you should see these errors.

```
src/components/DeployFlow/components/ServiceConfigCard.tsx:346:69 - error TS2345: Argument of type 'ProviderSchema' is not assignable to parameter of type '{ properties?: Record<string, { default?: unknown; }> | undefined; }'.
  Types of property 'properties' are incompatible.
    Type '{ [key: string]: ProviderSchemaProperty | undefined; model?: ProviderSchemaProperty | undefined; }' is not assignable to type 'Record<string, { default?: unknown; }>'.
      'string' index signatures are incompatible.
        Type 'ProviderSchemaProperty | undefined' is not assignable to type '{ default?: unknown; }'.
          Type 'undefined' is not assignable to type '{ default?: unknown; }'.

346               .filter(([key, value]) => shouldShowParam(key, value, schema))
                                                                        ~~~~~~

src/components/DeployFlow/components/ServiceConfigCard.tsx:421:47 - error TS2345: Argument of type 'ProviderSchema' is not assignable to parameter of type '{ properties?: Record<string, { default?: unknown; }> | undefined; }'.
  Types of property 'properties' are incompatible.
    Type '{ [key: string]: ProviderSchemaProperty | undefined; model?: ProviderSchemaProperty | undefined; }' is not assignable to type 'Record<string, { default?: unknown; }>'.
      'string' index signatures are incompatible.
        Type 'ProviderSchemaProperty | undefined' is not assignable to type '{ default?: unknown; }'.
          Type 'undefined' is not assignable to type '{ default?: unknown; }'.

421                   shouldShowParam(key, value, schema, excludeKeys),
                                                  ~~~~~~

src/components/DeployFlow/DeployFlow.tsx:249:9 - error TS2322: Type 'ProviderSchema' is not assignable to type 'Record<string, unknown>'.
  Index signature for type 'string' is missing in type 'ProviderSchema'.

249         providerParamsData[key] = cache.data;
            ~~~~~~~~~~~~~~~~~~~~~~~


Found 3 errors.
```

ryanmarc (dismissed):

Niranjan-J-Joshi (approved):

PR #1124, opened 2026-07-13
De-duplication support for Digitize

sats-23:
-Proposal - https://github.com/sats-23/project-ai-services/tree/dupIndex/docs/proposals/digitize-file-hash-dedup
-Why do we need this?
--With more data sources incoming to the digitize service, we have a necessity to hash files to be able to differentiate new duplicate files from already ingested files. 
--This will help us save huge time and resources by preventing re-ingestion of the same files
--This is the need of the hour considering multiple data sources can be connected to the same digitize instance and files could be common amongst sources

-Hash file (SHA256) when pipeline is triggered 
-If hash matches from PostgresDB, 
--If all files are duplicate, return 409, rejects the job
--If some files are duplicate, return 202, accepts the job and updates duplicate status
-Else, if non-duplicate
--Ingest as usual and save hash in DB 

Visuals:
-Some files are duplicate 
Screenshot 2026-07-17 at 11 51 32

-All files are duplicate (rejects job)
Screenshot 2026-07-17 at 11 51 09

### DIFF
--- a/docs/proposals/digitize-file-hash-dedup/file-hash-dedup-implementation.md
+++ b/docs/proposals/digitize-file-hash-dedup/file-hash-dedup-implementation.md
@@ -1,0 +1,1133 @@
+# File-Hash Duplicate Detection — Implementation Specification
+
+**Branch:** `dupIndex`
+**Scope:** Both `operation=ingestion` and `operation=digitization`.
+**Strategy:** Option B — lenient batch dedup. Strip already-existing files from a
+multi-file submission, process only novel files, and record skipped files directly in
+the job's document list with status `already_exists`.
+
+### Diagrams
+
+| Flowchart | Sequence |
+|---|---|
+| ![Decision flowchart](flowchart.svg) | ![Sequence diagram](sequence.svg) |
+
+---
+
+## Table of Contents
+
+1. [Overview](#1-overview)
+2. [New `ALREADY_EXISTS` Document Status](#2-new-already_exists-document-status)
+3. [New `JobCreatedResponse` Shape](#3-new-jobcreatedresponse-shape)
+4. [Touch-point 1 — Hash utility (`services/common/misc_utils.py`)](#4-touch-point-1--hash-utility)
+5. [Touch-point 2 — DB schema: `file_checksum_registry` table](#5-touch-point-2--db-schema-file_checksum_registry-table)
+6. [Touch-point 3 — `DatabaseManager.find_completed_document_by_hash` and `upsert_file_checksum`](#6-touch-point-3--databasemanager-hash-methods)
+7. [Touch-point 4 — `create_job` API handler (Option B gate)](#7-touch-point-4--create_job-api-handler-option-b-gate)
+8. [Touch-point 5 — `initialize_job_state` (record already-exists docs)](#8-touch-point-5--initialize_job_state-record-already-exists-docs)
+9. [Touch-point 6 — `_run_ingest` and `ingest()` (pass checksum dict)](#9-touch-point-6--_run_ingest-and-ingest-pass-checksum-dict)
+10. [Touch-point 7 — `_run_digitize` and `digitize()` (pass checksum dict)](#10-touch-point-7--_run_digitize-and-digitize-pass-checksum-dict)
+11. [Touch-point 8 — `create_indexing_handler` (persist hash on completion)](#11-touch-point-8--create_indexing_handler-persist-hash-on-completion)
+12. [Touch-point 9 — `_categorize_fields` (route `file_hash` to metadata)](#12-touch-point-9--_categorize_fields-route-file_hash-to-metadata)
+13. [Touch-point 10 — `_update_document` (register checksum on COMPLETED)](#13-touch-point-10--_update_document-register-checksum-on-completed)
+14. [Touch-point 11 — Deletion clears checksum registry](#14-touch-point-11--deletion-clears-checksum-registry)
+15. [Touch-point 12 — `get_all_documents` excludes `ALREADY_EXISTS` docs](#15-touch-point-12--get_all_documents-excludes-already_exists-docs)
+16. [Data Flow Diagram](#16-data-flow-diagram)
+17. [Timing Impact](#17-timing-impact)
+18. [Pros and Cons](#18-pros-and-cons)
+19. [Edge Cases](#19-edge-cases)
+
+---
+
+## 1. Overview
+
+When a caller submits a batch of files, each file's SHA-256 is computed
+**asynchronously (`asyncio.to_thread`) from the already-in-memory bytes** (zero
+additional I/O) before any staging or pipeline work begins. The hash is looked up
+via the dedicated `file_checksum_registry` table in PostgreSQL (indexed PK read,
+~2 ms).
+
+This feature applies to **both** `operation=ingestion` and `operation=digitization`.
+A digitization request is additionally blocked if the same file content has already
+been successfully ingested — there is no point re-digitizing content that is already
+in the vector store.
+
+**Option B behaviour:**
+
+| Scenario | Result |
+|---|---|
+| All files are new | Normal `202 Accepted` — full pipeline runs |
+| Some files already exist | `202 Accepted` — novel files processed, already-existing files recorded as `ALREADY_EXISTS` in the same job (visible via `GET /v1/jobs/{job_id}`) |
+| All files already exist | `409 Conflict` — no job created, no pipeline launched |
+
+A file is considered to already exist only if a **previous document** exists with:
+- `type = operation` (or `type = 'ingestion'` when the current operation is
+  `digitization`)
+- `status = 'completed'`
+- A matching entry in `file_checksum_registry`
+
+`failed` and `in_progress` records are **not** treated as already-existing —
+re-processing is allowed.
+
+---
+
+## 2. New `ALREADY_EXISTS` Document Status
+
+### 2a. `services/digitize/models.py`
+
+`ALREADY_EXISTS` is added to `DocStatus`:
+
+```python
+class DocStatus(str, Enum):
+    ACCEPTED = "accepted"
+    IN_PROGRESS = "in_progress"
+    DIGITIZED = "digitized"
+    PROCESSED = "processed"
+    CHUNKED = "chunked"
+    COMPLETED = "completed"
+    FAILED = "failed"
+    ALREADY_EXISTS = "already_exists"   # ← NEW
+```
+
+### 2b. `services/digitize/db/scripts/init_schema.sql`
+
+The `chk_doc_status` CHECK constraint includes `'already_exists'`:
+
+```sql
+CONSTRAINT chk_doc_status CHECK (
+    status IN ('accepted', 'in_progress', 'digitized', 'processed',
+               'chunked', 'completed', 'failed', 'already_exists')
+)
+```
+
+⚠️ This is a schema migration. On existing databases drop and re-create the
+constraint before deploying:
+
+```sql
+ALTER TABLE documents
+    DROP CONSTRAINT chk_doc_status,
+    ADD CONSTRAINT chk_doc_status CHECK (
+        status IN ('accepted', 'in_progress', 'digitized', 'processed',
+                   'chunked', 'completed', 'failed', 'already_exists')
+    );
+```
+
+### 2c. `services/digitize/db/models.py`
+
+The `__table_args__` `CheckConstraint` on `Document` mirrors the SQL:
+
+```python
+CheckConstraint(
+    "status IN ('accepted', 'in_progress', 'digitized', 'processed',"
+    " 'chunked', 'completed', 'failed', 'already_exists')",
+    name="chk_doc_status"
+),
+```
+
+---
+
+## 3. `AlreadyExistsFile` and `JobCreatedResponse` Shape
+
+### 3a. `services/digitize/models.py`
+
+The `DuplicateSkippedFile` model from the original proposal was renamed to
+`AlreadyExistsFile`. It carries `existing_doc_name` internally so the
+`JobDocumentSummary` can construct a human-readable `message` field —
+`"Already ingested as <filename>"` — without an extra API call.
+
+`JobCreatedResponse` contains only `job_id` — the `warnings` field was removed.
+Skipped files are recorded directly in the job's document list with
+`status = already_exists` and `message = "Already ingested as <name>"`,
+visible via `GET /v1/jobs/{job_id}`.
+
+```python
+class AlreadyExistsFile(BaseModel):
+    """Describes a single file that was skipped because it already exists."""
+    filename: str = Field(..., description="Original filename of the skipped file")
+    existing_doc_id: str = Field(..., description="doc_id of the already-ingested document")
+    existing_doc_name: str = Field(..., description="Name of the already-ingested document")
+    file_hash: str = Field(..., description="SHA-256 hash that matched, e.g. 'sha256:e3b0...'")
+
+
+class JobDocumentSummary(BaseModel):
+    """Compact per-document entry for job status responses."""
+    id: str
+    name: str
+    status: str
+    message: Optional[str] = Field(
+        default=None,
+        description="Human-readable message; set when status is 'already_exists', e.g. 'Already ingested as <filename>'",
+    )
+
+
+class JobCreatedResponse(BaseModel):
+    """Response model for job creation."""
+    job_id: str
+```
+
+---
+
+## 4. Touch-point 1 — Hash Utility
+
+**File:** `services/common/misc_utils.py`
+
+Rather than a dedicated `services/digitize/utils/hash.py`, the hash function lives in
+the shared utilities module as `generate_file_checksum`. It accepts either raw `bytes`
+(hashed in one shot) or a file path (read in 128-block chunks to support large files
+without loading them into memory):
+
+```python
+def generate_file_checksum(file) -> str:
+    """Compute a prefixed SHA-256 hex digest of a file.
+
+    Accepts either a file path (str or Path) or raw bytes.  When passed bytes
+    the entire content is hashed in one shot; when passed a path the file is
+    read in chunks so arbitrarily large files can be processed without loading
+    them entirely into memory.
+
+    The ``sha256:`` prefix makes the algorithm explicit in stored values so
+    the hash algorithm can be upgraded in future without ambiguity.
+
+    Returns:
+        String in the form ``'sha256:<64-char-hex>'``.
+    """
+    sha256 = hashlib.sha256()
+    if isinstance(file, (bytes, bytearray)):
+        sha256.update(file)
+    else:
+        with open(file, 'rb') as f:
+            for chunk in iter(lambda: f.read(128 * sha256.block_size), b''):
+                sha256.update(chunk)
+    return f"sha256:{sha256.hexdigest()}"
+```
+
+**Usage at the API layer** (`services/digitize/api/v1/jobs.py`):
+
+```python
+from common.misc_utils import generate_file_checksum
+...
+file_hash = await asyncio.to_thread(generate_file_checksum, content)
+```
+
+Hashing is offloaded to a thread pool via `asyncio.to_thread` to avoid blocking the
+async event loop (addressing the Con identified in the earlier proposal).
+
+**Usage in the digitize pipeline** (`services/digitize/pipeline/digitize.py`): when
+a pre-computed checksum is not available in `file_checksum_dict`, the function
+re-hashes the staged file from disk as a fallback:
+
+```python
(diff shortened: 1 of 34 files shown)


services/digitize/db/scripts/init_schema.sql
```diff
@@ -27,7 +27,7 @@ CREATE TABLE IF NOT EXISTS documents (
     error TEXT,
     metadata JSONB NOT NULL DEFAULT '{}',
     updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,  -- Last modification time (UTC)
-    CONSTRAINT chk_doc_status CHECK (status IN ('accepted', 'in_progress', 'digitized', 'processed', 'chunked', 'completed', 'failed')),
+    CONSTRAINT chk_doc_status CHECK (status IN ('accepted', 'in_progress', 'digitized', 'processed', 'chunked', 'completed', 'failed', 'duplicate_skipped')),
```
dharaneeshvrd: can we use `duplicate_skipped` better status here?
Something simpler like ALREADY_EXISTS can be a good alternative

services/digitize/db/scripts/init_schema.sql
```diff
@@ -63,3 +65,13 @@ END
 $$;
 
 -- Note: Using postgres superuser, no additional grants needed
+
+-- Migration: run this once on existing databases before deploying the new code
```
dharaneeshvrd: Not required

services/digitize/api/v1/jobs.py
```diff
@@ -221,7 +295,8 @@ async def create_job(
             )
             APIError.raise_error("INTERNAL_SERVER_ERROR", str(exc))
 
-        return {"job_id": job_id}
+        # Return job_id plus any duplicate warnings.
+        return {"job_id": job_id, "warnings": skipped_files}
```
dharaneeshvrd: I don't think we need to return this here

services/digitize/api/v1/jobs.py
```diff
@@ -195,22 +198,93 @@ async def create_job(
...
+        #
+        # A file is a duplicate only if a document with:
+        #   - type  = 'ingestion'
+        #   - status = 'completed'
+        #   - metadata->>'file_hash' = computed_hash
+        # already exists. Failed / in-progress records are NOT duplicates.
+
+        skipped_files: list[models.DuplicateSkippedFile] = []
+
+        if operation == models.OperationType.INGESTION:
```
dharaneeshvrd: Let's do this for both operation types
Digitize document flow also can use this logic to avoid digitizing again if the doc exists already

services/digitize/api/v1/jobs.py
```diff
@@ -195,22 +198,93 @@ async def create_job(
...
+
         # 6. Acquire semaphore slot.
         await concurrency_manager.acquire(op_key)
 
         # 7. Stage files and schedule background task.
         try:
+            # Build hash dict from the novel files so it travels to the indexing callback.
+            file_hash_dict: dict[str, str] = {}
+            if operation == models.OperationType.INGESTION:
+                file_hash_dict = {
```
dharaneeshvrd: why are we calculating the hash again here, lets cache above and use it here

services/digitize/api/v1/jobs.py
```diff
@@ -63,14 +65,15 @@ async def _run_ingest(
     job_id: str,
     filenames: List[str],
     doc_id_dict: dict,
+    file_hash_dict: Optional[dict] = None,  # filename -> "sha256:..."
```
manju956: can we rename to `file_checksum_dict` for better readability?

services/digitize/api/v1/jobs.py
manju956: +1

services/digitize/api/v1/jobs.py
```diff
@@ -195,22 +198,93 @@ async def create_job(
...
+            file_contents = novel_contents
+
         # 6. Acquire semaphore slot.
         await concurrency_manager.acquire(op_key)
 
         # 7. Stage files and schedule background task.
         try:
+            # Build hash dict from the novel files so it travels to the indexing callback.
+            file_hash_dict: dict[str, str] = {}
+            if operation == models.OperationType.INGESTION:
```
manju956: this applies to DIGITISE operation as well?

services/digitize/db/scripts/init_schema.sql
manju956: ALREADY_EXISTS looks better option

services/digitize/pipeline/ingest.py
```diff
@@ -122,7 +135,12 @@ def index_document_chunks(doc_id: str, chunks: list, path: str) -> bool:
 
     return index_document_chunks
 
-def ingest(directory_path: Path, job_id: Optional[str] = None, doc_id_dict: Optional[dict] = None):
+def ingest(
```
manju956, line 138: Make relevant changes in digitize too.

services/digitize/utils/hash.py
```diff
@@ -0,0 +1,32 @@
...
+
+Computes a stable SHA-256 digest from raw file bytes.
+The result is stored in document metadata as `file_hash` and used
+to detect duplicate ingestion submissions before pipeline work begins.
+"""
+
+import hashlib
+
+
+def compute_file_hash(content: bytes) -> str:
```
manju956: we already have this method `generate_file_checksum` in misc_utils.py

services/digitize/api/v1/jobs.py
sats-23: It is used on the UI side to fetch the current matching doc-name that is already ingested

services/digitize/db/scripts/init_schema.sql
sats-23: Removed

services/digitize/api/v1/jobs.py
sats-23: Reused the prev computation

services/digitize/api/v1/jobs.py
sats-23: Yes, updated it

services/digitize/api/v1/jobs.py
sats-23: Sure, updated it

services/digitize/api/v1/jobs.py
sats-23: Done

services/digitize/pipeline/ingest.py
sats-23: Done

services/digitize/utils/hash.py
sats-23: Didn't consider it as it involves a disk write in every usage. compute_file_hash does it in-memory and provides with format prefix. 

Let me know if we should still proceed with it

services/digitize/api/v1/jobs.py
dharaneeshvrd: ok in that case, swagger needs to be updated

sats-23: Done

sats-23:
@dharaneeshvrd @manju956 
Currently with support for digitization (along with ingestion), here's how it works:
-Existing digitized doc prevents duplicate digitization 
-Existing ingested doc prevents duplicate ingestion

Question: Should we add support for Existing ingested doc prevents duplicate digitization?

dharaneeshvrd:
>  Should we add support for Existing ingested doc prevents duplicate digitization?

Yes, we are handling the document as an independent entity, so cross checking file hash can be allowed.

services/digitize/db/scripts/init_schema.sql
```diff
@@ -36,6 +36,8 @@ CREATE TABLE IF NOT EXISTS documents (
 CREATE INDEX IF NOT EXISTS idx_jobs_submitted_at_status ON jobs(submitted_at DESC, status);
 CREATE INDEX IF NOT EXISTS idx_documents_job_id ON documents(job_id);
 CREATE INDEX IF NOT EXISTS idx_documents_submitted_at_status ON documents(submitted_at DESC, status);
+CREATE INDEX IF NOT EXISTS idx_documents_file_hash
```
sats-23: Planning to replace this with a separate table to use re-used by connectors
@dharaneeshvrd @manju956

services/digitize/utils/hash.py
manju956: sure, pls proceed

services/digitize/db/scripts/init_schema.sql
manju956: what would be the new table schema?

sats-23: The same one in the connectors proposal 

```
CREATE TABLE IF NOT EXISTS file_checksum_registry (
    sha256        TEXT        PRIMARY KEY,
    doc_id        TEXT        NOT NULL UNIQUE    -- one Document row per unique content
);
```

services/digitize/utils/hash.py
sats-23: Done

services/digitize/db/scripts/init_schema.sql
sats-23: Updated

services/digitize/api/v1/jobs.py
dharaneeshvrd: Now I feel this is bit extra.
Why can't we follow the same pattern where an existing file with new file is submitted.
I think we can admit the job and mark the document as already ingested.
I don't want to take a different UX here.

services/digitize/db/scripts/init_schema.sql
dharaneeshvrd: @sats-23 Wdyt here?

services/digitize/api/v1/jobs.py
```diff
@@ -207,10 +300,13 @@ async def create_job(
...
+                job_id, operation, output_format, filenames, job_name,
+                skipped_files=skipped_files,
             )
             if operation == models.OperationType.INGESTION:
-                background_tasks.add_task(_run_ingest, job_id, filenames, doc_id_dict)
+                background_tasks.add_task(
+                    _run_ingest, job_id, filenames, doc_id_dict, file_checksum_dict
+                )
             else:
                 background_tasks.add_task(_run_digitize, job_id, doc_id_dict, output_format)
```
dharaneeshvrd: I see we are utilising the computed checksum in ingest by passing the dict, but I dont see that happening for digitize, we are recomputing again, can we pass the computed hash from here and use it for the db update?

services/digitize/db/scripts/init_schema.sql
sats-23: Sure, let's go with already_Exists as it makes things easier

docs/proposals/digitize-file-hash-dedup/file-hash-dedup-implementation.md
```diff
@@ -0,0 +1,916 @@
+# File-Hash Duplicate Detection — Implementation Specification
+
+**Branch:** `dupIndex`  
+**Scope:** Ingestion pipeline only (`operation=ingestion`). Digitization is unaffected.
```
dharaneeshvrd: This needs to be updated.

services/digitize/utils/db.py
```diff
@@ -350,33 +353,45 @@ def create_document(
         output_format: Output format for the document
         operation: Type of operation (ingestion/digitization)
         submitted_at: ISO timestamp when document was submitted
+        initial_status: Initial document status (defaults to ACCEPTED)
```
manju956, line 360: what else can be `initial_status`

manju956: nevermind, i got it.

manju956 (dismissed):
LGTM

docs/proposals/digitize-file-hash-dedup/file-hash-dedup-implementation.md
sats-23: Done, updated proposal and all arch diagrams

services/digitize/api/v1/jobs.py
sats-23: Done, changed the method params

services/digitize/api/v1/jobs.py
sats-23: Yes, that is exactly what happens, we mark it as already_exists in DB after job is admitted, but just for the UI to show which doc_id is the duplicate ones, we are using this mechanism.

dharaneeshvrd: Ok, but I want to avoid displaying the warning message here.

dharaneeshvrd:
@pranithraoibm Can you please review the UI changes?

pranithraoibm:
@dharaneeshvrd Can you please assign my lead @ryanmarc for all the UI reviews going forward, he does all the UI PR reviews. Also please add the sqaud/ui label to the PR so that it shows up in the PR filters for Ryan to pick them up.

dharaneeshvrd:
@ryanmarc can you please review this PR and the reason for the change is mentioned in the description?
Since the changes are minimal on the UI side, I hope we can add this.

services/digitize/api/v1/jobs.py
```diff
@@ -203,6 +226,76 @@ async def create_job(
...
+        novel_filenames: list[str] = []
+        novel_contents:  list[bytes] = []
+
+        for filename, content in zip(filenames, file_contents):
+            file_hash = await asyncio.to_thread(generate_file_checksum, content)
+            existing = db_manager.find_completed_document_by_hash(file_hash, operation.value)
+
+            # A digitization request is also blocked by a completed ingestion of the
+            # same file — no point re-digitizing content that is already indexed.
+            if not existing and operation == models.OperationType.DIGITIZATION:
```
dharaneeshvrd, line 244: why are we involving operation type and calling the method twice?
just remove the operation type and do the check only once irrespective of the operation type

services/digitize/utils/jobs.py
```diff
@@ -131,6 +149,44 @@ def initialize_job_state(job_id: str, operation: str, output_format: OutputForma
...
+                extra_metadata={
+                    "existing_doc_id": skipped.existing_doc_id,
+                    "existing_doc_name": skipped.existing_doc_name,
+                    "file_hash": skipped.file_hash,
+                },
+            )
+            # Update job stats to include this resolved doc — no doc-level status
+            # write needed because it was already set correctly in create_document.
+            status_mgr.update_job_progress(
+                "",
```
dharaneeshvrd: dont we need to pass doc_id here?

services/digitize/api/v1/jobs.py
sats-23: That operation type check is strictly for the Ingestion, Digitize differentiation. We want 
1. Existing digitized doc to block duplicate digitize 
2. Existing ingested doc to block duplicate ingest
3. Existing ingested doc to block duplicate digitize

If we remove the operation type param, we will even end up with "Existing digitize doc blocks ingest doc" which is unfavorable as that doc will never reach the Vector store

dharaneeshvrd: Agree. thanks!

services/digitize/utils/jobs.py
sats-23: We want to skip db.py#_update_job (line 1051)
```
# Update document status if doc_id provided
        if doc_id:
            db_manager.update_document(doc_id, status=doc_status)
```
Since `create_document` has already changed the status to ALREADY_EXISTS, we don't want `update_document` to be called again. So to internally bypass, we are passing empty string

dharaneeshvrd: Then why are we updating the job status as in progress part of this method.

ui/digitize/src/constants/jobConstants.ts
```diff
@@ -15,6 +15,7 @@ export const DISPLAY_STATUS = {
   DIGITIZATION_ERROR: 'Digitization error',
   INGESTING: 'Ingesting...',
   DIGITIZING: 'Digitizing...',
+  ALREADY_EXISTS: 'already_exists',
```
ryanmarc, line 22: These are display value constants.. `already_exists` is not a display value.

services/digitize/utils/jobs.py
sats-23: It is just a status promotion, to show that we have moved past accepted, but still have some outstanding docs to be processed.

ui/digitize/src/constants/jobConstants.ts
sats-23: I have moved it out to a separate const called "DOC_STATUS"

services/digitize/utils/jobs.py
dharaneeshvrd: It should happen inside respective pipeline's(digitize or ingest), should not be part of POST handler.

sats-23: Done

dharaneeshvrd: Actually this line itself is not needed, since create_job will anyway set the status to ACCEPTED.

sats-23: But it is needed for the stats count. If we don't update_job_progress here, even though some docs (duplicates are recognized and resolved) will not update stats. Instead, we can make job_status optional param so that we make sure to update ALREADY_EXISTS in job status (Refer latest commit)

services/digitize/models.py
```diff
@@ -81,10 +91,14 @@ class DocumentContentResponse(BaseModel):
...
     """Compact per-document entry for job status responses."""
     model_config = ConfigDict(use_enum_values=True)
-    
+
     id: str
     name: str
     status: str
+    message: Optional[str] = Field(
+        default=None,
+        description="Human-readable message; set when status is 'already_exists'",
```
dharaneeshvrd: Can you please rewrite this description to make the messages reflect the status in general and not specific already_exists status?
Also please create a story to handle other status later?

services/digitize/tests/test_digitize_app_endpoints.py
dharaneeshvrd, line 1: I feel some more test cases can be added.

services/digitize/models.py
sats-23: Done, ticket [here](https://jsw.ibm.com/browse/AISERVICES-1495)

services/digitize/tests/test_digitize_app_endpoints.py
sats-23: Done

services/digitize/api/v1/jobs.py
```diff
@@ -123,16 +126,36 @@ async def _validate_files(
...
+        "of the same operation type already stored in the database.\n\n"
+        "| Scenario | Behaviour |\n"
+        "|---|---|\n"
+        "| All files already exist | **409 Conflict** — no job is created |\n"
+        "| Some files already exist | **202 Accepted** — only novel files are processed; "
+        "skipped files appear in the job's document list with `status=already_exists` and "
+        "a `message` field such as `\"Already ingested as <filename>\"` |\n"
+        "| No files already exist | **202 Accepted** |\n\n"
+        "Only *completed* documents trigger already-exists detection. Failed or in-progress records "
+        "are not considered to already exist."
```
dharaneeshvrd: I think it's too much of information regarding already_exists, let's keep it minimal and see this new response type can be added to the existing list.

services/digitize/db/models.py
```diff
@@ -150,4 +151,29 @@ class Document(Base):
...
+    The FK to documents(doc_id) ON DELETE CASCADE means registry entries are
+    automatically removed when the referenced document is deleted, preventing
+    orphaned hashes from blocking future re-ingestion of the same file.
+    """
+    __tablename__ = "file_checksum_registry"
+
+    sha256: Mapped[str] = mapped_column(Text, primary_key=True)
+    doc_id: Mapped[str] = mapped_column(
+        Text,
+        ForeignKey("documents.doc_id", ondelete="CASCADE"),
```
dharaneeshvrd, line 171: we should not cascade delete here right?
Hope you would take care in other PRs

services/digitize/api/v1/jobs.py
sats-23: Done, I have simplified it

services/digitize/db/models.py
sats-23: We need it in cases of when a doc is ingested and later deleted. It's stale entry can remain in the registry and end up blocking future ingests.

dharaneeshvrd: But what about same doc used by other jobs? There maybe a chance where the doc which is added the entry into the table gets deleted, but the hash entry should not be removed right? what do we need to do here?

dharaneeshvrd (approved):

services/digitize/db/models.py
dharaneeshvrd: I m merging this for now, since this can be addressed later too.

PR #1125, opened 2026-07-13
refactor: establish new folder structure, move flow-specific hooks, and delete dead code

pranithraoibm:
### Folder restructure
- `DeployFlow/` → `DeployFlow/DigitalAssistant/`
- `ServicesDeployFlow/` → `DeployFlow/Services/`
- `DeployFlow/Shared/` created with two extracted shared utils: `deployError.ts`, `displayHelpers.ts`

- `useDeployOptions`, `useProviderParams`, `useServiceParams` → `DigitalAssistant/hooks/`
- `useServiceDeployOptions`, `useProviderSchema` → `Services/hooks/`

- `formDataInitializer.ts`, `deploymentTransform.ts` (renamed to `digitalAssistantDeploymentTransform.ts`), `inferenceComponentHelper.ts`, `resourceSharing.ts` → `DigitalAssistant/utils/`
- `serviceDeploymentTransform.ts` → `Services/utils/`

### Dead code deleted
- `src/components/DynamicField/` — never imported anywhere
- `src/hooks/useResources.ts` — no consumers

### Import paths updated
- `components/index.ts`, `DigitalAssistants.tsx`, `Services.tsx`

---
> No logic changes — files are moved only.


ui/catalog/src/components/DeployFlow/Shared/utils/deployError.ts
```diff
@@ -0,0 +1,15 @@
...
+    const err = error as {
+      response?: {
+        data?: { detail?: string; message?: string; error?: string };
+      };
+      message?: string;
+    };
+    if (err.response?.data?.detail) return err.response.data.detail;
+    if (err.response?.data?.message) return err.response.data.message;
+    if (err.response?.data?.error) return err.response.data.error;
+    if (err.message) return err.message;
```
ryanmarc: This isn't a pure de-duplication, the Services flow's old handler used else if (error instanceof Error), so an axios error whose body had no error/detail/message field fell through to the friendly "Failed to deploy application". With the shared helper it now falls through to err.message, i.e. the Services deploy toast will show "Request failed with status code 500" instead. If that's intended, fine, but let's make it explicit; otherwise gate the last branch so raw axios strings don't leak to users

ui/catalog/src/components/DeployFlow/Services/components/ResourceRequirements.tsx
```diff
@@ -13,11 +13,11 @@ import type {
   ResourcesResponse,
   ServiceDeployOptions,
 } from "@/types/api.types";
-import type { ResourceItem } from "../../DeployFlow/types/StepTwo.types";
+import type { ResourceItem } from "../../DigitalAssistant/types/StepTwo.types";
 import {
   bytesToGB,
   getResourceStatus,
-} from "../../DeployFlow/utils/StepTwo.utils";
+} from "../../DigitalAssistant/utils/StepTwo.utils";
```
ryanmarc: The new DigitalAssistant/ - Services/ split only holds if the two siblings don't import each other's internals. ResourceItem, bytesToGB and getResourceStatus are flow-agnostic, since this PR already creates Shared/utils/, could we move them there (Shared/types/resources.ts, Shared/utils/resources.ts) and have both flows import from Shared/? Otherwise the folder boundary is cosmetic and DigitalAssistant/utils/StepTwo.utils.ts stays load-bearing for Services.

ui/catalog/src/hooks/useResources.ts
ryanmarc, line 1: ui/catalog/src/store/deploy.store.ts:55:useResources was this slice's only consumer, after its deletion in this PR, resources / resourcesLoading / resourcesError / resourcesFetchedAt, their four setters, isResourcesStale and RESOURCES_CACHE_DURATION are all unreachable (the two components that show resources use local useState). Since "delete dead code" is one of the PR's goals, could we drop the slice here too? tsc won't catch it because they're part of the exported store interface.

ui/catalog/src/components/DeployFlow/DigitalAssistant/steps/StepOne.tsx
```diff
@@ -26,21 +26,29 @@ export const StepOne: React.FC<StepProps> = ({
...
+      {
+        paramsMap: Record<string, import("@/types/api.types").ProviderSchema>;
+        isLoading: boolean;
+        errors: Record<string, string>;
+      }
+    > = {};
+    Object.entries(paramsByType).forEach(([componentType, paramsMap]) => {
+      result[componentType] = {
+        paramsMap,
+        isLoading: false,
```
ryanmarc: useMultiTypeProviderParams returns a real isLoading, but the adapter hardcodes false here, so the field can never be true and any future skeleton/spinner wired to it silently no-ops. Either thread the hook's value through (isLoading is a single flag for the whole batch, so isLoading from the destructure on line 31) or drop the key from the adapter shape entirely, since nothing reads it today.

ryanmarc (changes requested):

ui/catalog/src/components/DeployFlow/Shared/utils/deployError.ts
pranithraoibm: Fixed — gated the last branch on error instanceof Error so axios errors without a structured body fall through to the friendly message.

ui/catalog/src/components/DeployFlow/Services/components/ResourceRequirements.tsx
pranithraoibm: Moved ResourceItem to Shared/types.ts, bytesToGB and getResourceStatus to Shared/utils/resources.ts, and updated both ResourceRequirements files to import from Shared/

ui/catalog/src/hooks/useResources.ts
pranithraoibm: Removed them

ui/catalog/src/components/DeployFlow/DigitalAssistant/steps/StepOne.tsx
pranithraoibm: Dropped it

pranithraoibm:
@ryanmarc I have addressed the review comments PTAL

ui/catalog/src/components/DeployFlow/Shared/utils/deployError.ts
```diff
@@ -9,7 +9,7 @@ export function extractDeployError(error: unknown): string {
     if (err.response?.data?.detail) return err.response.data.detail;
     if (err.response?.data?.message) return err.response.data.message;
     if (err.response?.data?.error) return err.response.data.error;
-    if (err.message) return err.message;
+    if (error instanceof Error) return error.message;
```
ryanmarc: Thanks for tightening this up, but I think it still has the behaviour change I was worried about. Swapping err.message for error instanceof Error is better typing, but axios errors are Error instances, so this line still fires for them. The old Services handler used else if (error instanceof Error), so an axios error that had a response but no detail/message/error in the body skipped straight to the friendly fallback. Now it returns error.message, which for axios is the string "Request failed with status code 500". I checked with a real AxiosError from our own node_modules and the old code gives "Failed to deploy application" while this gives "Request failed with status code 500", so users will start seeing raw HTTP text in the Services deploy toast.

If showing the axios message is what we want, ok, If not, the fix is to only use the message when there is no response attached.

ui/catalog/src/components/DeployFlow/Services/steps/StepTwo.tsx
```diff
@@ -17,6 +17,7 @@ import type { StepProps, ServiceConfig } from "../types";
 import { ResourceRequirements } from "../components/ResourceRequirements";
 import { DynamicSchemaFields } from "../components/DynamicSchemaFields";
 import { ServiceCredentialDisplay } from "../components/ServiceCredentialDisplay";
+import { getDisplayName } from "../../Shared/utils/displayHelpers";
```
ryanmarc, line 20: Two small things now that Shared exists. The DA flow reaches these through this shim while the Services flow imports Shared/utils/resources directly, so we have two import paths for the same functions. There are only three call sites on the DA side (steps/StepTwo.tsx:4, components/ServiceConfigCard.tsx:14, types/StepTwo.types.ts:3), so pointing them at Shared and deleting the re-exports would leave one canonical path.

Also the doc comments got dropped in the move. getResourceStatus used to have "Determines if resources are sufficient, insufficient, or unknown" and bytesToGB had "Converts bytes to gigabytes (rounded)". Worth pasting those onto Shared/utils/resources.ts, since right now all four Shared files have no comments at all and they are the ones people will read first.

ryanmarc (changes requested):

ui/catalog/src/components/DeployFlow/Shared/utils/deployError.ts
pranithraoibm: Fixed by adding `&& !err.response` to only use error.message for network-level errors (no response attached). Axios errors with an empty response body now fall through to the friendly fallback. Also tightened the body field checks to typeof === "string" to avoid returning [object Object] if the backend sends a non-string value.


Tested it using a small script
```
import { extractDeployError } from "./deployError";

const results: { name: string; passed: boolean; got: string; expected: string }[] = [];

function test(name: string, got: string, expected: string) {
  results.push({ name, passed: got === expected, got, expected });
}

// 1. Axios error with a response but empty body → friendly fallback
test(
  "axios response with empty body",
  extractDeployError({ response: { data: {} }, message: "Request failed with status code 500" }),
  "Failed to deploy application",
);

// 2. Axios error with detail in body → use detail
test(
  "axios response with detail",
  extractDeployError({ response: { data: { detail: "Image not found" } }, message: "Request failed with status code 404" }),
  "Image not found",
);

// 3. Axios error with message in body → use message
test(
  "axios response with message",
  extractDeployError({ response: { data: { message: "Quota exceeded" } }, message: "Request failed with status code 403" }),
  "Quota exceeded",
);

// 4. Axios error with error in body → use error
test(
  "axios response with error field",
  extractDeployError({ response: { data: { error: "Namespace not found" } }, message: "Request failed with status code 400" }),
  "Namespace not found",
);

// 5. Network error (no response) → use error.message
const networkErr = Object.assign(new Error("Network Error"), { response: undefined });
test("network error (no response)", extractDeployError(networkErr), "Network Error");

// 6. Plain Error with no response property → use error.message
test("plain Error", extractDeployError(new Error("something broke")), "something broke");

// 7. Non-string detail (array) → friendly fallback, not "[object Object]"
test(
  "axios response with non-string detail",
  extractDeployError({ response: { data: { detail: ["error1", "error2"] } }, message: "Request failed with status code 400" }),
  "Failed to deploy application",
);

// 8. Non-string detail (object) → friendly fallback
test(
  "axios response with object detail",
  extractDeployError({ response: { data: { detail: { code: 400 } } }, message: "Request failed with status code 400" }),
  "Failed to deploy application",
);

// 9. Completely unknown value → friendly fallback
test("unknown string error", extractDeployError("oops"), "Failed to deploy application");

// 10. null → friendly fallback
test("null", extractDeployError(null), "Failed to deploy application");

// Report
let failed = 0;
for (const r of results) {
  if (r.passed) {
    console.log(`  ✓  ${r.name}  →  "${r.got}"`);
  } else {
    console.log(`  ✗  ${r.name}`);
    console.log(`       expected: "${r.expected}"`);
    console.log(`       got:      "${r.got}"`);
    failed++;
  }
}

console.log(`\n${results.length - failed}/${results.length} passed`);
if (failed > 0) process.exit(1);
```

And response was as follows
```
✓  axios response with empty body  →  "Failed to deploy application"
  ✓  axios response with detail  →  "Image not found"
  ✓  axios response with message  →  "Quota exceeded"
  ✓  axios response with error field  →  "Namespace not found"
  ✓  network error (no response)  →  "Network Error"
  ✓  plain Error  →  "something broke"
  ✓  axios response with non-string detail  →  "Failed to deploy application"
  ✓  axios response with object detail  →  "Failed to deploy application"
  ✓  unknown string error  →  "Failed to deploy application"
  ✓  null  →  "Failed to deploy application"

10/10 passed
```

ui/catalog/src/components/DeployFlow/Services/steps/StepTwo.tsx
pranithraoibm: Updated all the imports and added the comments

pranithraoibm:
Lint, compile and build checks and UI verification post changes (Since this is just code refactor change tested the deploy flow for both only, nothing specific)

https://github.com/user-attachments/assets/23797774-0a84-4950-b060-8df039985a7e

ryanmarc (dismissed):

Niranjan-J-Joshi (approved):

PR #1126, opened 2026-07-14
refactor: Create a shared foundation by extracting common types, utilities and styles

Niranjan-J-Joshi:
Create a shared foundation by extracting common types, utility functions, and styling that are currently duplicated across both tables. This provides a consistent base for the reusable hooks and components introduced in later phases.

**Implementation:**

* `shared/types.ts` — `RowStatus`, `TableHeaders`, `BaseTableState`
* `utils/reducerUtils.ts` — shared pure reducer helper functions
* `utils/tableUtils.ts` — `filterRowsBySearch`, `getVisibleHeaders`, `getToggleableHeaders`
* `shared/table.shared.module.scss` — shared CSS rules extracted from both `.module.scss` files: `.customToast`, `.deleteConfirmation`, `.overflowMenu*`, `.deleteMenuItem`, `.messageWithIcon`, `.messageIcon*`, `.statusTag*`


ui/catalog/src/components/Table/types.ts
```diff
@@ -0,0 +1,90 @@
...
+  isDeleteDialogOpen: boolean;
+  isConfirmed: boolean;
+  selectedRowId: string | null;
+
+  // Delete error toast
+  toastOpen: boolean;
+  deleteErrorMessage: string;
+  deleteErrorRowName: string;
+  isDeleting: boolean;
+  hasError: boolean;
```
ryanmarc, line 42: hasError and isLoading come straight from DeployedServicesTable's state, but DigitalAssistants' AppState has no hasError at all and calls the loading flag isLoadingApplications (pages/DigitalAssistants/types.ts:40). So AppState extends BaseTableState won't be a drop-in for the DA table, it needs a rename plus a field it never sets.

ui/catalog/src/components/Table/utils/reducerUtils.ts
```diff
@@ -0,0 +1,240 @@
...
+  switch (action.type) {
+    case "SHARED_SET_SEARCH":
+      return { ...state, ...setSearch(action.payload) };
+    case "SHARED_SET_PAGE":
+      return { ...state, ...setPage(action.payload) };
+    case "SHARED_SET_PAGE_SIZE":
+      return { ...state, ...setPageSize(action.payload) };
+    case "SHARED_OPEN_DELETE_DIALOG":
+      return { ...state, ...openDeleteDialog(action.payload) };
+    case "SHARED_CLOSE_DELETE_DIALOG":
```
ryanmarc, line 52: hasError and selectedRowId are already on state here, passing them through the action makes every dispatch site read state first, and a value captured in a stale closure would diverge from the reducer's own state. Can this be ...closeDeleteDialog(state.hasError, state.selectedRowId) and the action reduced to { type: "SHARED_CLOSE_DELETE_DIALOG" }? (closeDeleteDialog itself can stay as-is.)

ryanmarc (changes requested):

ui/catalog/src/components/Table/utils/reducerUtils.ts
Niranjan-J-Joshi: I have removed the unnecessary fields from SHARED_CLOSE_DELETE_DIALOG, so it's now just:
{ type: "SHARED_CLOSE_DELETE_DIALOG" }
The handler now reads hasError and selectedRowId directly from state

ui/catalog/src/components/Table/types.ts
Niranjan-J-Joshi: This PR only adds the shared files and doesn't change Digital Assistants or Deployed Services files yet, 

The differences mentioned (isLoadingApplications vs isLoading, and DA not having hasError) will be added in PR4 where we will be migrating Digital Assistants to use the shared state and actions. Deployed Services will be migrated in PR5.

ui/catalog/src/components/table/utils/reducerUtils.ts
```diff
@@ -0,0 +1,240 @@
+import type {
+  BaseTableState,
+  SharedTableAction,
+} from "@/components/table/types";
+
+export function handleSharedTableAction<S extends BaseTableState>(
```
ryanmarc: Any chance we can land one real consumer in this PR, even just the search/page/pageSize actions on DeployedServicesTable? Right now none of this is exercised by anything, and there are three things I cannot check without a caller.

First, the SHARED_* names do not match either table. We have SET_SEARCH on the DA page and DEPLOYED_SERVICES_SET_SEARCH on the services table, so adopting this means renaming about twenty constants plus every dispatch site in two big components. Worth deciding now whether the shared names win or whether the handler should accept the existing ones.

Second, a consumer's reducer takes its own AppAction union, which is not assignable to SharedTableAction, so everyone is going to write handleSharedTableAction(state, action as unknown as SharedTableAction). That cast silently defeats the union safety this whole file depends on. Exporting a type guard like isSharedTableAction(a): a is SharedTableAction would let callers narrow properly, and it also gives them something better than undefined to branch on.

Doing this with one caller now is a lot cheaper than after two tables have migrated.

ryanmarc (changes requested):

ui/catalog/src/components/table/utils/reducerUtils.ts
Niranjan-J-Joshi: - Wired up a Search consumer in DeployedServicesTable — TableToolbarSearch now dispatches SHARED_SET_SEARCH, which flows through isSharedTableAction → handleSharedTableAction.

- Shared actions will use the SHARED_* naming pattern, and dispatch calls in both tables will be updated accordingly.

- isSharedTableAction is exported from reducerUtils.ts and is used to identify shared actions before calling handleSharedTableAction

Niranjan-J-Joshi:
Added two more types/actions in shared folder

1. SHARED_SET_DELETING for tracking delete state.
2. SHARED_UPDATE_ROW_STATUS for updating row status

Also added a shared base row type so these actions can be reused by both tables.

I updated SHARED_SET_FETCH_ERROR as well, so clearing an old error no longer stops loading. Only a real error will do that.

These shared types and actions will be used by both the DA and DS table reducers, reducing duplicate code as more logic is moved into the shared layer.

ui/catalog/src/components/DeployedServicesTable/types.ts
```diff
@@ -235,15 +219,13 @@ export const INITIAL_STATE: AppState = {
   fetchError: null,
 };
 
-// Reducer
-export const appReducer = (state: AppState, action: AppAction): AppState => {
+// DS-specific cases only. SHARED_SET_SEARCH is handled by handleSharedTableAction.
```
ryanmarc, line 222: This comment says only SHARED_SET_SEARCH has been migrated, but the intent of this PR is to share all these transitions. Currently ownCases re-implements 15 handlers that are already implemented in reducerUtils.ts. The appReducer wires up handleSharedTableAction but no callsite (other than search) dispatches SHARED_* actions — so the shared path is dead code.

Is cleanup for this planned in a future PR? If not, this should addressed now.

Niranjan-J-Joshi: > Any chance we can land one real consumer in this PR, even just the search/page/pageSize actions on DeployedServicesTable? Right now none of this is exercised by anything, and there are three things I cannot check without a caller.
> 
> First, the SHARED_* names do not match either table. We have SET_SEARCH on the DA page and DEPLOYED_SERVICES_SET_SEARCH on the services table, so adopting this means renaming about twenty constants plus every dispatch site in two big components. Worth deciding now whether the shared names win or whether the handler should accept the existing ones.

In this PR we're only extracting the shared types, utilities, and actions, per the proposal. The actual consumer migration will be in later PRs — PR 4 (Digital Assistants) and PR 5 (Deployed Services).

As per the earlier review comment , I only wired up SHARED_SET_SEARCH in this PR to demonstrate the shared pattern end-to-end. The remaining dispatches will migrate the same way in PR 4 and PR 5.

Niranjan-J-Joshi:
- Replaced a.type.startsWith("SHARED_") with an explicit Set<SharedTableAction["type"]> lookup in isSharedTableAction.
The typed set is compiler-enforced — TypeScript errors if the set drifts from the union — removing the fragile prefix convention.

- Renamed table/ → Table/ to match the component name casing convention used across all other components in src/components/.

- Fixed SHARED_RESET_COLUMN_VISIBILITY to put its data under payload like every other action does, instead of a custom top-level field

ryanmarc (approved):

PR #1127, opened 2026-07-14
feat: Remove Summarize as part of Digital Assistant

mayuka-c:
- Removing Summarize service as part of Digital Assistant as it should not be part of it and not consumed by any other services.


yussufsh (dismissed):
lgtm

yussufsh (commented):
lgtm

yussufsh (approved):

PR #1138, opened 2026-07-15
Proposal for translation service

Niharika0306:


docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1191 @@
...
+
+### 2.3 Chunking Strategy
+
+Chunks are formed by a greedy paragraph-boundary algorithm:
+
+1. **Split on paragraph boundaries.** The digitized markdown is split on double-newline (`\n\n`) boundaries first. Each resulting block is a candidate unit (heading, paragraph, table, list block).
+2. **Greedy packing.** Starting from an empty chunk, blocks are appended one at a time. When appending the next block would exceed the configured chunk token budget, the current chunk is closed and a new one is started.
+3. **Sentence-level fallback.** If a single block is larger than the entire chunk budget (e.g., an unusually long paragraph), it is split further at sentence boundaries (`. `, `! `, `? `) to produce sub-blocks that fit.
+
+> **Implementation note:** The exact chunk token budget will be determined empirically during development, informed by testing with real documents and observing LLM translation quality and latency. This value should be configurable via an environment variable so it can be tuned without code changes.
```
manalilatkar: Can we introduce that variable here and decide the actual value by experimentation.

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1191 @@
...
+When a user uploads a contract, report, or invoice in German and asks for it in English, they expect to receive a document back — not a markdown file. The cases where markdown-only output is acceptable are narrow:
+
+- A developer integrating the API into their own pipeline.
+- A downstream service that will render or re-format the content itself.
+
+For the majority of real-world use cases, markdown is an intermediate representation that the user cannot use directly:
+
+- **Non-technical users cannot open or read `.md` files.** A translated contract delivered as raw markdown is useless to a lawyer, procurement officer, or business user.
+- **PDF is the universal document interchange format.** Contracts, invoices, regulatory filings, and reports are circulated as PDFs. Returning a translated PDF closes the loop — users can print, sign, share, or archive the output.
+- **DOCX is expected for editable documents.** When the input is a `.docx`, the user likely needs an editable output. A reconstructed PDF at minimum provides a readable result; DOCX reconstruction is noted in Future Enhancements.
```
manalilatkar: Is reconstructing .docx file hard, because of which we are pushing it into future enhancements? Why can't we do it in this release? It makes sense that we would return the translated document in the format that the user provided.

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1191 @@
...
+- **The quality bar is "usable", not "identical".** Users understand that a translated document will not have the exact same fonts, images, or pixel layout as the original. What they need is: all translated text present, correct structure (headings, paragraphs, tables), and a printable/shareable file. This is achievable.
+
+### 3.2 Reconstruction Strategy
+
+The reconstruction follows a two-step pipeline: **markdown → HTML → PDF**, using `weasyprint` — a pure-Python library with no binary dependency beyond system fonts, available on ppc64le / RHEL via `pip`.
```
manalilatkar: Have we tested whether this library re-renders table properly?

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1191 @@
...
+| 410 Gone | Job has `status='failed'`. Error details are on the job resource. |
+| 500 Internal Server Error | Failure reading result file. |
+
+**Sample response (200 — PDF job, with successful PDF reconstruction):**
+
+```json
+{
+    "job_id": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
+    "data": {
+        "translation": "# Q3 2024 Report\n\n## Executive Summary\n\nThe company achieved record revenue...\n\n| Quarter | Revenue | Growth |\n|---|---|---|\n| Q1 | €1.2M | +12% |\n| Q2 | €1.5M | +25% |",
```
manalilatkar, line 677: How is user getting the output PDF file?

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1235 @@
...
+5. Insert a row into `translate_jobs` with `status='accepted'`.
+6. Launch background processing via FastAPI `BackgroundTasks`.
+7. Return `202 Accepted` with `{ "job_id": "..." }`.
+
+**Background worker:**
+
+1. Acquire `job_limiter`; update row → `status='in_progress'`, `metadata.phase='digitizing'` (pdf/docx) or `'translating'` (txt/md).
+2. **PDF/DOCX path:** submit to digitize — `POST http://digitize-url/v1/jobs` with `output_format=md`. Store `digitize_job_id`. Poll `GET /v1/jobs/{digitize_job_id}` every `DIGITIZE_POLL_INTERVAL_SECS` (default 3) until `completed`/`failed` or `DIGITIZE_JOB_TIMEOUT_SECS` (default 300). Fetch text via `GET /v1/documents/{doc_id}/content`; store `digitize_doc_id`.
+   **TXT/MD path:** read the staged file directly (UTF-8 decode).
+3. Tokenize via `/tokenize`; run the hard context-window guard. On breach → `status='failed'`, `error='CONTEXT_LIMIT_EXCEEDED'`, diagnostics in `metadata`.
```
manalilatkar: Shouldn't this step trigger chunking ? Nowhere in the processing is chunking mentioned.

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1235 @@
...
+### 10.4 Prompt Token Overhead
+
+The system prompt above is approximately 90 tokens. The user prompt template (excluding the text) adds ~15 tokens. `PROMPT_OVERHEAD_TOKENS` defaults to `150` — a generous buffer that covers both templates with room to spare, avoiding the need to tokenize the template on every call.
+
+---
+
+## 11. Async Workflow Sequence Diagram
+
+```mermaid
+sequenceDiagram
```
manalilatkar, line 1074: sequence diagram does not show chunking process.

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1235 @@
...
+        Index('idx_translate_jobs_submitted_at_status', 'submitted_at', 'status'),
+    )
+```
+
+#### 12.2.2 `job_metadata` JSONB Shape
+
+Populated progressively as the job advances through phases:
+
+```json
+{
```
manalilatkar, line 1203: no need for any chunking info in job_metadata?

docs/proposals/translation_service_proposal.md
dharaneeshvrd: ~We can't use this library, this uses BSD-3 licence, we generally prefer pkgs with Apache or MIT license with widely used.~
Seems it's permissive, can be used.

docs/proposals/translation_service_proposal.md
dharaneeshvrd: +1

docs/proposals/translation_service_proposal.md
Niharika0306: Done.

dharaneeshvrd (commented):
I dont see info regarding how are we going to make LLM calls.
IMO that is the key part where we need to design it carefully.
There seems to be ways where you can programmatically group chunks and make calls to LLM to reduce the no of calls we are making. But we need to reconstruct the md programmatically, logic would be little complex, but worth doing.

docs/proposals/translation_service_proposal.md
Niharika0306: I understand. Wasn't sure considering the scope initially, hence kept it as a stretch goal. In fact docx reconstruction should be easier compared to pdf reconstruction. 
Have added it in the scope now.

docs/proposals/translation_service_proposal.md
Niharika0306: > Have we tested whether this library re-renders table properly?
Not yet Manali - will update once tested here.

docs/proposals/translation_service_proposal.md
Niharika0306: was planning to give the document url in this way:
"pdf_download_url": "/v1/translate/jobs/a1b2c3d4-e5f6-7890-abcd-ef1234567890/result/pdf"  ( its already in the response structure now)

something similar to how we query for the document in case of digitise job output

docs/proposals/translation_service_proposal.md
Niharika0306: yes, updated it now.

docs/proposals/translation_service_proposal.md
Niharika0306: updated to show it now.

docs/proposals/translation_service_proposal.md
Niharika0306: Updated. 
Added the phase, as well chunking_info.

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1391 @@
...
+| Supported document formats: TXT, PDF/DOCX via digitize           | `.txt`, `.md` files handled natively; `.pdf` and `.docx` delegated to the Digitize Documents service via HTTP                                                             |
+| Supported contents: texts and tables                             | Markdown tables from the digitize service are preserved structurally; the LLM prompt explicitly instructs cell/header translation while keeping `\| pipe \|` syntax intact |
+| Translation strategy for documents                               | Chunk-wise — markdown split on paragraph boundaries, each chunk translated in a separate LLM call, results concatenated (Section 2)                                       |
+| Output format for pdf/docx inputs                                | Translated markdown string + reconstructed output document: PDF (via `weasyprint`) for `.pdf` inputs, DOCX (via `python-docx`) for `.docx` inputs (Section 3)            |
```
dharaneeshvrd: @Niharika0306 @manalilatkar @sats-23 @manju956 
Whenever we introduce these new pkgs, let's document that explicitly with the respective licenses.
And loop in @pravin-dsilva for a quick review.
But both the pkgs seem ok to use.

docs/proposals/translation_service_proposal.md
dharaneeshvrd: Its new behaviour we are introducing here, IMO, providing endpoints `/v1/translate/jobs/{job_id}/result/docx` & `/v1/translate/jobs/{job_id}/result/pdf` is sufficient, no need to give a sample url like this.

manalilatkar:
> I dont see info regarding how are we going to make LLM calls. IMO that is the key part where we need to design it carefully. There seems to be ways where you can programmatically group chunks and make calls to LLM to reduce the no of calls we are making. But we need to reconstruct the md programmatically, logic would be little complex, but worth doing.

@dharaneeshvrd we are chunking in the first place because model's context length is not large enough. If we programmatically group chunks, as you say, won't the input_tokens value exceed the model's context length?

dharaneeshvrd:
If you read the chunking strategy, each heading, paragraph will consider as a separate chunk for reconstruction, at the end will have numerous tiny chunks and if you make LLM call for each chunk, we would flood the LLM with these calls.
Hence looking for optimisation here.

manalilatkar:
> If you read the chunking strategy, each heading, paragraph will consider as a separate chunk for reconstruction, at the end will have numerous tiny chunks and if you make LLM call for each chunk, we would flood the LLM with these calls. Hence looking for optimisation here.

I understand. Sounds good.

Niharika0306:
I have added more details to the chunking strategy and LLM call design. PTAL 
( mostly from section 2.3 to 2.6 , and 9.3)

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1662 @@
...
+
+1. Acquire `job_limiter`; update row → `status='in_progress'`, `metadata.phase='digitizing'` (pdf/docx) or `'translating'` (txt/md).
+2. **PDF/DOCX path:** submit to digitize — `POST http://digitize-url/v1/jobs` with `output_format=md`. Store `digitize_job_id`. Poll `GET /v1/jobs/{digitize_job_id}` every `DIGITIZE_POLL_INTERVAL_SECS` (default 3) until `completed`/`failed` or `DIGITIZE_JOB_TIMEOUT_SECS` (default 300). Fetch text via `GET /v1/documents/{doc_id}/content`; store `digitize_doc_id`.
+   **TXT/MD path:** read the staged file directly (UTF-8 decode).
+3. **Run lingua detection** on the full text (up to 500 chars sampled) to resolve `source_language` if `"auto"` (§10.2).
+4. **Chunk the text** (§2.3): split on paragraph boundaries, greedily pack blocks up to `CHUNK_TOKEN_BUDGET` tokens per chunk, apply sentence-level fallback for oversized blocks, treat tables as atomic units. Each block's token count is obtained via `/tokenize` during packing.
+5. `metadata.phase='translating'`: for each chunk in order:
+   a. Run the hard context-window guard on the chunk (§8). On breach → `status='failed'`, `error='CONTEXT_LIMIT_EXCEEDED'`, diagnostics in `metadata`.
```
manalilatkar: Why not apply the hard context-window guard while chunking the text in step 4, as in after a chunk is created. That way we can redo the chunking if the guard fails. No point in doing it at step 5.

Niharika0306: Guess this section was not updated post the latest chunking and LLM call strategies added. Its updated now.

Niharika0306:
@manalilatkar @dharaneeshvrd , updated the whole document according to the newly approved scope. PTAL

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+
+**One key design choice has been made for document inputs and is described in detail in Section 2:**
+- **Translation strategy:** Chunk-wise translation — the markdown is split into paragraph-boundary chunks and each chunk is translated in an independent LLM call, then reassembled using boundary-aware join logic.
+
+> **Scope note:** PDF and DOCX file inputs, the Digitize Documents service integration, and document reconstruction (PDF/DOCX output) are **not in scope for v1**. They are documented as future enhancements in §17.
+
+The service is a first-class member of the AI-Services platform and follows the same architectural patterns established by the summarize service:
+
+- FastAPI application running as a Python service in a Podman container (ppc64le / RHEL).
+- Semaphore-based concurrency limiting: a 4-slot job admission semaphore for async jobs, and a shared 32-slot semaphore in front of the vLLM inference endpoint.
```
dharaneeshvrd: Why are we limiting 4 slot here?
Digitize follows 4 slot due to the restriction with compute. I dont think we need to follow 4 slot here.

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+|:------------------------------------------------------------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Input: text to be translated                                      | Sync `text` field; async worker reads staged `.txt` / `.md` file as UTF-8                                                                                                |
+| Input: source language (e.g., "German")                          | `source_language` parameter — optional; defaults to `"auto"`. When `"auto"`, the service runs `detect_language()` from `common/lang_utils.py` (lingua) before the LLM call to resolve the actual language; falls back to LLM auto-detect if confidence is below threshold. When provided explicitly, validated against a fixed allowlist. |
+| Input: target language (e.g., "English")                         | `target_language` parameter — validated against a fixed allowlist; `"auto"` is not permitted                                                                              |
```
dharaneeshvrd, line 35: only German & English are allowed for current release.
let's mention explicitly

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+                client=http_client,
+                llm_endpoint=llm_endpoint,
+                messages=build_messages(chunk.text, resolved_source_language, target_language),
+                model=model_name,
+                max_tokens=compute_max_tokens(chunk),   # §8.2
+                temperature=0.0,
+            )
+    return chunk, translation, in_tok, out_tok
+
+tasks = [asyncio.create_task(translate_chunk(chunk)) for chunk in chunks]
```
manalilatkar, line 971: Line no 76 says the following:
` **Contained failure surface.** If one chunk fails, only that chunk is retried. Completed chunks are preserved, making partial recovery straightforward.`
but the implementation here, If any chunk fails (HTTP error, timeout, context guard breach), all sibling tasks are cancelled immediately and the job is marked failed. No partial results are written.
Suggestion: Add per-chunk retry with exponential backoff (2–3 attempts) inside translate_chunk() before propagating the failure. The TranslationChunk dataclass already carries index, so retrying a specific chunk is trivial. Only fail the job after retries are exhausted. This aligns the implementation with the design rationale and is the whole point of chunking vs. single-pass.

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+else:
+    close current chunk → emit as chunk[i]
+    start new chunk with this block
+    running_tokens = block_tokens
+```
+
+When the loop ends, the remaining open chunk is emitted as the final chunk.
+
+**Step 3 — Sentence-level fallback for oversized blocks.**
+If a single block's token count exceeds `CHUNK_TOKEN_BUDGET` on its own (e.g., a very long paragraph with no sub-structure), it cannot be placed as a unit. It is split at sentence boundaries (`.`, `!`, `?` followed by whitespace) and the resulting sentences are packed greedily using the same token-counting logic above.
```
manalilatkar: For German documents, splitting on .  / !  / ?  will incorrectly split on extremely common German abbreviations: z. B. (zum Beispiel / e.g.), Nr. 5, Abs. 2 (Absatz / paragraph), bzw., Dez. 2024, ca. 500, d. h. (das heißt / i.e.), u. a. (unter anderem / among others). These appear frequently in exactly the kind of legal and business German documents this service targets. The result is malformed sub-chunks and broken translations.
Suggestion: Replace the regex heuristic with a proper sentence tokenizer in case of german language.

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+| `target_language` | string | Yes      | Target language name. Must not be `"auto"`.                                                                                         |
+| `job_name`        | string | No       | Optional human-readable label for the job.                                                                                          |
+
+**Validation rules:**
+
+- Exactly one file per request.
+- Extension must be `.txt` or `.md`. Other file types (`.pdf`, `.docx`, `.xlsx`, etc.) are rejected with `415 UNSUPPORTED_FILE_TYPE`.
+- If `source_language` is omitted, it defaults to `"auto"`. If provided, it must be in the allowlist or `"auto"`.
+- `target_language` is required and must be in the allowlist; `"auto"` is not permitted.
+- No size limit enforced at submission — the context-window guard runs during chunking, when the true token count is known.
```
manalilatkar: The context-window guard runs after the file has been staged to disk (/var/cache/translate/staging/{job_id}/) and read into memory as UTF-8 (L502–L503). A malicious or accidental multi-GB text file upload will:
Fill the persistent volume at /var/cache/translate/staging/
Consume worker memory when .read() loads the entire file as a string
Potentially OOM the container before any guard fires
Suggestion: Add a MAX_UPLOAD_SIZE_MB env var (default: 50MB — generous for .txt/.md) and enforce it at the API boundary in the request handler, before staging.

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+If a single block's token count exceeds `CHUNK_TOKEN_BUDGET` on its own (e.g., a very long paragraph with no sub-structure), it cannot be placed as a unit. It is split at sentence boundaries (`.`, `!`, `?` followed by whitespace) and the resulting sentences are packed greedily using the same token-counting logic above.
+
+#### Parameters
+
+| Parameter | What it controls | How it's used |
+|:---|:---|:---|
+| `CHUNK_TOKEN_BUDGET` | Max input tokens per chunk | Packing decision threshold in steps 2 and 3 |
+| `OPENAI_BASE_URL` (shared) | `/tokenize` endpoint | Token count per block; one call per block, result cached |
+
+> The initial value of `CHUNK_TOKEN_BUDGET` will be determined empirically during development by testing with real German/English documents and observing translation quality and latency. A reasonable starting point is half of `MAX_MODEL_LEN` minus `PROMPT_OVERHEAD_TOKENS`, which reserves the other half for the translated output. See §13 for the env var.
```
dharaneeshvrd: German to English is fine.
But you need to consider the English to German word count ratio to give enough headroom for the extra words.

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+4. Stage the file to `/var/cache/translate/staging/{job_id}/`.
+5. Insert a row into `translate_jobs` with `status='accepted'`.
+6. Launch background processing via FastAPI `BackgroundTasks`.
+7. Return `202 Accepted` with `{ "job_id": "..." }`.
+
+**Background worker:**
+
+1. Acquire `job_limiter`; update row → `status='in_progress'`, `metadata.phase='chunking'`.
+2. Read the staged file as UTF-8 text.
+3. **Run lingua detection** on the full text (up to 500 chars sampled) to resolve `source_language` if `"auto"` (§10.2).
```
manalilatkar: Business documents frequently start with English-language content — reference numbers ("Ref: DE-2024-0847"), standardized headers ("CONFIDENTIAL", "Document Version 1.2"), client names, or bilingual letterheads — before the actual German body text begins. Sampling only the first 500 characters will detect English with high confidence, causing the prompt to tell the LLM "Translate the following text from English to English" for a German document, which is worse than letting the LLM auto-detect.
Maybe we should choose samples from multiple positions: beginning, middle, and end of the document — and take the majority/highest-confidence result.

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+
+Supported language values for `source_language` and `target_language` (case-insensitive): `English`, `German`. `source_language` additionally accepts `"auto"` (or may be omitted entirely).
+
+> **Scope note:** `French` and `Italian` are planned additions pending testing within the current release scope.
+
+**Processing logic:**
+
+1. Validate `text` is non-empty (else `400`).
+2. If `source_language` is omitted, default it to `"auto"`. If provided, validate against the allowlist (including `"auto"`); `"auto"` is never valid for `target_language` (else `400 INVALID_LANGUAGE`).
+3. Validate `target_language` is present and in the allowlist (else `400`).
```
manalilatkar, line 362: After both languages are validated against the allowlist, add a check: if source_language != "auto" and source_language.lower() == target_language.lower(), return 400 with a clear error like "source_language and target_language must be different". Skip the check when source is "auto" since the resolved language isn't known yet.
Currently, nothing in the validation logic prevents source_language: "English", target_language: "English". The LLM will process it (effectively paraphrasing the input), wasting compute and returning confusing results to the user.

docs/proposals/translation_service_proposal.md
```diff
@@ -0,0 +1,1430 @@
...
+4. Check the global vLLM semaphore; if all slots are occupied, return `429`.
+4. Tokenize the input via the vLLM `/tokenize` API — exact count.
```
manalilatkar: nit: steps are misnumbered.

manalilatkar (commented):
minor comments, rest LGTM.

docs/proposals/translation_service_proposal.md
dharaneeshvrd: IMO we can allow till the LLM's Max concurrent requests, both sync and async calls. In async job, you can make the call serially, so you would be still within your one slot allotted.

docs/proposals/translation_service_proposal.md
Niharika0306: Updated it.

docs/proposals/translation_service_proposal.md
Niharika0306: Was in the picture, but maybe somehow missed between iterations. Added it now.

docs/proposals/translation_service_proposal.md
Niharika0306: As discussed, this is handled by the retries made at the LLM calls.

docs/proposals/translation_service_proposal.md
Niharika0306: Done. using the existing sentence splitters from the lang utils in code for this purpose now.

docs/proposals/translation_service_proposal.md
Niharika0306: Done.

docs/proposals/translation_service_proposal.md
Niharika0306: Done.

docs/proposals/translation_service_proposal.md
Niharika0306: Handled.

docs/proposals/translation_service_proposal.md
Niharika0306: Increased it to 8 workers now. Making serial llm calls for the chunks seemed to be a lot of latency and under utilizing the capacity. 
so its now 8 async job slots, with 4 as chunk parallelism.

docs/proposals/translation_service_proposal.md
Niharika0306: Added it to be 40% of the max_model_len , to give headroom in case of German tokens.

docs/proposals/translation_service_proposal.md
dharaneeshvrd: Ok let's get started with this number for now.

docs/proposals/translation_service_proposal.md
dharaneeshvrd: Let's use the sentence splitter library instead of bringing a regex for these purposes in future too.

manalilatkar (approved):

PR #1142, opened 2026-07-15
Add global lock for digitize operation

dharaneeshvrd:
This document proposes adding a shared, Postgres-backed conversion queue **inside the existing
`digitize` service** so that `convert_doc()` runs under a globally capped, weighted semaphore
that is durable across restarts.


docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+Gate 2 passes if at least one free slot exists (`queued < QUOTA`). Admitted jobs with N > free slots get `min(N, free_slots)` tasks as `queued`, the rest as `pending`.
+
+| Scenario | Running weight | Free | New job weight | Ing queued | Dig queued | Decision |
+|---|---|---|---|---|---|---|
+| Ing A: 1 normal running. New dig: 1 normal | 1 | 3 | 1 | — | 0 | ✅ Accept (1 queued) |
+| Ing A: 1 normal running. New ing: 1L+1N | 1 | 3 | 3 | 0 | — | ✅ Accept (2 queued) |
+| Ing A: 1 large running. New ing: 2N | 2 | 2 | 2 | 0 | — | ✅ Accept (2 queued) |
+| Ing A: 2 normal running. New ing: 1L+1N | 2 | 2 | 3 | 0 | — | ❌ Reject (Gate 1: weight 3 > free 2) |
+| Ing A: 30 files (4 running, 6 queued). New dig: 1 normal | 4 | 0 | 1 | — | 0 | ❌ Reject (Gate 1: free=0) |
+| Ing A: 30 files (3 running, 7 queued). New dig: 1 normal | 3 | 1 | 1 | — | 0 | ✅ Accept (Gate 2: dig_queued 0 < 5) |
```
sats-23: Is it better to have a greedy algo fit for these cases?
Say in the remaining 7 files, is it better to sort it to find a Normal ingestion to fill the in_progress. 
Thinking from PoV of why to keep capacity unused?

sats-23: Maybe we can convert this queue into a weighted PriorityQueue data structure that orders the files by lesser pages = higher index along with weight coming from job age and it polls accordingly

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+| Ing A: 30 files (3 running, 7 queued). New dig: 6th waiting dig | 3 | 1 | 1 | — | 5 | ❌ Reject (Gate 2: dig_queued 5 >= 5) |
+| Ing A: 30 files (3 running, 7 queued). New ing: 1 normal | 3 | 1 | 1 | 10 | — | ❌ Reject (Gate 2: ing_queued 10 >= 10) |
+| Ing A: 30 files (3 running, 7 queued). New ing B: 1 normal | 3 | 1 | 1 | 7 | — | ✅ Accept (Gate 2: ing_queued 7 < 10) |
```
sats-23: What is the difference between this and prev one?
10 ingestion queue count is inclusive of the file running or exclusive?

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: Thanks for catching, the table did not get populated the way I intended, I ll repopulate it properly.

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+                    CHECK (operation IN ('ingestion','digitization')),
+    -- input
+    cached_file     TEXT          NOT NULL,        -- absolute path written at enqueue time
+    output_format   VARCHAR(10)   NOT NULL
+                    CHECK (output_format IN ('json','md','txt')),
+    page_count      INTEGER,                       -- filled in during enqueue
+    is_large        BOOLEAN       NOT NULL DEFAULT FALSE,
+    -- lifecycle
+    status          VARCHAR(50)   NOT NULL
+                    CHECK (status IN ('pending','queued','running','completed','failed')),
```
manju956, line 105: this means we will have to add additional state `queued` to existing doc status.

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+    result_path     TEXT,                          -- written on completion
+    error           TEXT,
+    queued_at       TIMESTAMPTZ   NOT NULL DEFAULT now(),
+    started_at      TIMESTAMPTZ,
+    completed_at    TIMESTAMPTZ,
+    updated_at      TIMESTAMPTZ   NOT NULL DEFAULT now()
+);
+
+CREATE INDEX idx_ct_status_op_queued ON conversion_tasks (status, operation, queued_at);
+CREATE INDEX idx_ct_job_id           ON conversion_tasks (job_id);
```
manju956: Do we need index on job_id as well? the index on doc_id alone will not suffice?

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+|---|---|---|
+| `ingestion` | **10** queued slots | A job can have many files; 10 gives enough buffer for large batches |
+| `digitization` | **5** queued slots | Each job is exactly 1 file; 5 concurrent waiting jobs is sufficient |
+
+```python
+# After the semaphore headroom check passes:
+ing_queued  = COUNT WHERE status='queued' AND operation='ingestion'
+dig_queued  = COUNT WHERE status='queued' AND operation='digitization'
+
+ING_QUOTA = 10   # settings.digitize.ingestion_queue_quota
```
manju956, line 157: Why do we have higher quita for ingestion as single request can carry multiple docs whereas digitize job can carry single file?

docs/proposals/docling_conversion_queue.md
manju956: Yeah, same doubt.

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+Two new consumers are planned:
+
+- **Data Source Connectors** — sync remote data sources into the vector DB by running the
+  ingestion pipeline. They have their own API endpoints, their own DB table, and their own
+  concurrency limits, but they call the same `processing/orchestrator.py` → `convert_doc()`
+  path that `POST /v1/jobs` (ingestion) uses today.
+- **Extract & Tag service** — runs the full digitization pipeline (including Docling
+  conversion) so it can attach a `doc_id` to a digitize document record. It therefore calls
+  the same `pipeline/digitize.py` → `convert_doc()` path that `POST /v1/jobs` (digitization)
+  uses today.
```
manalilatkar, line 26: Translate service can also be added here.

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+# Computed at admission time, inside a SELECT … FOR UPDATE transaction
+running_weight = SUM(weight FOR task IN conversion_tasks WHERE status = 'running')
+free_weight    = SEMAPHORE_CAPACITY - running_weight   # 4 - currently running weight
+new_job_weight = SUM(2 if is_large(f) else 1 FOR f in submitted_files)
+
+if new_job_weight > free_weight:
+    raise HTTP 429  # no free semaphore capacity for this job right now
+```
+
+This uses *running* weight only. A 30-file ingestion whose total weight fits the current
```
manalilatkar: How will this work? A 30-file ingestion job's total weight might be 30 (if all files are not large). Let's say nothing is running right now. Then 
```
free_weight = 4
new_job_weight = 30
if new_job_weight > free_weight:
    raise HTTP 429 # job is being rejected because 30>4
```

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+free capacity is accepted; the dispatcher drains the tasks in round-robin order.
+
+**Gate 2 — Per-operation queue quota (backstop)**
```
manalilatkar: If my understanding of Gate 2 and Gate 1 is correct, I feel there is no need for Gate 1. Its like the bouncer turning incoming requests away because the machine room is busy even though the waiting room is empty. 
The point of a durable queue is that work can wait, yet Gate 1 reintroduces reject-when-busy at admission — the same semantics as today's concurrency_manager.is_locked() check at jobs.py:185 that this design retires. Per the decision table (line 310), all-4-units-running rejects every job even with 15 queue slots empty, making the queue nearly vestigial. Suggest dropping Gate 1 and letting the Gate 2 quotas be admission control; running weight is the dispatcher's concern.

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+
+| Event | Input cache | Output file |
+|---|---|---|
+| Task completes | Deleted immediately after export | Kept until TTL (default 1 h) |
+| Task fails | Deleted | N/A |
+| Startup: `running` → `failed` | `chunks/` dir deleted; input deleted | N/A |
+| Startup: `queued`, file present | Kept (job will run) | N/A |
+| Startup: `queued`, file missing | N/A | N/A |
+
+A background `asyncio.create_task` TTL reaper deletes `result_path` and sets
```
manalilatkar: A TTL on conversion outputs conflicts with the existing contract: digitized documents are "permanent content files" (storage.py:31), served indefinitely via GET /v1/documents/{doc_id}/content and deleted only by user action (DELETE /v1/documents/{id}, api/v1/documents.py:148–210). External consumers depend on this — the translate service and the extract service store digitize_doc_id as a durable pointer and explicitly defer lifecycle to the digitize DELETE endpoint. Once outputs land in digitized_docs_dir , the reaper must not touch them. Scope the TTL to genuinely temporary artifacts only, or drop it.

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+    output_file, conversion_time = future.result()
+```
+
+Proposed:
+```python
+task_id = enqueue_conversion(
+    job_id=job_id, doc_id=doc_id,
+    file_path=file_path, output_format=output_format,
+)
+output_file, conversion_time = await poll_until_complete(task_id)
```
manalilatkar: This snippet seems unrunnable in the current scenario: digitize() is a synchronous function executed in a worker thread via asyncio.to_thread (jobs.py:46) — there's no event loop to await in. Either poll_until_complete must be a sync blocking helper (time.sleep poll loop — acceptable inside the thread), or the pipelines must be converted to async, which is a much larger change than this proposal scopes. Please specify the sync variant explicitly.

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+Currently submits `convert_document()` into a `ProcessPoolExecutor`.
+Under this proposal it enqueues each file and polls, replacing the
+`conversion_futures` dict with `task_id → path` tracking.
+The rest of the pipeline (process → chunk → index) is unchanged.
+
+#### Data Source Connectors
+
+Data Source Connectors drive the **ingestion pipeline** — they sync remote data sources into
+the vector DB via `processing/orchestrator.py`, the same path used by `POST /v1/jobs`
+(ingestion). They have their own entry-point, their own API surface, and their own
```
manalilatkar, line 914: Will the 2 gates still be applicable to data connectors workload?

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+
+---
+
+### 9. File change summary
+
+| File | Change |
+|---|---|
+| `db/models.py` | Add `ConversionTask` ORM model with `operation` column |
+| `db/manager.py` | Add `ConversionTask` CRUD + `_try_claim(operation, available)` + `get_running_weight()` + `get_queued_count_by_op()` |
+| `settings.py` | Add `conversion_queue_max`, `conversion_poll_interval`, `conversion_result_ttl` to `DigitizeConfig` |
```
manalilatkar: In line 684, we are removing `conversion_queue_max` and then in line no 694, we are adding it again. Why?

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+
+### 12. Risks and mitigations
+
+| Risk | Mitigation |
+|---|---|
+| Dispatcher claims the same task twice on concurrent invocations | `SELECT … FOR UPDATE SKIP LOCKED` is atomic; only one claim succeeds |
+| Semaphore drifts if `_run_conversion` crashes before `release` | `finally: await semaphore.release(weight)` — release always runs |
+| Race between admission check and INSERT (two jobs admitted simultaneously) | Both checks happen inside one `SELECT … FOR UPDATE` transaction; only one wins the race |
+| Cache volume fills with large PDFs | Emit a warning log when cache dir free space < threshold; TTL reaper prevents unbounded growth |
+| Caller polls forever on stuck task | Dispatcher sets a per-task wall-clock timeout (e.g. 30 min); marks timed-out tasks `failed` |
```
manalilatkar: This timeout is being mentioned for the first time unless I missed it somewhere — not in the design, settings , or implementation sequence. Maybe promote it to a setting and specify the mechanism: who detects it, and how is semaphore weight recovered?
Also, you can't just cancel a stuck worker process (it keeps burning a CPU even after you give up on it), and slow conversions legitimately retry 3 times with growing pauses, so an honest big document can take a while. Timeout has to account for this.

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+| Digitization task queues behind a large ingestion job's overflow | Queue is FIFO by `queued_at`; tasks from all jobs interleave naturally — a digitization task submitted after an ingestion job's overflow tasks will wait its fair turn |
+| Two concurrent ingestion jobs interleaving chunks into the vector DB | Each doc has a unique `doc_id` keying its chunks; concurrent ingestion jobs write independent document sets with no overlap |
+| `page_count` unavailable before enqueue (e.g. DOCX) | `get_document_page_count()` returns 0 for DOCX; treat 0 as `is_large=False` (weight 1) |
```
manalilatkar, line 1269: When does this scenario happen? Does the function estimate_docx_page_count on line no 293 of digitize/parsing/docx.py not work properly sometimes?

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1041 @@
...
+
+Gate 2 only requires **at least one free slot** in the queue — the caller is not rejected
+because N exceeds the remaining quota. Tasks that fit fill as `queued`; the rest are
+`pending`. The dispatcher promotes `pending → queued` as running tasks complete and quota
+headroom opens up. This lets a 30-file ingestion batch be accepted in one request without
+atomically reserving all 30 slots.
+
+The existing `has_active_jobs()` hard block at [`jobs.py:174`](services/digitize/api/v1/jobs.py:174)
+and the `ConcurrencyManager` semaphores at lines 185–199 are both removed. The
+`_run_ingest` and `_run_digitize` background tasks are removed — the dispatcher drives
```
manalilatkar, line 285: This Contradicts Section 7 (line 615 onward, enqueue + poll from within the pipelines) — and the code shows the pipelines can't be dropped: _run_digitize/_run_ingest invoke the entire sync pipelines via asyncio.to_thread(digitize, …) / to_thread(ingest, …) (jobs.py:46, 74), which own staging cleanup, StatusManager doc/job updates (pipeline/digitize.py:52–82), language detection, chunking, and indexing. If they're removed, nothing runs any of that, or are we planning to directly move this functionality?

manalilatkar (changes requested):

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: I have done major refactoring on this design.
But I did not use the weighted PriorityQueue to prevent large files not getting the turn on a connector like situation, current design follows the simple FIFO, let's get started with that for now, we can revisit this logic later if needed.

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: Removed this dup entry

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: yes

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: Agree, not required

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: quota here is for the docs behind these jobs, since ingestion allows multiple file per job wanted to keep a higher queue limit than digitize where only one file is allowed per job hence lower queue value.

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: I have replaced the weight based checking to queue checking, ptal

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: Yes agree on this, removed the running weight from the admission logic, now its happening only based on queue size.

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: Once the data connectors creates the job, it's upto the job how it handles, so the validations would apply for the job, but these gate logic is replaced with new queue only logic. ptal

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: Redone this part where process_documents would be depending the new task mgmt logic to update the task status wrt to the document available on the job.

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: Removed this. ptal

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: We can't have a time out for the conversion, since we are not putting a limit on the document page count, larger the page count it would take longer time to convert it. Removed now.

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: I have added the sample refactoring of process_documents where its dependent on the task status to invoke the next set of ingestion pipeline, ptal at the new logic.

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: Agree, this is not needed, added by bob, removed it

docs/proposals/docling_conversion_queue.md
dharaneeshvrd: Yes that could be the reason, since we can only do estimation on docx files. 
But I m not exactly sure on what scenario `estimate_docx_page_count ` method won't be able to count the pages.

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1269 @@
...
+                    CHECK (status IN ('pending','queued','running','completed','failed')),
+    result_path     TEXT,                          -- written on completion
+    error           TEXT,
+    queued_at       TIMESTAMPTZ   NOT NULL DEFAULT now(),
+    started_at      TIMESTAMPTZ,
+    completed_at    TIMESTAMPTZ,
+    updated_at      TIMESTAMPTZ   NOT NULL DEFAULT now()
+);
+
+CREATE INDEX idx_ct_status_op_queued ON conversion_tasks (status, operation, queued_at);
```
manalilatkar, line 114: maybe add one more index idx_ct_job_id so that the get_conversion_task_by_job_id function in Section 7 becomes easier, otherwise it's a sequential scan per tick per job.

`task = db_manager.get_conversion_task_by_job_id(job_id)`

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1269 @@
...
+        # convert_doc is CPU-bound Docling work that does not release the GIL.
+        # asyncio.to_thread would run it on a thread and cause GIL contention
+        # between concurrent conversions — they would not actually run in
+        # parallel.  run_in_executor with a ProcessPoolExecutor spawns a child
+        # process for each conversion, giving true CPU parallelism exactly as
+        # the existing pipeline/digitize.py and processing/orchestrator.py do.
+        # convert_doc must be a top-level importable function (picklable) —
+        # it already satisfies this requirement.
+        loop = asyncio.get_running_loop()
+        doc: DoclingDocument = await loop.run_in_executor(
```
manalilatkar: submitting bare convert_doc pickles the full DoclingDocument back across the process boundary — for a 1,000-page doc that's a massive IPC transfer into the API server's memory. Both existing wrappers deliberately export inside the child and return (path, time) (converter.py:181, 210). Submit an export-in-child wrapper instead of sending the whole document back.
context: A DoclingDocument for a 1,000-page PDF is enormous — it contains the parsed structure of every page, every table, every text block, all the metadata. We're potentially talking hundreds of megabytes.
That entire object gets serialized to bytes in the child, written through the IPC pipe, then deserialized and reconstructed in the parent process's memory.

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1269 @@
...
+    for task in running_tasks:
+        _safe_rmtree(Path(task.cached_file).parent / "chunks")
+        db_manager.update_task_status(
+            task.task_id, "failed",
+            error="Service restarted during conversion"
+        )
+        logger.warning(f"Recovery: task {task.task_id} running→failed")
+
+    # 2. queued — verify cached file
+    queued_tasks = db_manager.get_conversion_tasks(status="queued")
```
manalilatkar, line 695: what about pending tasks?

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1269 @@
...
+
+---
+
+### 12. Risks and mitigations
+
+| Risk | Mitigation |
+|---|---|
+| Dispatcher claims the same task twice on concurrent invocations | `SELECT … FOR UPDATE SKIP LOCKED` is atomic; only one claim succeeds |
+| Semaphore drifts if `_run_conversion` crashes before `release` | `finally: await semaphore.release(weight)` — release always runs |
+| Race between admission check and INSERT (two jobs admitted simultaneously) | Both checks happen inside one `SELECT … FOR UPDATE` transaction; only one wins the race |
```
manalilatkar: Correct `both checks` to check as one gate has been removed.

docs/proposals/docling_conversion_queue.md
```diff
@@ -0,0 +1,1269 @@
...
+| Two concurrent ingestion jobs interleaving chunks into the vector DB | Each doc has a unique `doc_id` keying its chunks; concurrent ingestion jobs write independent document sets with no overlap |
+| `page_count` unavailable before enqueue (e.g. DOCX) | `get_document_page_count()` returns 0 for DOCX; treat 0 as `is_large=False` (weight 1) |
+
+---
+
+## Implementation Sequence
+
+1. **ORM + migration** — add `ConversionTask` to `db/models.py`; `Base.metadata.create_all` creates the table on next startup.
+2. **`WeightedSemaphore`** — implement and unit-test in isolation (`workers/conversion_semaphore.py`).
+3. **`db/manager.py` CRUD** — `create_task`, `claim_queued_tasks`, `update_task_status`, `get_conversion_tasks`, `get_running_weight`, `get_queued_count`.
```
manalilatkar: Remove `get_running_weight` from this line.

manalilatkar (commented):
minor comments, otherwise changes look good.

manalilatkar (commented):
minor comments, otherwise changes look good.

docs/proposals/docling_conversion_queue.md
manju956: makes sense

dharaneeshvrd:
@manalilatkar Addressed all of your comments, ptal

manju956 (approved):
lgtm

manalilatkar:
@dharaneeshvrd all the changes LGTM. Rebase the code and I'll approve.

Niharika0306:
lgtm as well.

manalilatkar (approved):

PR #1144, opened 2026-07-15
adding bootstrap failures testcases

vijay-vin:


vijay-vin:
I have added the comments as how to run these tests and also the my test run outputs to the Jira task # aiservices-1365

ai-services/tests/e2e/cli/output.go
```diff
@@ -565,3 +565,259 @@ func ValidateCatalogUninstallOutput(output string) error {
...
+//   - "Error logging in to"
+func ValidateRegistryLoginFailureOutput(output string) error {
+	knownPatterns := []string{
+		"invalid username/password",
+		"unauthorized",
+		"authentication required",
+		"failed with status",
+		"Error logging in to",
+		"error logging in",
+		"401",
```
manju956: these error codes, do we see these in error msg patterns as well?

ai-services/tests/e2e/cli/output.go
```diff
@@ -565,3 +565,259 @@ func ValidateCatalogUninstallOutput(output string) error {
...
+		// ICR (IBM Container Registry) returns 400 Bad Request for invalid credentials
+		// rather than the more common 401 Unauthorized.
+		"400",
+		"bad request",
+		// Generic Podman/Skopeo auth failure prefix
+		"authenticating creds",
+		"requesting bearer token",
+	}
+
+	lowerOutput := strings.ToLower(output)
```
manju956: this block is common across multiple validates, can we create common method to do this?

ai-services/tests/e2e/bootstrap_failure_test.go
```diff
@@ -0,0 +1,519 @@
...
+								),
+							)
+						}
+
+						registryURL, _, _ := bootstrap.GetPodManCreds()
+						if registryURL == "" {
+							// Fall back to a well-known public registry so the test
+							// can still exercise authentication failure handling even
+							// when REGISTRY_URL is not exported.
+							registryURL = "docker.io"
```
manju956: our default registry is icr rt?

ai-services/tests/e2e/bootstrap_failure_test.go
```diff
@@ -0,0 +1,519 @@
...
+						// will still detect a failure.
+						serverURL, _, _ := bootstrap.GetCatalogCreds()
+						if serverURL == "" {
+							// Try to discover a running catalog instance.
+							infoOut, infoErr := cli.CatalogInfo(ctx, cfg, appRuntime)
+							if infoErr == nil {
+								serverURL = cli.ExtractCatalogBackendURL(infoOut)
+							}
+						}
+						if serverURL == "" {
```
manju956, line 204: I dont think we can default to any port if catalog info is not available, rather throw error and fail the test

ai-services/tests/e2e/bootstrap_failure_test.go
```diff
@@ -0,0 +1,519 @@
...
+									"[FAILURE-TEST][Spyre] Failed to remove temp chroot dir %s: %v",
+									emptyChrootDir,
+									removeErr,
+								)
+							}
+						}()
+
+						// Save and override GHW_CHROOT.  The CLI subprocess inherits
+						// the test process environment, so setting this here causes
+						// ghw.PCI() inside the CLI to read from the empty dir.
```
manju956: we can simply use `lspci` command to detect the cards and number of cards instead of ghw module.

ai-services/tests/e2e/cli/output.go
vijay-vin: Will modify the test code accordingly.

ai-services/tests/e2e/cli/output.go
vijay-vin: sure. Will create method for it.

ai-services/tests/e2e/bootstrap_failure_test.go
vijay-vin: agree. Will change it.

ai-services/tests/e2e/bootstrap_failure_test.go
vijay-vin: Agree with you, but my suggestion here would be to skip the test rather than fail. Failing will increase our test fail count when this actually is a environment/setup issue. Let me know if I should skip the test with proper message.

ai-services/tests/e2e/bootstrap_failure_test.go
vijay-vin: I agree with you on using lspci. 
I think the better approach here would be to use lspci to check whether Spyre cards are actually present on the machine. If there are no cards, run the test directly without GHW_CHROOT — the validator will naturally emit the "not attached" error. If there are cards, we still need GHW_CHROOT to mask them from the validator so the test remains deterministic. This way lspci drives the decision and GHW_CHROOT is only used as a fallback when real hardware is present.
I have modified the test code accordingly, and it seems to be working.

Let me know if you agree and I can push the new changes.

ai-services/tests/e2e/bootstrap_failure_test.go
manju956: If catalog itself not present at first place, is it wise to continue testing other tests? if yes, mark it failed and continue executing remaining ones.

ai-services/tests/e2e/bootstrap_failure_test.go
manju956: if there cards are not detected by `lspci` that means its CPU kind of setup. so, its good marke the test as failed. pls go ahead and update the code.

vijay-vin: done

ai-services/tests/e2e/bootstrap_failure_test.go
vijay-vin: done

ai-services/tests/e2e/cli/output.go
vijay-vin: done

ai-services/tests/e2e/bootstrap_failure_test.go
vijay-vin: done

ai-services/tests/e2e/cli/output.go
vijay-vin: done

manju956 (approved):
lgtm

Abhishek-Kushwaha1 (approved):

PR #1146, opened 2026-07-16
Schema registry

manalilatkar:
This PR contains:
1. DB init code
2. POST /v1/schemas
3. GET /v1/schemas (paginated, name filter)
4. GET /v1/schemas/{id}{}
5. DELETE /v1/schemas/{id} (RESTRICT on referencing jobs)
6. bulk DELETE /v1/schemas?confirm=true.


dharaneeshvrd (commented):
@manalilatkar 
Can we please follow the dir structure added by Manju recently for digitize service?

services/extract/app.py
```diff
@@ -0,0 +1,546 @@
...
+        schema_tokens=schema_tokens,
+        examples_tokens=examples_tokens,
+        custom_prompt_tokens=custom_prompt_tokens,
+        description=body.description,
+        examples=examples_raw,
+        custom_prompt=body.custom_prompt,
+    )
+    if row is None:
+        # A second concurrent request might have inserted the same name.
+        if db_repo.schema_name_exists(body.name):
```
dharaneeshvrd: I feel this check should happen before doing any validation, if we are not allowing dup name for schemas, why not keep it as primary key instead of an uid

services/extract/settings.py
```diff
@@ -0,0 +1,152 @@
...
+        description="Minimum reserved output tokens (floor for small schemas)",
+    )
+
+    max_output_tokens: int = Field(
+        default=4096,
+        ge=1,
+        description="Maximum reserved output tokens (ceiling for large schemas)",
+    )
+
+    prompt_overhead_tokens: int = Field(
```
dharaneeshvrd: why not calculate this dynamically during startup?

services/extract/settings.py
```diff
@@ -0,0 +1,152 @@
...
+        default=150,
+        ge=0,
+        description="Estimated token budget for the fixed system-prompt scaffold",
+    )
+
+    # vLLM
+    guided_decoding_enabled: bool = Field(
```
dharaneeshvrd, line 109: I am concerned about inference engines other than vllm like watsonx on whether it would support this
maybe we need to decide this based on the inference type if watsonx is not supported

services/extract/settings.py
```diff
@@ -0,0 +1,152 @@
...
+
+    extraction_temperature: float = Field(
+        default=0.0,
+        ge=0.0,
+        le=2.0,
+        description="Temperature for extraction generation (deterministic by default)",
+    )
+
+    # Concurrency
+    max_concurrent_jobs: int = Field(
```
dharaneeshvrd, line 122: why are we limiting to only 4?
per job won't we be making only one LLM call? if yes, we can have more

services/extract/settings.py
```diff
@@ -0,0 +1,152 @@
...
+        default=4,
+        ge=1,
+        description="Maximum number of async extraction jobs running in parallel",
+    )
+
+    # Digitize service integration
```
dharaneeshvrd: all digitize settings need to be removed

dharaneeshvrd (commented):
Remaining looks good to me mostly.

services/extract/settings.py
manalilatkar: I will create a task to test with Watsonx as well and this variable can be controlled accordingly. WDYT?

services/extract/settings.py
manalilatkar: @dharaneeshvrd per job we will be making 2 LLM calls, possibly 3 (one for extraction, one for validation and 1 for validation retry in case the first validation fails). Do you still want to increase this?

services/extract/settings.py
manalilatkar: yes, will do.

services/extract/app.py
manalilatkar: @dharaneeshvrd if you see line no 224, the first check happens there, in the beginning. This is the secod check as the comment mentions, in case a second concurrent request might have inserted the same name by now.

services/extract/settings.py
manalilatkar: implemented.

services/extract/app.py
dharaneeshvrd: can you please point to the new location since its moved now?

services/extract/settings.py
dharaneeshvrd: yes please create a JIRA task and post it here

services/extract/settings.py
dharaneeshvrd: yeah that is still a known number, and it would happen in serial manner, def we would be able to accommodate more. IMO we can keep the concurrency same as max_batch_size of vLLM

services/extract/app.py
manalilatkar: services/extract/api/v1/schema.py line no 81 and 127.

services/extract/api/v1/schema.py
```diff
@@ -0,0 +1,348 @@
...
+
+    # --- JSON Schema structural validation (against the normalized form) ---
+    validate_json_schema_structure(normalized)
+
+    # --- Validate example outputs against normalized schema ---
+    examples_raw = [ex.model_dump() for ex in body.examples] if body.examples else None
+    validate_examples(examples_raw, normalized)
+
+    # --- Conflict check (name uniqueness) ---
+    if db_repo.schema_name_exists(body.name):
```
dharaneeshvrd: Still this can be before any of the validation or normalization

services/extract/api/v1/schema.py
```diff
@@ -0,0 +1,348 @@
...
+        schema_tokens=schema_tokens,
+        examples_tokens=examples_tokens,
+        custom_prompt_tokens=custom_prompt_tokens,
+        description=body.description,
+        examples=examples_raw,
+        custom_prompt=body.custom_prompt,
+    )
+    if row is None:
+        # A second concurrent request might have inserted the same name.
+        if db_repo.schema_name_exists(body.name):
```
dharaneeshvrd: We are not following this convention in any of our APIs. Please see whether we really need this based on its possibility

dharaneeshvrd: I meant checking for concurrent requests for same details and the race between them

services/extract/api/v1/schema.py
manalilatkar: agree, I'll move this to the top.

services/extract/api/v1/schema.py
manalilatkar: @dharaneeshvrd , between the 2 checks, we are doing the following:
1. normalize schema
2. validate json schema
3. validate examples
4. calculate token counts for custom prompt and schema and examples
5. check schema share in context.

All these things take time and that's why this approach is used. Bob took this decision on purpose. 
All the other post requests mostly include basic validation and staging a file (this takes most time). Also name is important for schema unlike jobs where name is optional. 
If you still feel this is not necessary, I can remove the first check.

dharaneeshvrd: Except 4th step remaining should happen in microseconds. And from our past experience token count API returns response in no time(wrt to the ingestion pipeline where we call the tokenize for each and every line of the doc). Like you stated, if there are staging involved for multiple files, the request may take longer. IMO we can keep only the first check and reject earlier if the schema name already exists.
Btw did you get a chance to measure the time taken for this?

manalilatkar: No, I didn't get a chance to measure the time taken. Will remove the second check.

services/extract/settings.py
manalilatkar: you mean 32?

dharaneeshvrd: yes

manalilatkar: done

services/extract/api/v1/schema.py
manalilatkar: done

services/extract/api/v1/schema.py
manalilatkar: Removed the second check

dharaneeshvrd (approved):

PR #1148, opened 2026-07-16
docs: proposal document for remote podman setup

yussufsh:
Introduces a design proposal for enabling a single AI Services Catalog installation to act as a control plane that manages AI workload deployments across multiple LPARs.

Currently, every deployment is executed locally. The Catalog API Server can only drive the Podman socket on the same LPAR it runs on, requiring a separate Catalog instance per LPAR with no unified management. This proposal addresses that by introducing a lightweight agent daemon on each Worker LPAR that connects back to the control plane over a persistent bidirectional gRPC stream, allowing the control plane to dispatch Podman runtime commands to any registered worker without changing the existing runtime.Runtime interface or any of its callers.

Routing external traffic to individual worker nodes will be handled by Caddy on each worker LPAR.

Scope note: This proposal covers Podman container lifecycle operations only. It does not address a common LLM gateway (e.g. LiteLLM).

### DIFF
--- a/docs/proposals/catalog/remote-worker-agent-proposal.md
+++ b/docs/proposals/catalog/remote-worker-agent-proposal.md
@@ -1,0 +1,540 @@
+# Remote Worker Agent Proposal
+
+## Table of Contents
+
+1. [Executive Summary](#1-executive-summary)
+2. [Background and Motivation](#2-background-and-motivation)
+   - 2.1 [Current State](#21-current-state)
+   - 2.2 [Problem Statement](#22-problem-statement)
+   - 2.3 [Goals](#23-goals)
+   - 2.4 [Non-Goals](#24-non-goals)
+3. [System Topology](#3-system-topology)
+4. [Architecture Overview](#4-architecture-overview)
+   - 4.1 [Key Components](#41-key-components)
+   - 4.2 [Component Interaction Diagram](#42-component-interaction-diagram)
+5. [gRPC Protocol](#5-grpc-protocol)
+   - 5.1 [Service Definition](#51-service-definition)
+   - 5.2 [Command Types](#52-command-types)
+6. [Bootstrap and Registration Flow](#6-bootstrap-and-registration-flow)
+7. [Deployment Execution Flow](#7-deployment-execution-flow)
+8. [Agent Registry State Machine](#8-agent-registry-state-machine)
+9. [Database Schema](#9-database-schema)
+10. [New CLI Commands](#10-new-cli-commands)
+    - 10.1 [Control Plane Commands](#101-control-plane-commands)
+    - 10.2 [Worker Agent Commands](#102-worker-agent-commands)
+11. [Configuration Files](#11-configuration-files)
+12. [Code Structure](#12-code-structure)
+13. [Before vs After Comparison](#13-before-vs-after-comparison)
+14. [Security Design](#14-security-design)
+15. [Future Work](#15-future-work)
+
+---
+
+## 1. Executive Summary
+
+This proposal describes the design and implementation of a **Remote Worker Agent** system for the AI Services Catalog. It enables the Catalog API Server (running on a Control Plane LPAR) to dispatch Podman runtime commands over a persistent bidirectional gRPC stream to one or more **Worker Agent daemons** running on separate LPARs (e.g. Power10 nodes with Spyre AI accelerators).
+
+The existing `runtime.Runtime` interface and all its callers (`PodmanDeployer`, `SyncService`, `DeletionService`) are **unchanged**. The remote execution path is a new implementation of the same interface, selected at request time by specifying `runtime=remote`.
+
+---
+
+## 2. Background and Motivation
+
+### 2.1 Current State
+
+The AI Services Catalog deploys and manages AI service containers (RAG pipelines, chat, summarization, etc.) by invoking Podman commands **locally** — the Catalog API Server and the Podman socket must reside on the same LPAR.
+
+```
+┌───────────────────────────────────────────────────────┐
+│  Single LPAR                                          │
+│                                                       │
+│  Catalog API Server  ──────►  Podman  ──►  Container  │
+└───────────────────────────────────────────────────────┘
+```
+
+### 2.2 Problem Statement
+
+There is currently no way to use a single AI Services Catalog installation as a control plane that manages deployments across multiple LPARs. Every deployment operation is executed locally — the Catalog API Server can only drive the Podman socket on the same LPAR it is running on.
+
+This means that to deploy AI workloads on a second or third LPAR, an operator must install and maintain a completely independent Catalog instance on each one. There is no shared view of applications, no unified CLI, and no way to target a specific LPAR (for example, one carrying IBM Spyre AI accelerators) from a central management node.
+
+### 2.3 Goals
+
+1. Allow the Control Plane to deploy containers on **remote LPARs** without changing any existing deployer code.
+2. Keep the local Podman path (`--runtime podman`) working with **zero regression**.
+3. Support **multiple worker LPARs** with per-request agent selection via label selectors.
+4. Provide **operational visibility** — list, status, and delete agents from the control plane CLI.
+5. Provide a **secure bootstrap** process using single-use HMAC tokens (mTLS follow-up planned).
+
+### 2.4 Non-Goals
+
+- mTLS between control plane and agents (reserved for a follow-up; stubs are present).
+- OpenShift-based remote deployment.
+- Cross-LPAR container networking / storage management.
+
+---
+
+## 3. System Topology
+
+```
+┌───────────────────────────────────────────────────────┐
+│                    Control Plane LPAR                 │
+│                                                       │
+│   ai-services CLI.                                    │
+│         │  REST                                       │
+│         ▼                                             │
+│   Catalog API Server  :8080 (gin)                     │
+│         │                                             │
+│         ▼                                             │
+│   AgentDispatcher                                     │
+│         │                                             │
+│         ▼                                             │
+│   RemoteRuntime  ◄──────────────────────────┐         │
+│         │                                   │         │
+│         ▼                                   │         │
+│   AgentGateway  :9090 (gRPC)                │         │
+│         │  bidirectional stream             │         │
+│         │                          AgentRegistry      │
+│         │                          in-memory + DB     │
+└─────────┼──────────────────────────────────┼──────────┘
+          │                                  │
+  ┌───────┴──────────┐              ┌────────┴─────────┐
+  │  Worker LPAR-1   │              │  Worker LPAR-2   │
+  │                  │              │                  │
+  │  agent daemon    │              │  agent daemon    │
+  │  (gRPC client)   │              │  (gRPC client)   │
+  │       │          │              │       │          │
+  │  Podman Socket   │              │  Podman Socket   │
+  │       │          │              │       │          │
+  │  Spyre / GPU     │              │  Spyre / GPU     │
+  └──────────────────┘              └──────────────────┘
+```
+
+---
+
+## 4. Architecture Overview
+
+### 4.1 Key Components
+
+| Component | Location | Responsibility |
+|---|---|---|
+| **AgentGateway** | Control Plane `:9090` | gRPC server; accepts `Register` and `CommandStream` RPCs from worker daemons |
+| **AgentRegistry** | Control Plane | In-memory + PostgreSQL store; tracks agent state, heartbeats, and active command slots |
+| **AgentDispatcher** | Control Plane | Selects the best available READY agent for a request, creates a `RemoteRuntime` bound to it |
+| **RemoteRuntime** | Control Plane | Implements `runtime.Runtime`; serialises each method call into a `Command` proto, sends it via the agent's `CommandCh`, and awaits the `CommandResult` |
+| **agent daemon** | Each Worker LPAR | Persistent process; opens a `CommandStream`, executes incoming `Command` messages against the local Podman socket, returns `CommandResult` messages |
+| **agentbootstrap** | Worker LPAR (run once) | Reads `agent.conf`, calls `AgentGateway.Register` with the pre-shared token, receives and persists TLS material (future mTLS) |
+
+### 4.2 Component Interaction Diagram
+
+```mermaid
+graph TD
+    Client["Client"] -->|POST /applications| AH[ApplicationHandler]
+    AH --> AS[ApplicationService]
+    AS --> DE[DeploymentExecutor]
+    DE -->|runtime=remote| AD[AgentDispatcher]
+    AD -->|SelectAgent| REG[(AgentRegistry)]
+    AD -->|New| RR[RemoteRuntime]
+    RR -->|Command| CH[agent CommandCh]
+    CH --> GW[AgentGateway gRPC]
+    GW -->|stream.Send| DAEMON[agent daemon]
+    DAEMON -->|podman exec| POD[Podman Socket]
+    POD -->|result| DAEMON
+    DAEMON -->|stream.Send result| GW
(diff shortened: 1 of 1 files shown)


docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+
+- Allow AI application pods to be deployed on remote Worker LPARs from a single control plane.
+- Route external HTTPS traffic to deployed services via a per-worker Caddy proxy.
+- Allow the control plane to perform health checks and internal HTTP calls against remote pod endpoints over the gRPC stream, with no new exposed port.
+- Minimal operational overhead: a single daemon binary, one `configure` + `start` workflow per Worker.
+- All communication over a single outbound TCP connection from Worker → control plane (firewall-friendly).
+
+### 2.4 Non-Goals
+
+- OpenShift runtime support for worker Caddy (Podman only in this proposal).
```
mkumatag, line 60: I don't see any implementation for OpenShift in this proposal (not just Caddy), so I'm wondering if we can explore OpenShift support in a separate proposal altogether.

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+- Streaming/chunked HTTP proxy responses (single request/response only).
+- mTLS between agent and gateway (reserved fields would be present; treated as future work).
```
mkumatag, line 62: Are we creating another proposal for mTLS, or will this same document be updated?

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+  string command_id   = 1;
+  bool   success      = 2;
+  bytes  data         = 3; // JSON-encoded response
+  string error        = 4;
+  bool   is_heartbeat = 5;
+  string agent_name   = 6;
+}
+```
+
+### 5.2 Command Types
```
mkumatag, line 179: Is there a possibility of reducing these command types? As they grow, it will be very difficult to manage.

mkumatag: and also not everything will work for both podman and opeshift.

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+## 6. Bootstrap and Registration Flow
+
+### 6.1 Control-plane setup
+
+```bash
+# Start the AgentGateway on the control plane
+ai-services catalog configure --runtime podman --agentgateway-port 9090
+
+# Issue a single-use bootstrap token (24-hour expiry)
+ai-services catalog agent issue-token
```
mkumatag, line 243: may be we need to rename the agent -> worker, let us brainstorm and find all suitable names for these commands.

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+```bash
+# Step 1 — bootstrap the Worker LPAR (installs Podman, SELinux policies, etc.)
+ai-services bootstrap configure --runtime podman
+
+# Step 2 — deploy the worker Caddy proxy pod (run once)
+ai-services agent configure --runtime podman [--https-port 443] [--domain-name example.com] \
+    [--ssl-cert /path/cert.pem --ssl-key /path/key.pem]
+
+# Step 3 — register with the control plane and start the daemon
+ai-services agent start \
```
mkumatag, line 260: can we rename this to `ai-services worker join` does above steps - 1 and 2 for better UX

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+    GW-->>W: RegisterResponse{agent_name: "lpar-1"}
+
+    W->>GW: CommandStream — Send(CommandResult{is_heartbeat:true, agent_name:"lpar-1"})
+    GW->>REG: SetWorkerIP("lpar-1", "192.168.1.5")
+    GW->>REG: SetDomainSuffix("lpar-1", "192.168.1.5.nip.io")
+    GW->>REG: MarkReady("lpar-1")
+    Note over W,GW: stream open — awaiting Commands
+```
+
+**Worker IP capture:** The gateway would read the TCP source IP from the gRPC peer info on the stream context. This would be the address of the Worker LPAR as seen from the control plane.
```
mkumatag, line 290: I'm not sure how this works when multiple IP addresses are set - such as internal and external IPs in PowerVS infrastructure. I'm wondering if we can add an explicit field to advertise the IP address during the join command.

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+---
+
+## 7. Worker Caddy Proxy
+
+Each Worker LPAR would run its own Caddy instance for externally-visible HTTPS routing to deployed service pods. Routes would be registered dynamically by the agent daemon after each successful pod deployment.
+
+### 7.1 Design principles
+
+- **Same pattern as catalog Caddy.** The catalog configure command deploys a Caddy pod; the proposed `agent configure` command would do the same for the worker.
+- **Admin port always random.** Setting `adminPort: "0"` in `values.yaml` would cause the OS to assign a random loopback port, resolved at runtime by inspecting the running pod.
```
mkumatag, line 303: Is there any advantage to choosing a random port? Could this be a fixed port instead?

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+
+:8080 {
+    respond /health "OK" 200
+}
+```
+
+### 7.3 `agent configure` command
+
+```
+ai-services agent configure \
```
mkumatag, line 352: as mentioned about, please explore command like `worker join`

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+When `agent start` runs, it would call an `injectCaddyManager` function:
+
+```go
+adminURL, err := BuildAdminURL(pc)  // pod inspect → random port
+pm := NewCaddyManager(adminURL, AgentCaddyServerName)
+caddyMgr := NewLocalCaddyManagerAdapter(pm)
+pc.SetCaddyManager(caddyMgr)
+```
+
+If the Caddy pod is not running, a warning would be logged and the daemon would start normally — all other runtime operations would continue; only route registration would be unavailable until `agent configure` is run.
```
mkumatag, line 398: This will be problematic, wondering if we can avoid starting the worker itself if something is not configured properly.

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+
+    READY --> DISCONNECTED : heartbeat timeout (90s)
+    BUSY --> DISCONNECTED : heartbeat timeout (90s)
+    DISCONNECTED --> DISCONNECTED : stream closed
+
+    DISCONNECTED --> PENDING : agent re-registers
+    REJECTED --> PENDING : agent re-registers
+```
+
+Proposed status values: `pending`, `ready`, `busy`, `draining`, `disconnected`, `rejected`.
```
mkumatag, line 605: Can we elaborate on the `busy` state? What happens when a worker's state is `busy` - does it prevent deploying another service? Clarifying the expected behavior in this state would help us understand the repercussions of introducing it.

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+
+### 10.3 Agent selection
+
+`registry.SelectAgent(selector)` would iterate all in-memory agents. An agent would match if:
```
mkumatag, line 609: Who is the consumer of this function? Is this chosen during the deployment of a service, or will it be used somewhere else?

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+
+## 11. Database Schema
+
+A new `agents` table would persist agent state across catalog restarts:
+
+```sql
+CREATE TABLE agents (
+    agent_name     TEXT PRIMARY KEY,
+    labels         JSONB        NOT NULL DEFAULT '{}',
+    capabilities   JSONB        NOT NULL DEFAULT '{}',
```
mkumatag, line 628: information is missing for the `capabilities` field.

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+);
+```
+
+The in-memory `Registry` would be the primary source of truth for live routing decisions. The database would provide durability so that `ai-services catalog agent list` works correctly after a catalog restart.
+
+---
+
+## 12. CLI Commands
+
+### 12.1 Control-plane commands
```
mkumatag, line 642: rename the agent -> worker

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+
+**Delete an agent:**
+```
+ai-services catalog agent delete --name <agent-name>
+```
+Would remove the agent from the in-memory registry and PostgreSQL.
+
+**Start the AgentGateway:**
+```
+ai-services catalog configure --runtime podman --agentgateway-port 9090
```
mkumatag, line 664: can we start the agentgateway by default and allow them to override if `--agentgateway-port` flag mentioned.

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+```
+
+### 12.2 Worker agent commands
+
+**`agent configure`** — Deploy the worker Caddy proxy pod (run once before `agent start`):
```
mkumatag, line 669: let us see if we can club both configure and start into `worker join` command

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+    --token  <uuid>                     # required
+    --runtime podman                    # required; no default
+    [--tls-dir /var/lib/ai-services/agent/tls]
+```
+
+- Would load `DomainSuffix` from `~/.config/ai-services/agent.json` and send it as `labels["domain_suffix"]`.
+- Would inject a `LocalCaddyManager` into the Podman runtime by inspecting the running Caddy pod.
+- Would reconnect with exponential backoff (5s base, 120s max) on stream failure.
+
+**`agent status`** — Show current agent state:
```
mkumatag, line 699: wondering why somebody wants to run this command on a every single worker node, rather I prefer to check the status in the control plane

docs/proposals/catalog/remote-worker-agent-proposal.md
```diff
@@ -0,0 +1,879 @@
...
+
+---
+
+## 13. Configuration Files and Assets
+
+### 13.1 Worker agent config
+
+Proposed path: `~/.config/ai-services/agent.json`
+
+```json
```
mkumatag, line 712: find a way to specify these in the command line itself

mkumatag (changes requested):

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: Yes we are planning a PoC on OCP as well, which is not possible until we complete the "Catalog on OCP" story.
But on that side, we think a similar approach should also be possible in OCP environment. We will create a different proposal for sure.

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: Yes see https://github.com/IBM/project-ai-services/pull/1193 which is still under review.

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: These are things that the deployer will use to talk to Podman directly today and similarly, we have an interface common for both runtime, which is sub-set for this.
Maybe we can think of having a different Command Type for each runtime. I would defer it until we have OCP remote deployment.

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: Sure we can go with `worker` for now.

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: You mean running all below steps in come command?
- bootstrap configure (LPAR setup and verification)
- worker configure (Caddy)
- join/start (Registration)

Doable, but will have to have consolidated parameters for it and sequential phased execution.

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: This is the TCP source IP that the gRPC gateway can detect. So this is information that we are getting from the stream connection.
Also note that right now we are not using this data anywhere other than for identification purposes. We might not want it.

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: Could be a fixed one, but kept it the same as the Catalog Caddy. I think the original intention was to make it configurable via values.yaml.

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: You are right, in the PoC I made it kind of required, but the proposal was not updated.
Will take care of this.

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: Right now, I think we will consider only ready and disconnected. In case we are taking worker name as in the bootstrap token generation step a `pending` status would make sense. Others I will clear up.

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: Yes, during application deployment, the control plane will select the worker for deployment, deciding the runtime as remote.

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: Was thinking of adding the h/w resource like # of cards and other details. Now I feel we might not need this during registration.

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: I kept this to differentiate based on a flag. Thinking may we can change it later once the gateway flow is stable.

docs/proposals/catalog/remote-worker-agent-proposal.md
yussufsh: More of a convenience to check from the other side, rather than a need. I am with you on this; it makes sense to check it from the control plane.

mkumatag:
@yussufsh can you please update the proposal with review addressed and merge the PR..

yussufsh:
> @yussufsh can you please update the proposal with review addressed and merge the PR..

Sorry this has been sitting on my stash work. I will update the proposal with the latest implementation details and request for review.

mkumatag (approved):

mkumatag:
I'm merging this PR now, will revisit later if needed. Thanks.

PR #1150, opened 2026-07-17
AISERVICES-1350: consolidate duplicate YAML and term libraries

yussufsh:
1. Two term libraries — golang.org/x/term and github.com/charmbracelet/x/term were both imported solely for ReadPassword. The charmbracelet one was only used in one file with a trivially different call signature (uintptr vs int).

2. Two YAML libraries — gopkg.in/yaml.v3 and go.yaml.in/yaml/v3 have an identical API. gopkg.in/yaml.v3 is the original, but its author [explicitly marked it unmaintained](https://github.com/go-yaml/yaml) in April 2025. go.yaml.in/yaml/v3 is the active successor maintained by the official YAML organization at github.com/yaml/go-yaml.

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.179
+TAG?=v0.0.180
 CONTAINER_BUILDER?=podman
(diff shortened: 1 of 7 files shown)


mayuka-c (approved):
LGTM

Can we verify once if the about call is working fine with this pkg?
GET /api/v1/services/:id

Shubhamag12 (approved):
lgtm

yussufsh:
> LGTM
> 
> Can we verify once if the about call is working fine with this pkg? GET /api/v1/services/:id

Yes, verified APIs and even reset password is working fine.

PR #1153, opened 2026-07-17
feat: add catalog configure command for openshift

Shubhamag12:
This PR introduces OpenShift support for the ai-services catalog configure command. Users can now deploy the catalog service (PostgreSQL + catalog backend + catalog UI) to an OpenShift cluster Helm chart and OpenShift Routes.

```
──> curl https://catalog-api-ai-services.apps.ais-orch-1.example.com/health --insecure              
{"message":"ok"}%
```

Screenshot 2026-07-17 at 20 06 31


ai-services/cmd/ai-services/cmd/catalog/configure.go
```diff
@@ -388,15 +399,38 @@ func buildFlagValidator() *flagvalidator.FlagValidator {
...
 func runResetPassword() error {
-	return catalogPodman.ResetCatalogPassword()
+	rt := vars.RuntimeFactory.GetRuntimeType()
+	switch rt {
+	case types.RuntimeTypePodman:
+		return catalogPodman.ResetCatalogPassword()
+
+	case types.RuntimeTypeOpenShift:
+		return catalogOpenShift.ResetCatalogPassword()
+	}
```
yussufsh: Please add a default case.

ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	return values, nil
+}
+
+func generateArgParams(rt *runtimeOpenshift.OpenshiftClient, passwordHash string) (map[string]string, error) {
+	argParams := make(map[string]string)
+	argParams["backend.adminPasswordHash"] = passwordHash
+
+	dbSecretExists, err := rt.SecretExists("catalog-db-secret")
```
yussufsh: Thinking if we really need this? Then we should do this in podman as well. cc @mayuka-c

yussufsh: Please use a constant for "catalog-db-secret".

ai-services/internal/pkg/catalog/utils/password.go
```diff
@@ -29,4 +35,74 @@ func HashPasswordPBKDF2(password string, iteration int) (string, error) {
...
+	return promptAndHashPassword()
+}
+
+// PromptAndHashPassword prompts for a new password and returns its hash.
+// Used for resets where the secret already exists.
+func PromptAndHashPassword() (string, error) {
+	return promptAndHashPassword()
+}
+
+func promptAndHashPassword() (string, error) {
```
yussufsh: If it is just calling a private function, rename the existing one.

ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
```diff
@@ -0,0 +1,189 @@
...
+		logger.Infof("failed to display next steps: %v\n", err)
+
+		return nil //nolint:nilerr // intentionally swallow error for non-critical step
+	}
+
+	return nil
+}
+
+func getOperationTimeout(ctx context.Context, tp templates.Template, timeout time.Duration) (time.Duration, error) {
+	s := spinner.New("Setting the operation timeout...")
```
yussufsh: Why do we need spinner here? This operation should be pretty fast.

ai-services/assets/catalog/openshift/templates/catalog-db-statefulset.yaml
```diff
@@ -44,6 +45,10 @@ spec:
                 secretKeyRef:
                   name: catalog-db-secret
                   key: db-password
+            # By default, PostgreSQL is designed to block everything unless explicitly told otherwise. 
+            # The POSTGRES_INITDB_ARGS variable acts as a key to override this default behavior.                  
+            - name: POSTGRES_INITDB_ARGS
+              value: "--auth-host=scram-sha-256"
```
yussufsh: Can you please elaborate on this, why is this needed?

yussufsh: Also fix the trailing spaces.

yussufsh (changes requested):

ai-services/cmd/ai-services/cmd/catalog/configure.go
Shubhamag12: we are already checking it in PreRunE, will add it here as well

ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
Shubhamag12: >Thinking if we really need this? 

yes, because with helm we do installOrUpgrade, so when tries to upgrade, the secret is generated again, but the openshift secret remains same, creating discrepancy

ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
Shubhamag12: makes sense, I added what was there for application create for openshift

ai-services/assets/catalog/openshift/templates/catalog-db-statefulset.yaml
Shubhamag12: I was facing this error

> 172.30.40.138:5432 (catalog-db): server error: FATAL: no pg_hba.conf entry for host "10.130.1.41", user "admin", database "postgres", no encryption (SQLSTATE 28000)


checked online and this was the answer

>the postgres image defaults to requiring SSL for remote connections. It is rejecting the connection because the client is connecting without TLS and pg_hba.conf has no rule that permits unencrypted connections from that host. This env var tells initdb to write a pg_hba.conf with:
`host  all  all  all  scram-sha-256`
host accepts both encrypted and unencrypted connections, password auth via scram-sha-256 either way.

This entry was missing in the image, hence it was written using the env var, below are references

References: 
1. https://stackoverflow.com/questions/1406025/no-pg-hba-conf-entry-for-host (explain about the issue)
2. https://hub.docker.com/_/postgres, search for pg_hba.conf (fixes it by creating default file)

This is what I understood, let me know if this requires any further discussion :)

mayuka-c: Any reason for adding this?
We already have intergrated postgresDB for digitize and there I do not see it being used: https://github.com/IBM/project-ai-services/blob/main/ai-services/assets/applications/rag/openshift/templates/postgres-statefulset.yaml#L41

ai-services/assets/catalog/openshift/values.yaml
```diff
@@ -9,7 +9,7 @@ ui:
       cpu: "500m"
 
 backend:
-  image: icr.io/ai-services-cicd/ai-services:v0.0.181
+  image: icr.io/ai-services-private/ai-services:test
```
mayuka-c: ANy reason for changing this image?

ai-services/assets/catalog/openshift/templates/catalog-db-statefulset.yaml
Shubhamag12: image is different, maybe that's the reason?
But it kept failing with 18-3, until above was added explicitly.

ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare values: %w", err)
+	}
+
+	return values, nil
+}
+
+func generateArgParams(rt *runtimeOpenshift.OpenshiftClient, passwordHash string) (map[string]string, error) {
+	argParams := make(map[string]string)
+	argParams["backend.adminPasswordHash"] = passwordHash
```
mayuka-c: Can we have constant for supported argParams and then use it for both podman and OpenShift?

I would say to have a separate file with Podman and OpenShift argParams or in common.go

mayuka-c (commented):
Overall LGTM

ai-services/assets/catalog/openshift/templates/catalog-db-statefulset.yaml
Shubhamag12: I will use 18-4 and check

mayuka-c: Try deploying with same image

ai-services/assets/catalog/openshift/values.yaml
Shubhamag12: need to revert this
it was a testing image

ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
Shubhamag12: you mean all params defined in common.go, or just the params which are common?

mayuka-c: Maybe having constant var like constantOpenShiftParams, constantPodmanParams, constantCommonParams

ai-services/assets/catalog/openshift/values.yaml
Shubhamag12: done

ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
Shubhamag12: done

ai-services/assets/catalog/openshift/templates/catalog-db-statefulset.yaml
Shubhamag12: working with 18-4, ptal

ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/project-ai-services/ai-services/internal/pkg/cli/templates"
+	"github.com/project-ai-services/ai-services/internal/pkg/helm"
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+	runtimeOpenshift "github.com/project-ai-services/ai-services/internal/pkg/runtime/openshift"
+	"github.com/project-ai-services/ai-services/internal/pkg/spinner"
+	"github.com/project-ai-services/ai-services/internal/pkg/utils"
+)
+
+const (
+	catalogDBSecretName = "catalog-db-secret"
```
yussufsh: Please move it to a common const.
I guess @prachikurade1 also can use the same in catalog B&R.

mayuka-c (dismissed):
LGTM

yussufsh (approved):

mayuka-c (approved):

PR #1158, opened 2026-07-20
Digitize focussed DS proposal 

sats-23:
-Keeping the catalog side of changes abstract, this proposal focuses on Digitize side of changes required with a Multithreading approach


docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1768 @@
...
+
+**Error handling policy for the document-deletion loop (step 3):**
+
+| HTTP status from `DELETE /v1/documents/{doc_id}` | Action |
+| --- | --- |
+| `200 OK` / `204 No Content` | Success — proceed to next entry |
+| `404 Not Found` | Document already absent — treat as success; proceed |
+| `5xx` / network error | Log warning with `doc_id` and `remote_path`; continue loop (best-effort cleanup) |
+
+Because step 5's `ON DELETE CASCADE` removes the registry unconditionally, a document-deletion failure means the document may remain in the VDB/storage but its registry row is gone. This is a known trade-off: a hard delete of the connector cannot be held hostage by a flaky document-deletion call. Operators can find and clean up orphaned documents via the `GET /v1/documents` listing filtered by `connector_id`.
```
manalilatkar: you are mentioning a step 5 but step 5 is missing above.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1768 @@
...
+
+The `connection_details` JSONB column stores all connector-type-specific fields — both non-sensitive config and encrypted credential blobs. The exact keys differ by `type`:
+
+| `type` | Keys stored in `connection_details` |
+| --- | --- |
+| `ssh_sftp` | `port`, `username`, `remote_path`, `private_key_ciphertext`, `encrypted_dek` |
+| `s3` | `region`, `bucket_name`, `access_key_id`, `secret_access_key_ciphertext`, `encrypted_dek` |
+
+```sql
+CREATE TABLE IF NOT EXISTS active_connectors (
```
manalilatkar: Shouldn't this contain last_sync_error as well?

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1768 @@
...
+| `access_key_id` | `string` | ✅ | IAM Access Key ID used for API authentication. Stored in plaintext (not a secret). |
+| `secret_access_key_ciphertext` | `string (base64)` | ✅ | AES-256-GCM encrypted IAM Secret Access Key. Encrypted under the per-connector DEK. |
+| `encrypted_dek` | `string (base64)` | ✅ | AES-256-GCM encrypted 32-byte DEK. Encrypted under the pod's `digitize_KEK` secret mount. |
+
+**Response:**
+
+| Status | Body | Meaning |
+| --- | --- | --- |
+| `202 Accepted` | `{ "connector_id": "c7f3a2d1-..." }` | Connector row created; worker thread started asynchronously via `BackgroundTask`. |
+| `409 Conflict` | Error detail | A connector with this `connector_id` already exists. Catalog should use `PUT` to update. |
```
manalilatkar: This line contradicts line no 860:
`| `upsert_active_connector(config: dict)` | `active_connectors` | `INSERT … ON CONFLICT (id) DO UPDATE` — idempotent re-push from catalog. `connection_details` is stored as-is from `config["connection_details"]`. |`

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1768 @@
...
+    │     before download begins, so that a crash mid-batch does not      │
+    │     leave the registry in a permanently inconsistent state.          │
+    │                                                                     │
+    │     NOTE: only files in needs_ingest reach this step — deduped      │
+    │     files already have a registry row from the connector that        │
+    │     originally ingested them.                                        │
+    │                                                                     │
+    │     For each file in needs_ingest:                                  │
+    │         upsert_pending_checksum(id, file.checksum)                  │
+    │           ← lightweight write; doc_id is NULL until ingest succeeds │
```
manalilatkar: How is this possible if line no 728 says:
` doc_id        TEXT        NOT NULL UNIQUE    `

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1768 @@
...
+
+**Key zeroization:** After calling `from_private_key()`, the `privkey_pem` variable is overwritten with `"\x00" * len(privkey_pem)` and set to `None`. The DEK is not retained between ticks — re-derived fresh on every tick to minimise in-memory exposure.
+
+**KEK load timing (Decision D-6 — recommended):** The KEK is read fresh inside `SFTPScanner.connect()` on every tick. This keeps KEK residence time in memory minimal and allows future KEK rotation without a pod restart.
+
+### 7.2 File Scanning — Streaming SHA-256, Filtered by Extension
+
+File type filtering is applied at scan time using `allowed_extensions` from the connector config. This avoids SFTP transfers for files the ingest pipeline would reject. The extension list is populated by catalog (Decision D-8).
+
+`scan()` performs four phases:
```
manalilatkar: nit: it's actually 5 steps below.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Not exactly, it is stored in connector_sync_history and can be retrieved with a query

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Fixed

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1774 @@
...
+
+**Example — `ssh_sftp`:**
+
+```json
+{
+  "connector_id":          "c7f3a2d1-...",
+  "type":                  "ssh_sftp",
+  "host":                  "sftp.example.com",
+  "allowed_extensions":    [".pdf", ".docx", ".xlsx"],
+  "sync_interval_seconds": 300,
```
dharaneeshvrd: for this release let's configure this as env and use it for now.
let's not expose via post payload

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1774 @@
...
+  "connection_details": {
+    "port":                   22,
```
dharaneeshvrd: this is not configurable.
can be removed now

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1774 @@
...
+    "username":               "sync_user",
+    "remote_path":            "/exports/reports",
+    "private_key_ciphertext": "<base64-AES-256-GCM ciphertext of Ed25519 PEM>",
+    "encrypted_dek":          "<base64-AES-256-GCM ciphertext of 32-byte DEK under pod KEK>"
```
dharaneeshvrd: we can remove this.
and remove all logic related to kek and dek at the proposal level
seems like we would be receiving the private key as plain text, let's keep only that.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1774 @@
...
+**Request body (`application/json`):**
+
+Fields common to **all** connector types are at the top level. All connector-specific credential and connection fields are nested inside `connection_details`.
+
+**Example — `ssh_sftp`:**
+
+```json
+{
+  "connector_id":          "c7f3a2d1-...",
+  "type":                  "ssh_sftp",
```
dharaneeshvrd: we can simply keep it as ssh.
sftp is an underlying detail on how are we doing this.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1774 @@
...
+  "host":                  "sftp.example.com",
+  "allowed_extensions":    [".pdf", ".docx", ".xlsx"],
```
dharaneeshvrd: remove xlsx

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1774 @@
...
+**Example — `s3`:**
+
+```json
+{
+  "connector_id":          "a1b2c3d4-...",
+  "type":                  "s3",
+  "host":                  "s3.amazonaws.com",
+  "allowed_extensions":    [".pdf", ".docx", ".xlsx"],
+  "sync_interval_seconds": 300,
+  "connection_details": {
```
dharaneeshvrd, line 151: need to add prefix and delimiter

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1774 @@
...
+| --- | --- | --- |
+| `202 Accepted` | `{ "connector_id": "c7f3a2d1-..." }` | Connector row created; worker thread started asynchronously via `BackgroundTask`. |
+| `409 Conflict` | Error detail | A connector with this `connector_id` already exists. Catalog should use `PUT` to update. |
+| `401 Unauthorized` | Error detail | Missing or invalid bearer token. |
+
+---
+
+### B. `PUT /v1/connectors/{connector_id}` — Update an existing connector
+
+**Purpose:** Called by catalog when the configuration of an existing connector changes — for example, credentials are rotated, the remote path or bucket changes, the sync interval is adjusted, or `allowed_extensions` is updated. Digitize stops the existing sync worker, applies the updated config via `upsert_active_connector`, restarts a fresh worker with the new settings, and **immediately triggers a manual sync tick** on the new worker so that any changes take effect at once.
```
dharaneeshvrd: as per our discussion, we won't trigger a manual sync on this, we would throw an error if there is an existing sync happening on the source. we would allow only if the connector is in synced/error state

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: So it will not be configurable on UI also?

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Done

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Done, no more manual triggers. Not hot-update for PUT call either, we will wait for current sync to finish and at start of each tick, fetch from DB

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Updated, PTAL

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: SHA is computed over stream during file walk. 
Used in memory for lookup and diff computation. 
Batch ingestion is done. 
Only after successful ingest, the checksums are stored and table populated.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Fixed

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,892 @@
...
+Each PR is independently testable — no PR leaves things in a broken or untestable state.
+
+---
+
+### PR 1 — DB Schema + ORM Models + Settings
+
+**Files touched:** `init_schema.sql`, `db/models.py`, `config/settings.py`
+
+**What's built:**
+- 4 new tables: `active_connectors`, `file_checksum_registry`, `connector_file_membership`, `connector_sync_history`
```
sats-23: Some of these details will update once de-dup PR is merged

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,892 @@
...
+> **Scope:** Internal `digitize` behavior after catalog sends connector payloads. Catalog-side concerns such as key generation, connector CRUD, deployment wiring, TLS provisioning, and bearer-token enforcement remain out of scope and are treated as infrastructure-level prerequisites.
+
+---
+
+## 1. Preconditions
+
+Before any `digitize` connector endpoint is called:
+
+- Catalog has already validated the remote connector configuration.
+- Catalog sends secret material in plaintext over TLS:
```
dharaneeshvrd: TLS is till agent or catalog backend.
From there to digitize is just internal podman network without TLS.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,892 @@
+# Digitize — Data Source Connector Processing Proposal
+
+> **Scope:** Internal `digitize` behavior after catalog sends connector payloads. Catalog-side concerns such as key generation, connector CRUD, deployment wiring, TLS provisioning, and bearer-token enforcement remain out of scope and are treated as infrastructure-level prerequisites.
```
dharaneeshvrd:
```suggestion
> **Scope:** Internal `digitize` behavior after catalog sends connector payloads. Catalog-side concerns such as key management, connector CRUD, deployment wiring and TLS provisioning remain out of scope and are treated as infrastructure-level prerequisites.
```

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,892 @@
...
+
+Before any `digitize` connector endpoint is called:
+
+- Catalog has already validated the remote connector configuration.
+- Catalog sends secret material in plaintext over TLS:
+  - `ssh`: `private_key`
+  - `s3`: `secret_access_key`
+- `digitize` encrypts those secret fields at rest using `/run/secrets/connector_encryption_key` before persisting them.
+- `/run/secrets/connector_api_token` and `/run/secrets/connector_encryption_key` are mounted before pod start.
+- TLS and bearer-token enforcement are already provided by infrastructure.
```
dharaneeshvrd: this line can be removed

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,892 @@
...
+
+Common fields:
+
+| Field | Type | Required | Notes |
+| --- | --- | --- | --- |
+| `connector_id` | `string (UUID)` | ✅ | Stable catalog ID |
+| `type` | `string` | ✅ | `ssh` or `s3` |
+| `host` | `string` | ✅ | SFTP host or S3 endpoint |
+| `allowed_extensions` | `array[string]` | ✅ | Non-matching files are ignored |
+| `sync_interval_seconds` | `integer` | ✅ | Poll interval |
```
dharaneeshvrd: let's remove this and configure via env in this release.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,892 @@
...
+}
+```
+
+#### Response codes
+
+| Status | Meaning |
+| --- | --- |
+| `202 Accepted` | Connector created; worker start scheduled in a background task |
+| `409 Conflict` | Connector already exists |
+| `401 Unauthorized` | Missing or invalid token |
```
dharaneeshvrd: when is this possible?

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,892 @@
...
+
+### 3.2 `PUT /v1/connectors/{connector_id}`
+
```
dharaneeshvrd, line 202: Mention there would be validation against the last sync status, if it is in progress, request need to be rejected

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,892 @@
...
+}
+```
+
+Response codes:
+
+| Status | Meaning |
+| --- | --- |
+| `200 OK` | Connector updated; worker picks up changes on next tick |
+| `404 Not Found` | Connector does not exist |
+| `401 Unauthorized` | Missing or invalid token |
```
dharaneeshvrd: here too

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,892 @@
...
+
+Response codes:
+
+| Status | Meaning |
+| --- | --- |
+| `204 No Content` | Connector removed |
+| `404 Not Found` | Connector does not exist |
+| `401 Unauthorized` | Missing or invalid token |
+
+### 3.4 `GET /v1/connectors`
```
dharaneeshvrd, line 279: Please add json payload sample for this and remaining GET calls

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,892 @@
...
+
+Lists active connectors with non-secret configuration and current sync state.
+
+Returned fields include:
+
+- connector identity and config
+- `sync_status`, `last_sync_at`, `last_sync_error`, `attached_at`
+- non-secret `connection_details`
```
dharaneeshvrd: connection_details can be part of `/v1/connectors/{connector_id}` no need in GET all call

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,892 @@
...
+| `404 Not Found` | Connector does not exist |
+| `401 Unauthorized` | Missing or invalid token |
+
+### 3.4 `GET /v1/connectors`
+
+Lists active connectors with non-secret configuration and current sync state.
+
+Returned fields include:
+
+- connector identity and config
```
dharaneeshvrd, line 285: what configs are you planning to include?

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,892 @@
...
+    allowed_extensions      JSONB       NOT NULL DEFAULT '[]',
+    sync_interval_seconds   INTEGER     NOT NULL DEFAULT 300,
+    attached_at             TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    last_sync_at            TIMESTAMPTZ,
+    sync_status             TEXT        NOT NULL DEFAULT 'idle',
+    CONSTRAINT chk_connector_type CHECK (type IN ('ssh', 's3'))
+);
+```
+
+### 4.3 `file_checksum_registry`
```
dharaneeshvrd: Will connector or sync worker try to add an entry into this table?

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
dharaneeshvrd: yes

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
dharaneeshvrd: I think we need to revisit the rejection part.
Maybe we can allow the db update and next sync onwards it would point to the new details.

dharaneeshvrd: Do you see any problems with that?
@sats-23 @manju956

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Removed it. Both Boto and Paramiko rely on throwing exceptions in case credentials fails

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Yes, I thought that was the plan, keep the db update async, because either way we are polling from DB before starting every fresh sync tick

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Removed

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Done, reworded it

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Removed

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Done

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Removed

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Planned was connector_id, type, host, allowed_extensions, sync_interval_seconds

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Nope, the batch job will take care of adding entries to file_checksum_registry. 
The Sync Worker thread will only work with connector_file_membership.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,978 @@
...
+    sha256        TEXT NOT NULL,
+    PRIMARY KEY (connector_id, sha256),
+    FOREIGN KEY (connector_id) REFERENCES active_connectors(id) ON DELETE CASCADE,
+    FOREIGN KEY (sha256) REFERENCES file_checksum_registry(sha256)
+);
+```
+
+Deduplication and deletion rules:
+
+- Existing `sha256` in `file_checksum_registry` => reuse document and add membership only.
```
sats-23: @dharaneeshvrd @manju956 
I feel like we should keep the insert operations to connector_file_membership as part of the main thread (instead of individual worker thread) and executed along with insert operations to file_checksum_registry (in same transaction).

This comes from PoV that a file membership can be owned by a manual ingest by user as well (which we never store). So if a connector is deleted, we risk removing the files that the user intended to manually ingest as well. 

Minor change, but I feel its good to keep the pass the owner/connectorID through the create_job loop so the main thread can execute this DB operation 
Needed because:
1. Stores membership for files ingested manually as "user" or "manual" to prevent being deleted when connector is dropped. 
2. Prevents race condition and PK violation when multiple threads are adding/removing entries from membership table

dharaneeshvrd: Take care of below things:

What is the field you are thinking to add to job properties?
What would be the value of it in case of user triggered job?
We should not show this property on GET calls of job.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,989 @@
...
+```
+
+Response codes:
+
+| Status | Meaning |
+| --- | --- |
+| `200 OK` | Connector updated; worker picks up changes on next tick |
+| `404 Not Found` | Connector does not exist |
+
+### 3.3 `DELETE /v1/connectors/{connector_id}`
```
dharaneeshvrd, line 233: Please mention that DELETE would be allowed only if there are no active sync happening.
Since stopping active sync requires, force stopping the jobs which is not there currently, mention that as a stretch, if that's accomplished, DELETE can be allowed even when a sync is happening.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,989 @@
...
+#### Example response
+
+```json
+[
+  {
+    "connector_id": "c7f3a2d1-4e5b-4c6d-8f9a-0b1c2d3e4f5a",
+    "type": "ssh",
+    "host": "sftp.example.com",
+    "allowed_extensions": [".pdf", ".docx"],
+    "sync_interval_seconds": 300,
```
dharaneeshvrd: no need.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,989 @@
...
+    "attached_at": "2025-01-10T08:00:00Z",
+    "last_sync_at": "2025-01-15T14:32:10Z",
+    "sync_status": "idle",
+    "last_sync_error": null,
```
dharaneeshvrd, line 299: we discussed to attach last sync's details as well right?
to show the file stats.

dharaneeshvrd: nevermind.
sry that's on GET of a specific connector.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Done

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Sure, all seem to be implementation specifics. I shall not mention in proposal

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1170 @@
...
+| S3 multi-part | S3 ETag = `MD5(MD5(p₁)‖…‖MD5(pₙ))-N` — hex + `-N` suffix | Unique per (file content + part size) |
+| SFTP | `md5sum` output from remote host — 32-char hex | Unique per file content, protocol-independent |
+
+### 4.4 `connector_file_membership`
+
+Maps connectors to the `sha256` values they currently own.
+
+```sql
+CREATE TABLE IF NOT EXISTS connector_file_membership (
+    connector_id  TEXT NOT NULL,
```
dharaneeshvrd: this field should be generic to accommodate the user submitted jobs

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -547,23 +615,24 @@ Base scanner responsibilities:
 
 Subclass responsibilities:
 
-- remote listing
-- SHA-256 computation (strategy is transport-specific — see §7.1 and §7.2)
-- file download
+- remote listing — yields `(remote_path, checksum)` pairs
```
sats-23: Why path?

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -501,30 +552,47 @@ The DB layer stores and returns ciphertext only. Encryption happens in the API l
...
+| `insert_checksum_registry(checksum, source, doc_id)` | register newly ingested content |
+| `delete_checksum_registry(checksum)` | defensive cleanup |
+| `list_connector_checksums(connector_id)` | current checksums owned by this connector |
+| `insert_connector_membership(connector_id, checksum)` | add ownership row |
+| `delete_connector_membership_atomic(connector_id, checksum)` | remove ownership, return remaining ref count |
 | `insert_sync_history()` | create tick row |
 | `update_sync_history()` | finalize tick row |
 | `update_sync_history_files_syncing()` | live progress updates |
 | `list_sync_history()` | paginated history query |
+| `set_document_metadata(doc_id, metadata)` | write `source_checksum` + S3 key into `documents.metadata` |
```
sats-23, line 612: Why S3 key?

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Added details for user submitted in latest commit

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
manju956: this is basically the s3 bucket path to trace where the document is sourced from. I think we dont need, since we already have connector_id->checksum mapping in `connector_membership` table.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -38,32 +38,34 @@ PUT
...
+  → connector_file_membership is never written for user-submitted docs
 ```
 
 ### 2.3 Main Components
 
 - `active_connectors`: current connector configuration and top-level sync state
-- `file_checksum_registry`: global content registry keyed by `sha256` — stores a content fingerprint and a reference to the ingested `doc_id`. **Already implemented** — table and ORM model exist.
-- `connector_file_membership`: which connector currently references which checksum
+- `file_checksum_registry`: **user-submitted documents only** — keyed by `sha256`, stores a content fingerprint and the `doc_id`. **Already implemented** — table and ORM model exist. Connector code must never write to this table.
+- `connector_file_membership`: **connector-sourced documents only** — keyed by `checksum` (PK), carries the list of connector IDs that reference this file and the `doc_id` for deletion. User-submitted docs are never written here.
```
manju956: So, we will have SHA for user submitted files and md5 for connector sourced files?

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -458,63 +503,91 @@ class FileChecksumRegistry(Base):
...
-| S3 multi-part | S3 ETag = `MD5(MD5(p₁)‖…‖MD5(pₙ))-N` — hex + `-N` suffix | Unique per (file content + part size) |
-| SFTP | `md5sum` output from remote host — 32-char hex | Unique per file content, protocol-independent |
+**Connector code must not write here.** The connector path uses `connector_file_membership` exclusively.
 
 ### 4.4 `connector_file_membership`
 
-Maps connectors to the checksums they currently own.
+**Connector-sourced documents only.** This table is the sole dedup and reference-counting store for all content ingested via connectors. User-submitted code must never write to this table.
+
+The schema uses `checksum` as the primary key. The `connector_ids` column is a JSON array of all connector IDs that currently reference this file (cross-connector shared-content tracking). `doc_id` is stored here — not in `file_checksum_registry` — so that deletion can proceed without a registry join.
```
manju956: why does `connector_ids` have to be JSON array not a simple array?

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: No, everything will be MD5, I shall make that change

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: It just chosen for cosmetics reason for some operations like removing element by value.
Here's a comparison:

| Operation | TEXT[] Equivalent | JSONB Equivalent |
| :--- | :--- | :--- |
| Remove element by value | `array_remove(col, val)` | `col - val` |
| Containment check | `col @> ARRAY[val]` | `col @> jsonb_build_array(val)` |
| Append element | `col \|\| ARRAY[val]` | `col \|\| jsonb_build_array(val)` |
| Length check | `array_length(col, 1)` | `jsonb_array_length(col)` |

The performance difference is negligible. 

What do you think? Should we switch to TEXT[] or keep JSONB

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1748 @@
...
+
+Before any `digitize` connector endpoint is called:
+
+- Catalog has already validated the remote connector configuration.
+- Catalog sends secret material in plaintext via API calls:
+  - `ssh`: `private_key`
+  - `s3`: `secret_access_key`
+- `digitize` encrypts those secret fields at rest using `/run/secrets/connector_encryption_key` before persisting them.
+- `/run/secrets/connector_api_token` and `/run/secrets/connector_encryption_key` are mounted before pod start.
+- The `file_checksum_registry` table and the `FileChecksumRegistry` ORM model already exist in the codebase (merged in the de-duplication PR). This table is **user-submitted documents only** — connector code must never read from or write to it. Connector dedup is handled exclusively via `connector_file_membership` (see §4.4).
```
dharaneeshvrd: this is not required, we can remove all mention of file_checksum_registry table

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1334 @@
...
+- `/run/secrets/connector_api_token` and `/run/secrets/connector_encryption_key` are mounted before pod start.
+- The `file_checksum_registry` table and `FileChecksumRegistry` ORM model are already implemented (user-submitted documents only). Connector code must never read from or write to it — connector dedup is handled exclusively via `connector_file_membership` (see §4).
+
+---
+
+## 2. System Overview
+
+### 2.1 Architecture Diagram
+
+![Architecture Diagram](architecture-diagram.svg)
```
dharaneeshvrd, line 25: Arch diagram need some rewording 
SHA256 to md5
f.c.r table need to be removed
c.f.m table need to be renamed

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1334 @@
...
+
+### 2.2 Runtime Flow
+
+```text
+── Attach (POST /v1/connectors) ─────────────────────────────────────
+  → validate + encrypt credentials
+  → INSERT active_connectors row
```
dharaneeshvrd: why did we start calling `active_connectors`?
should it be just connectors similar to jobs, documents..

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1334 @@
...
+  → worker thread starts and runs first tick immediately
+
+── Config update (PUT /v1/connectors/{id}) ──────────────────────────
+  → merge + re-encrypt changed fields
+  → UPDATE active_connectors row
+  → running worker re-reads config from DB before its next tick
+
+── Each sync tick (connector-sourced) ───────────────────────────────
+  Step 1 — load state
+    → SELECT known_checksums FROM connector_file_membership
```
dharaneeshvrd: how about?
connector_file_membership - connector_document_checksum
file_checksum_registry - document_checksum

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Done

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1371 @@
...
+
+```json
+{
+  "connector_id": "b5c6d7e8-...",
+  "connector_name": "prod-cos-ai-services",
+  "type": "s3",
+  "allowed_extensions": [".pdf", ".docx"],
+  "connection_details": {
+    "host": "s3.us.cloud-object-storage.appdomain.cloud",
+    "endpoint_url": "https://s3.us.cloud-object-storage.appdomain.cloud",
```
dharaneeshvrd: why do we need two properties?
let's keep only endpoint_url

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1371 @@
...
+## 1. Preconditions
+
+Before any `digitize` connector endpoint is called:
+
+- Catalog has already validated the remote connector configuration.
+- Catalog sends secret material in plaintext via API calls:
+  - `ssh`: `private_key`
+  - `s3`: `secret_access_key`
+- `digitize` encrypts those secret fields at rest using `/run/secrets/connector_encryption_key` before persisting them.
+- `/run/secrets/connector_api_token` and `/run/secrets/connector_encryption_key` are mounted before pod start.
```
dharaneeshvrd: remove connector_api_token

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1371 @@
...
+| --- | --- | --- | --- |
+| `host` | `string` | ✅ | S3 endpoint hostname (e.g. `s3.us-east-1.amazonaws.com`) |
+| `endpoint_url` | `string` | ✅ | Full S3 endpoint URL. AWS S3: `https://s3.<region>.amazonaws.com`. IBM COS: `https://s3.<region>.cloud-object-storage.appdomain.cloud`. Provider and region are auto-detected from this URL — no separate `region` field needed. |
+| `bucket_name` | `string` | ✅ | |
+| `access_key_id` | `string` | ✅ | IAM key ID (AWS) or HMAC key ID (IBM COS) |
+| `secret_access_key` | `string` | ✅ | IAM secret (AWS) or HMAC secret (IBM COS) |
+| `prefix` | `string` | ❌ | Key prefix to scope listing — empty means bucket root |
+| `delimiter` | `string` | ❌ | Set `"/"` for non-recursive (immediate children only) |
+
+> **Checksum-based dedup:** For S3 connectors, `list_objects_v2` returns the object ETag at no extra API cost — stored as `checksum` in `connector_document_checksum`. If the checksum is already present for this connector the file is **never downloaded**. The checksum is also stored in `documents.metadata.source_checksum` for traceability.
```
dharaneeshvrd:
```suggestion
> **Checksum-based dedup:** For S3 connectors, `list_objects_v2` returns the object ETag at no extra API cost — stored as `checksum` in `connector_document_checksum`. If the checksum is already present for this connector the file is **never downloaded**.
```

dharaneeshvrd: Hope the removed line is not valid

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1371 @@
...
+    "last_sync_error": null,
+    "total_files": 42
+  },
+  {
+    "connector_id": "a1b2c3d4-5e6f-7a8b-9c0d-1e2f3a4b5c6d",
+    "connector_name": "prod-s3-rag-docs",
+    "type": "s3",
+    "attached_at": "2025-01-12T09:15:00Z",
+    "last_sync_at": "2025-01-15T14:30:00Z",
+    "sync_status": "failed",
```
dharaneeshvrd: can we use out of sync instead of failed?

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -0,0 +1,1371 @@
...
+  "items": [
+    {
+      "sync_id": 3,
+      "started_at": "2025-01-15T14:32:00Z",
+      "finished_at": "2025-01-15T14:32:10Z",
+      "total_files": 42,
+      "new_files": 0,
+      "removed_files": 0,
+      "failed_files": 0,
+      "sync_status": "up to date"
```
dharaneeshvrd:
```suggestion
      "id": 3,
      "started_at": "2025-01-15T14:32:00Z",
      "finished_at": "2025-01-15T14:32:10Z",
      "total_files": 42,
      "new_files": 0,
      "removed_files": 0,
      "failed_files": 0,
      "status": "up to date",
      "error": ""
```

let's remove redundant sync prefix
and we need to add error field here similar to the one available at connector level.
also now I m thinking the status at connector level should not be used as is here.
It is at per sync level, hence thinking started, completed & failed would make more sense here, wdyt?
cc @manju956

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Yes, just stale reference, removed

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Yes, makes sense to keep it simple. 
Error message (if any) on the latest sync will be the last_sync_error on GET connectors

manju956: makes sense to me.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
dharaneeshvrd: > Error message (if any) on the latest sync will be the last_sync_error on GET connectors

Yes, we can update both connectors and sync table on every tick respectively

dharaneeshvrd (approved):

PR #1163, opened 2026-07-21
Add services and components helm charts for openshift

Shubhamag12:
This PR adds helm charts for all services and components.


ai-services/assets/components/embedding/vllm-cpu/openshift/templates/embedding-inferenceservice.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: serving.kserve.io/v1beta1
+kind: InferenceService
+metadata:
+  name: embedding
```
mayuka-c, line 4: Each application will have its own embedding instance. This will not work if all are deployed in same namespace

mayuka-c: Similarly everywhere for services and ocmponents

ai-services/assets/services/chat/openshift/values.schema.json
```diff
@@ -0,0 +1,28 @@
...
+          "description": "Custom initial system prompt for conversational RAG",
+          "default": "You are a helpful, conversational AI assistant. Engage naturally with users across multiple turns of conversation. Provide clear, accurate, and contextually relevant responses. Reference previous exchanges when appropriate to maintain conversation flow.",
+          "minLength": 10,
+          "maxLength": 5000,
+          "x-ui-controlled-by": "editSystemPrompt"
+        }
+      }
+    }
+  }
+}
```
mayuka-c: Add a new line

ai-services/assets/services/chat/openshift/metadata.yaml
```diff
@@ -0,0 +1,7 @@
+name: chat
+version: "1.0.0"
+
+# Resource requirements (UI + Backend)
+resources:
+  cpu: 1 # 50m (UI) + 1 (Backend) = ~1 core
+  memory: 1610612736 # 512Mi (UI) + 1Gi (Backend) = 1.5Gi in bytes
```
mayuka-c: Here too

ai-services/assets/components/embedding/vllm-cpu/openshift/templates/embedding-inferenceservice.yaml
Shubhamag12: but each application will have a different namespace right?

mayuka-c: Yeah as I said if we have same namespace it will not work. So what would be the namespace for each application?

ai-services/assets/components/llm/vllm-spyre/openshift/templates/instruct-inferenceservice.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: serving.kserve.io/v1beta1
+kind: InferenceService
+metadata:
+  name: instruct
```
mayuka-c: Also lets use llm similar to what we follow in new podman structure. Also in other places

ai-services/assets/components/embedding/vllm-cpu/openshift/templates/embedding-inferenceservice.yaml
Shubhamag12: I was thinking appid (first few digits) would be the namespace for each application, wdys?

ai-services/assets/components/llm/vllm-spyre/openshift/templates/instruct-inferenceservice.yaml
Shubhamag12: done

yussufsh (approved):
overall lgtm.
Missing pieces like ServingRuntime and other things will be taken care in the create flow.

PR #1167, opened 2026-07-22
refactor application layer to make it runtime agnostic

Shubhamag12:
Splits the application_service.go into a repository/application_service/ subpackage with three focused files:

1. common.go: ApplicationServiceBase struct with shared DB operations (UpdateApplication, GetApplicationByID, InsertDeploymentRecords, and all private helpers)
2. podman.go: PodmanApplicationService with Podman-specific methods (ListApplications, DeleteApplication, CreateApplication, GetApplicationResources, ApplicationsPs) moved verbatim
3. openshift.go: OpenShiftApplicationService stub where every method returns a clear "OpenShift runtime is not yet supported" error, so the apiserver starts cleanly on --runtime openshift instead of panicking
4. Factory in application_service.go reduced to routing only, explicit switch on RuntimeTypePodman, RuntimeTypeOpenShift, and default (panic, programmer error)
5. Extracts ApplicationServiceInterface into a dedicated interface.go; deletes errors.go with constants moved into common.go
6. Public types (ValidationError, ListApplicationsRequest, DeleteApplicationResponse, ValidatePaginationParams) re-exported from the factory file as aliases, no changes needed in any caller

Testing:
Created and deleted application using UI


ai-services/internal/pkg/catalog/apiserver/repository/application_service/podman.go
```diff
@@ -0,0 +1,574 @@
...
+
+// PodmanApplicationService implements ApplicationServiceInterface for the Podman runtime.
+// It embeds ApplicationServiceBase for all shared DB operations and adds Podman-specific
+// deployment, pod-status, and resource-inspection logic.
+type PodmanApplicationService struct {
+	ApplicationServiceBase
+}
+
+// ListApplications retrieves a paginated list of applications with filters.
+func (s *PodmanApplicationService) ListApplications(ctx context.Context, req ListApplicationsRequest) (*types.ApplicationListResponse, error) {
```
mayuka-c, line 36: Wouldnt this be same for both podman and openshift. This is not runtime dependent. So we should be using same for both

ai-services/internal/pkg/catalog/apiserver/repository/application_service/podman.go
```diff
@@ -0,0 +1,574 @@
...
+			TotalPages: totalPages,
+			HasNext:    req.Page < totalPages,
+			HasPrev:    req.Page > 1,
+		},
+	}, nil
+}
+
+// DeleteApplication initiates async deletion of an application and returns immediately.
+func (s *PodmanApplicationService) DeleteApplication(ctx context.Context, id uuid.UUID, user string, keepData bool) (*DeleteApplicationResponse, error) {
+	app, err := s.AppRepo.GetByID(ctx, id)
```
mayuka-c, line 84: Even here most of the logic is same, the only different would be during pod deletion where you will use helm and rest remains same. Only that part should be runtime specific and rest should be same across both

ai-services/internal/pkg/catalog/apiserver/repository/application_service/podman.go
```diff
@@ -0,0 +1,574 @@
...
+	return &DeleteApplicationResponse{
+		ID:      id.String(),
+		Status:  string(models.ApplicationStatusDeleting),
+		Message: "Deletion initiated successfully",
+	}, nil
+}
+
+// CreateApplication validates, plans, persists, and asynchronously deploys a new application
+// using the Podman runtime executor.
+func (s *PodmanApplicationService) CreateApplication(ctx context.Context, req apimodels.CreateApplicationRequest) (*apimodels.CreateApplicationResponse, error) {
```
mayuka-c, line 134: Here too, we should not be duplicating the same logic to openshift

ai-services/internal/pkg/catalog/apiserver/repository/application_service/podman.go
```diff
@@ -0,0 +1,574 @@
...
+	}
+
+	// Phase 5: async deployment
+	go s.executeDeploymentAsync(ctx, plan, req)
+
+	return &apimodels.CreateApplicationResponse{ID: plan.ApplicationID.String()}, nil
+}
+
+// executeDeploymentAsync runs the Podman deployment in a background goroutine.
+func (s *PodmanApplicationService) executeDeploymentAsync(parentCtx context.Context, plan *deployment.DeploymentPlan, req apimodels.CreateApplicationRequest) {
```
mayuka-c, line 170: This entire file needs to be properly refactored to ensure we reuse everything for openshift, most of these are not podman specific.

Only min runtime specific we need to segregate

ai-services/internal/pkg/catalog/apiserver/repository/application_service/common.go
```diff
@@ -0,0 +1,560 @@
...
+
+	// ErrMsgUserNotOwner is returned when a user does not own the application.
+	ErrMsgUserNotOwner = "user does not own this application"
+
+	// ErrMsgApplicationAlreadyDeleting is returned when an application is already being deleted.
+	ErrMsgApplicationAlreadyDeleting = "application is already being deleted"
+
+	// ErrMsgApplicationNameExists is returned when an application with the given name already exists.
+	ErrMsgApplicationNameExists = "application with name '%s' already exists"
+)
```
mayuka-c: This can still be in common errors.go file

ai-services/internal/pkg/catalog/apiserver/repository/application_service/podman.go
Shubhamag12: yes, most of it would be same. I was thinking when we implement delete for openshift, we can move logic then. wdyt?

ai-services/internal/pkg/catalog/apiserver/repository/application_service/podman.go
Shubhamag12: same

ai-services/internal/pkg/catalog/apiserver/repository/application_service/common.go
Shubhamag12: will move into error.go inside service folder

ai-services/internal/pkg/catalog/apiserver/repository/application_service/podman.go
mayuka-c: Pls ensure we are reusing and lets avoid duplicates

mayuka-c (dismissed):
LGTM

ai-services/internal/pkg/catalog/apiserver/repository/application_service/podman.go
Shubhamag12: will handle it when we implement ListApplications then

ai-services/internal/pkg/catalog/apiserver/repository/application_service/common.go
Shubhamag12: done

mayuka-c (approved):

PR #1176, opened 2026-07-23
fix: normalize uploaded file extensions to lowercase before processing 

### LINKED ISSUE #1172: File ingestion fails if the file extension is not lower case
## 🐛 Description
Discovered that attempting to ingest pdf files with a .PDF will fail with an unclear message that reads "Ingestion failed: division by zero".

## ✅ Expected Behavior
The error message should be more clear or the file name and extension should be normalized at the time of the import.  Failure message should more clear with something like "Unsupported or unreadable file format" or the file name i.e file.PDF should be automatically normalized to be read as file.pdf for a successful ingest operation.

## ❌ Actual Behavior
Noticed that the operation was failing and attempted to manually alter the file extension and was then able to successfully ingest the files.

## 🔁 Steps to Reproduce

1. Modify a sample .pdf file to an extension of .PDF - it should fail at the time of ingest with an unclear message
- Behavior is easily reproduced

## 🖥️ Environment Info
- RHEL Version: 9.8
- AI Services Version:
Version: v0.3.0
GitCommit: 991ae85f
BuildDate: 2026-06-30T15:23:17Z

## 🧪 Diagnostic Commands & Output

Output failure message is: 
"Ingestion failed: division by zero".

## 📸 Screenshots / Logs
N/A

## 📎 Additional Context
 This is a probably a quick fix in the ingestion code with something like:
extension = file.filename.rsplit(".", 1)[-1].lower() # normalize before matching

dharaneeshvrd:
Fixes #1172

This change is required to accommodate the valid capital case file extensions(PDF, DOCX)

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.190
+TAG?=v0.0.191
 CONTAINER_BUILDER?=podman
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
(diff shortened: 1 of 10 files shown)


manju956 (approved):
lgtm

PR #1177, opened 2026-07-23
npm CVE fixes

pranithraoibm:
This fixes all the npm cves mentioned here: [Dependabot list](https://github.com/IBM/project-ai-services/security/dependabot?before=Y3Vyc29yOnYyOpLLP_kzMzMzMzPPAAAAAaymfqI%3D&q=is%3Aopen+ecosystem%3Anpm)

Ran `npm audit fix` across all 3 UI folders and ensured `found 0 vulnerabilities` was returned


yussufsh:
@pranithraoibm please update the version ai-services Makefile as well.

Also don't we need to update `package.json` everywhere?

pranithraoibm:
> Also don't we need to update package.json everywhere?

@yussufsh package.json declares the allowed range (e.g. ^7.16.0), while package-lock.json records the exact installed version. npm audit fix only updates the lockfile to install a safe version within that range — package.json only needs changing if the fix requires a version outside the declared range (e.g. a new major version), which isn't the case here. Fixes to these CVEs were present within the range specified in package.json so only package.lock.json was updated and no changes were needed in the package.json file.

pranithraoibm:
Just checked the depandabot PR https://github.com/IBM/project-ai-services/pull/1162
There it updates the package.json file as well. This is a good to have update but there is no issue if its missed as when we run npm install versions are always picked from package-lock.json and it already had the right versions. Updated for these packages in this PR and will handle the version updates in package.json for other packages to keep them in sync with the package-lock.json file versions in a separate PR.

yussufsh:
Thanks @pranithraoibm I will merge this. Please update the other packages as well.
Also, I would think adding commits to dependabot PRs would be great instead of a new PR?

yussufsh (approved):

PR #1178, opened 2026-07-23
Proposal: MCP Sidecar for AI Services

Bleama02:
Summary
Adds the MCP sidecar proposal document for AI Services.

Changes:

- Added docs/proposals/mcp-sidecar-proposal.md

Jira
[AISERVICES-1455](https://jsw.ibm.com/browse/AISERVICES-1455)


docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+**Tool selection boundary.** The sidecar never decides which tool to call. It only receives an
+already-decided `tools/call` request from the connected agent and executes it. All reasoning
+about *which* tool fits a user's request happens client-side, inside the agent's own LLM, using
+nothing but the tool names and descriptions we expose. This is worth stating explicitly: a reader
+unfamiliar with agentic tool-use will otherwise reasonably ask whether this component is making
+autonomous decisions about customer data. It is not.
+
+The sidecar is fully integrated into the existing `ai-services` CLI. Customers who already use
+`ai-services bootstrap` and `ai-services application start` to set up their environment can opt
+into MCP through flags they already know, no separate tooling to learn or install.
```
mayuka-c: Where exactly would this MCP flag be set and stored?

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+no standardised way for an AI agent or ISV application to discover capabilities and call them
+without detailed knowledge of the underlying HTTP layer.
+
+All four microservices already expose a well-documented `/openapi.json` with rich descriptions
+written for Swagger UI. This proposal surfaces that existing documentation as callable MCP tools
+with no changes to any service.
+
+## Proposed Solution
+
+A single generic MCP sidecar container image, deployed once per service, configured entirely by
```
mayuka-c, line 100: As per the statement here, we will be deploying side car container for each service for a given application. So if the MCP is not set, we will not be deploying this side car container across all applications?

Again during application deployment how do we know whether MCP is set or not?

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+
+```
+AI Agent  (any MCP-compatible client)
+  │
+  │  MCP tools/list  →  receives typed tool definitions + descriptions
+  │  MCP tools/call  →  calls tool with structured arguments
+  ▼
+MCP Sidecar  (one per service, generic image, env-var config)
+  │  startup: fetches OPENAPI_URL → parses spec → registers tools
+  │  runtime: reconstructs HTTP request → forwards to SERVICE_URL
```
mayuka-c: I guess this sidecar is exposed externally for each service?

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+
+The same image runs for every service. Each instance is configured by two required environment
+variables and an optional tag filter that hides internal endpoints (health, metrics) from the agent.
+
+| Service | `OPENAPI_URL` | `SERVICE_URL` | Recommended `TAGS` |
+|---|---|---|---|
+| digitize | `http://digitize:4000/openapi.json` | `http://digitize:4000` | `ingestion,jobs` |
+| chatbot | `http://chatbot:5000/openapi.json` | `http://chatbot:5000` | `chat,retrieval` |
+| summarize | `http://summarize:6000/openapi.json` | `http://summarize:6000` | `summarization,jobs` |
+| similarity | `http://similarity:7000/openapi.json` | `http://similarity:7000` | `similarity` |
```
mayuka-c: I guess these values will be hardcoded in the pod template file

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+
+Customers deploy sidecars only for the services they run. Two services = two sidecar containers.
```
mayuka-c, line 230: Isnt it automatically gets deployed whenever the service is deployed?
Customers will not be deploying it manually right?

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+restructures a request), the tool just quietly changes shape the next time the sidecar happens to
+restart. No warning, nothing to check beforehand. An ISV's agent built around the old shape could
+break with no notice.
+
+Not something we need to fully solve for v1, but at minimum, we should log when a tool's shape
+changed between restarts, so it's visible instead of a silent surprise.
+
+### Keeping the Generated Config in Sync
+
+`application mcp config` is a manual, on-demand command today. This bites in both directions:
```
mayuka-c: How exactly this works with remote deployment?

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+
+- **Adding a service:** `application start rag --mcp` brings the new sidecar up automatically:
```
mayuka-c: Isnt side car container always running?.
If not what is its lifecycle?

mayuka-c: Didnt quite get the use of this if its running always

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+  `OPENAPI_URL`/`SERVICE_URL` are resolved from the application's own known config, no manual env
+  vars. But the agent host's `~/mcp.json` still only lists whatever was there before. The new
+  capability is invisible to the agent until someone remembers to re-run `mcp config` and hand the
+  updated file over again.
+- **Removing a service:** the mirror problem. See Sidecar Lifecycle on Delete below.
+
+**Recommendation:** have `application start --mcp` and `application mcp stop` auto-regenerate
```
mayuka-c: Again start, stop doesnt have any APIs today, so how exactly would it work for remote deployments?

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+`~/mcp.json` as part of the same command, rather than leaving regeneration as a separate manual
+step. This resolves both directions with one design decision instead of two, and removes a step a
+customer could otherwise easily forget.
+
+### Sidecar Lifecycle on Delete
+
+`application delete <name>` already performs a cascade delete: it lists every pod carrying that
```
mayuka-c: Cascade meaning?
If the side car container is part of same pod, it gets deleted because we delete the service pod.

How exactly would this side car run?

mayuka-c (commented):
These are my intial comments

Also can we have architecture diagram on how it works with low level designs?

I see lot of new commands being added and we need to see how all these works with remote deployment

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+### Files Modified (removing IBM Cloud-specific logic)
+
+| File | Change |
+|---|---|
+| `cmd/ibmcloud-api-mcp/main.go` | Remove IBM hostname validation; rename binary |
+| `internal/openapi/interface.go` | Remove IBM name-prefixing + region-server logic |
+| `internal/tool/provider.go` | Remove the `region` input parameter it adds to every tool |
+| `internal/server/http.go` | Swap IAM JWT validation for Bearer token passthrough; remove stdio entry point |
+| `internal/authenticator/` | Remove `cli.go`, `op.go`; keep `env.go`, `passthrough.go`, `api_key.go` |
+| `go.mod` | Rename module; drop `github.com/IBM/go-sdk-core/v5` |
```
mayuka-c, line 496: Where exactly does this exist and where it is removed?
Didnt quite get it

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+### Components Reused (verbatim, no changes)
+
+| File | What it does |
+|---|---|
+| `internal/openapi/loader.go` | Fetches + parses any `/openapi.json` |
+| `internal/openapi/convert.go` | Converts OpenAPI schemas → MCP tool JSON Schema |
+| `internal/tool/aggregator.go` | Tool registry + call routing |
+| `internal/tool/provider.go` | Per-operation HTTP executor → `SERVICE_URL` |
+| `internal/server/http.go` | HTTP transport, rate limiting, CORS, `/health` |
+| `internal/config/config.go` | Emits MCP client config JSON |
```
mayuka-c, line 485: Are these new files which we would be writing and implementing?

docs/proposals/mcp-sidecar-proposal.md
Bleama02: Correct, the sidecar only gets deployed if --mcp is passed when running application start. If it's not passed, no sidecar gets deployed for that application.

To answer your question above as well, as for how we know whether MCP is set during deployment, that's actually something Ryan asked as a gap during his look at the proposal. A thought is to add an mcp_endpoint column to the services table in the catalog database. If the column is null, no sidecar is running for that service. If it's populated, the sidecar is up and that's its address. That way the CLI always has a reliable way to check the MCP state per service. Let me know if you have any ideas.

docs/proposals/mcp-sidecar-proposal.md
Bleama02: The sidecar is not exposed externally, my understanding is that it sits inside the deployment environment and is only accessible by the agent running in that same environment.

docs/proposals/mcp-sidecar-proposal.md
Bleama02: The customer wouldn't be deploying it manually. When --mcp is passed to application start, the sidecar gets deployed automatically alongside the service. It only deploys sidecars for the services that are part of that customer's application, so if they only have two services, only two sidecars get deployed.

docs/proposals/mcp-sidecar-proposal.md
Bleama02: These are actually files from the existing ibmcloud-api-mcp codebase that @ryanmarc provided us that we're keeping as is, no changes needed. The whole point of starting from that codebase is that we don't have to write this from scratch, most of the core logic is already there and working.

docs/proposals/mcp-sidecar-proposal.md
Bleama02: The sidecar is not always running, its lifecycle is tied to the service it sits alongside.

docs/proposals/mcp-sidecar-proposal.md
noahmasoud: Yes, that's right, and it matches how the rest of the chart already works. backend-deployment.yaml hardcodes CHATBOT_SIMILARITY_SERVICE_URL as a literal string today, so the sidecar's OPENAPI_URL/SERVICE_URL would follow the same pattern in its own template rather than being dynamically resolved. Nothing new being introduced here, just extending the existing convention to the sidecar.

docs/proposals/mcp-sidecar-proposal.md
noahmasoud: We don't persist a flag anywhere. The idea is application ps gets a new MCP Endpoint column that reports live whether a sidecar is running for that pod, its URL or null, computed the same way ps already reports pod status. The running sidecar container is the state, there's nothing separate to keep in sync.

docs/proposals/mcp-sidecar-proposal.md
noahmasoud: Cascade delete sweeps every pod with the app's AppName label and force-deletes them, not per-resource, that part's unchanged. On placement, the sidecar should be its own pod, not a container merged into the service's pod. 

Reason: application mcp stop needs to kill just the sidecar while the service keeps running, and that's not clean if they share a pod. A separate pod with the same label still gets swept by cascade delete for free.

docs/proposals/mcp-sidecar-proposal.md
noahmasoud: These files exist today in the _ibmcloud-api-mcp_ repo. Phase 1 forks that into a new mcp/ directory here. The table is what changes after the fork, original repo untouched.

docs/proposals/mcp-sidecar-proposal.md
noahmasoud: There's no remote-deploy API today, confirmed that in the code. On openshift, application create runs helm install/upgrade and blocks until pods are ready, so creation already is the running state. That's why start/stop are no-ops there, not because openshift isn't supported. Given that, my thought is --mcp should hook into create for openshift instead of start, and mcp stop would become a helm upgrade toggling the sidecar off. But this touches a gap that exists independent of this proposal, so I'd rather get your take before locking something in

is that the right approach, or is there already a plan for how start/stop should work on openshift that this should line up with instead

docs/proposals/mcp-sidecar-proposal.md
noahmasoud: yeah, I can work on a diagram for this. Do you wnat something now or after the final proposal?

dharaneeshvrd: Generally it would be part of the proposal, If you add one, that would be great.

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+written for Swagger UI. This proposal surfaces that existing documentation as callable MCP tools
+with no changes to any service.
+
+## Proposed Solution
+
+A single generic MCP sidecar container image, deployed once per service, configured entirely by
+environment variables. At startup it fetches the service's `/openapi.json`, registers one MCP tool
+per API operation, and listens for agent connections. The agent calls tools; the sidecar forwards
+requests to the real service. **Most of the core logic needs no new code.** This is a targeted
+adaptation of the existing `ibmcloud-api-mcp` repository, not a from-scratch build. That said, this
```
dharaneeshvrd: can you please provide the link to the repository?

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+is a starting point, not a constraint: where our services' behavior doesn't fit what that repo
+already handles (see Async Operations below), writing new logic is the right call, not something
```
dharaneeshvrd: > where our services' behavior doesn't fit what that repo
already handles (see Async Operations below), writing new logic is the right call

where are we planning to implement this?

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,424 @@
...
+### Request Flow
+
+```
+AI Agent  (any MCP-compatible client)
+  │
+  │  MCP tools/list  →  receives typed tool definitions + descriptions
+  │  MCP tools/call  →  calls tool with structured arguments
+  ▼
+MCP Sidecar  (one per service, generic image, env-var config)
+  │  startup: fetches OPENAPI_URL → parses spec → registers tools
```
dharaneeshvrd: If I understand this correctly, we would use the ibmcloud-api-mcp tool to parse the openapi spec and dynamically generate tools at runtime and register them.
please correct me if I m wrong

docs/proposals/mcp-sidecar-proposal.md
dharaneeshvrd: Is the sidecar going to be the existing `ibmcloud-api-mcp` tool?
how are we planning to generate the source image for this sidecar? is it already available for power?

Bleama02: The sidecar is adapted from the ibmcloud-api-mcp codebase that Ryan gave us, so not used directly as is but forked and modified to remove the IBM Cloud specific parts that don't apply to our setup.

As for the image, it's not available for Power yet. I believe that is what Phase 2 of the proposal covers, writing the Containerfile with ppc64le support for Power/Spyre and building it locally before pushing to ICR.

docs/proposals/mcp-sidecar-proposal.md
noahmasoud: Ok, I was not aware of any conventions on our team @dharaneeshvrd

docs/proposals/mcp-sidecar-proposal.md
Bleama02: That's correct. At startup the sidecar fetches the service's /openapi.json, parses it, and dynamically registers one MCP tool per operation. The core logic for that comes from ibmcloud-api-mcp with the IBM Cloud-specific pieces removed.

docs/proposals/mcp-sidecar-proposal.md
ryanmarc: The repo for the base mcp code I wrote is currently private under https://github.com/IBM-Cloud/ibmcloud-api-mcp. I gave you read access to it, but I can not make the repo public.

docs/proposals/mcp-sidecar-proposal.md
ryanmarc: One thing to consider about this is that it will register every endpoint as a tool - The MCP servers should only be for end user consumption and not the management of the services (at least that's my understanding.)

The ibmcloud-api-mcp code does allow for filtering the endpoints that have tools generated by using a tag in the openapi spec though - so we may want to identify specifically which endpoints on each service should actually have tools and create tags in the openapi docs so we can target just those endpoints for tool creation.

docs/proposals/mcp-sidecar-proposal.md
ryanmarc: We should probably discuss this further - the mcp server would likely need to be exposed so that agents can connect to it. I don't think that agents are going to be running on the same machine / environment.

docs/proposals/mcp-sidecar-proposal.md
ryanmarc: Looks like these tag suggestions are already added around line 95 here.

docs/proposals/mcp-sidecar-proposal.md
ryanmarc: @mayuka-c by "remote deployment" do you mean the changes being made for the PMC integration where things will be deployed to workers?

If so, I think it would be a good idea to get a meeting setup to discuss this and make sure things are coordinated and that planning for the MCP work is able to account for what that look like.

docs/proposals/mcp-sidecar-proposal.md
ryanmarc: Maybe i'm missing something here - why couldn't the sidecar just be a container inside the service pod? Does it really have to be its own pod?

docs/proposals/mcp-sidecar-proposal.md
ryanmarc: The existing mcp repo is written in Go, it can easily be complied for power.

docs/proposals/mcp-sidecar-proposal.md
Bleama02: You're right, I had the wrong understanding. The sidecar does need to be externally reachable since the agent connects over the network. We have an open question in the proposal around securing that exposure that will need to be discussed though.

docs/proposals/mcp-sidecar-proposal.md
dharaneeshvrd: > The existing mcp repo is written in Go, it can easily be complied for power.

Ok fine then

Keeping the thread unresolved since @mayuka-c started it.

docs/proposals/mcp-sidecar-proposal.md
dharaneeshvrd: @noahmasoud Are you planning to create one?
Seems orchestration team wants a meeting to discuss this further, it would be easy to explain over a diagram something like https://github.com/IBM/project-ai-services/blob/main/docs/proposals/docling_conversion_queue.md#11b-full-lifecycle--digitization-1-file

mayuka-c:
As discussed in today's call,
lets have mcp sidecar running as a side-car container in the same service pods. Let it be long running container and let us introduce subcommands of mcp in future releases (if required).
cc. @ryanmarc @dharaneeshvrd @yussufsh

docs/proposals/mcp-sidecar-proposal.md
dharaneeshvrd: I see this line still exists, which means we would have some logic specific to services inside mcp/ dir at root level.

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,682 @@
...
+
+| File | Purpose |
+|---|---|
+| `Containerfile` | Renamed + adapted from the upstream `Dockerfile` (which already exists and works, amd64/Alpine), retargeted to UBI9 base and `ppc64le` for Power/Spyre, IBM-specific `IAM_ENDPOINT` env default dropped |
+| `.golangci.yml` | Same linters as `ai-services/.golangci.yml` |
+| `Makefile` | Adapted upstream Makefile, `ppc64le` cross-compile target |
+
+### CLI additions to `ai-services`
+
+- **`application start`:** modified to automatically launch the MCP sidecar container alongside
```
dharaneeshvrd: We discussed not to touch any of the commands. Do we still need this?

dharaneeshvrd (commented):
Since we decided to make only manifest changes to deploy this as a sidecar, It would be good to have those detailed in the proposal.

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,682 @@
...
+  each service. No new flag required.
+- **`application stop`:** modified to always tear the sidecar down alongside the service.
+- **`application info`:** add `--format mcp-json` and `--output <path>` flags to the existing
+  command, reusing its existing pod-resolution logic to emit the `mcpServers` JSON instead of
+  human-readable text.
+- **`application ps`:** add `MCP ENDPOINT` column to output, reporting the sidecar address for
+  each service or null if no sidecar is running.
+
+### Note on OpenAPI Description Quality
```
dharaneeshvrd, line 557: Agree, currently all the APIs have valid summary and description, but in case any improvement or minor adjustments needed for these descriptions, hope that would be taken care part of this effort :)

dharaneeshvrd (commented):
I can give an approval from my side once we take care of the manifest related changes detailed and cli changes reverted.
Other things lgtm.

docs/proposals/mcp-sidecar-proposal.md
Bleama02: I was under the impression that there was still work that needs to be done in the cli. After talking to Ryan, i see that is not the case and will update by the end of the day.

docs/proposals/mcp-sidecar-proposal.md
Bleama02: @dharaneeshvrd, Changes have been made, which is, taking out cli changes and adding manifest related changes

dharaneeshvrd (dismissed):
@mayuka-c 
Please take a look at the CLI-related things once

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,664 @@
...
+to a future release (see Future Work).
+
+### OpenShift: a different attach point
+
+Podman is the primary target for v1 (see Delivery Phases), but the automatic attach point above
+doesn't carry over to OpenShift as-is, worth stating directly rather than implying one mechanism
+covers both runtimes.
+
+On OpenShift, `application start`/`application stop` are no-ops today, they log and return without
+doing anything. `application create` is what actually deploys and waits for a running app, via
```
mayuka-c: Even for podman it will be the same, application create would be the one actually deploying

docs/proposals/mcp-sidecar-proposal.md
```diff
@@ -0,0 +1,664 @@
...
+The sidecar runs in **HTTP transport mode only** (`--http` flag). Stdio transport — the default in
+the upstream binary — requires the MCP client to fork the process directly, which is not viable for
+a containerised sidecar. HTTP transport is the only mode that makes sense here: the sidecar binds
+to a port and the agent connects over the network. The stdio code path is removed entirely rather
+than left as an unused option.
+
+### MCP Sidecar Exposure via Caddy
+
+External agents reach the sidecar through Caddy. When `application start` deploys sidecars
+automatically, the CLI must also automatically register each sidecar route with Caddy's Admin API
```
mayuka-c: `application start` will not be deploying the pods!, as said application create already takes care of deploying the pod which also deploys sidecar containers

mayuka-c (commented):
Minor comments, Rest LGTM

docs/proposals/mcp-sidecar-proposal.md
Bleama02: Changes made

docs/proposals/mcp-sidecar-proposal.md
Bleama02: Changes made

mayuka-c (approved):
LGTM

PR #1180, opened 2026-07-24
Implement resources API for openshift

prachikurade1:


prachikurade1:
Dev testing screenshots:
Missing auth:
missing_token
Invalid auth:
invalid_token
Cards available:
resources_api_openshift
Cards in use:
no_cards_available

ai-services/internal/pkg/runtime/openshift/openshift.go
```diff
@@ -462,14 +466,180 @@ func (kc *OpenshiftClient) VolumeExists(nameOrID string) (bool, error) {
...
+// (non-terminating) pods cluster-wide.
+func (kc *OpenshiftClient) sumSpyreInUse() (int64, error) {
+	podList, err := kc.KubeClient.CoreV1().Pods("").List(kc.Ctx, metav1.ListOptions{})
+	if err != nil {
+		return 0, fmt.Errorf("failed to list running pods: %w", err)
+	}
+
+	var used int64
+
+	for i := range podList.Items {
```
yussufsh: I am afraid this could take a long time to check for all the pods and containers under each of them.
What is the time taken in your tests?
Also, we need to think if there is a better way to do this.
cc @mayuka-c @Shubhamag12

prachikurade1: Approx 1-1.2 seconds. Will try to find some better mechanism to check for pods using spyre cards.

prachikurade1: Switched to Thanos query to fetch the spyre cards being used for pods.

yussufsh: is this well-tested?

yussufsh (dismissed):
lgtm
@prachikurade1 please bump up the version.
@mayuka-c @Shubhamag12 this works for me but please ack if you do not have an issue with this implementation.

ai-services/assets/catalog/openshift/templates/rbac.yaml
```diff
@@ -46,3 +50,19 @@ roleRef:
...
+kind: ClusterRoleBinding
+metadata:
+  name: ai-services-monitoring-view-binding
+subjects:
+  - kind: ServiceAccount
+    name: ai-services-sa
+    namespace: ai-services
+roleRef:
+  kind: ClusterRole
+  name: cluster-monitoring-view
```
Shubhamag12, line 67: one question, who creates this role? thanos?

also is thanos already part of cluster or do we deploy it separately?

Shubhamag12 (commented):
minor comment

ai-services/assets/catalog/openshift/templates/rbac.yaml
prachikurade1: The cluster-monitoring-view role is an in-built OCP role that grants read access to core platform monitoring metrics, Prometheus, and alerting APIs.
Thanos is already part of the cluster, not required to deploy separately. 
```
Thanos Querier and Thanos Ruler are available and deployed by default in the core monitoring stack of OpenShift Container Platform (OCP) clusters
```

prachikurade1:
> lgtm @prachikurade1 please bump up the version. @mayuka-c @Shubhamag12 this works for me but please ack if you do not have an issue with this implementation.

Bumped the version

ai-services/internal/pkg/runtime/openshift/thanos.go
```diff
@@ -0,0 +1,148 @@
...
+		return 0, fmt.Errorf("failed to parse thanos response: %w", err)
+	}
+
+	if result.Status != "success" {
+		return 0, fmt.Errorf("thanos returned non-success status %q: %s", result.Status, result.Error)
+	}
+
+	var total float64
+
+	for _, r := range result.Data.Result {
```
yussufsh, line 99: IMO you should check for `Data` and `Result` for `nil` or empty list (`[]`).
In such cases it means Thanos did not return proper values.

ai-services/internal/pkg/runtime/openshift/thanos.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("failed to build thanos request: %w", err)
+	}
+
+	token := loadThanosToken()
+	if token != "" {
+		req.Header.Set("Authorization", "Bearer "+token)
+	}
+
+	resp, err := buildThanosHTTPClient().Do(req)
```
yussufsh: This function `queryThanos` is called multiple times loading the token and building new client every time.

yussufsh: Forgot to mention that the token is read from FS every time. This definitely should be cached.

ai-services/internal/pkg/runtime/openshift/openshift.go
```diff
@@ -506,14 +509,136 @@ func (kc *OpenshiftClient) VolumeExists(nameOrID string) (bool, error) {
...
+		AvailableBytes: int64(availMem),
+	}
+
+	// --- Spyre cards ---
+	// Read directly from node capacity/allocatable so the count is accurate
+	// even when Thanos has not yet scraped the custom resource metric.
+	spyreInfo, err := kc.getSpyreCardInfo()
+	if err != nil {
+		// Non-fatal: log and return an empty accelerator map.
+		logger.ErrorfCtx(kc.Ctx, "Could not retrieve Spyre card info: %v", err)
```
yussufsh: We should fail here instead of giving a false result.

yussufsh (changes requested):
Also make use of Context for Thanos and logger.

ai-services/internal/pkg/runtime/openshift/openshift.go
prachikurade1: done

ai-services/internal/pkg/runtime/openshift/thanos.go
prachikurade1: done

ai-services/internal/pkg/runtime/openshift/thanos.go
prachikurade1: done

ai-services/internal/pkg/runtime/openshift/openshift.go
```diff
@@ -506,14 +509,135 @@ func (kc *OpenshiftClient) VolumeExists(nameOrID string) (bool, error) {
...
+		AvailableBytes: int64(availMem),
+	}
+
+	// --- Spyre cards ---
+	// Read directly from node capacity/allocatable so the count is accurate
+	// even when Thanos has not yet scraped the custom resource metric.
+	spyreInfo, err := kc.getSpyreCardInfo()
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve Spyre card info: %w", err)
+	} else if spyreInfo != nil {
```
yussufsh:
```suggestion
	} 
    if spyreInfo != nil {
```

yussufsh (commented):
minor

yussufsh (dismissed):

yussufsh (approved):

PR #1181, opened 2026-07-24
fix(catalog): resolve double logout/me API calls and minor housekeeping

pranithraoibm:
- Fix double logout api call: useSessionTimeout manual path now just navigates to /logout instead of calling logout() itself; the page owns the single API call
- Fix double /me call: added in-flight promise guards on logout() and getUserInfo() so concurrent calls (StrictMode double-effect fire) deduplicate to one request (observed this post vite 8.x upgrade hence the fix)
- Deprecation fix: replaced deprecated <Tag filter onClose> with <DismissibleTag> in CatalogBrowseLayout (observed this in the browser console when clicking on the filter checkboxes in the usecasereferences page)
- Cleanup: removed sleep(5000) delay after delete in DigitalAssistants as the intermittent error was observed only locally and works completely fine on the deployed UI, added session.constants to the constants barrel, updated direct imports to go through @/constants

- Added s to 'Digital Assistant' in the backend to match the digital assistant page heading to the navigation as requested [here](https://ibm-systems.slack.com/archives/C0ACKMZ3L0G/p1784121142617599)


ui/catalog/src/hooks/useSessionTimeout.ts
```diff
@@ -70,24 +70,28 @@ export const useSessionTimeout = (): UseSessionTimeoutReturn => {
         showWarningRef.current = false;
       }
 
+      if (logoutType === "manual") {
```
ryanmarc: performLogout("manual") is only reachable via handleManualLogout → the returned handleLogout (line 244), and nothing consumes it — SessionManager.tsx destructures only {showWarning, timeRemaining, extendSession}, and the logout button in AppHeader.tsx has its own handler that navigates to /logout directly. I checked the merge base too, so it was already dead before this PR, which means the double-/logout fix is entirely the guard in auth.ts. Since this PR is doing housekeeping anyway, could we either drop handleManualLogout / handleLogout / the logoutType parameter from the hook (performLogout becomes the auto path), or point AppHeader at the hook so there's one logout entry point? Right now the diff reads like the fix and the invariant in the comment is documenting something nothing exercises.

ui/catalog/src/services/auth.ts
```diff
@@ -32,37 +32,64 @@ export const login = async (payload: LoginRequest): Promise<LoginResponse> => {
...
-    // Clear all service deploy store data
-    useServiceDeployStore.getState().clearAllCache();
+export const logout = async (): Promise<void> => {
+  if (logoutPromise) {
+    return logoutPromise;
   }
+
+  const refreshToken = useAuthStore.getState().refreshToken;
+
+  logoutPromise = (async () => {
```
ryanmarc: Two things here.

(1) Ordering: the finally block runs synchronously inside the async body, so if the IIFE settles before its first await (a synchronous throw out of api.post), logoutPromise = null fires before the assignment on this line and the guard is left holding a permanently-rejected promise, after which every logout() short-circuits at line 38 and never clears tokens again. Narrow with axios, but it costs nothing to avoid.

(2) We already have this helper: dedupe() in src/utils/requestManager.ts does exactly this and gets the ordering right by using .finally() (a microtask, so it can't race the set). This PR would be the third copy of the pattern, api/axios.ts has one for token refresh too.

ui/catalog/src/pages/DigitalAssistants/DigitalAssistants.tsx
```diff
@@ -208,8 +206,6 @@ const DigitalAssistantsPage = () => {
     try {
       await deleteApplication(state.selectedRowId);
       dispatch({ type: ACTION_TYPES.CLOSE_DELETE_DIALOG });
-      // Await the delayed reload to ensure error handling
-      await sleep(5000);
```
ryanmarc, line 212: Agreed that await sleep(5000) had to go, it pinned the modal's Delete button in the disabled "Deleting…" state for five seconds. But it was papering over a real gap: deletion is async server-side (hence the "Deleting..." row status), so this immediate refetch usually still returns the row, and the only thing that refreshes afterwards is the 2-minute interval, which is additionally gated on state.rowsData.length > 0. Net effect for the user: confirm delete → dialog closes → row stays put for up to 120s. Could we add a short poll while any row is in a transitional state? Something like:

```
useEffect(() => {
  if (!state.rowsData.some(r => r.status === "Deleting..." || r.status === "Deploying...")) return;
  const id = setInterval(loadApplications, 5000);
  return () => clearInterval(id);
}, [state.rowsData]);
```

That covers deploys too, and lets the 2-minute interval stay as the idle refresh.

ryanmarc (changes requested):

ui/catalog/src/hooks/useSessionTimeout.ts
pranithraoibm: Dropped handleManualLogout, handleLogout, and the logoutType param, AppHeader already handles manual logout independently.

ui/catalog/src/services/auth.ts
pranithraoibm: Replaced both logout and getUserInfo with dedupe() from requestManager

ui/catalog/src/pages/DigitalAssistants/DigitalAssistants.tsx
pranithraoibm: Added a 5 s poll that fires whenever any row is in a "Deploying..." or "Deleting..." state and clears itself once all rows resolve. For now the status values are hardcoded but as @Niranjan-J-Joshi is already working on the table code refactoring, asked him to remove these hardcoded values as and when he is working on the status related changes.

pranithraoibm:
@ryanmarc I have addressed all the review comments and added a small change to address [this](https://ibm-systems.slack.com/archives/C0ACKMZ3L0G/p1784121142617599) as well, PTAL

ryanmarc (approved):

PR #1182, opened 2026-07-27
[Design Proposal] Cross-VM Service communication with mTLS using Caddy

mayuka-c:
- Enabling cross service VM communication (between control plane and services on worker nodes) using Caddy as reverse_proxy.
- Uses mTLS for secure communication.
- Defines public, private ingress and egress blocks for proxy.


mayuka-c:
Moving this to draft as this will be work for 4Q26

PR #1183, opened 2026-07-27
fix npm CVEs and address lint issues in catalog

pranithraoibm:
- brace-expansion DoS — caused by eslint@9 pulling in minimatch@3 → fixed by upgrading to eslint@10 which uses safe minimatch@10. `eslint-plugin-react` had to be removed as it doesn't support eslint@10 yet and CVEs would be present if kept. eslint@10 support addition is tracked here [#3979](https://github.com/jsx-eslint/eslint-plugin-react/pull/3979), so once it gets merged we can add it back to our repo.

- react-router CSRF + 3 other CVEs — no patched v7 release exists, fix only in v8 → migrated from `react-router-dom@7` to `react-router@8.3.0`, updated all imports accordingly

- Synced all package.json declared versions to match package-lock.json

- Fixed lint errors surfaced by stricter eslint-plugin-react-hooks@7.1.1 rules

- Verified all the changes but running `npm i`, `npm audit`, `npm run fix && npm run check` and `npm run build`

- Once this PR is approved will raise the same fix for `digitize` and `chatbot` repos as well


ui/catalog/eslint.config.js
```diff
@@ -1,6 +1,5 @@
 import js from '@eslint/js';
 import tseslint from 'typescript-eslint';
-import reactPlugin from 'eslint-plugin-react';
```
ryanmarc, line 3: This PR says it addresses lint issues. Is it addressing them by deleting the use of the react plugin? This removed several checks.

Why is this being removed?

ui/catalog/src/hooks/useSessionTimeout.ts
```diff
@@ -115,8 +115,9 @@ export const useSessionTimeout = (): UseSessionTimeoutReturn => {
 
   useEffect(() => {
     if (showWarning) {
-      updateCountdown();
-      countdownTimerRef.current = setInterval(updateCountdown, 1000);
+      const tick = () => updateCountdown();
+      tick();
+      countdownTimerRef.current = setInterval(tick, 1000);
```
ryanmarc: tick() is the same call as updateCountdown(), it just hides it from react-hooks/set-state-in-effect. The cascading render the rule is warning about is still there, minus the warning. This PR already uses the honest form in DeploymentDetails.tsx:71; could we do the same here?

ui/catalog/eslint.config.js
pranithraoibm: Explained the reason behind removal in the first point of the PR description

ui/catalog/package.json
```diff
@@ -16,31 +16,31 @@
...
+    "@eslint/js": "^10.0.1",
+    "@types/node": "^24.12.0",
+    "@types/react": "^19.2.14",
     "@types/react-dom": "^19.2.3",
     "@vitejs/plugin-react": "^5.2.0",
-    "eslint": "^9.39.1",
-    "eslint-plugin-react": "^7.37.5",
-    "eslint-plugin-react-hooks": "^7.0.1",
-    "eslint-plugin-react-refresh": "^0.4.24",
+    "esbuild": "^0.28.1",
```
ryanmarc: Is this needed? Nothing in the repo imports esbuild directly, and both main's lockfile and this branch's resolve a single hoisted node_modules/esbuild@0.28.1 (only dependent: vite, range ^0.27.0 || ^0.28.0), so the resolved tree is identical with and without it. If the goal is to floor a transitive version, overrides is the mechanism that actually guarantees it (there's already an overrides block just below for react-table); as a direct devDependency it's a phantom entry that nobody will remember to keep current.

ui/catalog/src/pages/Login/Login.tsx
```diff
@@ -30,30 +30,25 @@ const LoginPage = () => {
   const [credentialError, setCredentialError] = useState<boolean>(false);
   const [networkError, setNetworkError] = useState<boolean>(false);
   const [loading, setLoading] = useState<boolean>(false);
+
+  const locationState = location.state as LoginLocationState | null;
+  const isInactivityLogout =
```
ryanmarc: Two things worth noting alongside the set-state-in-effect fix.

(1) sessionStorage.getItem now runs on every render, and the effect on line 44 removes that key, so the component reads mutable external state during render, which is the pattern the react-hooks 7 rules are steering us away from. A lazy initializer (useState(() => isInactivityLogout)) reads it once at mount and keeps render pure. 

(2) The banner is now decided only at mount, whereas the old effect could turn it on when location changed. I couldn't find a live path that regresses (we only ever land on /login fresh), but a one-line comment saying "mount-only by design" would stop someone re-introducing the effect later.

ryanmarc (changes requested):

ui/catalog/eslint.config.js
pranithraoibm: @ryanmarc Do let me know if this is fine or if I need to follow any other approach...

ui/catalog/src/hooks/useSessionTimeout.ts
pranithraoibm: Done

ui/catalog/package.json
pranithraoibm: Removed it. I guess it was added because when I ran npm install multiple times during the eslint/react-router upgrade iterations, esbuild got dropped from node_modules and the build failed with `Cannot find package 'esbuild'`. Checked fresh run of `npm install` after removing it now and it succeeded.

ui/catalog/src/pages/Login/Login.tsx
pranithraoibm: Moved sessionStorage.getItem into a lazy initializer so it runs once at mount and cleans up in the same call. Added the mount-only comment to make the intent explicit.

pranithraoibm:
@ryanmarc I have addressed all the review comments PTAL

pranithraoibm:
npm CVEs for all the UI folders are handled in [this](https://github.com/IBM/project-ai-services/pull/1207) PR now hence closing this

PR #1185, opened 2026-07-28
refactor : Create reusable components for common table functionalities

Niranjan-J-Joshi:
**Implementation:**

* `table/components/TableToolbarActions.tsx` — toolbar with search, refresh, export, and column visibility
* `table/components/CellRenderers.tsx` — `ActionCell`, `NameCell`, `StatusCell`, `MessageCell`, and the merged `STATUS_CONFIG` map with keys matching exact API-returned strings (no ellipsis)
* `table/components/ExportModal.tsx` — CSV filename input modal with text input and inline validation error
* `table/components/DeleteModal.tsx` — delete confirmation modal with warning message and confirmation checkbox
* `table/components/TableToasts.tsx` — renders both the delete-error and export status
* `table/components/TableEmptyStates.tsx` — renders the no-data, no-search-results, and fetch-error empty states


ui/catalog/src/components/Table/components/CellRenderers.tsx
```diff
@@ -0,0 +1,177 @@
...
+  return (
+    <div className={sharedStyles.messageWithIcon}>
+      <MessageIcon size={16} className={iconClassName} />
+      <span className={sharedStyles.messageText}>{message}</span>
+    </div>
+  );
+};
+
+export const ActionCell = ({ rowId, rowData, onDelete }: ActionCellProps) => {
+  const isDeleteEnabled =
```
ryanmarc: This is the rule #1198 is currently replacing. That PR changes both call sites to allow delete during mid-deployment states: DA becomes rowData?.status !== "Deleting" and services becomes rowData?.status?.toLowerCase() !== "deleting...".

Because the two PRs touch different files, nothing catches the collision. git merge-tree is clean in both orders and main + 1185 type-checks fine. The revert only surfaces later, in whichever PR swaps the per-table ActionCell for this one, and by then it will read as a mystery regression rather than a merge-order problem.

Rather than copying #1198's rule in here (which would leave the same trap if the rule changes again), I would make the predicate a prop, which this component needs anyway to serve two tables that use different status strings:

`isDeleteEnabled?: (status: string | undefined) => boolean;`

with the current behavior as the default. Then the adoption PR passes each table's own rule and #1198 survives.

ryanmarc (changes requested):

ui/catalog/src/components/Table/components/CellRenderers.tsx
Niranjan-J-Joshi: Added isDeleteEnabled prop to shared ActionCell component.

ui/catalog/src/components/Table/components/CellRenderers.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  return (
+    <div className={sharedStyles.messageWithIcon}>
+      <MessageIcon size={16} className={iconClassName} />
+      <span className={sharedStyles.messageText}>{message}</span>
+    </div>
+  );
+};
+
+const defaultIsDeleteEnabled = (status: string | undefined) =>
```
ryanmarc: Thanks for adding the prop, that is the right shape. One catch though: the default still carries the old rule, so this does not resolve the conflict with #1198 yet.

`defaultIsDeleteEnabled is status === "Running" || status === "Error"`, which is verbatim the condition #1198 replaces. #1198 goes the other way: `rowData?.status?.toLowerCase() !== "deleting..."`, so delete is enabled for every status except a row that is already deleting. Its description puts it as "enabled for all the statuses except deleting itself."

So any table that adopts ActionCell without passing isDeleteEnabled quietly gets the pre-#1198 behaviour back. Nothing consumes these components yet, so this is a trap for whoever does the adoption rather than a live bug, but the default is exactly what they will hit if they are not looking for it.

Two ways out:

1. Flip the default: `const defaultIsDeleteEnabled = (status?: string) => status?.toLowerCase() !== "deleting...";`
2. Drop the default and make `isDeleteEnabled` required, so each table states its own rule at the call site.

I would lean towards 2. The DA page and the services table genuinely disagree about this, and an implicit default is what let the mismatch go unnoticed in the first place.

ryanmarc (changes requested):

ui/catalog/src/components/Table/components/CellRenderers.tsx
Niranjan-J-Joshi: Removed default and made isDeleteEnabled required, so now each table passes its own delete eligibility rule.

ryanmarc (changes requested):
this fails `npm run format:check` now.

ryanmarc (dismissed):

pranithraoibm (approved):

PR #1186, opened 2026-07-28
Initial PR for the db schema, models and skeleton

Niharika0306:
Translation service Scaffold, DB Schema & Container.

[Jira link](https://jsw.ibm.com/browse/AISERVICES-1458)

### DIFF
--- a/services/translate/db/scripts/init_schema.sql
+++ b/services/translate/db/scripts/init_schema.sql
@@ -1,0 +1,49 @@
+-- Translation service — idempotent schema initialization
+-- Safe to run multiple times (CREATE IF NOT EXISTS / CREATE OR REPLACE throughout).
+
+CREATE TABLE IF NOT EXISTS translate_jobs (
+    job_id              VARCHAR(255) PRIMARY KEY,
+    job_name            VARCHAR(500),
+    source_language     VARCHAR(100) NOT NULL DEFAULT 'auto',
+    target_language     VARCHAR(100) NOT NULL,
+    input_type          VARCHAR(20)  NOT NULL,
+    document_name       VARCHAR(500),
+    document_word_count INTEGER,
+    status              VARCHAR(50)  NOT NULL,
+    submitted_at        TIMESTAMP WITH TIME ZONE NOT NULL,
+    completed_at        TIMESTAMP WITH TIME ZONE,
+    error               TEXT,
+    job_metadata        JSONB,
+    updated_at          TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    CONSTRAINT chk_translate_job_status
+        CHECK (status IN ('accepted','in_progress','completed','failed')),
+    CONSTRAINT chk_translate_input_type
+        CHECK (input_type IN ('text','txt','md'))
+);
+
+-- Composite index: job listing with status filter and boot-time zombie scan
+CREATE INDEX IF NOT EXISTS idx_translate_jobs_submitted_at_status
+    ON translate_jobs(submitted_at DESC, status);
+
+-- Trigger function to keep updated_at current on every UPDATE
+CREATE OR REPLACE FUNCTION update_translate_jobs_updated_at_column()
+RETURNS TRIGGER AS $$
+BEGIN
+    NEW.updated_at = CURRENT_TIMESTAMP;
+    RETURN NEW;
+END;
+$$ language 'plpgsql';
+
+-- Conditionally create the trigger (idempotent; requires PostgreSQL 14+ for IF NOT EXISTS on triggers)
+DO $$
+BEGIN
+    IF NOT EXISTS (
+        SELECT 1 FROM pg_trigger WHERE tgname = 'update_translate_jobs_updated_at'
+    ) THEN
+        CREATE TRIGGER update_translate_jobs_updated_at
+            BEFORE UPDATE ON translate_jobs
+            FOR EACH ROW
+            EXECUTE FUNCTION update_translate_jobs_updated_at_column();
+    END IF;
+END
+$$;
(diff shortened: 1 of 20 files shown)


services/translate/workers/concurrency.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __init__(self) -> None:
+        self._job_limiter = asyncio.BoundedSemaphore(
+            settings.translate.max_concurrent_jobs
+        )
+        self._chunk_semaphore = asyncio.BoundedSemaphore(
+            settings.translate.chunk_parallelism
+        )
+        self._vllm_semaphore = asyncio.BoundedSemaphore(
+            settings.common.llm.max_batch_size
```
manalilatkar, line 50: asyncio.BoundedSemaphore() is instantiated inside ConcurrencyManager.__init__, which runs at module import time (line 36) — before any event loop exists. In Python 3.10+, asyncio.BoundedSemaphore() no longer auto-attaches to a running loop at creation time, but it will bind to the loop on first acquire(). If uvicorn creates a different loop after import (which it does with --loop uvloop as specified in the Containerfile), the semaphore may be bound to the wrong loop or to no loop, causing RuntimeError: ... attached to a different loop.
Fix: Add another method called initialize in which you perform steps in line 27 to 34 and call it in lifespan of app.py.

manalilatkar (commented):
1 minor thing. rest LGTM.

services/translate/workers/concurrency.py
Niharika0306: Addressed.

manalilatkar:
@Niharika0306 rebase and I'll approve.

dharaneeshvrd:
I m looking too, can we please hold?

services/translate/db/scripts/init_schema.sql
```diff
@@ -0,0 +1,49 @@
...
+-- Safe to run multiple times (CREATE IF NOT EXISTS / CREATE OR REPLACE throughout).
+
+CREATE TABLE IF NOT EXISTS translate_jobs (
+    job_id              VARCHAR(255) PRIMARY KEY,
+    job_name            VARCHAR(500),
+    source_language     VARCHAR(100) NOT NULL DEFAULT 'auto',
+    target_language     VARCHAR(100) NOT NULL,
+    input_type          VARCHAR(20)  NOT NULL,
+    document_name       VARCHAR(500),
+    document_word_count INTEGER,
```
dharaneeshvrd: I feel word count can be part of metadata itself, no need for dedicated field here.

services/translate/settings.py
```diff
@@ -0,0 +1,138 @@
...
+    """Translation-specific settings."""
+
+    # Cache / storage
+    cache_dir: Path = Field(
+        default=Path("/var/cache/translate"),
+        description="Base cache directory for staging and results",
+    )
+
+    # Context-window budget
+    chunk_token_budget: int = Field(
```
dharaneeshvrd, line 23: this needs to be dynamic, since we would be supporting other models like granite 4h small from watsonx and new models from spyre this release.

services/translate/settings.py
```diff
@@ -0,0 +1,138 @@
...
+        ge=1,
+        description=(
+            "Maximum input tokens per translation chunk (async path). "
+            "Default is 13107 (≈ 40% of 32768, the granite-3.3-8b-instruct context window). "
+            "Can be overridden via CHUNK_TOKEN_BUDGET env var. "
+            "Updated at runtime if a different MAX_MODEL_LEN is detected."
+        ),
+    )
+
+    prompt_overhead_tokens: int = Field(
```
dharaneeshvrd, line 34: let's compute at runtime and use it.

services/translate/db/scripts/init_schema.sql
Niharika0306: Can do that. Though I'm having second thoughts on this. 
Computing the word count has never been accurate in our services. So since we anyways have input and output tokens count in the metadata, wondering if adding this is needed? cc @dharaneeshvrd @manalilatkar

dharaneeshvrd: I m ok with it, you can decide.

services/translate/settings.py
Niharika0306: Right. So this value was just added as a fallback. The true value will be dynamically fetched in the job implementation. 
Is updated in the description:

`        description=(
            "Maximum input tokens per translation chunk (async path). "
            "Default is 13107 (≈ 40% of 32768, the granite-3.3-8b-instruct context window). "
            "Can be overridden via CHUNK_TOKEN_BUDGET env var. "
            "Updated at runtime if a different MAX_MODEL_LEN is detected."`

services/translate/db/scripts/init_schema.sql
Niharika0306: okay, removing it then. Thanks!

services/translate/settings.py
Niharika0306: Yes this will be computed once the fastapi cycle starts, in the implementation PR next, once the prompt is also available. This is just a default value.

services/translate/models.py
```diff
@@ -0,0 +1,146 @@
...
+    )
+    target_language: str = Field(
+        ..., description="Target language name (e.g. 'English'). Must not be 'auto'."
+    )
+
+
+class SyncTranslateResponse(BaseModel):
+    """Response body for POST /v1/translate (200)."""
+
+    data: Dict[str, Any]   # translation, source_language, target_language, word counts
```
manalilatkar: @Niharika0306 now that we have removed document_word_count, will this data variable still contain word count as mentioned in the end?

Niharika0306: No, it won't. Changed it here, as well as in async job result.

manalilatkar (approved):
LGTM

PR #1193, opened 2026-07-29
[Design proposal] mtls bootstrapping of agents with the gateway

hasan4791:
- Enabling secure communication between worker lpars and control plane agents

### DIFF
--- a/docs/proposals/catalog/mtls-bootstrapping-proposal.md
+++ b/docs/proposals/catalog/mtls-bootstrapping-proposal.md
@@ -1,0 +1,358 @@
+# Agent Gateway mTLS Bootstrapping — Design Proposal
+
+**Version:** 1.0
+**Date:** July 2026
+**Status:** Draft / Proposal
+
+---
+
+## Table of Contents
+
+1. [Overview](#1-overview)
+2. [Goals](#2-goals)
+3. [Architecture](#3-architecture)
+   - [3.1 Component Roles](#31-component-roles)
+   - [3.2 Ports and Listeners](#32-ports-and-listeners)
+   - [3.3 gRPC Service Definition](#33-grpc-service-definition)
+4. [Certificate Authority and PKI](#4-certificate-authority-and-pki)
+   - [4.1 Root CA](#41-root-ca)
+   - [4.2 Gateway Server Certificate](#42-gateway-server-certificate)
+   - [4.3 Agent Client Certificate](#43-agent-client-certificate)
+   - [4.4 Control Plane PKI Bootstrap](#44-control-plane-pki-bootstrap)
+5. [Agent Bootstrap — Certificate Provisioning](#5-agent-bootstrap--certificate-provisioning)
+   - [5.1 Bootstrap Flow](#51-bootstrap-flow)
+   - [5.2 Credential Validation on Restart](#52-credential-validation-on-restart)
+6. [TLS Configuration Details](#6-tls-configuration-details)
+   - [6.1 Gateway Server (Control Plane)](#61-gateway-server-control-plane)
+   - [6.2 Agent Client (Worker Daemon)](#62-agent-client-worker-daemon)
+7. [Hybrid TLS — Bootstrap vs mTLS Mode](#7-hybrid-tls--bootstrap-vs-mtls-mode)
+8. [Key Design Decisions](#8-key-design-decisions)
+9. [Security Considerations](#9-security-considerations)
+
+---
+
+## 1. Overview
+
+This proposal describes how the **Agent Gateway** (control plane) and **agent daemons** (worker LPARs) establish mutual TLS (mTLS) for all gRPC communication.
+
+There are two distinct phases:
+
+- **Bootstrap (first contact):** A new agent daemon has no certificate yet. It connects to the gateway using a single-use pre-shared token, submits a CSR, and receives both a CA-signed client certificate and the server's `ca.crt` in return. This first connection uses TLS (server-only) with an optional CA verification fallback.
+- **Operational (post-bootstrap):** All subsequent connections — including the long-lived `CommandStream` — use strict mTLS. The gateway requires and verifies the agent's client certificate. The agent verifies the gateway's server certificate against `ca.crt`.
+
+---
+
+## 2. Goals
+
+| Goal | Approach |
+|---|---|
+| Encrypted agent-to-gateway communication | TLS on all gRPC connections; mTLS after bootstrap |
+| Authenticated agents only on `CommandStream` | Client certificate required and verified against CA pool |
+| Zero pre-distributed secrets except token | Agent generates its own key locally; only a short-lived pre-shared token is distributed |
+| Credential freshness on every restart | `hasValidTLSCredentials` checks expiry and probes the gateway live before skipping re-registration |
+| Cert lifecycle tied to agent onboarding | Bootstrap generates ECDSA key + CSR; gateway signs and delivers the cert |
+| Fail-safe token guard | Registration is rejected immediately if no valid cert exists and no token is provided |
+
+---
+
+## 3. Architecture
+
+### 3.1 Component Roles
+
+```mermaid
+graph TB
+  subgraph CP["Control Plane (VM-CP)"]
+    GW["AgentGateway\n:9090 gRPC"]
+    CA["Internal CA\n(caCert + caKey)"]
+    TS["TokenStore\n(pre-shared tokens)"]
+    REG["Registry\n(agent state)"]
+    GW --> CA
+    GW --> TS
+    GW --> REG
+  end
+
+  subgraph W1["Worker LPAR 1"]
+    AD1["Agent Daemon"]
+    TLS1["/etc/ai-services/agent-tls/\ntls.crt · tls.key · ca.crt"]
+    AD1 --> TLS1
+  end
+
+  subgraph W2["Worker LPAR 2"]
+    AD2["Agent Daemon"]
+    TLS2["/etc/ai-services/agent-tls/\ntls.crt · tls.key · ca.crt"]
+    AD2 --> TLS2
+  end
+
+  W1 -- "gRPC mTLS (:9090)" --> CP
+  W2 -- "gRPC mTLS (:9090)" --> CP
+```
+
+### 3.2 Ports and Listeners
+
+| Port | Purpose | TLS Mode |
+|------|---------|----------|
+| `:9090` | Agent Gateway gRPC — `Register` + `CommandStream` | Hybrid: TLS-only for `Register` (bootstrap), strict mTLS for `CommandStream` |
+
+### 3.3 gRPC Service Definition
+
+```protobuf
+service AgentGateway {
+  // Called once at bootstrap — TLS only (no client cert required).
+  rpc Register(RegisterRequest) returns (RegisterResponse);
+
+  // Long-lived bidirectional stream — strict mTLS required.
+  rpc CommandStream(stream CommandResult) returns (stream Command);
+}
+```
+
+---
+
+## 4. Certificate Authority and PKI
+
+The PKI is rooted in a single internal CA created once on the control plane during `catalog configure`. All files are persisted under `/var/lib/ai-services/gateway-pki/` and loaded into the `Gateway` at startup.
+
+### 4.1 Root CA
+
+Generated once during `catalog configure --agentgateway-port` and persisted on the control plane. The public `ca.crt` is returned to every worker LPAR in `RegisterResponse` so agents can verify the gateway's server certificate on all future connections.
+
+```
+/var/lib/ai-services/gateway-pki/ca.key   — CA private key  (0600, control plane only)
+/var/lib/ai-services/gateway-pki/ca.crt   — CA public cert  (0644, returned to agents via RegisterResponse)
+```
+
+### 4.2 Gateway Server Certificate
+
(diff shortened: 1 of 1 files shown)


docs/proposals/catalog/mtls-bootstrapping-proposal.md
```diff
@@ -0,0 +1,358 @@
...
+
+---
+
+## 1. Overview
+
+This proposal describes how the **Agent Gateway** (control plane) and **agent daemons** (worker LPARs) establish mutual TLS (mTLS) for all gRPC communication.
+
+There are two distinct phases:
+
+- **Bootstrap (first contact):** A new agent daemon has no certificate yet. It connects to the gateway using a single-use pre-shared token, submits a CSR, and receives both a CA-signed client certificate and the server's `ca.crt` in return. This first connection uses TLS (server-only) with an optional CA verification fallback.
```
yussufsh, line 40: I do not think the first connection can be TLS since the server will create a self-signed certificate. Unless we add the server crt to worker trust store this needs to be insecure.

docs/proposals/catalog/mtls-bootstrapping-proposal.md
```diff
@@ -0,0 +1,358 @@
...
+## 2. Goals
+
+| Goal | Approach |
+|---|---|
+| Encrypted agent-to-gateway communication | TLS on all gRPC connections; mTLS after bootstrap |
+| Authenticated agents only on `CommandStream` | Client certificate required and verified against CA pool |
+| Zero pre-distributed secrets except token | Agent generates its own key locally; only a short-lived pre-shared token is distributed |
+| Credential freshness on every restart | `hasValidTLSCredentials` checks expiry and probes the gateway live before skipping re-registration |
+| Cert lifecycle tied to agent onboarding | Bootstrap generates ECDSA key + CSR; gateway signs and delivers the cert |
+| Fail-safe token guard | Registration is rejected immediately if no valid cert exists and no token is provided |
```
yussufsh, line 54: I did not understand this. Token provision is not part of mTLS process.

docs/proposals/catalog/mtls-bootstrapping-proposal.md
```diff
@@ -0,0 +1,358 @@
...
+The provided material is validated before any files are written. All four checks must pass or `catalog configure` exits with an error:
+
+| # | Check | How |
+|---|---|---|
+| ① | Cert and key correspond to each other | `tls.X509KeyPair(certPEM, keyPEM)` — errors if the private key doesn't match the public key in the cert |
+| ② | Certificate is a CA | `cert.IsCA == true && cert.BasicConstraintsValid == true` |
+| ③ | Certificate is currently valid | `time.Now()` is between `cert.NotBefore` and `cert.NotAfter` |
+| ④ | Certificate can sign other certs | `cert.KeyUsage & x509.KeyUsageCertSign != 0` |
+
+**Mode B — Auto-generated CA (default for fresh deployments):**
```
yussufsh, line 170: This is preferred since we will not allow other traffic to flow for this. The application lifecycle will happen via Caddy.

docs/proposals/catalog/mtls-bootstrapping-proposal.md
```diff
@@ -0,0 +1,358 @@
...
+  // Long-lived bidirectional stream — strict mTLS required.
+  rpc CommandStream(stream CommandResult) returns (stream Command);
+}
+```
+
+---
+
+## 4. Certificate Authority and PKI
+
+The PKI is rooted in a single internal CA created once on the control plane during `catalog configure`. All files are persisted under `/var/lib/ai-services/gateway-pki/` and loaded into the `Gateway` at startup.
```
mayuka-c, line 112: I guess based on recent discussion we would not be running catalog configure on control plane. So we would need to create the internal CA during catalog pod startup and it should be idempotent

docs/proposals/catalog/mtls-bootstrapping-proposal.md
```diff
@@ -0,0 +1,358 @@
...
+
+### 4.1 Root CA
+
+Generated once during `catalog configure --agentgateway-port` and persisted on the control plane. The public `ca.crt` is returned to every worker LPAR in `RegisterResponse` so agents can verify the gateway's server certificate on all future connections.
+
+```
+/var/lib/ai-services/gateway-pki/ca.key   — CA private key  (0600, control plane only)
+/var/lib/ai-services/gateway-pki/ca.crt   — CA public cert  (0644, returned to agents via RegisterResponse)
```
mayuka-c, line 120: Again CA private key and public cert we may need to setup during pod start

docs/proposals/catalog/mtls-bootstrapping-proposal.md
```diff
@@ -0,0 +1,358 @@
...
+```
+
+### 4.2 Gateway Server Certificate
+
+Generated during `catalog configure` immediately after the root CA is available. Signed by the root CA and persisted alongside it. The gateway loads it at startup and presents it to every connecting agent.
+
+```
+/var/lib/ai-services/gateway-pki/server.key  — server private key (0600)
+/var/lib/ai-services/gateway-pki/server.crt  — server public cert (0644, serverAuth EKU)
```
mayuka-c, line 129: Same here

docs/proposals/catalog/mtls-bootstrapping-proposal.md
```diff
@@ -0,0 +1,358 @@
...
+The PKI is initialised as part of `ai-services catalog configure --agentgateway-port <port>`. The process supports two modes:
+
+**Mode A — Operator-provided CA (recommended for production):**
+The operator supplies an existing CA key and certificate via flags. The gateway uses these to sign agent certificates. This allows integration with an existing PKI.
+
+```
+ai-services catalog configure --runtime podman \
+  --agentgateway-port 9090 \
+  --gateway-ca-cert /path/to/ca.crt \
+  --gateway-ca-key  /path/to/ca.key
```
mayuka-c, line 158: Where exactly would this be ran. Didnt quite get it

docs/proposals/catalog/mtls-bootstrapping-proposal.md
```diff
@@ -0,0 +1,358 @@
...
+  A([Agent daemon starts]) --> B{hasValidTLSCredentials?}
+  B -- "yes — cert not expired\n+ live mTLS probe succeeds" --> Z([Skip registration\nuse existing credentials])
+  B -- "no" --> C{PreSharedToken\nprovided?}
+  C -- "no" --> ERR([Error: cannot register\nwithout token])
+  C -- "yes" --> D[Generate ECDSA P-256\nprivate key locally]
+  D --> E["Build CSR\nCN=AgentName · Org=system:agents\nAlgo=ECDSAWithSHA256"]
+  E --> F["Resolve TLS config via buildTLSConfig\nca.crt present → RootCAs pool\nca.crt absent  → InsecureSkipVerify"]
+  F --> G[Dial AgentGateway :9090\nTLS only — no client cert]
+  G --> H["Call Register RPC\n(AgentName · PreSharedToken · CSR · Labels · Capabilities)"]
+  H --> I{Token valid?\nCSR CN == AgentName?\nCSR signature OK?}
```
mayuka-c, line 219: CN == AgentName validation? how can control plane gateway validate this as it doesnt even know the agentName?

docs/proposals/catalog/mtls-bootstrapping-proposal.md
```diff
@@ -0,0 +1,358 @@
...
+  I -- "no" --> REJECT([Registration rejected])
+  I -- "yes" --> J["Gateway signs CSR with rootCA\nSerial = rand 128-bit\nNotAfter = now + 1 year"]
+  J --> K["Write tls.crt → 0644\nWrite tls.key → 0600\nWrite ca.crt  → 0644\nto /etc/ai-services/agent-tls/"]
```
mayuka-c, line 222: Where exactly is this being stored?
Dont the gateway need to respond back to agent with all these 3 and it needs to be stored in agents LPAR?

docs/proposals/catalog/mtls-bootstrapping-proposal.md
```diff
@@ -0,0 +1,358 @@
...
+
+### 5.2 Credential Validation on Restart
+
+When the agent daemon restarts, `hasValidTLSCredentials` runs **before** any registration attempt. It performs three checks in order and returns `false` at the first failure, triggering fresh registration:
+
+```mermaid
+flowchart TD
+  A([Agent restarts]) --> B["Load tls.crt + tls.key\nfrom /etc/ai-services/agent-tls/"]
+  B --> C{Key pair loads OK?}
+  C -- "no — files missing\nor mismatched" --> FAIL([false → proceed to Register])
```
mayuka-c, line 235: How can the agent register here, since it doesnt have the bootstrap token?

docs/proposals/catalog/mtls-bootstrapping-proposal.md
hasan4791: The connection is still TLS but untrusted(InsecureSkipVerify=true). We accept the cert whatever server provides and proceed with the registration.

docs/proposals/catalog/mtls-bootstrapping-proposal.md
hasan4791: The scenario here is by some reason the certs on clients is invalid, the flow will proceed with the registration process. So it checks whether token is there, if not it will error out.

docs/proposals/catalog/mtls-bootstrapping-proposal.md
hasan4791: Ok, so we are generating all the certs during the initial deployment and going to persist in the storage.

docs/proposals/catalog/mtls-bootstrapping-proposal.md
hasan4791: Don't we have any configure or initial setup cli command for control plane?
If its not, we should handle it during initial pod startup time only and persist it.
@yussufsh please clarify

docs/proposals/catalog/mtls-bootstrapping-proposal.md
hasan4791: this is for control plane, if we don't have configure command, this should be ignored as well.

docs/proposals/catalog/mtls-bootstrapping-proposal.md
hasan4791: Are we not generating the preshared token based on agenName?

docs/proposals/catalog/mtls-bootstrapping-proposal.md
hasan4791: yes, this happens on client after it receives the response. Will update the flow chart.
Response have tls.cert for client and ca.crt.

docs/proposals/catalog/mtls-bootstrapping-proposal.md
hasan4791: Yes, that would be caught in the registration flow.

docs/proposals/catalog/mtls-bootstrapping-proposal.md
yussufsh: @mayuka-c today's discussion is only for that special case where we will need to write some kind of bootstrap steps in the script to bring up the catalog on the control plane.
For normal remote deployment we will still need catalog and caddy with custom cert. So we would still need configure command.

hasan4791: So in control plane we don't have configure command?

docs/proposals/catalog/mtls-bootstrapping-proposal.md
```diff
@@ -0,0 +1,330 @@
...
+| Single CA signs both server and agent certs | Simplifies trust model — one root to distribute, no intermediate CAs required |
+| PKI persisted under `gateway-pki/` | Survives process restarts without regeneration; loaded at startup if all four files are present |
+
+---
+
+## 9. Security Considerations
+
+- **Root CA private key** (`caKey`) is generated once and stored in a persistent, access-controlled location on the control plane's disk. It is loaded into the gateway process at startup and is never transmitted over the wire. It must never be distributed to worker LPARs.
+- **Pre-shared token** is single-use. `TokenStore.Validate` consumes it on first use, preventing replay attacks.
+- **CSR subject** is accepted as-submitted by the agent; the gateway does not enforce or validate a specific `AgentName` field. Agent identity on the gateway side is derived from `VerifiedChains[0][0].Subject.CommonName` after the mTLS handshake, not from any application-layer name claim in the request.
```
mayuka-c: What would be the CommonName used here?

mayuka-c (commented):
Rest LGTM

docs/proposals/catalog/mtls-bootstrapping-proposal.md
hasan4791: For now we can use uuid of the machine.

mayuka-c: What would the UUID be and where is the commonName would get validated?

hasan4791: We don't validate it currently. It can be used in future if required.

yussufsh:
@hasan4791 let's update and merge this one.

hasan4791:
@yussufsh If changes are good, lets merge this PR as well

yussufsh (approved):
merging this as implementation is also completed.

yussufsh:
As per Chandra all comments are resolved. If any changes, we can update this later.

PR #1194, opened 2026-07-29
removed mention of digitize from extract proposal

manalilatkar:
1. removed mention of digitize from extract proposal
2. made json schema optional in post /schema

### DIFF
--- a/docs/proposals/entity_extraction_service_proposal.md
+++ b/docs/proposals/entity_extraction_service_proposal.md
@@ -302,7 +302,7 @@ Returns the full schema record including the normalized `json_schema`, `examples
 | 404 Not Found | No schema with this ID. |
 | 500 Internal Server Error | Database failure. |
 
-**Sample response (201):**
+**Sample response (201) with user provided json schema:**
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


dharaneeshvrd (commented):
Can you please add a sample response for inferred schema as true? both post and get.

manalilatkar:
> Can you please add a sample response for inferred schema as true? both post and get.

@dharaneeshvrd for post, I have already added on line no 178, for get, is it really required? As the only difference would be that the field "inferred_schema" (on line 313) would be true instead of false. IMO, it would be overkill, but if you want I can add.

dharaneeshvrd:
I m looking for sample payload & responses depicting the json_examples and the respective extracted schema with inferred schema as true.
I think post example is added already, but get is pointing to the generic schema with few more fields.
It's better to differentiate with a new text and fields, also multiple examples to show case how we identify the required field.
I am stressing this because, users would add schemas based on examples mostly.

manalilatkar:
@dharaneeshvrd added a request with multiple examples to show how "required" works.

dharaneeshvrd (approved):

PR #1197, opened 2026-07-30
feat: stop in-flight deployment on application deletion

shwetha-s-poojary:
Allows deletion of an application at any point during deployment with no orphaned resources.

## Features
- **DeploymentRegistry** (`deployment_registry.go`) — new; tracks in-flight deployment goroutines by app ID so `DeleteApplication` can cancel them before transitioning to Deleting

## Fixes
- **`RunContainerWithSpec`** — now accepts `ctx`; races `containers.Wait` against `ctx.Done()` and calls `containers.Stop` on cancellation so the model download container is terminated immediately instead of blocking
- **`DownloadModelContainer`** — threads `ctx` through to `RunContainerWithSpec` so the cancellation signal reaches the running container
- **`HandleDeploymentStepError`** — exits silently if `ctx.Err() != nil` so a cancelled deployment step doesn't overwrite the Deleting status with Error
- **`RunConcurrently`** — same silent-exit guard added for concurrent component/service deployment steps
- **`ExecuteDeployment`** — skips the final `Running` status update if `ctx.Err() != nil`; leaves DeletionService as sole status owner
- **`RegisterRoute` / `createRoute`** — ctx now propagated to Caddy HTTP calls so route registration is also interrupted on cancel
- **`Helm Install/Upgrade`** — `RunWithContext` replaces `Run` so in-flight
  Helm operations on OpenShift abort immediately on delete rather than
  blocking until the 10-minute timeout expires

## Refactor
- **`DeleteApplication`** — promoted from Podman-only to shared base method in `ApplicationServiceBase`; wires `DeploymentRegistry.Cancel` before marking app as Deleting
- **`executeDeploymentAsync`** — moved to base; registers/deregisters with `DeploymentRegistry` and exits silently on `ctx.Err()`
- **`CreatePod`** — signature updated to accept `ctx`; uses new `podmanCtx()` helper to bridge caller cancellation with the Podman connection context
- **`prepareDeployment`** — extracted from `ExecuteDeployment` to group image pull + model download + status transition into one step
- **`NewOpenShiftApplicationService`** — constructor added, wiring
  `DeploymentRegistry` into the base for full OCP cancellation parity

## Tested
- DELETE during model download → container stopped within stop timeout, app deleted ✅
- DELETE during image pull / pod creation → no orphaned pods, volumes or secrets ✅
- DELETE during Helm install/upgrade (OCP) → operation aborted at Helm level, namespace cleaned up ✅
- Normal deployment unaffected ✅

## Notes
- UI delete button during deployment deferred to follow-up PR

### DIFF
--- a/ai-services/internal/pkg/application/openshift/create.go
+++ b/ai-services/internal/pkg/application/openshift/create.go
@@ -121,11 +121,11 @@ func deployApp(ctx context.Context, chart chart.Charter, timeout time.Duration, 
 	if !isAppExist {
 		// if App does not exist then perform install
 		logger.Infof("App: %s does not exist, proceeding with install...", app)
-		err = helmClient.Install(app, chart, &helm.InstallOpts{Values: values, Timeout: timeout})
+		err = helmClient.Install(context.Background(), app, chart, &helm.InstallOpts{Values: values, Timeout: timeout})
 	} else {
 		// if App exists, perform upgrade so that the actual state of the app meets the desired state
 		logger.Infof("App: %s already exist, proceeding with reconciling...", app)
-		err = helmClient.Upgrade(app, chart, &helm.UpgradeOpts{Values: values, Timeout: timeout})
+		err = helmClient.Upgrade(context.Background(), app, chart, &helm.UpgradeOpts{Values: values, Timeout: timeout})
 	}
 	if err != nil {
 		s.Fail("failed to create application")
--- a/ai-services/internal/pkg/application/openshift/create.go
+++ b/ai-services/internal/pkg/application/openshift/create.go
@@ -121,11 +121,11 @@ func deployApp(ctx context.Context, chart chart.Charter, timeout time.Duration, 
 	if !isAppExist {
 		// if App does not exist then perform install
 		logger.Infof("App: %s does not exist, proceeding with install...", app)
-		err = helmClient.Install(context.Background(), app, chart, &helm.InstallOpts{Values: values, Timeout: timeout})
+		err = helmClient.Install(ctx, app, chart, &helm.InstallOpts{Values: values, Timeout: timeout})
 	} else {
 		// if App exists, perform upgrade so that the actual state of the app meets the desired state
 		logger.Infof("App: %s already exist, proceeding with reconciling...", app)
-		err = helmClient.Upgrade(context.Background(), app, chart, &helm.UpgradeOpts{Values: values, Timeout: timeout})
+		err = helmClient.Upgrade(ctx, app, chart, &helm.UpgradeOpts{Values: values, Timeout: timeout})
 	}
 	if err != nil {
 		s.Fail("failed to create application")
--- a/ai-services/internal/pkg/catalog/apiserver/repository/application_service/deployment_registry.go
+++ b/ai-services/internal/pkg/catalog/apiserver/repository/application_service/deployment_registry.go
@@ -48,10 +48,17 @@ func (r *DeploymentRegistry) Cancel(appID uuid.UUID) {
 	}
 }
 
-// Deregister removes a completed deployment from the registry without
-// cancelling it. Called by the goroutine when deployment finishes normally.
+// Deregister removes a completed deployment from the registry and calls
+// cancel() to release the child context node from the parent's tree.
+// Not calling cancel() would leak the context until the parent (context.Background)
+// is itself cancelled — which never happens.
 func (r *DeploymentRegistry) Deregister(appID uuid.UUID) {
 	r.mu.Lock()
+	cancel, ok := r.entries[appID]
 	delete(r.entries, appID)
 	r.mu.Unlock()
+
+	if ok {
+		cancel()
+	}
 }
--- a/ai-services/internal/pkg/catalog/apiserver/repository/application_service/deployment_registry.go
+++ b/ai-services/internal/pkg/catalog/apiserver/repository/application_service/deployment_registry.go
@@ -1,0 +1,57 @@
+package applicationservice
+
+import (
+	"context"
+	"sync"
+
+	"github.com/google/uuid"
+)
+
+// DeploymentRegistry tracks in-flight deployments so they can be cancelled
+// when a delete request arrives mid-deployment.
+type DeploymentRegistry struct {
+	mu      sync.Mutex
+	entries map[uuid.UUID]context.CancelFunc
+}
+
+// NewDeploymentRegistry creates a new registry.
+func NewDeploymentRegistry() *DeploymentRegistry {
+	return &DeploymentRegistry{
+		entries: make(map[uuid.UUID]context.CancelFunc),
+	}
+}
+
+// Register stores a cancel function for the given application and returns
+// a cancellable context derived from the provided parent.
+func (r *DeploymentRegistry) Register(parent context.Context, appID uuid.UUID) context.Context {
+	ctx, cancel := context.WithCancel(parent)
+
+	r.mu.Lock()
+	r.entries[appID] = cancel
+	r.mu.Unlock()
+
+	return ctx
+}
+
+// Cancel signals the in-flight deployment for appID to stop and removes it
+// from the registry. It is a no-op if appID is not registered.
+func (r *DeploymentRegistry) Cancel(appID uuid.UUID) {
+	r.mu.Lock()
+	cancel, ok := r.entries[appID]
+	if ok {
+		delete(r.entries, appID)
+	}
+	r.mu.Unlock()
+
+	if ok {
+		cancel()
+	}
+}
+
+// Deregister removes a completed deployment from the registry without
+// cancelling it. Called by the goroutine when deployment finishes normally.
+func (r *DeploymentRegistry) Deregister(appID uuid.UUID) {
+	r.mu.Lock()
+	delete(r.entries, appID)
+	r.mu.Unlock()
+}
--- a/ai-services/internal/pkg/catalog/apiserver/repository/application_service/openshift.go
+++ b/ai-services/internal/pkg/catalog/apiserver/repository/application_service/openshift.go
@@ -12,10 +12,16 @@ import (
 
 // OpenShiftApplicationService implements ApplicationServiceInterface for the OpenShift runtime.
 // It embeds ApplicationServiceBase for all shared DB operations.
-//
-// TODO(OCP): set base.DeploymentRegistry = NewDeploymentRegistry() in the OpenShift constructor when CreateApplication is implemented.
 type OpenShiftApplicationService struct {
 	ApplicationServiceBase
+}
+
+// NewOpenShiftApplicationService creates a new OpenShiftApplicationService with a fresh DeploymentRegistry
+// wired into the base. This makes in-flight deployments cancellable by a concurrent DeleteApplication.
+func NewOpenShiftApplicationService(base ApplicationServiceBase) *OpenShiftApplicationService {
+	base.DeploymentRegistry = NewDeploymentRegistry()
+
+	return &OpenShiftApplicationService{ApplicationServiceBase: base}
 }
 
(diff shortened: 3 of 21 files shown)


ai-services/internal/pkg/catalog/apiserver/repository/application_service/deployment_registry.go
```diff
@@ -0,0 +1,57 @@
+package applicationservice
+
```
mayuka-c, line 2: Let this file be along with deployer.go as it is related to deployment

ai-services/internal/pkg/catalog/apiserver/repository/application_service/podman.go
```diff
@@ -191,6 +209,13 @@ func (s *PodmanApplicationService) executeDeploymentAsync(parentCtx context.Cont
 
 	err := s.DeploymentExecutor.ExecuteWithPlan(ctx, plan, req, runtimeTypes.RuntimeTypePodman)
 	if err != nil {
+		// Context was cancelled — deletion is in charge, exit silently.
+		if ctx.Err() != nil {
+			logger.InfofCtx(ctx, "Deployment cancelled for application %s (deletion in progress)", plan.ApplicationName)
+
+			return
```
mayuka-c: Why do we have this context error cancellation check only fot ExecuteWithPlan and not for others?

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -132,14 +117,19 @@ func (d *PodmanDeployer) ExecuteDeployment(
...
 
 		return fmt.Errorf("failed to register application routes: %w", err)
 	}
 
-	// Step 6: Update application status to Running
+	// Step 5: Update application status to Running.
+	// Skip if the context was cancelled — deletion is now in charge of the status.
+	if ctx.Err() != nil {
+		return nil
+	}
```
mayuka-c: This can go in the below error block similar to handleDeploymentStepError

ai-services/internal/pkg/runtime/podman/podman.go
```diff
@@ -115,7 +115,28 @@ func (pc *PodmanClient) ListPods(filters map[string][]string) ([]types.Pod, erro
...
+	// Child of pc.Context so the podman connection value is present.
+	mergedCtx, cancel := context.WithCancel(pc.Context)
+
+	// Mirror cancellation from the caller into the merged context.
+	go func() {
+		select {
+		case <-callerCtx.Done():
+			cancel()
+		case <-mergedCtx.Done():
+			// merged context was cancelled by someone else — nothing to do
```
mayuka-c: `cancelled by someone` Meaning?

mayuka-c: Do we need mergedCtx?

ai-services/internal/pkg/catalog/apiserver/repository/application_service/podman.go
shwetha-s-poojary: ExecuteWithPlan is the only call in this function that can actually fail right , it wraps the whole deployment pipeline. The ctx.Err() check is just figuring out whether the error came from a cancellation or a real failure. Is there a specific other place you think this check should be?

ai-services/internal/pkg/runtime/podman/podman.go
shwetha-s-poojary: poor wordings , will update. What it meant was the only thing that cancels mergedCtx other than callerCtx is the defer cancel() in CreatePod itself when the function returns.
Yes. mergedCtx.Done() is the exit condition for the goroutine in the normal (no cancellation) case. When CreatePod returns, defer cancel() closes mergedCtx, which unblocks the goroutine and lets it exit. Without it the goroutine would leak on every normal deployment

ai-services/internal/pkg/catalog/apiserver/repository/application_service/deployment_registry.go
shwetha-s-poojary: deployer.go already imports from applicationservice (to use AppRepo, ServiceRepo etc). If you also move DeploymentRegistry into the deployer package, then applicationservice would need to import from the deployer package to use it. That creates a circular import

ai-services/internal/pkg/runtime/podman/podman.go
```diff
@@ -115,7 +115,31 @@ func (pc *PodmanClient) ListPods(filters map[string][]string) ([]types.Pod, erro
...
 }
 
-func (pc *PodmanClient) CreatePod(body io.Reader, opts map[string]string) ([]types.Pod, error) {
+// podmanCtx derives a child of pc.Context (which carries the podman connection) that
+// is cancelled when the caller's ctx is cancelled.  This is necessary because the
+// podman bindings SDK requires its own connection-carrying context, but we still want
+// cancellation signals from the caller to propagate (e.g. mid-deployment deletion).
+func (pc *PodmanClient) podmanCtx(callerCtx context.Context) (context.Context, context.CancelFunc) {
+	// Child of pc.Context so the podman connection value is present.
+	mergedCtx, cancel := context.WithCancel(pc.Context)
```
mayuka-c: Why do we call this as mergedCtx, let us just ctx

mayuka-c (commented):
Rest LGTM

mayuka-c:
Also @shwetha-s-poojary please coordinate with @pranithraoibm to enable deletion on UI during deployment

shwetha-s-poojary:
> Also @shwetha-s-poojary please coordinate with @pranithraoibm to enable deletion on UI during deployment

@mayuka-c , @pranithraoibm has already raised the PR: #1198 . I think he just needs to validate it once we merge this PR

ai-services/internal/pkg/catalog/apiserver/repository/application_service/openshift.go
```diff
@@ -14,6 +14,8 @@ var errOpenShiftNotSupported = errors.New("OpenShift runtime is not yet supporte
 
 // OpenShiftApplicationService implements ApplicationServiceInterface for the OpenShift runtime.
 // It embeds ApplicationServiceBase for all shared DB operations.
+//
+// TODO(OCP): set base.DeploymentRegistry = NewDeploymentRegistry() in the OpenShift constructor when CreateApplication is implemented.
```
yussufsh: Is it done ot NOT?

shwetha-s-poojary: DeleteApplication for OCP still returns errOpenShiftNotSupported, so there's nothing to wire into yet.This TODO tracks what's needed when OCP deletion is implemented.That will be going in as follow-up.

shwetha-s-poojary:
@yussufsh  just checking if you have any further comments on the PR. If everything looks good, would it be possible to merge it? It’s been stale for a few days, and I’m having to rebase frequently to resolve conflicts. Thanks!

ai-services/internal/pkg/runtime/openshift/openshift.go
```diff
@@ -208,7 +208,7 @@ func (kc *OpenshiftClient) ListPods(filters map[string][]string) ([]types.Pod, e
 }
 
 // CreatePod creates a pod from YAML manifest.
-func (kc *OpenshiftClient) CreatePod(body io.Reader, opts map[string]string) ([]types.Pod, error) {
+func (kc *OpenshiftClient) CreatePod(ctx context.Context, body io.Reader, opts map[string]string) ([]types.Pod, error) {
```
yussufsh:
```suggestion
func (kc *OpenshiftClient) CreatePod(_ context.Context, body io.Reader, opts map[string]string) ([]types.Pod, error) {
```

ai-services/internal/pkg/application/openshift/create.go
```diff
@@ -121,11 +121,11 @@ func deployApp(ctx context.Context, chart chart.Charter, timeout time.Duration,
 	if !isAppExist {
 		// if App does not exist then perform install
 		logger.Infof("App: %s does not exist, proceeding with install...", app)
-		err = helmClient.Install(app, chart, &helm.InstallOpts{Values: values, Timeout: timeout})
+		err = helmClient.Install(context.Background(), app, chart, &helm.InstallOpts{Values: values, Timeout: timeout})
```
yussufsh: Why is this from the background? Should not we use `ctx` parameter to let a cancellation signal reach the Helm operations?

ai-services/internal/pkg/application/openshift/create.go
```diff
@@ -121,11 +121,11 @@ func deployApp(ctx context.Context, chart chart.Charter, timeout time.Duration,
...
 	} else {
 		// if App exists, perform upgrade so that the actual state of the app meets the desired state
 		logger.Infof("App: %s already exist, proceeding with reconciling...", app)
-		err = helmClient.Upgrade(app, chart, &helm.UpgradeOpts{Values: values, Timeout: timeout})
+		err = helmClient.Upgrade(context.Background(), app, chart, &helm.UpgradeOpts{Values: values, Timeout: timeout})
```
yussufsh: Same here.

yussufsh (changes requested):

yussufsh (dismissed):
lgtm

ai-services/internal/pkg/cli/helpers/models.go
```diff
@@ -105,8 +105,9 @@ func DownloadModelContainer(ctx context.Context, model, targetDir string) error
 		},
 	}
 
-	// Run container with spec
-	exitCode, err := runtimeClient.RunContainerWithSpec(s)
+	// Run container with spec, passing ctx so cancellation (e.g. mid-deployment delete)
+	// stops the download container immediately instead of blocking until it finishes.
+	exitCode, err := runtimeClient.RunContainerWithSpec(ctx, s)
```
mayuka-c, line 110: Did you validate where if lets say there was a model deployment happening and it got cancelled, will it be able to download complete model in next run and works fine?

ai-services/internal/pkg/catalog/utils/application_utils.go
```diff
@@ -16,7 +16,15 @@ import (
 )
 
 // HandleDeploymentStepError updates the application status to Error and logs the failure.
+// If the context has already been cancelled (e.g. a mid-deployment delete), it exits
+// silently so the deletion goroutine retains ownership of the application status.
 func HandleDeploymentStepError(ctx context.Context, appRepo dbrepo.ApplicationRepository, appID uuid.UUID, stepContext string, err error) {
+	if ctx.Err() != nil {
+		logger.InfofCtx(ctx, "Deployment step %q for %s cancelled (deletion in progress)\n", stepContext, appID)
```
mayuka-c: This could be a warning level

mayuka-c (commented):
Rest LGTM

ai-services/internal/pkg/cli/helpers/models.go
shwetha-s-poojary: Yes

ai-services/internal/pkg/catalog/utils/application_utils.go
shwetha-s-poojary: Will change it to warn

ai-services/internal/pkg/cli/helpers/models.go
mayuka-c: How exactly does it know?. Is it based on size it determines whether to download or not?

shwetha-s-poojary: The changes here only stop the model download when the deployment is cancelled. On the next run, we rely on hf download to handle the existing local download state. I’m not completely sure about the exact logic at the Hugging Face level, but from what I’ve looked up, it doesn’t appear to be size-based and uses download metadata to determine what needs to be downloaded.

mayuka-c (approved):

PR #1198, opened 2026-07-30
allow delete during mid-deployment states

pranithraoibm:
Previously the delete action was only enabled for Running and Error statuses. Now it's enabled for all the statuses except deleting itself.


shwetha-s-poojary:
This should go in once backend code #1197 is merged

ui/catalog/src/pages/DigitalAssistants/CellRenderers.tsx
```diff
@@ -60,8 +60,7 @@ interface CellRendererProps {
 }
 
 export const ActionCell = ({ rowId, dispatch, rowData }: CellRendererProps) => {
-  const canDelete =
-    rowData?.status === "Running" || rowData?.status === "Error";
+  const canDelete = rowData?.status !== "Deleting";
```
ryanmarc, line 63: The backend sends "Deleting" with no dots (db/models/application.go:20-26) and this page passes the API value straight through at applications.api.ts:386. So the behaviour is correct.

The problem is the declared type disagrees. pages/DigitalAssistants/types.ts:7 says the union is "Deploying..." | "Deleting..." | ... with dots, and only the as cast on applications.api.ts:386 keeps that from erroring. So this line is comparing against a value TypeScript thinks is impossible, and it only compiles because rowData is loosely typed as { status?: string } on line 59. If someone later "corrects" this to match the type, delete gets disabled on every row and nothing catches it.

This is not your bug, but this PR is the first code that depends on the un-dotted value being right, so it would be a good place to fix the union to match the backend enum and drop the cast. It is biting elsewhere too: #1181 just added a poll on this same page keyed on "Deploying..." and "Deleting..." from the same union, and it never fires.

ryanmarc (changes requested):

ryanmarc:
Question rather than an objection: this makes deleting during Deploying and Downloading a normal thing users will do, and both deletion and deployment run as background goroutines on the server (application_service/podman.go kicks off PerformDeletion, and the deployer runs its own goroutine). The API's only guard is "already Deleting" (podman.go), so nothing stops a delete from tearing down resources while the deploy is still creating them.

Has this been tried against a real mid-deploy application? If the deployer already checks for a Deleting status between steps then this is fine and worth a line in the description. If not, the guard probably belongs on the server rather than in the button's disabled prop, since the API is the thing that has to be safe.

ui/catalog/src/pages/DigitalAssistants/CellRenderers.tsx
pranithraoibm: Totally agree, even I thought to standardize these status values in this PR or https://github.com/IBM/project-ai-services/pull/1181, but as mentioned [here](https://github.com/IBM/project-ai-services/pull/1181#discussion_r3680932428) @Niranjan-J-Joshi will be covering this standardization of the status values in his 4th PR of table code refactor and has already started with the implementation with types moving into a single file and everything. So I have asked him to cover this and the one in https://github.com/IBM/project-ai-services/pull/1181 as well. So is it fine if Niranjan covers this in his unified PR or you want me to do it here itself? I'm fine with anything do let me know.

pranithraoibm:
I think the backend changes are yet to go in correct @shwetha-s-poojary? 

Request from the backend team for us was to enable the delete button and initiate the deletion on click when the deployment is in progress which I have verified on the UI. Apart from that if there is anything, that will be something the backend team would have to answer.

shwetha-s-poojary:
> I think the backend changes are yet to go in correct @shwetha-s-poojary?

Yes

> Question rather than an objection: this makes deleting during Deploying and Downloading a normal thing users will do, and both deletion and deployment run as background goroutines on the server (application_service/podman.go kicks off PerformDeletion, and the deployer runs its own goroutine). The API's only guard is "already Deleting" (podman.go), so nothing stops a delete from tearing down resources while the deploy is still creating them.
Has this been tried against a real mid-deploy application? If the deployer already checks for a Deleting status between steps then this is fine and worth a line in the description. If not, the guard probably belongs on the server rather than in the button's disabled prop, since the API is the thing that has to be safe.

@ryanmarc the PR linked [here](https://github.com/IBM/project-ai-services/pull/1198#issuecomment-5130252994) might answer most of your questions.This UI PR is a  follow-up.

pranithraoibm:
UI Verification with the updated backend

https://github.com/user-attachments/assets/1c612af6-39cf-4d47-b799-bae97ce5db32

ryanmarc (approved):

ryanmarc (dismissed):

Niranjan-J-Joshi (approved):

shwetha-s-poojary (approved):
LGTM

PR #1202, opened 2026-07-31
[Design Proposal] Support for Custom Asset onboarding

mayuka-c:
- Proposal for onboarding customer services.
- Proposes API endpoints with DB proposal to store the metadata information.
- The assets are being stored inside volumes (works both in podman and openshift).
- Also works seemlessly for remote deployments (Refer Section 9)
- Addition of CLI commands for bundles

### DIFF
--- a/docs/proposals/catalog/custom-service-templates-proposal.md
+++ b/docs/proposals/catalog/custom-service-templates-proposal.md
@@ -17,14 +17,14 @@
     - 4.3 [Custom service assets (proposed)](#43-custom-service-assets-proposed)
     - 4.4 [Custom component assets (proposed)](#44-custom-component-assets-proposed)
 5. [CatalogProvider Integration](#5-catalogprovider-integration)
-6. [API Upload](#6-api-upload)
+6. [API Create Bundle](#6-api-create-bundle)
     - 6.1 [Design goals](#61-design-goals)
     - 6.2 [New API endpoints](#62-new-api-endpoints)
     - 6.3 [Bundle format](#63-bundle-format)
     - 6.4 [Server-side processing pipeline](#64-server-side-processing-pipeline)
     - 6.5 [New database migration](#65-new-database-migration)
     - 6.6 [Storage per runtime](#66-storage-per-runtime)
-    - 6.7 [Upload flow diagrams](#67-upload-flow-diagrams)
+    - 6.7 [Create Bundle flow diagrams](#67-create-bundle-flow-diagrams)
     - 6.8 [Handler and service (Go)](#68-handler-and-service-go--as-implemented)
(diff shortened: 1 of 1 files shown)


mayuka-c:
@yussufsh @Shubhamag12 Have addressed the comments that we discussed today in the call. PTAL.

Also regarding validate, looks like best/standard practice (if we have robust validation) is to do automatically in both POST and PUT and also have a separate API to be able to validate it independentely. So I have kept the API and added the same validation process in both POST and PUT.

yussufsh:
> @yussufsh @Shubhamag12 Have addressed the comments that we discussed today in the call. PTAL.
> 
> Also regarding validate, looks like best/standard practice (if we have robust validation) is to do automatically in both POST and PUT and also have a separate API to be able to validate it independentely. So I have kept the API and added the same validation process in both POST and PUT.

Fine with me. But are you saying you will also validate the bundle in POST/PUT every time? Do we know how much time does it take to extract and read the FS?

mayuka-c:
> > @yussufsh @Shubhamag12 Have addressed the comments that we discussed today in the call. PTAL.
> > Also regarding validate, looks like best/standard practice (if we have robust validation) is to do automatically in both POST and PUT and also have a separate API to be able to validate it independentely. So I have kept the API and added the same validation process in both POST and PUT.
> 
> Fine with me. But are you saying you will also validate the bundle in POST/PUT every time? Do we know how much time does it take to extract and read the FS?

I did not verify, let me try to implement Validations and see how much time it takes like basic file structuring, yaml formatting and labels etc and get back on the latency

mayuka-c:
> > > @yussufsh @Shubhamag12 Have addressed the comments that we discussed today in the call. PTAL.
> > > Also regarding validate, looks like best/standard practice (if we have robust validation) is to do automatically in both POST and PUT and also have a separate API to be able to validate it independentely. So I have kept the API and added the same validation process in both POST and PUT.
> > 
> > 
> > Fine with me. But are you saying you will also validate the bundle in POST/PUT every time? Do we know how much time does it take to extract and read the FS?
> 
> I did not verify, let me try to implement Validations and see how much time it takes like basic file structuring, yaml formatting and labels etc and get back on the latency

The Validate API took around 1sec (Added both podman and Openshift validations) - validating basic things as said above

image

The entire Create API took around `1.156s`. So extract + DB updates takes rougly around 0.2s

image

So I guess overall should be fine. Also noting I had done both podman and openshift validations synchronously. If we have a gorutine for each then it would reduce few milliseconds of time too.

yussufsh (approved):
Overall proposal looks good from my side to start with this feature.

mayuka-c:
Going ahead merging to start with implementation. If anything is missed/required we could update along with implementations

PR #1203, opened 2026-07-31
docs: add steps to rotate Watsonx API keys on Podman runtime

shwetha-s-poojary:
Adds a doc with steps to rotate a Watsonx API key for an AI Services application on the Podman runtime (PowerVS), using application delete `delete+ backup + application create+ restore` to preserve data volumes while replacing the key.

### DIFF
--- a/docs/rotate-watsonx-keys.md
+++ b/docs/rotate-watsonx-keys.md
@@ -5,13 +5,28 @@ How to rotate an IBM Watsonx API key for an AI Services application running on t
 ## Prerequisites
 
 - The Catalog UI is deployed and accessible. See [Configuring with PowerVS and IBM watsonx](https://www.ibm.com/docs/en/aiservices/2026.06.0?topic=podman-configuring-powervs-watsonx).
-- You have the new Watsonx API key, project ID, and Watsonx URL ready.
+- You have the new Watsonx API key ready.
 
-## Before you begin
+## Approach 1 — Script (no data loss)
 
-Rotating a key requires deleting and recreating the application. Each new application gets a new internal ID, so all Podman volumes (ingested documents, vector store) are created under a new name and the old data becomes unreachable.
+Rotates the key by replacing only the LiteLLM pod. All ingested data is preserved.
 
-**Back up your data before proceeding.** The backup and restore steps apply to the following deployments:
+```bash
+export WATSONX_APIKEY_NEW="<your-new-api-key>"
+./hack/rotate-watsonx-key/rotate-watsonx-key.sh <app-name>
+```
+
+The script must be run from within the repository. After it completes, access the Q&A interface and send a test query to confirm the application is working correctly with the new key.
+
+> **Note:** This approach is not available in the Catalog UI.
+
+## Approach 2 — Delete and recreate (data loss without backup)
+
+Deletes and recreates the entire application. Use this if the script approach is not available.
+
+### Before you begin
+
+Each new application gets a new internal ID so all data volumes become unreachable after deletion. Back up your data first if you have ingested documents to preserve.
 
 | Deployment | Backup needed | Targets |
 |---|---|---|
@@ -20,48 +35,38 @@ Rotating a key requires deleting and recreating the application. Each new applic
 | `summarize` standalone | ⚠️ No backup target available — data will be lost | — |
 | `chat` standalone | No persistent data | — |
 
-If you have no ingested data to preserve, skip to [Step 3](#3-delete-the-application).
+### Steps
 
-> **Note:** Backup and restore for the `summarize` standalone service is not currently supported. Job history stored in its Postgres database cannot be recovered after deletion.
-
-## Steps
-
-### 1. Back up your data
-
-Use the CLI to back up both OpenSearch and digitize data. Backup is not available in the Catalog UI.
+**1. Back up your data** (CLI only — skip if no data to preserve)
 
 ```bash
 ai-services application backup <app-name> --target opensearch --runtime podman
 ai-services application backup <app-name> --target digitize --runtime podman
 ```
 
-Note the generated `.tar.gz` filenames — you will need them to restore.
+**2. Set the new credentials**
 
-### 2. Set the new credentials
+Export your new Watsonx API key, project ID, and URL before proceeding. If using the Catalog UI, have these values ready to enter in the deploy form.
 
-Export your new Watsonx API key, project ID, and URL in the terminal before proceeding. This is required for the CLI path. If using the Catalog UI to redeploy, have these values ready to enter in the deploy form.
-
-### 3. Delete the application
+**3. Delete the application**
 
 **CLI:** Delete the application using the CLI.
 
 **Catalog UI:** Open the application, click the overflow menu, and select **Delete**.
 
-### 4. Recreate the application with the new key
+**4. Recreate the application with the new key**
 
 **CLI:** Create the application using the new exported credentials.
 
 **Catalog UI:** Deploy the application again from the Catalog and enter the new Watsonx API key in the deploy form.
 
-### 5. Restore your data
-
-Use the CLI to restore both targets into the newly created application. Restore is not available in the Catalog UI.
+**5. Restore your data** (CLI only — skip if no backup was taken)
 
 ```bash
 ai-services application restore <app-name> --target opensearch --filename <opensearch-backup>.tar.gz --runtime podman --yes
 ai-services application restore <app-name> --target digitize --filename <digitize-backup>.tar.gz --runtime podman --yes
 ```
 
-### 6. Verify
+**6. Verify**
 
 Access the Q&A interface and send a test query to confirm the application is working correctly with the new key.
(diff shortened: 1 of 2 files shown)


docs/rotate-watsonx-keys.md
```diff
@@ -0,0 +1,67 @@
...
+Use the CLI to restore both targets into the newly created application. Restore is not available in the Catalog UI.
+
+```bash
+ai-services application restore <app-name> --target opensearch --filename <opensearch-backup>.tar.gz --runtime podman --yes
+ai-services application restore <app-name> --target digitize --filename <digitize-backup>.tar.gz --runtime podman --yes
+```
+
+### 6. Verify
+
+Confirm the new key is active by checking the mounted secret file inside the LiteLLM container.
```
mayuka-c: This would not be the right way to verify.
For verification they can access Q&A and see if its working fine. I feel this would be right way instead of them checking secret (they will not have access to VM and not the correct way to do)

mayuka-c (commented):
Rest LGTM

@shwetha-s-poojary We need not have to merge as this is only for docs. Please work with @irapandey to get it added into the docs

docs/rotate-watsonx-keys.md
shwetha-s-poojary: Yeah you are right that's something I used to check the change of keys locally :)
Will remove that and add "access and query Q&A to verify"

hack/rotate-watsonx-key/rotate-watsonx-key.sh
```diff
@@ -0,0 +1,138 @@
...
+echo "→ Looking up TemplateID from catalog DB..."
+TEMPLATE_ID=$(podman exec ai-services--db-postgresql \
+    psql -U postgres -d ai_services -tAc \
+    "SELECT DISTINCT c.id FROM components c
+     JOIN service_dependencies sd ON sd.dependency_id = c.id
+     JOIN services s ON s.id = sd.service_id
+     JOIN applications a ON a.id = s.app_id
+     WHERE a.name = '${APP_NAME}' AND c.type = 'llm' AND c.provider = 'watsonx'
+     LIMIT 1;")
+
```
shwetha-s-poojary, line 88: @mayuka-c .The script takes <app-name> but there's no app-name label on the pod - the mapping only exists in the catalog DB. Without the DB, we can only match by llm-* prefix which breaks with multiple apps. Options: (1) take pod name as input instead (2) use the DB for the lookup. Which is preferred?

hack/rotate-watsonx-key/rotate-watsonx-key.sh
```diff
@@ -0,0 +1,138 @@
...
+#   export WATSONX_APIKEY_NEW="<your-new-api-key>"
+#   ./hack/rotate-watsonx-key/rotate-watsonx-key.sh rag-test
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+TEMPLATE_DIR="${SCRIPT_DIR}/../../ai-services/assets/components/llm/watsonx/podman/templates"
+SECRET_TMPL="${TEMPLATE_DIR}/watsonx-secret.yaml.tmpl"
+SERVER_TMPL="${TEMPLATE_DIR}/watsonx-server.yaml.tmpl"
+
```
shwetha-s-poojary, line 21: @yussufsh The alternative to using template files is hardcoding the pod spec inline, but that diverges from what the catalog actually deploys and breaks silently if the template changes. Is the repo dependency the concern, or is there a different reason to avoid the template files?

hack/rotate-watsonx-key/rotate-watsonx-key.sh
mayuka-c: You should be getting the podname and labels from the existing pod and simply reuse. Is there a problem with doing this?

mayuka-c: Here is what you have to do
- Read the appName from the user while executing the script
- Do application ps call using the same appName -> It will list all the pods for a given app
- Look for llm pod with name (llm-) prefix and ensure its deployed using watsonx (Validate by doing inspect) -> if not fail
- If it is watsonx provider, then you gather the details (like podname, labels etc)
- Update the secret (Again secret name will be watsonx-secret-{{ InstanceSlug }}). InstanceSlug you already have as part of step 3
- Once secret is updated, redeploy the llm pod with same podname and labels

shwetha-s-poojary: application ps calls the catalog API which reads from the DB underneath.You didn't want DB dependency right? Or did you want it to be cleaner and not reference the DB directly?

shwetha-s-poojary: >You should be getting the podname and labels from the existing pod and simply reuse. Is there a problem with doing this?

No question was rather directed towards which option /approach is preferred.

- Read the appName from the user while executing the script -- script already does that
- Do application ps call using the same appName -> It will list all the pods for a given app -- instead of listing we are directly querying DB(the application ps does that internally)
- Look for llm pod with name (llm-) prefix and ensure its deployed using watsonx (Validate by doing inspect) -> if not fail  -- script already does that
- If it is watsonx provider, then you gather the details (like podname, labels etc) -- script already does that
- Update the secret (Again secret name will be watsonx-secret-{{ InstanceSlug }}). InstanceSlug you already have as part of step 3 -- (I'm assuming this means podman secret rm + podman secret create) script already does that
- Once secret is updated, redeploy the llm pod with same podname and labels. -- script already does that

But the concern you raised earlier was to fetch template ID using podman inspect instead of  fetching it from DB, that in when I had the above questions :)

mayuka-c: I mean with appName you will not be able to determine pod. Obviously we need to determine the exact pod and for that we need to use subcommands. We should not be making db calls directly by exec into the postgres container

mayuka-c: From the flow that I see you're doing is wrong. Discover of pod cannot be done with appName

hack/rotate-watsonx-key/rotate-watsonx-key.sh
```diff
@@ -0,0 +1,138 @@
...
+    exit 1
+fi
+
+# ── Discover slug from running pod ────────────────────────────────────────────
+echo "→ Discovering LLM pod for application '${APP_NAME}'..."
+LLM_POD=$(podman pod ps --format "{{.Name}}" | grep "^llm-" | head -1)
+if [[ -z "${LLM_POD}" ]]; then
+    echo "Error: No running LLM pod found. Is the application deployed?" >&2
+    exit 1
+fi
```
mayuka-c: You're getting the first llm pod (What if we have multiple applications). How do we know this is the pod belonging to the exact application which we want to rotate?

This is incorrect. You need to first make DB call and then determine

mayuka-c: Use the approach that I have suggested

hack/rotate-watsonx-key/rotate-watsonx-key.sh
shwetha-s-poojary: Got it. I see the issue now - the ordering in my script is incorrect. I'll first run `application ps` with the `appName` to identify the correct `llm-*` pod and then proceed with the rest of the flow.

My confusion earlier was around the DB dependency. I understand now that the expectation is to use `application ps` for pod discovery rather than making direct DB queries.

shwetha-s-poojary:
Successful script run:

```
# ./rotate-watsonx-keys.sh rag-test
→ Discovering LLM pod for application 'rag-test'...
  Pod:       llm-b27801fca9
  Slug:      b27801fca9
  Secret:    watsonx-secret-b27801fca9
  Container: llm-b27801fca9-litellm
→ Reading TemplateID from pod label...
  TemplateID: ***
→ Reading configuration from running container 'llm-b27801fca9-litellm'...
  Image:              ***/litellm:***
  WATSONX_PROJECT_ID: ***
  WATSONX_URL:        https://us-south.ml.cloud.ibm.com
  INSTRUCT_MODEL:     ibm/granite-4-h-small
→ Rendering pod templates...
→ Stopping pod 'llm-b27801fca9'...
llm-b27801fca9
→ Removing pod 'llm-b27801fca9'...
3ea542338a1aba3210a8b872c26db5d6a02991521e2398c00e6c6d3c089a4d07
→ Removing old Podman secret 'watsonx-secret-b27801fca9'...
50c091b7df1993de7e6f39ff9
→ Replaying pod with new key (podman kube play)...
Secrets:
2722ef711770902067c763524
Pod:
e92993945a1c6be7e23b8cc26c0292bf4879359828dc357819e18ca42d4a6a66
Container:
7f0db37cf3e6f818db3a0263313058d87281d1917817fae72df896368226186c

→ Waiting for container to start...
✓ Key rotated successfully (mounted key starts with: ***...).

✓ Done. Verify end-to-end functionality via the Q&A interface.
  To confirm LiteLLM is healthy:
    podman exec llm-b27801fca9-litellm curl -s http://localhost:8000/health/readiness
# podman exec llm-b27801fca9-litellm curl -s http://localhost:8000/health/readiness
{"status":"healthy","db":"Not connected"}
```

shwetha-s-poojary:
Closing this PR since we are not moving ahead with the script.But using it as reference point for documentation.

PR #1205, opened 2026-08-03
AISERVICES-1382 added CRUD endpoints for extract jobs

manalilatkar:
In this PR:
- POST /v1/extract/jobs
- GET /v1/extract/jobs
- GET /v1/extract/jobs/{job_id}
- DELETE /v1/extract/jobs
- DELETE /v1/extract/jobs/{job_id}
- GET /v1/extract/jobs/{job_id}/result
- unit tests

### DIFF
--- a/services/extract/api/v1/jobs.py
+++ b/services/extract/api/v1/jobs.py
@@ -229,7 +229,7 @@ async def create_extract_job(
     """Validate, stage, record, and enqueue an async extraction job."""
     _check_job_admission()
     filename, source_type = _validate_and_resolve_file(file)
-    _validate_file_content(file)
+    await _validate_file_content(file)
     _resolve_schema(schema_id)
 
     job_id = str(uuid.uuid4())
--- a/services/extract/api/v1/jobs.py
+++ b/services/extract/api/v1/jobs.py
@@ -76,7 +76,7 @@ def _validate_and_resolve_file(file: UploadFile) -> tuple[str, str]:
             415, "UNSUPPORTED_FILE_TYPE",
             f"Only .txt and .md files are accepted. Received: {raw_ext}",
         )
-    return filename, ext.lstrip(".")
+    return filename, (ext or "").lstrip(".")
 
 
 def _resolve_schema(schema_id: str):
@@ -306,8 +306,8 @@ async def list_extract_jobs(
             schema_id=row.schema_id,
             status=row.status,
             document_name=row.document_name,
-            submitted_at=fmt_dt(row.submitted_at),
-            completed_at=fmt_dt(row.completed_at),
+            submitted_at=fmt_dt(row.submitted_at) or "",
+            completed_at=fmt_dt(row.completed_at) or "",
         )
         for row in rows
     ]
@@ -354,8 +354,8 @@ async def get_extract_job(job_id: str) -> JobDetailResponse:
             source_type=row.source_type
         ),
         metadata=row.job_metadata,
-        submitted_at=fmt_dt(row.submitted_at),
-        completed_at=fmt_dt(row.completed_at),
+        submitted_at=fmt_dt(row.submitted_at) or "",
+        completed_at=fmt_dt(row.completed_at) or "",
         error=row.error,
     )
 
--- a/services/extract/api/v1/jobs.py
+++ b/services/extract/api/v1/jobs.py
@@ -6,22 +6,56 @@ Exposes one router:
 - ``router`` → mounted at ``/v1/extract``
 """
 
+import uuid
+from datetime import datetime, timezone
+from typing import Optional
 
+from fastapi import APIRouter, BackgroundTasks, File, Form, Query, UploadFile
+from fastapi.responses import JSONResponse, Response
 
-from fastapi import APIRouter
+from common.error_utils import http_error_responses
+from common.misc_utils import cleanup_staging_directory, get_logger
 
-from common.misc_utils import get_logger
-
-from extract.utils.schema import (
-    SchemaValidationError
+from extract.db.manager import db_repo
+from extract.models import (
+    DocumentInfo,
+    JobCreatedResponse,
+    JobDetailResponse,
+    JobListItem,
+    JobResultResponse,
+    JobsListResponse,
+    PaginationInfo,
 )
-
+from extract.settings import settings
+from extract.state import job_limiter
+from extract.utils.job import (
+    delete_all_job_files,
+    delete_job_files,
+    read_result_file,
+    stage_uploaded_file,
+    validate_file_extension,
+)
+from extract.utils.schema import SchemaValidationError
 
 router = APIRouter()
 logger = get_logger("jobs_router")
 
+
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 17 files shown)


services/extract/app.py
```diff
@@ -2,24 +2,26 @@
...
 
-from extract.utils.schema import SchemaValidationError
+
+from extract.utils.schema import (
+    SchemaValidationError,
+    check_schema_share_in_context,
+    compute_token_counts,
+    normalize_schema,
+    validate_examples,
+    validate_json_schema_structure
```
dharaneeshvrd: why are we adding unused imports?

dharaneeshvrd (commented):
can we fix some of the warnings from bob for app.py & jobs.py?

manalilatkar:
> can we fix some of the warnings from bob for app.py & jobs.py?

@dharaneeshvrd sure, I'll work on it.

services/extract/api/v1/jobs.py
```diff
@@ -30,33 +53,414 @@ async def extract_sync():
...
+async def create_extract_job(
+    background_tasks: BackgroundTasks,
+    file: UploadFile = File(...),
+    schema_id: str = Form(...),
+    job_name: Optional[str] = Form(None),
+) -> JobCreatedResponse:
+    """Validate, stage, record, and enqueue an async extraction job."""
+    # --- Job admission check ---
+    if job_limiter.locked():
+        raise SchemaValidationError(
```
dharaneeshvrd: why are we returning SchemaValidationError on all the places?

services/extract/api/v1/jobs.py
```diff
@@ -30,33 +53,414 @@ async def extract_sync():
...
+            "RATE_LIMIT_EXCEEDED",
+            "Job concurrency limit reached. Please try again later.",
+            status=429,
+        )
+
+    # --- File extension validation ---
+    filename = file.filename or ""
```
dharaneeshvrd: let's normalise the file extension to allow capital file extension like .TXT or .MD

dharaneeshvrd: Can we do some validation on the content to make sure it has readable text or md content?
Just thinking to handle the case where user passes unsupported content with allowed extension

services/extract/api/v1/jobs.py
```diff
@@ -30,33 +53,414 @@ async def extract_sync():
...
-@router.post("/jobs", status_code=202, tags=["jobs"], include_in_schema=True)
-async def create_extract_job():
-    """Async extraction job — implementation in follow-up iteration."""
-    raise SchemaValidationError("NOT_IMPLEMENTED", "POST /v1/extract/jobs not yet implemented.", status=501)
+# ---------------------------------------------------------------------------
+# POST /v1/extract/jobs — Submit an async extraction job
+# ---------------------------------------------------------------------------
+
+@router.post(
+    "/jobs",
```
dharaneeshvrd, line 201: Is nt this?
```suggestion
    "/v1/extract/jobs",
```
if yes, please correct in all the endpoint handlers below

services/extract/api/v1/jobs.py
```diff
@@ -30,33 +53,414 @@ async def extract_sync():
...
+    return JobDetailResponse(
+        job_id=row.job_id,
+        job_name=row.job_name,
+        schema_id=row.schema_id,
+        status=row.status,
+        document=DocumentInfo(
+            name=row.document_name,
+            source_type=row.source_type,
+            digitize_job_id=row.digitize_job_id,
+            digitize_doc_id=row.digitize_doc_id,
```
dharaneeshvrd: needs cleanup

services/extract/api/v1/jobs.py
```diff
@@ -30,33 +53,414 @@ async def extract_sync():
...
+async def delete_extract_job(job_id: str) -> Response:
+    row = db_repo.get_job_by_id(job_id)
+    if row is None:
+        raise SchemaValidationError(
+            "RESOURCE_NOT_FOUND",
+            f"Job {job_id!r} not found.",
+            status=404,
+        )
+
+    if row.status in ("accepted", "in_progress"):
```
dharaneeshvrd: Is it possible for the job to be in accepted state? Won't it move to in_progress as soon as we start the bg job?
Instead we can check for final status completed or failed

services/extract/api/v1/jobs.py
manalilatkar: @dharaneeshvrd the extract part gets added in the app.py here: https://github.com/IBM/project-ai-services/blob/f3d79023c19741eeb40167d029897137af0ecccf/services/extract/app.py#L170

services/extract/api/v1/jobs.py
manalilatkar: Fair enough.

services/extract/api/v1/jobs.py
dharaneeshvrd: Ok

manalilatkar:
> can we fix some of the warnings from Bob for app.py & jobs.py?

Fixed the following things in `def create_extract_job` in extract/api/v1/jobs.py :

1. (file.filename or "").lower() 
2. except around db_repo.create_job ( Correctness (prevent orphan staging dirs)
3. Report actual bad extension in 415 message
4. Extract _check_job_admission / _validate_and_resolve_file / _resolve_schema / _stage_and_persist 

Fixed the following in `def lifespan` in app.py:

1. Move deferred imports to module top
2. Extract DB block into _initialize_database()
3. Flatten nested try/except

manalilatkar:
Following files were used to test the validate_content section:

```
# Empty file
touch empty.txt

# Binary file disguised as .txt (a small PNG)
printf '\x89PNG\r\n\x1a\n\x00\x00\x00' > binary.txt

# File with null bytes
printf 'Hello\x00World\x00Test' > nullbytes.txt

# PDF disguised as .txt
printf '%%PDF-1.7 fake pdf content here' > fakepdf.txt

# File with excessive control characters
python3 -c "
import sys
content = 'normal text ' + ''.join(chr(i) for i in range(1,9)) * 20
sys.stdout.buffer.write(content.encode('utf-8'))
" > controlchars.txt

# Unsupported extension
echo "some content" > document.csv

# Non-UTF-8 encoding (Latin-1 with high bytes)
python3 -c "
import sys
sys.stdout.buffer.write(b'Stra\xdfe in M\xfcnchen\xff\xfe\xfd\xfc')
" > latin1.txt
```

services/extract/api/v1/jobs.py
manalilatkar: my bad it slipped through. cleaned up.

services/extract/api/v1/jobs.py
manalilatkar: filename case fixed, validation added.

services/extract/api/v1/jobs.py
manalilatkar: Fixed it to raise `ExtractException` everywhere.

services/extract/app.py
manalilatkar: removed.

dharaneeshvrd (approved):

PR #1207, opened 2026-08-03
npm CVE fixes & sync package.json versions

pranithraoibm:
- Fixed the npm CVEs across all the UI folders
- Synced package.json version ranges across ui/catalog, ui/chatbot, and ui/digitize to match resolved versions in their lock files
- Addressed the review comment from [here](https://github.com/IBM/project-ai-services/pull/1183#discussion_r3665620201)

https://github.com/user-attachments/assets/1a82f888-25dc-44ae-9bd1-5f74524a2641


pranithraoibm:
@ryanmarc 
The brace-expansion CVE was fixed in brace-expansion version 1.1.17 which was published on July 29th with the security [patch](https://github.com/juliangruber/brace-expansion/releases/tag/v1.1.17).

Since minimatch@3 depends on brace-expansion@^1.1.11, a fresh npm install now automatically pulls in 1.1.18 (the latest safe version), so the vulnerability is gone without touching eslint at all. So eslint remains at v9 and all the changes are undone from this PR https://github.com/IBM/project-ai-services/pull/1183 and same fix is applied to all the UI repos.

ui/catalog/src/pages/Login/Login.tsx
```diff
@@ -30,30 +30,25 @@ const LoginPage = () => {
...
+      const locationState = location.state as LoginLocationState | null;
+      const isInactivityLogout =
+        locationState?.logoutReason === LogoutReason.INACTIVITY ||
+        sessionStorage.getItem(SESSION_STORAGE_KEYS.LOGOUT_REASON) ===
+          LogoutReason.INACTIVITY;
+
+      if (isInactivityLogout) {
+        sessionStorage.removeItem(SESSION_STORAGE_KEYS.LOGOUT_REASON);
+        sessionStorage.removeItem(SESSION_STORAGE_KEYS.LOGOUT_MESSAGE);
         navigate(location.pathname, { replace: true, state: null });
```
ryanmarc: This navigate() call never actually runs. useNavigate in react-router 8.3.0 keeps an activeRef that only flips to true in a layout effect (dist/development/lib/hooks.js:281-287), and bails out early if it is still false. A useState initializer runs during the first render, before any layout effect, so the ref is false and the navigation is dropped. You should see You should call navigate() in a React.useEffect(), not when your component is first rendered. in the console when this path fires.

I rendered this component shape and the old useEffect version side by side under jsdom with a MemoryRouter entry carrying state: { logoutReason: "INACTIVITY" }. The old version ends with location.state === null; this version ends with location.state still holding { logoutReason: "INACTIVITY" }.

The visible effect: sessionStorage is cleared fine, so the banner does not repeat through that path, but the /login history entry keeps the flag. Log out for inactivity, sign back in, hit browser Back to that entry, and you get the inactivity banner again on a session that never timed out.

Simplest fix is to keep the derivation in the initializer and move only the two side effects back into an effect:

```
const [showInactivityNotification] = useState<boolean>(() => /* just the boolean */);

useEffect(() => {
  if (!showInactivityNotification) return;
  sessionStorage.removeItem(SESSION_STORAGE_KEYS.LOGOUT_REASON);
  sessionStorage.removeItem(SESSION_STORAGE_KEYS.LOGOUT_MESSAGE);
  if (location.state) navigate(location.pathname, { replace: true, state: null });
}, []);
```

That keeps the mount-only behaviour you were going for and gets the history cleanup back.

ryanmarc (changes requested):

ui/catalog/src/pages/Login/Login.tsx
pranithraoibm: Done

ryanmarc (dismissed):

Niranjan-J-Joshi (approved):

ryanmarc (approved):

PR #1214, opened 2026-08-04
enable sync service for openshift runtime

Shubhamag12:
1. Removed unnecessary fields from reranker schema.json.
2. Introduced a RuntimeSync interface with separate Podman and OpenShift backends, allowing the sync service to switch behaviour based on the active runtime.
3. OpenShift resource validation reads the deployed Helm release manifest, extracts expected workloads/secrets/PVCs from that manifest, and checks that those resources still exist in the application namespace.

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/services/sync/openshift.go
+++ b/ai-services/internal/pkg/catalog/apiserver/services/sync/openshift.go
@@ -1,0 +1,64 @@
+package sync
+
+import (
+	"fmt"
+
+	"github.com/project-ai-services/ai-services/internal/pkg/constants"
+	"github.com/project-ai-services/ai-services/internal/pkg/runtime"
+	"github.com/project-ai-services/ai-services/internal/pkg/runtime/common"
+	runtimetypes "github.com/project-ai-services/ai-services/internal/pkg/runtime/types"
+)
+
+// openShiftSync implements RuntimeSync for the OpenShift runtime.
+type openShiftSync struct{}
+
+func newOpenShiftSync() *openShiftSync {
+	return &openShiftSync{}
+}
+
+// FetchPodStatuses fetches all pods labelled with the given templateID using the OpenShift runtime.
+func (s *openShiftSync) FetchPodStatuses(rt runtime.Runtime, templateID string) ([]*PodStatus, error) {
+	filteredPods, err := common.FetchFilteredPods(rt, templateID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch pods: %w", err)
+	}
+
+	if len(filteredPods) == 0 {
+		return nil, fmt.Errorf("no pod found with template ID: %s", templateID)
+	}
+
+	podStatuses := make([]*PodStatus, 0, len(filteredPods))
+	for _, pod := range filteredPods {
+		health := derivePodHealth(pod.Containers)
+		podStatuses = append(podStatuses, &PodStatus{
+			State:   pod.Status,
+			Health:  health,
+			PodName: pod.Name,
+		})
+	}
+
+	return podStatuses, nil
+}
+
+// derivePodHealth determines pod-level health from its container statuses.
+// A pod is healthy only when all containers report "healthy".
+func derivePodHealth(containers []runtimetypes.Container) string {
+	if len(containers) == 0 {
+		return string(constants.NotReady)
+	}
+
+	for _, c := range containers {
+		if c.Health != string(constants.Ready) {
+			return string(constants.NotReady)
+		}
+	}
+
+	return string(constants.Ready)
+}
+
+// ShouldValidateResources returns false — OpenShift uses Helm charts; all secrets, volumes,
+// and pods are managed by Helm so there are no .tmpl pod specs to count.
+func (s *openShiftSync) ShouldValidateResources() bool { return false }
+
+// Ensure openShiftSync implements RuntimeSync at compile time.
+var _ RuntimeSync = (*openShiftSync)(nil)
--- a/ai-services/internal/pkg/catalog/apiserver/services/sync/openshift.go
+++ b/ai-services/internal/pkg/catalog/apiserver/services/sync/openshift.go
@@ -93,10 +93,9 @@ func (s *openShiftSync) ValidateResources(ctx context.Context, input ResourceVal
 
 	var errorMessages []string
 
-	// Workload resources in the manifest imply how many pods/controllers we expect to exist.
-	if input.ActualPodCount < expectedCounts.Pods {
-		errorMessages = append(errorMessages,
-			fmt.Sprintf("Pod count mismatch: expected %d, found %d", expectedCounts.Pods, input.ActualPodCount))
+	// OpenShift workloads may be manually scaled, so require at least one pod when a workload exists.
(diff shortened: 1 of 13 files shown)


ai-services/internal/pkg/catalog/apiserver/services/sync/openshift.go
```diff
@@ -0,0 +1,64 @@
...
+			return string(constants.NotReady)
+		}
+	}
+
+	return string(constants.Ready)
+}
+
+// ShouldValidateResources returns false — OpenShift uses Helm charts; all secrets, volumes,
+// and pods are managed by Helm so there are no .tmpl pod specs to count.
+func (s *openShiftSync) ShouldValidateResources() bool { return false }
```
mayuka-c: What if someone manually deletes a secret/volume?. Then dont we need to update Status in DB to error?

ai-services/internal/pkg/catalog/apiserver/services/sync/podman.go
```diff
@@ -0,0 +1,105 @@
...
+// ShouldValidateResources returns true — Podman uses .tmpl pod specs and external secrets/volumes.
+func (s *podmanSync) ShouldValidateResources() bool { return true }
+
+// ValidateResources checks that the expected Podman secrets and volumes exist.
+func (s *podmanSync) ValidateResources(ctx context.Context, expectedSecretNames, expectedVolumeNames []string, rt runtime.Runtime) string {
+	return validateResources(ctx, expectedSecretNames, expectedVolumeNames, rt)
+}
+
+// validateResources is shared by both syncer implementations.
+func validateResources(ctx context.Context, expectedSecretNames, expectedVolumeNames []string, rt runtime.Runtime) string {
```
mayuka-c: I see this method is being called by above public method, why not combine and one single public one

ai-services/internal/pkg/catalog/apiserver/services/sync/openshift.go
Shubhamag12: I was wondering how should we calculate these resources the same way as we do for podman?

mayuka-c: Yes we might need to do like podman or can you verify if deleting a secret or volume makes pod to go in Error state?

Shubhamag12: tested it and it does go into error state.

mayuka-c: Yes then we need to check for those like podman

Shubhamag12: added

ai-services/internal/pkg/catalog/apiserver/services/sync/podman.go
Shubhamag12: removed

Shubhamag12:
Screenshot 2026-08-11 at 17 33 22

ai-services/internal/pkg/catalog/apiserver/services/sync/openshift.go
```diff
@@ -0,0 +1,249 @@
...
+	return counts, nil
+}
+
+// countManifestResource records only resource kinds relevant to sync validation.
+func (s *openShiftSync) countManifestResource(resource *unstructured.Unstructured, counts *ResourceCounts) {
+	switch resource.GetKind() {
+	case "Deployment", "DeploymentConfig", "StatefulSet":
+		// These controller kinds represent workload instances we expect to be present.
+		counts.Pods++
+		s.extractManifestPodResources(resource, counts)
```
mayuka-c, line 141: What if there are replicas added for deployment/StatefulSet?

ai-services/internal/pkg/catalog/apiserver/services/sync/openshift.go
```diff
@@ -0,0 +1,249 @@
...
+		return ""
+	}
+
+	var errorMessages []string
+
+	// Workload resources in the manifest imply how many pods/controllers we expect to exist.
+	if input.ActualPodCount < expectedCounts.Pods {
+		errorMessages = append(errorMessages,
+			fmt.Sprintf("Pod count mismatch: expected %d, found %d", expectedCounts.Pods, input.ActualPodCount))
+	}
```
mayuka-c: In OpenShift there might be a situation where user can change the replica count manually, so I feel its better we ensure there exists atleast one pod for a given service/component. Dont need to be the exact.

Shubhamag12: makes sense, will add that logic

ai-services/internal/pkg/catalog/apiserver/services/sync/openshift.go
Shubhamag12: based on below comment this should be fine now

ai-services/internal/pkg/catalog/apiserver/services/sync/openshift.go
Shubhamag12: added, ptal

mayuka-c (dismissed):
LGTM

Please ensure we have thoroughly validated with all negative and postive test cases

mayuka-c (approved):

Shubhamag12:
> LGTM
> 
> Please ensure we have thoroughly validated with all negative and postive test cases

Validated below cases 
1. healthy deployments
2. manual scaling
3. missing pod
4. deleted secret 
5. deleted PVC

mayuka-c:
> > LGTM
> > Please ensure we have thoroughly validated with all negative and postive test cases
> 
> Validated below cases
> 
> 1. healthy deployments
> 2. manual scaling
> 3. missing pod
> 4. deleted secret
> 5. deleted PVC

Can you test couple more?
- Deletion of namespace manually
- Deletion of deployment/Statefulset

Shubhamag12:
> Can you test couple more?
> 
>     * Deletion of namespace manually
> 
>     * Deletion of deployment/Stxatefulset

1. For namespace deletion:
Screenshot 2026-08-12 at 15 04 51

2. Deletion of deployment/Statefulset would result in no pod present, which is already covered as part of `missing pod`

mayuka-c:
> > Can you test couple more?
> > ```
> > * Deletion of namespace manually
> > 
> > * Deletion of deployment/Stxatefulset
> > ```
> 
> 1. For namespace deletion:
> 
> Screenshot 2026-08-12 at 15 04 51
> 2. Deletion of deployment/Statefulset would result in no pod present, which is already covered as part of `missing pod`

We would need to handle namespace better and mention with namespace name that is not found. Maybe can be done as part of separate PR

Shubhamag12:
> > > Can you test couple more?
> > > ```
> > > * Deletion of namespace manually
> > > 
> > > * Deletion of deployment/Stxatefulset
> > > ```
> > 
> > 
> > 
> > 1. For namespace deletion:
> > 
> > Screenshot 2026-08-12 at 15 04 51
> > 2. Deletion of deployment/Statefulset would result in no pod present, which is already covered as part of `missing pod`
> 
> We would need to handle namespace better and mention with namespace name that is not found. Maybe can be done as part of separate PR

sure, will create a PR for it

PR #1217, opened 2026-08-05
Jira 3 in a box workflow

tsambari2:


docs/product-tracking/jira-3-in-a-box-workflow.md
yussufsh, line 1: May be another folder under docs would be helpful. `jira-3-in-a-box-workflow` does not make sense at a high level.

It could be under `docs/product-tracking`/`docs/tracking-guide`/`docs/project-management` 😄

PR #1219, opened 2026-08-05
add application ps api for openshift

Shubhamag12:
Extends the GET /applications/{id}/ps endpoint to support the OpenShift runtime.

Previously ApplicationsPs was implemented only on PodmanApplicationService. This PR refactors the implementation into ApplicationServiceBase so both runtimes share the pod/container status logic, and adds runtime-specific overrides that pass the correct namespace to the runtime client:

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/repository/application_service/common.go
+++ b/ai-services/internal/pkg/catalog/apiserver/repository/application_service/common.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/google/uuid"
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog"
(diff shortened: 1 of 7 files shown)


ai-services/internal/pkg/catalog/apiserver/repository/application_service/common.go
```diff
@@ -672,4 +677,154 @@ func (s *ApplicationServiceBase) executeDeploymentAsync(parentCtx context.Contex
...
+	services []models.Service,
+) ([]types.Pod, error) {
+	componentMap := make(map[string][]types.Pod)
+
+	for _, service := range services {
+		serviceDependencies, err := s.ServiceDependencyRepo.GetDependenciesByServiceID(ctx, service.ID)
+		if err != nil {
+			logger.ErrorfCtx(ctx, "Failed to get dependencies for service %s: %v", service.ID, err)
+
+			continue
```
yussufsh: Not related to your code, but continue statements here is kind of silent log.
cc @mayuka-c what was the purpose if you remember?

Shubhamag12: done

mayuka-c (dismissed):
LGTM

mayuka-c:
Pls take care of version bumps

mayuka-c (approved):

PR #1222, opened 2026-08-06
DB updates to support Remote Deployment

mayuka-c:
- Introduces a new table `workers` to store worker info.
- Introduces worker_id as FK inside applications table (as a single worker can have multiple applications table).
- Introduces runtime_type to worker to know whether its podman or openshift.
- Db handlers for workers table and changes in application handlers to support worker_id.
- Add metadata column store domain_suffix and other infos.


yussufsh:
Will take up once the remote framework is ready.

yussufsh (approved):

PR #1223, opened 2026-08-06
Update Data Source Connectors Proposal 

sats-23:
-Moved away from multithreading to APScheduler based approach


docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -766,7 +852,7 @@ Phase 5 — close tick record  [close_sync_log]
...
+**Concurrency guard:** same-connector overlap is prevented entirely by the DB-level `sync_status` check. The APScheduler `max_instances=1` policy ensures a new scheduled tick is not dispatched while the connector's `sync_status` is already `'syncing'`. `POST /sync` uses `try_acquire_sync_lock()` — an atomic `UPDATE … WHERE sync_status != 'syncing' RETURNING id` — before dispatching `_run_tick`; if no row is returned, the request is a no-op. Cross-connector duplicate ingestion of the same brand-new checksum is handled at the DB level by `ON CONFLICT (checksum, connector_id) DO NOTHING` on every `INSERT INTO connector_document_checksum` row, so at most one `doc_id` is ever registered per `(checksum, connector_id)` pair.
```
dharaneeshvrd: Let's handle the multiple replica scenarios as well
With the help of example we saw in yesterday's discussion by registering AP scheduler itself with db.

dharaneeshvrd (commented):
Let's consider the cancelling the sync tick scenario as well

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

sats-23:
> Let's consider the cancelling the sync tick scenario as well

I have proposed a new approach where we allow instant deletion irrespective of any stage of the sync tick by implementing a cooperative cancellation via asyncio.CancelledError, PTAL

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -270,11 +282,9 @@ Response codes:
 
 | Status | Meaning |
 | --- | --- |
-| `204 No Content` | Connector removed |
+| `204 No Content` | Connector removed (was idle or tick was cancelled successfully) |
 | `404 Not Found` | Connector does not exist |
-| `409 Conflict` | A sync tick is currently running; retry after the tick completes |
-
-> **Future:** when job cancellation is introduced, the `409` guard will be replaced by an interrupt call so that DELETE can succeed at any stage of the lifecycle.
+| `408 Request Timeout` | Tick did not acknowledge cancellation within the 10 s window |
```
dharaneeshvrd: How would you cancel the underlying job?
Even if you let the current job complete, it won't be able to complete within 10secs

sats-23: Every delete call now return 204, cancel job will be implemented later

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
dharaneeshvrd, line 1: Not able to add comments inline, adding it at file level.

- We can get started with this background task based deletion for now, but before sending 204, connector needs to be marked as `delete_pending` or similar.
- Need to alter other parts of the doc like `2.2 Runtime Flow` to reflect this new deletion behaviour.
- 3.6 GET is using syncs, 3.7 DELETE is using sync

sats-23: Addressed all 3, PTAL

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -556,294 +86,324 @@ CREATE TABLE IF NOT EXISTS connector_sync_logs (
...
+        └─ Return 204 No Content immediately
 ```
 
-**Ordering guarantee:** Phase 4b (orphan removal) runs only after Phase 4a (all ingest jobs) completes.
+#### Teardown Flow
 
-**Concurrency lock:** A process-wide `ingest_lock` (e.g. `threading.Lock`) **must be held continuously from the `SELECT checksum` DB read in Phase 2 through the end of Phase 4a** (i.e. until all `INSERT INTO connector_document_checksum` rows for newly created sessions are committed). Without this lock, two connector workers racing on the same brand-new file would both observe it absent from `all_checksums`, both call `create_job`, and produce duplicate documents. The lock is released before Phase 4b begins so orphan removal does not block other ticks unnecessarily.
+```text
+_run_teardown(connector_id) (Background asyncio.Task)
+  │
```
dharaneeshvrd, line 253: before removing connector job, need to ensure, the last sync is in one of the final state(cancelled, failed or completed)

sats-23: Done, added. 
There is a recovery phase where the last sync status could be stuck in started in case of a mid-tick crash. 
I have handled that as well

docs/proposals/data-source-connectors/delete-flow.svg
dharaneeshvrd, line 1: Please add a step in _run_teardown to check for the last sync status to be in one of the final state.

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
dharaneeshvrd: I dont see this

docs/proposals/data-source-connectors/delete-flow.svg
sats-23: Sorry, order was swapped

sats-23: Updated diagram as well, PTAL

dharaneeshvrd (approved):

PR #1226, opened 2026-08-10
chore: spyre provider logic in selecting default and refactor the code

yussufsh:
Refactoring in the `selectProviderFromDeployOptions` code by reducing the number of for loops.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/create.go
+++ b/ai-services/cmd/ai-services/cmd/application/create.go
@@ -928,88 +928,66 @@ func extractProviderParams(prefix string, allParams map[string]string, providerP
 
 // selectProviderFromDeployOptions determines the provider ID for a component using deploy options.
 // Priority:
-// 1. If user provided provider-specific params (e.g., llm.vllm-cpu.model), use that provider.
-// 2. For LLM and reranker components: Use vllm-spyre by default.
-// 3. Default provider marked in deploy options.
+// 1. User-specified provider (matched against deploy options).
+// 2. Default-marked provider.
+// 3. Provider containing "spyre" in its ID.
 // 4. First available provider.
-// Returns an error if multiple providers are specified for the same component type.
+// Returns an error if multiple providers are explicitly selected for the same component type.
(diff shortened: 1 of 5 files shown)


ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -928,88 +928,66 @@ func extractProviderParams(prefix string, allParams map[string]string, providerP
...
+	if defaultProvider != "" {
+		return defaultProvider, make(map[string]string), nil
 	}
 
-	// Use default provider if marked.
-	if defaultID := findDefaultProvider(compDeployOpt); defaultID != "" {
-		return defaultID, make(map[string]string), nil
+	if spyreProvider != "" {
+		return spyreProvider, make(map[string]string), nil
 	}
```
mayuka-c, line 952: Just a query, do we need spyreProvider?
Where exactly will this be used?

mayuka-c: I see we kind of checking spyre in the id and considering it to be SpyreProvider, but do we really need to do it?

yussufsh: It is more of a fallback mechanism that will come in handy with BYO service for validation.
It will consider the priorities below...
1. Check the given provider
2. Check for default provider
3. Check for Spyre provider
4. Consider the first in the list

mayuka-c: Check for Spyre provider provides a hard check on id where we are checking for word spyre. I feel we need not do spyre check as it is applicable only for llm, reranker and embedding and we need not have to return spyre by default. Having other 3 should be fine IMO but either way I'm fine

mayuka-c (dismissed):
Rest LGTM

ai-services/cmd/ai-services/cmd/application/create.go
yussufsh: I updated it to check for "vllm-spyre" to consider only vllm components.

Shubhamag12 (approved):
lgtm

mayuka-c (approved):
LGTM

PR #1232, opened 2026-08-10
Data Source Connectors: Sync Tick Logic 

sats-23:
--New API Endpoints
POST /v1/connectors/{id}/sync — triggers a manual sync, returns sync_seq
POST /v1/connectors/{cid}/syncs/{sync_seq}/stop — cancels an in-progress sync
GET /v1/connectors/{cid}/syncs/{sync_seq} — fetch status of a specific sync

--Core Sync tick logic 
Contains batch processing, acquire DB lock, dispatch_sync (common for both), allow partial failures on ingestion

--Connector Lifecycle and Deletion 
New status for connector (DELETE_PENDING) and sync cancel (CANCEL_PENDING) that is used to intercept and initiate necessary tear down where applicable 

--Updated tests

### DIFF
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -10,6 +10,7 @@ Endpoints:
   GET    /v1/connectors
   GET    /v1/connectors/{connector_id}
   GET    /v1/connectors/{connector_id}/syncs
+  GET    /v1/connectors/{connector_id}/syncs/{sync_seq}
   POST   /v1/connectors/{connector_id}/syncs
   POST   /v1/connectors/{connector_id}/syncs/{sync_seq}/stop
 """
@@ -27,6 +28,7 @@ from digitize.connectors.models import (
     ConnectorDetailResponse,
     ConnectorListItem,
     ConnectorUpdateRequest,
+    SyncLogDetailResponse,
     SyncLogItem,
     SyncLogResponse,
     SyncStatus,
@@ -640,6 +642,58 @@ async def get_sync_history(
     except Exception as exc:
         logger.error(
             f"Unexpected error fetching sync log for {connector_id}: {exc}",
+            exc_info=True,
+        )
+        APIError.raise_error(ErrorCode.INTERNAL_SERVER_ERROR, str(exc))
+
+
+# ---------------------------------------------------------------------------
+# GET /v1/connectors/{connector_id}/syncs/{sync_seq}
+# ---------------------------------------------------------------------------
+
+@router.get(
+    "/{connector_id}/syncs/{sync_seq}",
+    response_model=SyncLogDetailResponse,
+    responses={
+        404: http_error_responses[404],
+        500: http_error_responses[500],
+    },
+    summary="Get a specific sync log entry",
+    description="Returns one sync log entry identified by its sequence number.",
+    response_description="Sync log entry",
+)
+async def get_sync(connector_id: str, sync_seq: int):
+    try:
+        connector = db_ops.get_active_connector(connector_id)
+        if connector is None:
+            APIError.raise_error(
+                ErrorCode.RESOURCE_NOT_FOUND,
+                f"Connector {connector_id!r} not found",
+            )
+
+        log = db_ops.get_sync_log(connector_id, sync_seq)
+        if log is None:
+            APIError.raise_error(
+                ErrorCode.RESOURCE_NOT_FOUND,
+                f"Sync {sync_seq} not found for connector {connector_id!r}",
+            )
+
+        return SyncLogDetailResponse(
+            seq=log.seq,
+            started_at=get_utc_timestamp(log.started_at) or "",
+            finished_at=get_utc_timestamp(log.finished_at),
+            total_files=log.total_files,
+            new_files=log.new_files,
+            removed_files=log.removed_files,
+            status=log.status,
+            error=log.error or "",
+        )
+
+    except HTTPException:
+        raise
+    except Exception as exc:
+        logger.error(
+            f"Unexpected error fetching sync {sync_seq} for {connector_id}: {exc}",
             exc_info=True,
         )
         APIError.raise_error(ErrorCode.INTERNAL_SERVER_ERROR, str(exc))
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -30,6 +30,7 @@ from digitize.connectors.models import (
     SyncLogItem,
     SyncLogResponse,
     SyncStatus,
+    SyncTriggerResponse,
 )
 from digitize.connectors.encryption import (
     encrypt_secrets,
@@ -429,6 +430,7 @@ async def get_connector(connector_id: str):
 @router.post(
     "/{connector_id}/sync",
     status_code=status.HTTP_202_ACCEPTED,
+    response_model=SyncTriggerResponse,
     responses={
         404: http_error_responses[404],
         500: http_error_responses[500],
@@ -439,7 +441,9 @@ async def get_connector(connector_id: str):
         "Safe and idempotent: if a tick is already running the request is "
         "accepted without starting a duplicate (no-op 202). "
         "The tick runs asynchronously; this endpoint returns as soon as the "
-        "task has been dispatched."
+        "task has been dispatched. "
+        "Always returns the sync_seq of the active sync — either the newly "
+        "dispatched one or the already-running one."
     ),
     response_description="Sync dispatched (or already in progress)",
 )
@@ -459,14 +463,44 @@ async def trigger_sync(connector_id: str):
         if acquired:
             asyncio.create_task(run_tick(connector_id))
             logger.info(f"Manual sync dispatched for connector {connector_id!r}")
-        
-        return Response(status_code=202)
+            # open_new_sync_log hasn't been called yet (that happens inside run_tick),
+            # so fetch the seq of the row that the background task will open shortly.
+            # Since we hold the lock we query for the highest existing seq + 1 via
+            # the active row that run_tick will create. Instead, we read it after a
+            # brief yield so the task can open the log first.
+            # Simpler and race-free: open_new_sync_log is the very first thing
+            # run_tick does — schedule a wait-for-seq helper.
+            sync_seq = await _wait_for_sync_seq(connector_id)
+        else:
+            sync_seq = db_ops.get_active_sync_seq(connector_id)
+            if sync_seq is None:
+                raise RuntimeError(
+                    f"Sync lock held but no active sync-log row found for connector {connector_id!r}"
+                )
+            logger.info(f"Sync already in progress for connector {connector_id!r}, seq={sync_seq}")
+
+        return SyncTriggerResponse(sync_seq=sync_seq)
 
     except HTTPException:
         raise
     except Exception as exc:
         logger.error(f"Unexpected error triggering sync for {connector_id}: {exc}", exc_info=True)
         APIError.raise_error(ErrorCode.INTERNAL_SERVER_ERROR, str(exc))
+
+
+async def _wait_for_sync_seq(connector_id: str, attempts: int = 10, interval: float = 0.1) -> int:
+    """Poll until run_tick's open_new_sync_log creates the active sync-log row.
+
+    Returns the seq as soon as it appears, or raises RuntimeError if it never does.
+    """
+    for _ in range(attempts):
+        await asyncio.sleep(interval)
+        seq = db_ops.get_active_sync_seq(connector_id)
+        if seq is not None:
+            return seq
+    raise RuntimeError(
+        f"Timed out waiting for sync-log row to appear for connector {connector_id!r}"
+    )
 
 
 # ---------------------------------------------------------------------------
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -32,7 +32,8 @@ from digitize.connectors.models import (
     SyncLogDetailResponse,
     SyncLogItem,
     SyncLogResponse,
-    SyncStatus,
+    ConnectorStatus,
+    SyncLogStatus,
     SyncTriggerResponse,
 )
 from digitize.connectors.encryption import (
@@ -242,7 +243,7 @@ async def delete_connector(connector_id: str):
 
         db_ops.mark_sync_delete_pending(connector_id)
 
-        if connector.sync_status != SyncStatus.SYNCING:
+        if connector.sync_status != ConnectorStatus.SYNCING:
             # No tick running — kick off teardown ourselves.
             asyncio.create_task(_run_teardown(connector_id))
 
@@ -274,7 +275,7 @@ async def _run_teardown(connector_id: str) -> None:
     try:
         # Step 1: guard — verify the latest sync-log is terminal before
         # proceeding.  If it is still 'started' (stuck), cancel it now.
-        _TERMINAL_STATUSES = {SyncStatus.CANCELLED, SyncStatus.FAILED, SyncStatus.COMPLETED}
+        _TERMINAL_STATUSES = {SyncLogStatus.CANCELLED, SyncLogStatus.FAILED, SyncLogStatus.COMPLETED}
         logs, _ = db_ops.list_sync_logs(connector_id, limit=1)
         if logs:
             latest_log = logs[0]
@@ -287,7 +288,7 @@ async def _run_teardown(connector_id: str) -> None:
                 db_ops.close_sync_log(
                     connector_id=connector_id,
                     seq=latest_log.seq,
-                    status=SyncStatus.CANCELLED,
+                    status=SyncLogStatus.CANCELLED,
                     error="Cancelled by connector deletion",
                 )
 
@@ -567,7 +568,7 @@ async def cancel_sync(connector_id: str, sync_seq: int):
                 ErrorCode.RESOURCE_NOT_FOUND,
                 f"Connector {connector_id!r} not found",
             )
-        if connector.sync_status != SyncStatus.SYNCING:
+        if connector.sync_status != ConnectorStatus.SYNCING:
             APIError.raise_error(
                 ErrorCode.RESOURCE_LOCKED,
                 "No sync is currently running for this connector.",
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -152,18 +152,24 @@ async def update_connector(connector_id: str, body: ConnectorUpdateRequest):
         if body.connector_name is None and body.allowed_extensions is None and body.connection_details is None:
             return Response(status_code=200)
 
+        existing = db_ops.get_active_connector(connector_id)
+        if existing is None:
+            APIError.raise_error(
+                ErrorCode.RESOURCE_NOT_FOUND,
+                f"Connector {connector_id!r} not found",
+            )
+        if existing.sync_status == ConnectorStatus.DELETE_PENDING:
+            APIError.raise_error(
+                ErrorCode.RESOURCE_LOCKED,
+                f"Connector {connector_id!r} is pending deletion and cannot be updated",
+            )
+
         key_path = _get_key_path()
 
         # If connection_details is being updated, we need to merge with existing
         # encrypted details so untouched keys stay encrypted and intact.
         merged_details: Optional[dict] = None
         if body.connection_details is not None:
-            existing = db_ops.get_active_connector(connector_id)
-            if existing is None:
-                APIError.raise_error(
-                    ErrorCode.RESOURCE_NOT_FOUND,
-                    f"Connector {connector_id!r} not found",
-                )
             merged_details = merge_and_encrypt_partial(
                 existing.type,
                 existing.connection_details,
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -376,6 +376,7 @@ async def list_connectors():
                 last_sync_at=get_utc_timestamp(c.last_sync_at),
                 sync_status=c.sync_status,
                 last_sync_error=c.last_sync_error,
+                error=c.error,
                 total_files=c.total_files,
             )
             for c in connectors
@@ -424,6 +425,7 @@ async def get_connector(connector_id: str):
             last_sync_at=get_utc_timestamp(connector.last_sync_at),
             sync_status=connector.sync_status,
             last_sync_error=connector.last_sync_error,
+            error=connector.error,
             connection_details=strip_secrets(connector.type, connector.connection_details or {}),
             total_files=connector.total_files,
         )
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -559,6 +559,11 @@ async def cancel_sync(connector_id: str, sync_seq: int):
                 ErrorCode.RESOURCE_NOT_FOUND,
                 f"Connector {connector_id!r} not found",
             )
+        if connector.sync_status != SyncStatus.SYNCING:
+            APIError.raise_error(
+                ErrorCode.RESOURCE_LOCKED,
+                "No sync is currently running for this connector.",
+            )
 
         active_seq = db_ops.get_active_sync_seq(connector_id)
         if active_seq is None or active_seq != sync_seq:
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -621,7 +621,7 @@ async def get_sync_history(
 
         items = [
             SyncLogItem(
-                id=log.id,
+                seq=log.seq,
                 started_at=get_utc_timestamp(log.started_at) or "",
                 finished_at=get_utc_timestamp(log.finished_at),
                 total_files=log.total_files,
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -11,6 +11,7 @@ Endpoints:
   GET    /v1/connectors/{connector_id}
   GET    /v1/connectors/{connector_id}/syncs
   POST   /v1/connectors/{connector_id}/sync
+  DELETE /v1/connectors/{connector_id}/sync
 """
 
 from typing import List, Optional
@@ -444,6 +445,54 @@ async def trigger_sync(connector_id: str):
         raise
     except Exception as exc:
         logger.error(f"Unexpected error triggering sync for {connector_id}: {exc}", exc_info=True)
+        APIError.raise_error(ErrorCode.INTERNAL_SERVER_ERROR, str(exc))
+
+
+# ---------------------------------------------------------------------------
+# DELETE /v1/connectors/{connector_id}/sync
+# ---------------------------------------------------------------------------
+
+@router.delete(
+    "/{connector_id}/sync",
+    status_code=status.HTTP_204_NO_CONTENT,
+    responses={
+        404: http_error_responses[404],
+        409: http_error_responses[409],
+        500: http_error_responses[500],
+    },
+    summary="Stop a running sync",
+    description=(
+        "Signals the active sync tick to stop at its next cancellation checkpoint. "
+        "Returns 204 immediately; the tick exits asynchronously and the sync log "
+        "is marked 'cancelled'. The connector remains and resumes its normal "
+        "schedule on the next interval. "
+        "Returns 409 if no sync is currently running."
+    ),
+    response_description="Stop signal sent",
+)
+async def cancel_sync(connector_id: str):
+    try:
+        connector = db_ops.get_active_connector(connector_id)
+        if connector is None:
+            APIError.raise_error(
+                ErrorCode.RESOURCE_NOT_FOUND,
+                f"Connector {connector_id!r} not found",
+            )
+
+        signalled = db_ops.mark_sync_cancel_pending(connector_id)
+        if not signalled:
+            APIError.raise_error(
+                ErrorCode.RESOURCE_LOCKED,
+                "No sync is currently running for this connector.",
+            )
+
+        logger.info(f"Cancel-sync signal sent for connector {connector_id!r}")
+        return Response(status_code=204)
+
+    except HTTPException:
+        raise
+    except Exception as exc:
+        logger.error(f"Unexpected error cancelling sync for {connector_id}: {exc}", exc_info=True)
         APIError.raise_error(ErrorCode.INTERNAL_SERVER_ERROR, str(exc))
 
 
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -14,6 +14,7 @@ Endpoints:
   DELETE /v1/connectors/{connector_id}/sync
 """
 
+import asyncio
 from typing import List, Optional
 
 from fastapi import APIRouter, HTTPException, Query, Response, status
@@ -199,27 +200,28 @@ async def update_connector(connector_id: str, body: ConnectorUpdateRequest):
     status_code=status.HTTP_204_NO_CONTENT,
     responses={
         404: http_error_responses[404],
-        409: http_error_responses[409],
         500: http_error_responses[500],
     },
     summary="Detach and delete a connector",
     description=(
-        "Removes a connector and all its associated state. "
-        "Rejected with 409 if a sync tick is currently in progress. "
-        "Documents owned exclusively by this connector are deleted. "
-        "Staging directories are cleaned up best-effort."
+        "Non-blocking detachment. Always returns 204 immediately. "
+        "If a sync tick is running the connector is marked delete_pending and "
+        "the tick handles teardown; otherwise teardown runs as a background task."
     ),
-    response_description="No content on successful deletion",
+    response_description="No content — teardown proceeds in the background",
 )
 async def delete_connector(connector_id: str):
     """
-    DELETE stub for PR3 (no worker yet):
-    1. Check connector exists → 404 if not
-    2. Guard: check if sync is in progress (sync_status == 'syncing') → 409
-    3. Snapshot checksums owned by this connector
-    4. Remove ownership rows and delete documents when last owner
-    5. Delete the connector row
-    6. Best-effort staging dir cleanup
+    Fast, non-blocking DELETE:
+
+    Case A — sync_status == 'syncing':
+        Mark DELETE_PENDING. The running tick will hit _check_delete_pending at
+        its next checkpoint, cancel itself, and dispatch teardown.
+        Return 204 immediately.
+
+    Case B — sync_status != 'syncing':
+        Mark DELETE_PENDING, dispatch asyncio.create_task(_run_teardown(...)),
+        return 204 immediately.
     """
     try:
         connector = db_ops.get_active_connector(connector_id)
@@ -229,18 +231,35 @@ async def delete_connector(connector_id: str):
                 f"Connector {connector_id!r} not found",
             )
 
-        # Guard: reject if a tick is currently running
-        if connector.sync_status == SyncStatus.SYNCING:
-            APIError.raise_error(
-                ErrorCode.RESOURCE_LOCKED,
-                "A sync tick is currently running for this connector. "
-                "Retry after the tick completes.",
-            )
+        db_ops.mark_sync_delete_pending(connector_id)
 
-        # Snapshot checksums owned by this connector
+        if connector.sync_status != SyncStatus.SYNCING:
+            # No tick running — kick off teardown ourselves.
+            asyncio.create_task(_run_teardown(connector_id))
+
+        return Response(status_code=204)
+
+    except HTTPException:
+        raise
+    except Exception as exc:
+        logger.error(f"Unexpected error deleting connector {connector_id}: {exc}", exc_info=True)
+        APIError.raise_error(ErrorCode.INTERNAL_SERVER_ERROR, str(exc))
+
+
+async def _run_teardown(connector_id: str) -> None:
+    """
+    Background teardown task for connector deletion (Case B: no tick running).
+
+    Steps mirror the spec teardown flow:
+      C. Snapshot checksums owned by this connector
+      D. Remove ownership rows; delete documents when last owner
+      E. Delete the connector row (cascades to sync_logs)
+      F. Sweep residual batch staging directories
+    """
+    logger.info(f"Starting teardown for connector {connector_id!r}")
+    try:
+        # Step C+D: remove checksum ownership; delete orphaned documents
         owned_checksums = db_ops.list_connector_checksums(connector_id)
-
-        # Remove ownership rows; delete documents when last owner
         for checksum in owned_checksums:
             try:
                 remaining, doc_id = db_ops.remove_connector_checksum_entry(connector_id, checksum)
@@ -253,7 +272,7 @@ async def delete_connector(connector_id: str):
                     exc_info=True,
                 )
 
-        # Delete the connector row (cascades to connector_sync_logs)
+        # Step E: delete the connector row (cascades to connector_sync_logs)
         deleted = db_ops.delete_active_connector(connector_id)
         if not deleted:
             logger.warning(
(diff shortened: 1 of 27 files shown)


dharaneeshvrd:
Update PR description

services/digitize/connectors/models.py
```diff
@@ -21,23 +21,28 @@ class SyncStatus:
...
+        UP_TO_DATE    ──► SYNCING       ──► UP_TO_DATE    (tick completed cleanly)
+                            └──► OUT_OF_SYNC              (tick finished with errors)
+        UP_TO_DATE    ──► DELETE_PENDING                  (DELETE, no active sync)
+        SYNCING       ──► DELETE_PENDING                  (DELETE arrived mid-sync)
 
     ConnectorSyncLog.status lifecycle:
         STARTED ──► COMPLETED  (all files processed successfully)
-                └──► FAILED    (fatal tick error or partial failure)
+                ├──► FAILED    (fatal tick error or partial failure)
+                └──► CANCELLED (tick was cancelled via DELETE_PENDING)
```
dharaneeshvrd: how difficult/complicated is it to add DELETE /v1/connectors/{id}/sync to stop the running sync and adding cancel_pending status to ConnectorSyncLog.status?

dharaneeshvrd: If it is a low hanging fruit, we can add it.

services/digitize/connectors/sync_tick.py
```diff
@@ -0,0 +1,344 @@
...
+        scanned_files: list[tuple[str, str]] = await asyncio.to_thread(scanner.scan)
+
+        known_checksums: set[str] = set(list_connector_checksums(connector_id))
+        all_checksums: set[str] = set(list_all_checksums())
+
+        ingest_list, orphan_checksums = _classify(
+            connector_id, scanned_files, known_checksums, all_checksums
+        )
+
+        update_sync_log(sync_seq, total_files=len(scanned_files))
```
dharaneeshvrd: dont we need to update the new files and files to remove as well now itself?

services/digitize/connectors/sync_tick.py
```diff
@@ -0,0 +1,344 @@
...
+    The caller is responsible for acquiring the sync lock (sync_status='syncing')
+    before calling this coroutine.  open_new_sync_log() is called here to open
+    the sync-log row.
+    """
+    config = get_active_connector(connector_id)
+    if config is None:
+        logger.error(f"Connector {connector_id!r} not found; tick aborted")
+        return
+
+    sync_seq: int = open_new_sync_log(connector_id)
```
dharaneeshvrd: we need to mark the connector's status as syncing

services/digitize/connectors/sync_tick.py
```diff
@@ -0,0 +1,344 @@
...
+                job_name=job_name,
+            )
+
+            for checksum, filename in checksum_to_filename.items():
+                add_connector_checksum_entry(connector_id, checksum, doc_id_dict[filename])
+
+            await asyncio.to_thread(ingest, batch_dir, job_id, doc_id_dict)
+
+            new_count += len(batch)
+            update_sync_log(sync_seq, new_files=new_count)
```
dharaneeshvrd: updating new files should be just one time after classify
if there are some files goes to failed state during ingestion, it should be part of failed files list

services/digitize/connectors/sync_tick.py
```diff
@@ -0,0 +1,344 @@
...
+    """Download and ingest *ingest_list* in batches of up to 10 files.
+
+    Each batch of up to 10 files is downloaded into a single staging directory,
+    registered as a single job, and passed to ``ingest()`` together.
+
+    Returns (new_count, failed_count).  Staging directories are removed after
+    each batch regardless of success.
+    """
+    staging_base = settings.digitize.staging_dir / "connectors"
+    new_count = 0
```
dharaneeshvrd: I dont think this is required
only failed count is enough

services/digitize/connectors/sync_tick.py
```diff
@@ -0,0 +1,344 @@
...
+        checksum_to_filename: dict[str, str] = {}
+
+        try:
+            for remote_path, checksum in batch:
+                filename = Path(remote_path).name
+                local_checksum = await asyncio.to_thread(
+                    scanner.download_to, remote_path, batch_dir / filename
+                )
+                # Cancellation checkpoint: bail after each blocking download.
+                _check_delete_pending(connector_id)
```
dharaneeshvrd: dont need this for every file
maybe after all the 10 files are downloaded

services/digitize/connectors/sync_tick.py
```diff
@@ -0,0 +1,344 @@
...
+                operation=OperationType.INGESTION,
+                output_format=OutputFormat.JSON,
+                documents_info=filenames,
+                job_name=job_name,
+            )
+
+            for checksum, filename in checksum_to_filename.items():
+                add_connector_checksum_entry(connector_id, checksum, doc_id_dict[filename])
+
+            await asyncio.to_thread(ingest, batch_dir, job_id, doc_id_dict)
```
dharaneeshvrd, line 332: are we not modifying the ingest method to handle connector related changes?
if there are not a lot of changes, we can do in this PR itself

services/digitize/connectors/sync_tick.py
```diff
@@ -0,0 +1,344 @@
...
+
+    removed = 0
+    for checksum in orphan_checksums:
+        try:
+            remaining, doc_id = remove_connector_checksum_entry(connector_id, checksum)
+            removed += 1
+            if remaining == 0 and doc_id:
+                await asyncio.to_thread(_best_effort_delete_document, doc_id)
+        except Exception as exc:
+            logger.error(
```
dharaneeshvrd, line 370: dont we need to propagate the exception and mark the sync as failed?

dharaneeshvrd (commented):
Let's add post /v1/connectors/{id}/sync endpoint as well
It would just call the run_sync_tick in a background task and come out.

services/digitize/connectors/models.py
sats-23: Added! But we will need to handle cancel job more carefully in this case!

services/digitize/connectors/sync_tick.py
sats-23: Yes, I have already sandwiched it into a single method for ease of use. 
Refer - https://github.com/IBM/project-ai-services/blob/3d6233c9e33b9712ea8a59d3c16ab28572e5e8cf/services/digitize/db/manager.py#L1067-L1071

services/digitize/connectors/sync_tick.py

services/digitize/connectors/sync_tick.py
sats-23: I guess there is no need. We are already handling adding to connector document checksum separately. 
Also reused polling (wait for job status). Is there any other scenario where we may need this?

dharaneeshvrd: we need to mark the document as connector document right? to not return on the GET call to avoid displaying the digitise service UI.

services/digitize/connectors/sync_tick.py
dharaneeshvrd: It's better to do it explicitly, so when we reuse this method and set the syncing status even when its not needed.

services/digitize/api/v1/connectors.py
```diff
@@ -399,6 +401,101 @@ async def get_connector(connector_id: str):
...
+                "No sync is currently running for this connector.",
+            )
+
+        logger.info(f"Cancel-sync signal sent for connector {connector_id!r}")
+        return Response(status_code=204)
+
+    except HTTPException:
+        raise
+    except Exception as exc:
+        logger.error(f"Unexpected error cancelling sync for {connector_id}: {exc}", exc_info=True)
```
manju956, line 654: Do we need retry here?

services/digitize/connectors/sync_tick.py
sats-23: Done, used "raise" which will be caught and mark sync failed

services/digitize/connectors/sync_tick.py
sats-23: Removed failed count, ensured to update count of others after classify

services/digitize/connectors/sync_tick.py
sats-23: Removed failed count, ensured to update count of others after classify

services/digitize/connectors/sync_tick.py
sats-23: Removed failed count, ensured to update count of others after classify

services/digitize/connectors/sync_tick.py
```diff
@@ -0,0 +1,342 @@
...
+        # checksum → filename for post-ingest checksum registration
+        checksum_to_filename: dict[str, str] = {}
+
+        try:
+            for remote_path, checksum in batch:
+                filename = Path(remote_path).name
+                local_checksum = await asyncio.to_thread(
+                    scanner.download_to, remote_path, batch_dir / filename
+                )
+                scanner.verify_integrity(local_checksum, checksum)
```
manju956: What action do we need to take if integrity of few files in the batch is compromised? ignore them?

services/digitize/connectors/sync_tick.py
manju956: +1

services/digitize/api/v1/connectors.py
sats-23: I guess we can add retry at multiple phases, most importantly at scanner connect, download and other phases. 
I suggest to create a ticket and take it later with detailed analysis after the happy path is achieved!

sats-23: Created - https://jsw.ibm.com/browse/AISERVICES-1586

services/digitize/connectors/sync_tick.py
```diff
@@ -0,0 +1,342 @@
...
+
+# ---------------------------------------------------------------------------
+# Phase 5 helpers
+# ---------------------------------------------------------------------------
+
+def _complete_tick(sync_seq: int, connector_id: str) -> None:
+    close_sync_log(
+        connector_id=connector_id,
+        seq=sync_seq,
+        status="completed",
```
manju956: better to use enums for various status.

services/digitize/api/v1/connectors.py
manju956: Sure, can be done as separate activity.

services/digitize/connectors/sync_tick.py
sats-23: Yes, I've added changes to skip for now, but in future, we will need to handle it better by keeping track of failed count and retry logic if download is failed and so on

manju956: Ok.

services/digitize/api/v1/connectors.py
```diff
@@ -399,6 +401,101 @@ async def get_connector(connector_id: str):
...
+        connector = db_ops.get_active_connector(connector_id)
+        if connector is None:
+            APIError.raise_error(
+                ErrorCode.RESOURCE_NOT_FOUND,
+                f"Connector {connector_id!r} not found",
+            )
+
+        acquired = db_ops.try_acquire_sync_lock(connector_id)
+        if acquired:
+            asyncio.create_task(run_tick(connector_id))
```
dharaneeshvrd: let's use background_tasks.add_task to keep it uniform with other bg job usage.

services/digitize/connectors/models.py
dharaneeshvrd: how different is to handle the connector's delete_pending status?
I thought on the check_delete_pending method, this cancel_pending would go as an additional status check

services/digitize/api/v1/connectors.py
dharaneeshvrd: Retry for this API is not needed right, its just updating the db with this status

services/digitize/connectors/models.py
```diff
@@ -21,23 +21,29 @@ class SyncStatus:
...
     """
 
     # Connector.sync_status values
     UP_TO_DATE = "up to date"
     SYNCING = "syncing"
     OUT_OF_SYNC = "out of sync"
+    DELETE_PENDING = "delete pending"
+    CANCEL_PENDING = "cancel pending"
```
dharaneeshvrd: why its here?
it should be part of `ConnectorSyncLog.status`

services/digitize/connectors/sync_tick.py
sats-23: Updated

services/digitize/db/manager.py
```diff
@@ -1034,6 +1057,54 @@ def delete_connector_checksum(
...
+        Returns True if the update landed (a tick was running and is now
+        signalled to cancel), False if no tick was running or connector not found.
+        """
+        try:
+            with get_db_session() as session:
+                result = session.execute(
+                    update(Connector)
+                    .where(
+                        Connector.id == connector_id,
+                        Connector.sync_status == SyncStatus.SYNCING,
```
dharaneeshvrd: it should change respect sync's status

services/digitize/api/v1/connectors.py
```diff
@@ -10,6 +10,8 @@
   GET    /v1/connectors
   GET    /v1/connectors/{connector_id}
   GET    /v1/connectors/{connector_id}/syncs
+  POST   /v1/connectors/{connector_id}/sync
+  DELETE /v1/connectors/{connector_id}/sync
```
dharaneeshvrd:
```suggestion
  POST   /v1/connectors/{connector_id}/syncs
  POST /v1/connectors/{connector_id}/syncs/{sync_seq}/stop
```
Let's use plural

services/digitize/api/v1/connectors.py
```diff
@@ -399,6 +401,101 @@ async def get_connector(connector_id: str):
...
+
+    except HTTPException:
+        raise
+    except Exception as exc:
+        logger.error(f"Unexpected error triggering sync for {connector_id}: {exc}", exc_info=True)
+        APIError.raise_error(ErrorCode.INTERNAL_SERVER_ERROR, str(exc))
+
+
+# ---------------------------------------------------------------------------
+# DELETE /v1/connectors/{connector_id}/sync
```
dharaneeshvrd: Let's use
```suggestion
# POST /v1/connectors/{connector_id}/syncs/{sync_seq}/stop
```

services/digitize/connectors/models.py
dharaneeshvrd: I see check_delete_pending already handling this status but checking in connectors status, any other things required other than checking from sync log

services/digitize/connectors/sync_tick.py
dharaneeshvrd: @manju956 we need to brainstorm on where to add the documents that are triggered from connector to the conversion tasks to add them into the digitize queue

services/digitize/connectors/sync_tick.py
```diff
@@ -0,0 +1,343 @@
...
+
+            for checksum, filename in checksum_to_filename.items():
+                add_connector_checksum_entry(connector_id, checksum, doc_id_dict[filename])
+
+            await asyncio.to_thread(ingest, batch_dir, job_id, doc_id_dict)
+
+            await _wait_for_job(job_id, connector_id)
+
+        except Exception as exc:
+            logger.warning(
```
dharaneeshvrd, line 340: > dont we need to propagate the exception and mark the sync as failed?

here too

services/digitize/connectors/sync_tick.py
sats-23: We are using a different filter on the GET for documents, where we are excluding connector checksums 
https://github.com/IBM/project-ai-services/blob/23b228afc912b18e443af5698864901d8824c61a/services/digitize/db/manager.py#L467-L473

services/digitize/api/v1/connectors.py
sats-23: How will the user know the sync sequence?

dharaneeshvrd: POST /syncs should return and via GET all too

services/digitize/connectors/models.py
sats-23: Because the "check_delete_pending" which is now made generic for both delete connector and cancel sync is used at connector table level to check if it is in syncing state/not. And us setting state at this table is picked by check_delete_pending to proceed to raise CancelledError

dharaneeshvrd: yeah but connector should not reflect the individual sync's status right?
we need to alter the logic to check in both the tables.

services/digitize/db/manager.py
sats-23: Can you please rephrase? Didn't understand

dharaneeshvrd: I meant, it should update the  in connector sync log table's respective sync entry's status as cancel_pending.

services/digitize/api/v1/connectors.py
sats-23: I suggest let's keep it simple like /syncs/stop. Easy for us also to process.

dharaneeshvrd: Agree it would be easy, but it won't follow the REST standards.

services/digitize/connectors/sync_tick.py
dharaneeshvrd: ok hope no other change is required

sats-23: Yes, no other

services/digitize/connectors/models.py
sats-23: But sync_status on connector reflects the latest sync status right?
We could read from the other table as well, but need to consider if there is no problems from blind read on last entry

dharaneeshvrd: > But sync_status on connector reflects the latest sync status right?

But the status lifecycle is different, it's not one to one mapping.
It won't be blind if user stops on the specific sync seq via POST /v1/connectors/{cid}/syncs/{sync_seq}/stop

services/digitize/connectors/sync_tick.py
sats-23: Then, if a specific batch fails, we will end up skipping the entire tick and marking it failed. 
Should we do it that way or allow partial failure?

services/digitize/api/v1/connectors.py
sats-23: Updated endpoint

services/digitize/api/v1/connectors.py
sats-23: Done, updated

services/digitize/connectors/models.py
sats-23: Done, cancel sync updates on ConnectorSyncLog and delete connector updates on Connectors Table respectively

services/digitize/connectors/sync_tick.py
dharaneeshvrd: ok we can continue the remaining batches, but the error should be propagated even if one batch fails and sync log status should be failed and connector status should be out of sync

services/digitize/api/v1/connectors.py
sats-23: Bob says:

Cons of switching to BackgroundTasks.add_task
1. run_tick may be cut short mid-sync
BackgroundTasks runs tasks inside the ASGI request scope. Once the scope closes (client disconnects, server timeout, graceful shutdown), Starlette stops waiting. A long-running sync tick — crawling files, hashing, ingesting — can easily outlast a normal HTTP connection lifetime. With asyncio.create_task the task is owned by the event loop and keeps running regardless.

2. _wait_for_sync_seq polling would still be needed — and might get worse
_wait_for_sync_seq polls the DB because the task and the endpoint run concurrently. BackgroundTasks runs after the response is sent, so you'd have the opposite problem: the endpoint can't poll at all because the task hasn't started yet when the response goes out. You'd have to return a sync_seq of null or restructure the whole response contract.

3. No way to get the sync_seq back to the caller
Right now the flow is:

create_task → task starts immediately → _wait_for_sync_seq polls → seq returned in 202

With BackgroundTasks, the task starts after the 202 is already sent. You lose the ability to return a meaningful sync_seq in the response — the SyncTriggerResponse would always be missing it at dispatch time.

4. Cancellation signal timing changes
cancel_sync writes a DB flag that run_tick polls at checkpoints. This works fine today because the task is already running when the cancel request arrives. With BackgroundTasks, if the cancel request arrives before the response is fully flushed (rare but possible under load), the tick hasn't started yet and the cancel signal would be a no-op or cause a stale-seq 409.

5. The lock acquired before the task is released without a running task
At line 462, try_acquire_sync_lock acquires the lock before the task is scheduled. With asyncio.create_task the task starts on the next await (the poll), so the lock is valid immediately. With BackgroundTasks there's a window — between the 202 going out and the task actually starting — where the lock is held but no task is running. A concurrent trigger request would see acquired=False (lock held) but get_active_sync_seq returns None, hitting the RuntimeError at line 477–479.

Bottom line: the asyncio.create_task choice here is load-bearing. The entire sync_seq return contract, the lock/poll flow, and the cancellation mechanism all assume the task is running concurrently with the endpoint, not after it. Switching would require rethinking all three.

dharaneeshvrd: Let's go with asyncio.create_task itself, it seems to be the appropriate option to run the background tasks that we have.

dharaneeshvrd: Let's change it in other usage as well

services/digitize/api/v1/connectors.py
```diff
@@ -227,18 +232,35 @@ async def delete_connector(connector_id: str):
...
-            APIError.raise_error(
-                ErrorCode.RESOURCE_LOCKED,
-                "A sync tick is currently running for this connector. "
-                "Retry after the tick completes.",
-            )
+        db_ops.mark_sync_delete_pending(connector_id)
 
-        # Snapshot checksums owned by this connector
-        owned_checksums = db_ops.list_connector_checksums(connector_id)
+        if connector.sync_status != SyncStatus.SYNCING:
```
dharaneeshvrd: Like we discussed there is no need for this

services/digitize/api/v1/connectors.py
```diff
@@ -227,18 +232,35 @@ async def delete_connector(connector_id: str):
...
+    Background teardown task for connector deletion (Case B: no tick running).
+
+    Steps mirror the spec teardown flow:
+      C. Snapshot checksums owned by this connector
+      D. Remove ownership rows; delete documents when last owner
+      E. Delete the connector row (cascades to sync_logs)
+      F. Sweep residual batch staging directories
+    """
+    logger.info(f"Starting teardown for connector {connector_id!r}")
+    try:
```
dharaneeshvrd, line 280: missing a step related to checking the last sync log to be in final state before proceeding to the tear down

services/digitize/api/v1/connectors.py
```diff
@@ -399,6 +423,118 @@ async def get_connector(connector_id: str):
...
+        connector = db_ops.get_active_connector(connector_id)
+        if connector is None:
+            APIError.raise_error(
+                ErrorCode.RESOURCE_NOT_FOUND,
+                f"Connector {connector_id!r} not found",
+            )
+
+        acquired = db_ops.try_acquire_sync_lock(connector_id)
+        if acquired:
+            sync_seq = db_ops.open_new_sync_log(connector_id)
```
dharaneeshvrd: `db_ops.open_new_sync_log` let's rename this, to reflect that it would set the connector's status also to syncing.

services/digitize/db/manager.py
```diff
@@ -1034,6 +1057,124 @@ def delete_connector_checksum(
...
+                    .values(sync_status=SyncStatus.SYNCING)
+                    .returning(Connector.id)
+                ).one_or_none()
+                return result is not None
+        except SQLAlchemyError as e:
+            logger.error(f"DB error in try_acquire_sync_lock({connector_id}): {e}", exc_info=True)
+            raise
+
+    @staticmethod
+    def mark_sync_cancel_pending(connector_id: str) -> bool:
```
dharaneeshvrd, line 1126: why are we involving connector here?
user is passing the sync seq anyway in the stop request, just need to update the connector sync log's respective sync entry

services/digitize/api/v1/connectors.py
sats-23: But we can't directly tear down mid-tick right? We need that tick to exit before being able to delete connector. Else we risk dangling resources

services/digitize/api/v1/connectors.py
sats-23: Done, added

services/digitize/api/v1/connectors.py
sats-23: I suggest we document it in comments. 
Something like "update_status_to_syncing" would be less descriptive of fresh entries in new table

services/digitize/db/manager.py
sats-23: Connectors is used to verify if the connector is in "syncing" state. 
It is a form of 2 step lock:
1. Check if the sync seq provided by user is pointing to the latest sync (handles wrong sync_seq)
2. Check if connector is in syncing state to update the sync_seq's state

services/digitize/connectors/models.py
sats-23: Updated with check_interrupt_call

services/digitize/connectors/sync_tick.py
sats-23: But it will end up requiring 2 calls everywhere that shouldn't be forgotten by future maintainers. 
Is there any specific case where we may need to call this separately?

services/digitize/db/manager.py
sats-23: Done, updated

services/digitize/db/manager.py
dharaneeshvrd: The check should happen at the API handler level, not on this method

services/digitize/api/v1/connectors.py
dharaneeshvrd: I would prefer the method name itself reflects that, when a user reads a code that calls this method, he would think this would open only a new sync log.

services/digitize/api/v1/connectors.py
dharaneeshvrd: yes but that should be checked on the tear down method right?

services/digitize/connectors/sync_tick.py
dharaneeshvrd: I m ok with doing 2 transactions in the same method, but as per this [thread](https://github.com/IBM/project-ai-services/pull/1232#discussion_r3773219813), let's rename the method accordingly

services/digitize/db/manager.py
sats-23: Done

services/digitize/api/v1/connectors.py
sats-23: In the _handle_interrupt, we are directly calling tear down method.
Here, we need it only for case when it is not syncing

services/digitize/connectors/sync_tick.py
sats-23: Done, updated

services/digitize/api/v1/connectors.py
sats-23: How about renaming it to "init_sync_log_and_set_syncing" or "open_sync_log_with_status"?

services/digitize/api/v1/connectors.py
sats-23: Updated

services/digitize/api/v1/connectors.py
dharaneeshvrd: Ok this is a different logic than what's discussed. But this also looks good. Let's go ahead with this.

services/digitize/api/v1/connectors.py
```diff
@@ -227,18 +232,59 @@ async def delete_connector(connector_id: str):
...
+        logs, _ = db_ops.list_sync_logs(connector_id, limit=1)
+        if logs:
+            latest_log = logs[0]
+            if latest_log.status not in _TERMINAL_STATUSES:
+                logger.warning(
+                    f"Connector {connector_id!r} latest sync-log (seq={latest_log.seq}) "
+                    f"has non-terminal status {latest_log.status!r}; "
+                    "marking it cancelled before teardown"
+                )
+                db_ops.close_sync_log(
```
dharaneeshvrd: I dont think we should close the sync blindly right?
We might need to wait for it to automatically move to terminal status.

But now since the triggering point is different for syncing case, this check itself may not be required, wdyt?

sats-23: End of the day, if the status is changed or not, we will remove all entries. 
Any thread dangling also is not possible coz that only will interrupt and call tear down.

services/digitize/api/v1/connectors.py
```diff
@@ -72,14 +79,15 @@ def _get_key_path() -> str:
     response_description="Connector created; worker start scheduled",
 )
 async def create_connector(body: ConnectorCreateRequest):
+    connector_id = body.connector_id or str(uuid.uuid4())
     try:
         key_path = _get_key_path()
         encrypted_details = encrypt_secrets(body.type, body.connection_details, key_path)
 
         sync_interval = settings.digitize.connector.sync_interval_seconds
```
dharaneeshvrd, line 89: we need to add a validation for the connection details to check the provided details are valid or not by simply creating the scanner and check the connection. Maybe you can take care part of the APScheduler changes.

services/digitize/connectors/models.py
```diff
@@ -7,37 +7,57 @@
...
+from pydantic import BaseModel, Field, field_validator
 
 
 # ---------------------------------------------------------------------------
 # Connector / sync-log status constants
 # ---------------------------------------------------------------------------
 
-class SyncStatus:
-    """String constants for connector sync_status and sync-log status columns.
+class SyncStatus(str, Enum):
```
dharaneeshvrd: I feel it would be more clear if we create two classes for ConnectorStatus & SyncStatus
Can be part of follow up PR

services/digitize/api/v1/connectors.py
dharaneeshvrd: > End of the day, if the status is changed or not, we will remove all entries.

That's not a correct behaviour imo, we cannot simply remove last sync entry if it is in syncing state.

services/digitize/api/v1/connectors.py
dharaneeshvrd, line 289: If there is a exception in removing the documents, then we should nt remove the connector, an error message needs to be updated and keep the entry in delete_pending

services/digitize/api/v1/connectors.py
```diff
@@ -251,28 +305,23 @@ async def delete_connector(connector_id: str):
...
     except Exception as exc:
-        logger.error(f"Unexpected error deleting connector {connector_id}: {exc}", exc_info=True)
-        APIError.raise_error(ErrorCode.INTERNAL_SERVER_ERROR, str(exc))
+        logger.error(
+            f"Unexpected error during teardown for connector {connector_id!r}: {exc}",
+            exc_info=True,
+        )
 
 
 def _best_effort_delete_document(doc_id: str) -> None:
```
dharaneeshvrd: why are we calling it _best_effort?

services/digitize/api/v1/connectors.py
```diff
@@ -294,20 +343,27 @@ def _best_effort_delete_document(doc_id: str) -> None:
...
     from pathlib import Path
 
     base = Path(staging_connectors_dir)
     if not base.exists():
         return
-    prefix = f"{connector_id}-"
+    prefix = f"{connector_id}-{sync_seq}-" if sync_seq is not None else f"{connector_id}-"
     for entry in base.iterdir():
         if entry.is_dir() and entry.name.startswith(prefix):
             cleanup_staging_directory(entry.name, base, ignore_errors=True)
```
dharaneeshvrd: why are we passing base when the iterator is entry?

sats-23: cleanup_staging_directory works by taking the base directory and actual entry separately. Checks on the base path existing, forms final path and deletes. 
Entry is full path, entry.name is just the directory path

services/digitize/api/v1/connectors.py
sats-23: Just that it tries to delete and in case of failures, does not throw any exception

services/digitize/api/v1/connectors.py
sats-23: But that case, issue will be VDB related, why hold back on connector deletion?

services/digitize/connectors/models.py
sats-23: Done, added

services/digitize/api/v1/connectors.py
sats-23: But as part of delete connector, we end up removing the sync log entries for that specific connector as well right?

services/digitize/api/v1/connectors.py
sats-23: Dropping this as per discussion

services/digitize/api/v1/connectors.py
sats-23: Updated 
open_new_sync_log -> init_sync_log_and_update_connector
close_sync_log -> finalize_sync_log_and_update_connector

services/digitize/api/v1/connectors.py
sats-23: Removed as per discussion

services/digitize/connectors/sync_tick.py

services/digitize/api/v1/connectors.py
dharaneeshvrd: no, let's throw the exception and populate the error message. 
Please check all similar usage and avoid silencing the errors like this, let's propagate the error message to the connector.
Also we need to ensure clean up the earlier error messages after a successful sync.

services/digitize/api/v1/connectors.py
dharaneeshvrd: we need to, we can t leave the stale docs in vdb, the user managing the connector should be aware of this.

services/digitize/api/v1/connectors.py
```diff
@@ -227,18 +240,38 @@ async def delete_connector(connector_id: str):
...
             )
 
-        # Guard: reject if a tick is currently running
-        if connector.sync_status == SyncStatus.SYNCING:
-            APIError.raise_error(
-                ErrorCode.RESOURCE_LOCKED,
-                "A sync tick is currently running for this connector. "
-                "Retry after the tick completes.",
-            )
+        db_ops.mark_sync_delete_pending(connector_id)
```
dharaneeshvrd: rename this to reflect that you are updating connector table
no need to check for the syncing status as part of the db transaction also, let's utilise the if check below and call this method appropriately

dharaneeshvrd: as per the discussion, let's update the delete_pending immediately after validating the connector id irrespective of the sync status.

services/digitize/api/v1/connectors.py
```diff
@@ -227,18 +240,38 @@ async def delete_connector(connector_id: str):
...
+      3. Delete the connector row (cascades to sync_logs)
+      4. Sweep residual batch staging directories
+    """
+    logger.info(f"Starting teardown for connector {connector_id!r}")
+    try:
+        # Steps 1+2: remove checksum ownership; delete orphaned documents
+        owned_checksums = db_ops.list_connector_checksums(connector_id)
         for checksum in owned_checksums:
             try:
                 remaining, doc_id = db_ops.remove_connector_checksum_entry(connector_id, checksum)
```
dharaneeshvrd, line 285: as per the discussion,
please rename this method that it returns the remaining docs as well.
also let's avoid making multiple db transaction in same method under db/manager.py, to coordinate these you can use the utils/db.py

services/digitize/api/v1/connectors.py
```diff
@@ -197,27 +209,28 @@ async def update_connector(connector_id: str, body: ConnectorUpdateRequest):
...
-        "Documents owned exclusively by this connector are deleted. "
-        "Staging directories are cleaned up best-effort."
+        "Non-blocking detachment. Always returns 204 immediately. "
+        "If a sync tick is running the connector is marked delete_pending and "
+        "the tick handles teardown; otherwise teardown runs as a background task."
     ),
-    response_description="No content on successful deletion",
+    response_description="No content — teardown proceeds in the background",
 )
 async def delete_connector(connector_id: str):
```
dharaneeshvrd, line 229: as per the discussion and earlier comments, lets update the error message of connector table to reflect any failure in detaching the connector.

dharaneeshvrd: ok let's update this in a follow up PR.

services/digitize/api/v1/connectors.py
sats-23: Done, updated

services/digitize/api/v1/connectors.py
sats-23: Sure

services/digitize/api/v1/connectors.py
sats-23: Done, segregated all 2 callers into separate methods

services/digitize/api/v1/connectors.py
```diff
@@ -227,87 +240,125 @@ async def delete_connector(connector_id: str):
...
+        if deletion_failed:
+            db_ops.set_connector_error(connector_id, "Documents deletion failed")
+            logger.warning(
+                f"Skipping connector row deletion for {connector_id!r} "
+                "due to document deletion failures"
+            )
+            return
+
+        # Step 3: delete the connector row (cascades to connector_sync_logs)
         deleted = db_ops.delete_active_connector(connector_id)
```
dharaneeshvrd, line 307: Let's move this to the end of this method
So that in case of a failure in `_sweep_staging_dir`, the deletion can be retried.
If we remove this now itself, the user won't be able to retry the deletion, let's design like that

services/digitize/api/v1/connectors.py
```diff
@@ -399,6 +452,148 @@ async def get_connector(connector_id: str):
...
+        "dispatched one or the already-running one."
+    ),
+    response_description="Sync dispatched (or already in progress)",
+)
+async def trigger_sync(connector_id: str):
+    import asyncio
+    from digitize.connectors.sync_tick import run_tick
+
+    try:
+        connector = db_ops.get_active_connector(connector_id)
```
dharaneeshvrd: this is not really getting the active connector, you need to add a filter to check the connector is not in terminal state like delete pending.
we should not trigger sync or update the connector via PUT on a delete_pending connector.

services/digitize/api/v1/connectors.py
```diff
@@ -399,6 +452,148 @@ async def get_connector(connector_id: str):
...
+        if connector is None:
+            APIError.raise_error(
+                ErrorCode.RESOURCE_NOT_FOUND,
+                f"Connector {connector_id!r} not found",
+            )
+
+        acquired = db_ops.try_acquire_sync_lock(connector_id)
+        if acquired:
+            asyncio.create_task(run_tick(connector_id))
+            logger.info(f"Manual sync dispatched for connector {connector_id!r}")
```
dharaneeshvrd: When we implement the APScheduler, what would be the entry point, IMO, if its going to call this endpoint, we dont need to manage two different flows.

services/digitize/api/v1/connectors.py
```diff
@@ -399,6 +452,148 @@ async def get_connector(connector_id: str):
...
+            # init_sync_log_and_update_connector hasn't been called yet (that happens inside run_tick),
+            # so fetch the seq of the row that the background task will open shortly.
+            # Since we hold the lock we query for the highest existing seq + 1 via
+            # the active row that run_tick will create. Instead, we read it after a
+            # brief yield so the task can open the log first.
+            # Simpler and race-free: init_sync_log_and_update_connector is the very first thing
+            # run_tick does — schedule a wait-for-seq helper.
+            sync_seq = await _wait_for_sync_seq(connector_id)
```
dharaneeshvrd: if we address [this](https://github.com/IBM/project-ai-services/pull/1232/changes#r3794288261), then we dont need to have this logic of getting the sync seq like this, we can generate the sync seq and pass it as an arg

dharaneeshvrd (commented):
I see in lot of places where we raise the exception, we are not logging the error messages
Can we please log the error message as well?

services/digitize/api/v1/connectors.py
sats-23: Done

services/digitize/api/v1/connectors.py
sats-23: Added guard for both delete pending and sync cancel pending

services/digitize/api/v1/connectors.py
sats-23: Yes, the core run_tick is going to be called commonly for both flows.
Just that APScheduler will not be dealing with the HTTP exception handling as the endpoint. Everything else is same

services/digitize/api/v1/connectors.py
sats-23: Do-able, but we will end up moving init_sync_log_and_update_connector out of the run_tick which we didn't wanna do earlier. 

Either way, for the APScheduler, we just need to call run_tick, we don't need to await or store sync_seq. So this is a clean approach. 

If we strictly want to call trigger_sync for both, we will have to move init_sync_log_and_update_connector out into trigger_sync and pass sync_seq as a param to run_tick

dharaneeshvrd: > If we strictly want to call trigger_sync for both, we will have to move init_sync_log_and_update_connector out into trigger_sync and pass sync_seq as a param to run_tick

I feel this would be the straightforward way and easy to maintain as well. Let's do this. 
I would want this for the job submission actually, so we dont maintain different flow for different consumers.

sats-23: Done, updated both 
Now acquires lock and opens sync log outside of run_tick 
And also has a new dispatch_sync that can be called by both APScheduler and the API

services/digitize/api/v1/connectors.py
dharaneeshvrd: How about the PUT /connectors call?

sats-23: That is async and allows only connection details to be affected. So it shouldn't really affect us

dharaneeshvrd (approved):

PR #1236, opened 2026-08-10
adding similarity failure tests and exclude from e2e suite

vijay-vin:
Adding Similarity Failure testcases
Adding support to exclude all Failure testcases from e2e test suite.
We should pass "run-failure-tests" flag when running any failure testcases.   By default it is false, so e2e test can exclude all failure testcases. 
README.md file is updated showing how to use this new flag.

Ran all failure testcases (bootstrap, catalog and similarity) again with the new flag and also the e2e test suite. The result file is attached. I did not attach full e2e test file output, but the failure testcase runs are the complete output.

### DIFF
--- a/ai-services/tests/e2e/bootstrap_failure_test.go
+++ b/ai-services/tests/e2e/bootstrap_failure_test.go
@@ -9,15 +9,24 @@
 //
 // Labels
 //
-//	failure-test   – all tests in this file
-//	registry       – Test 1
-//	catalog        – Test 2 (both sub-cases)
-//	validation     – Test 3 and Test 4
-//	spyre          – Test 4 specifically
+//	failure-test       – all tests in this file (umbrella label, shared with all failure suites)
+//	bootstrap-failure  – all tests in this file (domain label)
+//	registry           – Test 1
+//	catalog            – Test 2 (both sub-cases)
+//	validation         – Test 3 and Test 4
+//	spyre              – Test 4 specifically
 //
-// Running only failure tests:
+// Running ALL failure tests together (all three failure suites):
 //
 //	ginkgo -r --label-filter="failure-test" ./tests/e2e
+//
+// Excluding ALL failure tests from the normal run:
+//
+//	ginkgo -r --label-filter="!failure-test" ./tests/e2e
+//
+// Running only bootstrap failure tests:
+//
+//	ginkgo -r --label-filter="bootstrap-failure" ./tests/e2e
 //
 // Running a specific category:
 //
@@ -25,10 +34,6 @@
 //	ginkgo -r --label-filter="failure-test && catalog"    ./tests/e2e
 //	ginkgo -r --label-filter="failure-test && validation" ./tests/e2e
 //	ginkgo -r --label-filter="failure-test && spyre"      ./tests/e2e
-//
-// Excluding failure tests from the normal run:
-//
-//	ginkgo -r --label-filter="!failure-test" ./tests/e2e
 package e2e
 
 import (
@@ -78,6 +83,19 @@ var _ = ginkgo.Describe("Bootstrap Failure Scenarios",
 	// self-contained and must not depend on the result of a preceding test.
 	func() {
 
+		// ── Default-exclusion guard ───────────────────────────────────────────
+		//
+		// Failure tests are skipped unless --run-failure-tests is explicitly
+		// passed.  This mirrors the --app-name guard used by Language Support
+		// Tests and prevents accidental execution during a normal suite run.
+		ginkgo.BeforeEach(func() {
+			if !runFailureTests {
+				ginkgo.Skip(
+					"[FAILURE-TEST][Bootstrap] Skipping — pass --run-failure-tests to opt in to failure test execution",
+				)
+			}
+		})
+
 		// ── Test 1: Invalid Registry Credentials ─────────────────────────────
 		//
 		// Rationale: An operator who sets the wrong REGISTRY_PASSWORD should
@@ -95,7 +113,7 @@ var _ = ginkgo.Describe("Bootstrap Failure Scenarios",
 			func() {
 				ginkgo.It(
 					"fails gracefully with invalid registry credentials",
-					ginkgo.Label("failure-test", "registry", "spyre-independent"),
+					ginkgo.Label("failure-test", "bootstrap-failure", "registry", "spyre-independent"),
 					func() {
 						// This test has only been validated on the podman runtime.
 						// Skip explicitly on openshift until ported and verified.
@@ -176,7 +194,7 @@ var _ = ginkgo.Describe("Bootstrap Failure Scenarios",
 				// 2a ── Invalid catalog credentials ───────────────────────────
 				ginkgo.It(
 					"fails gracefully with invalid catalog credentials",
-					ginkgo.Label("failure-test", "catalog", "spyre-independent"),
+					ginkgo.Label("failure-test", "bootstrap-failure", "catalog", "spyre-independent"),
 					func() {
 						if appRuntime != "podman" {
 							ginkgo.Skip(
@@ -250,7 +268,7 @@ var _ = ginkgo.Describe("Bootstrap Failure Scenarios",
 				// 2b ── Unreachable catalog server ─────────────────────────
 				ginkgo.It(
 					"fails gracefully when catalog server is unreachable",
-					ginkgo.Label("failure-test", "catalog", "spyre-independent"),
+					ginkgo.Label("failure-test", "bootstrap-failure", "catalog", "spyre-independent"),
 					func() {
 						if appRuntime != "podman" {
 							ginkgo.Skip(
@@ -315,7 +333,7 @@ var _ = ginkgo.Describe("Bootstrap Failure Scenarios",
 			func() {
 				ginkgo.It(
 					"bootstrap validate rejects an invalid --runtime flag value",
-					ginkgo.Label("failure-test", "validation", "spyre-independent"),
+					ginkgo.Label("failure-test", "bootstrap-failure", "validation", "spyre-independent"),
 					func() {
 						ctx, cancel := context.WithTimeout(
 							context.Background(),
@@ -378,7 +396,7 @@ var _ = ginkgo.Describe("Bootstrap Failure Scenarios",
 				//     GHW_CHROOT is restored in a deferred cleanup.
 				ginkgo.It(
 					"bootstrap validate reports missing Spyre accelerator card",
-					ginkgo.Label("failure-test", "validation", "spyre"),
+					ginkgo.Label("failure-test", "bootstrap-failure", "validation", "spyre"),
 					func() {
 						if appRuntime != "podman" {
 							ginkgo.Skip(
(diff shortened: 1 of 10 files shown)


Abhishek-Kushwaha1:
@vijay-vin  can you attach screen shot of the test run
@Abhishek-Kushwaha1 I have added the test run results in the Jira when I opened this PR.

ai-services/tests/e2e/rag/similarity.go
```diff
@@ -0,0 +1,218 @@
...
+	"time"
+
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+)
+
+// similaritySearchPath is the POST endpoint on the similarity service.
+const similaritySearchPath = "/v1/similarity-search"
+
+// HTTP status codes used by the similarity service validate functions.
+const (
```
manju956: We need to declare HTTP status codes in a common utility and reuse.

ai-services/tests/e2e/rag/similarity.go
```diff
@@ -0,0 +1,218 @@
...
+//
+// Reuses buildPostJSONRequest (evaluator.go) for consistent TLS, JSON marshalling,
+// and Bearer-token handling, then drives sharedRAGClient directly so the body is
+// not discarded on non-200 responses.
+func PostSimilaritySearchRaw(
+	ctx context.Context,
+	baseURL string,
+	body map[string]any,
+) (statusCode int, rawBody string, err error) {
+	req, err := buildPostJSONRequest(ctx, baseURL, similaritySearchPath, body)
```
manju956, line 46: looks like regular http requests. can we reuse the existing http client here?

ai-services/tests/e2e/bootstrap_failure_test.go
```diff
@@ -95,7 +113,7 @@ var _ = ginkgo.Describe("Bootstrap Failure Scenarios",
 			func() {
 				ginkgo.It(
 					"fails gracefully with invalid registry credentials",
-					ginkgo.Label("failure-test", "registry", "spyre-independent"),
+					ginkgo.Label("failure-test", "bootstrap-failure", "registry", "spyre-independent"),
```
manju956, line 116: bootstrap is spyre-dependent as we configure cards rt?

manju956:
> Adding Similarity Failure testcases Adding support to exclude all Failure testcases from e2e test suite. We should pass "run-failure-tests" flag when running any failure testcases. By default it is false, so e2e test can exclude all failure testcases. README.md file is updated showing how to use this new flag.
> 
> Ran all failure testcases (bootstrap, catalog and similarity) again with the new flag and also the e2e test suite. The result file is attached. I did not attach full e2e test file output, but the failure testcase runs are the complete output.

@Abhishek-Kushwaha1 Can you PTAL?

ai-services/tests/e2e/rag/similarity.go
```diff
@@ -0,0 +1,218 @@
...
+// ─────────────────────────────────────────────────────────────────────────────
+// Private helpers
+// ─────────────────────────────────────────────────────────────────────────────
+
+// similarityErrorEnvelope mirrors the JSON error shape returned by every
+// similarity service endpoint:
+//
+//	{"error":{"code":"...","message":"...","status":N}}
+type similarityErrorEnvelope struct {
+	Error struct {
```
Abhishek-Kushwaha1: `similarityErrorEnvelope` duplicates the existing `SimilarityErrorResponse` / `ErrorResponse` types in `similarity/similarity.go`.
Since `PostSimilaritySearchRaw` is moving to the `similarity` package, we can reuse those types and remove this struct.

ai-services/tests/e2e/rag/similarity.go
```diff
@@ -0,0 +1,218 @@
...
+	return nil
+}
+
+// ValidateSimilarityInvalidTopKError asserts that the service correctly rejected
+// a top_k value below the minimum (ge=1) with HTTP 422 (Pydantic validation).
+//
+// FastAPI's 422 body uses {"detail":[...]} rather than the custom error envelope,
+// so this validator checks for the HTTP status and the presence of "top_k" in
+// the raw body — a shape-agnostic assertion that works for both response formats.
+func ValidateSimilarityInvalidTopKError(statusCode int, rawBody string) error {
```
Abhishek-Kushwaha1: Tests 1 and 2 check the error code and message, but Test 3 only checks the status and `"top_k"`.
It would be better to check the 422 response fields too, especially `top_k` and the validation message.

ai-services/tests/e2e/rag/similarity.go
```diff
@@ -0,0 +1,218 @@
...
+	keywords := []string{"connection", "dial", "no such host", "i/o timeout", "context deadline exceeded"}
+	for _, kw := range keywords {
+		if strings.Contains(msg, kw) {
+			return nil
+		}
+	}
+
+	return fmt.Errorf(
+		"expected transport error to contain one of %v — got: %s",
+		keywords, err.Error(),
```
Abhishek-Kushwaha1: The errors.Is(err, ErrNonRetriable) check can never be true here because PostSimilaritySearchRaw doesn't wrap errors with ErrNonRetriable.

ai-services/tests/e2e/similarity_failure_test.go
```diff
@@ -0,0 +1,508 @@
...
+
+	ginkgo "github.com/onsi/ginkgo/v2"
+	gomega "github.com/onsi/gomega"
+)
+
+// similarityFailureTestTimeout caps how long any single similarity failure test
+// may run.  Tests 1–3 exercise pure HTTP validation (no vectorstore I/O), and
+// Test 4 explicitly uses an unreachable URL — both should resolve in well under
+// 30 seconds.  Critically, this timeout MUST be shorter than sharedRAGClient's
+// global httpClientTimeout (10 minutes) so that TC-4 never hangs waiting for a
```
Abhishek-Kushwaha1: The comment says `httpClientTimeout` is 10 minutes, but it is actually 25 minutes in `evaluator.go`.
Please update the comment to 25 minutes so it accurately explains why the 30-second timeout is used.

ai-services/tests/e2e/similarity_failure_test.go
```diff
@@ -0,0 +1,508 @@
...
+						// (10 minutes) so the test never hangs on DNS/TCP failure.
+						ctx, cancel := withTimeout(similarityFailureTestTimeout)
+						defer cancel()
+
+						logger.Infof(
+							"[FAILURE-TEST][Similarity] Attempting request to unreachable URL: %s",
+							unreachableSimilarityURL,
+						)
+
+						_, _, transportErr := rag.PostSimilaritySearchRaw(
```
Abhishek-Kushwaha1, line 362: TC-1, TC-2, and TC-3 are using `rag.PostSimilaritySeahRawrc`, but the `similarity` package already has helpers for these API calls.
It would be better to use those helpers, or move `PostSimilaritySearchRaw` there so it uses the correct HTTP client.

This comment is valid for all places where  `rag.PostSimilaritySeahRawrc` is occured in this file

ai-services/tests/e2e/rag/similarity.go
```diff
@@ -0,0 +1,218 @@
+// Package rag — similarity search helpers for E2E failure tests.
```
Abhishek-Kushwaha1, line 1: I am not getting why this  new `similarity.go` file is under  rag folder we already have `similarity.go` file in the similarity folder.

ai-services/tests/e2e/rag/similarity.go
vijay-vin: agree.
Replaced — all four usages with the standard net/http package constants

ai-services/tests/e2e/rag/similarity.go
vijay-vin: PostSimilaritySearchRaw already reuses buildPostJSONRequest and sharedRAGClient from evaluator.go. It can't reuse PostJSON because that calls handlePostJSONResponse which discards non-200 bodies via io.Discard — failure tests need to inspect that body. 
I now updated the doc comment to make this explicit.

ai-services/tests/e2e/bootstrap_failure_test.go
vijay-vin: The spyre-independent label is correct for TC-1 through TC-3 because those tests exercise CLI flag validation and network error handling — code paths that never reach SpyreRule or any PCI detection. Only TC-4 carries spyre because it specifically validates the "no card attached" failure path inside bootstrap validate.

ai-services/tests/e2e/rag/similarity.go
vijay-vin: Good catch on the duplication. 
Moving PostSimilaritySearchRaw to the similarity package would break its dependencies on buildPostJSONRequest and sharedRAGClient, which are unexported in the rag package. Instead, I've deleted similarityErrorEnvelope and updated parseSimilarityErrorBody to unmarshal directly into the already-exported similarity.SimilarityErrorResponse type
Also, based on your last comment below, this similarity.go file has been modified. I am just adding these comments for continuity sake.

ai-services/tests/e2e/rag/similarity.go
vijay-vin: agree. Done.

ai-services/tests/e2e/rag/similarity.go
vijay-vin: agree. Done

ai-services/tests/e2e/similarity_failure_test.go
vijay-vin: agree. Modified all occurrences of 10 minutes.

ai-services/tests/e2e/similarity_failure_test.go
vijay-vin: PostSimilaritySearchRaw must stay in the rag package for TC-4 — the connectivity test needs the raw transport error to check DNS/dial keywords, and similarity.SimilaritySearchExpectingError wraps that error as "request failed: %w", obscuring them. Since the rag import cannot be dropped, switching TC-1/2/3/5 to similarity.SimilaritySearchExpectingError would leave both imports in the test file anyway — net result is churn with no functional gain. The validators themselves have been moved to the similarity package (addressed in comment below)

ai-services/tests/e2e/rag/similarity.go
vijay-vin: Valid point. 
All five ValidateSimilarity* functions and parseSimilarityErrorBody have been moved to similarity/similarity.go where they belong. rag/similarity.go now contains only PostSimilaritySearchRaw, which must stay in the rag package because it depends on buildPostJSONRequest and sharedRAGClient — both unexported in rag/evaluator.go.

NOTE: Because of this change, some of my reply comments above may not be valid any more. Code is refactored properly.

ai-services/tests/e2e/rag/similarity.go
manju956: ok

ai-services/tests/e2e/bootstrap_failure_test.go
manju956: ok

manju956 (dismissed):
lgtm

Abhishek-Kushwaha1 (dismissed):
LGTM

manju956 (approved):

PR #1237, opened 2026-08-11
AISERVICES-1381 added sync extract endpoint

manalilatkar:
In this PR:

1. sync extraction endpoint: POST /v1/extract/
2. helper methods
3. unit tests

### DIFF
--- a/services/extract/api/v1/jobs.py
+++ b/services/extract/api/v1/jobs.py
@@ -32,7 +32,7 @@ from extract.models import (
     JobsListResponse,
     PaginationInfo,
 )
-from extract.state import concurrency_limiter
+from extract.state import concurrency_limiter, job_limiter
 from extract.settings import settings
 from extract.utils.exceptions import ExtractException
 from extract.utils.request import check_request_body_size
@@ -135,11 +135,6 @@ async def extract_sync(request: Request) -> JSONResponse:
             "Server is at maximum vLLM concurrency. Please retry later.",
         )
 
-    # ------------------------------------------------------------------
-    # Resolve LLM config once — cached after the first call, but kept
-    # outside the semaphore to avoid holding the slot during any cold-start
-    # network probes.
-    # ------------------------------------------------------------------
     llm_model_dict = get_llm_endpoint()
     llm_endpoint: str = llm_model_dict.get("llm_endpoint", "")
     llm_model: str = llm_model_dict.get("llm_model", "")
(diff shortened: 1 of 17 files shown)


services/extract/api/v1/jobs.py
```diff
@@ -1,49 +1,261 @@
...
+        body_dict = json.loads(raw_body)
+    except Exception:
+        raise ExtractException(400, "INVALID_JSON", "Request body is not valid JSON.")
+
+    try:
+        body = ExtractionRequest(**body_dict)
+    except Exception as exc:
+        raise ExtractException(400, "INVALID_REQUEST", f"Invalid request body: {exc}")
+
+    schema_row = _resolve_schema(body.schema_id)
```
dharaneeshvrd, line 102: Sry I dont know how did we miss this in the proposal, but did nt we discuss to accept the schema part of the extraction request also, so user can avoid making two calls for the requests which does nt need to reuse the schema for multiple extraction requests.
If schema id is not supplied, with the help of schema details passed on the extraction request should be passed to schema workflow for validation and infer schema if required and use it for the actual extraction request.

manalilatkar: @dharaneeshvrd no this was not discussed before, I'm hearing this for the first time, but it can be considered. Do you want me to make that change in the same PR or in a separate PR?

dharaneeshvrd: ok somehow this was my assumption :)
Yeah we need to consider this. Let's do it in a follow up PR, please create a story for this

services/extract/utils/vllm.py
```diff
@@ -0,0 +1,259 @@
...
+
+    payload: Dict[str, Any] = {
+        "model": llm_model,
+        "messages": messages,
+        "max_tokens": max_tokens,
+        "temperature": settings.extract.extraction_temperature,
+        "stream": False,
+    }
+
+    if settings.extract.guided_decoding_enabled:
```
dharaneeshvrd, line 87: Is this working?
Let's ensure to test on cpu and WatsonX as well. 
Hope we have a JIRA story to track this.

services/extract/api/v1/jobs.py
```diff
@@ -1,49 +1,261 @@
...
+@router.post(
+    "",
+    response_model=ExtractionResponse,
+    status_code=200,
+    tags=["extraction"],
+    summary="Synchronous extraction",
+    description=(
+        "Extract structured data from plain text against a registered schema in a "
+        "single blocking call.  Returns validated, schema-conformant JSON.\n\n"
+        "**Processing steps:**\n"
```
dharaneeshvrd: This is not required and going to be confusing as well, swagger description should help user and the AI(once mcp is there) when and how to use this API.

services/extract/api/v1/jobs.py
manalilatkar: done

services/extract/api/v1/jobs.py
```diff
@@ -1,49 +1,261 @@
...
+    # ------------------------------------------------------------------
+    # 1. Parse & basic field validation
+    # ------------------------------------------------------------------
+    try:
+        body_dict = json.loads(raw_body)
+    except Exception:
+        raise ExtractException(400, "INVALID_JSON", "Request body is not valid JSON.")
+
+    try:
+        body = ExtractionRequest(**body_dict)
```
dharaneeshvrd: not sure this would work when we have optional fields(since schema details can be optional)

services/extract/api/v1/jobs.py
```diff
@@ -1,49 +1,261 @@
...
     delete_all_job_files,
     delete_job_files,
     read_result_file,
     stage_uploaded_file,
     validate_file_extension,
 )
-from extract.utils.exceptions import ExtractException
-from extract.utils.schema import fmt_dt
+from extract.utils.schema import (
+    SchemaValidationError,
```
dharaneeshvrd: unused

services/extract/api/v1/jobs.py
```diff
@@ -1,49 +1,261 @@
...
+    if concurrency_limiter.locked():
+        raise ExtractException(
+            429, "RATE_LIMIT_EXCEEDED",
+            "Server is at maximum vLLM concurrency. Please retry later.",
+        )
+
+    llm_model_dict = get_llm_endpoint()
+    llm_endpoint: str = llm_model_dict.get("llm_endpoint", "")
+    llm_model: str = llm_model_dict.get("llm_model", "")
+    max_model_len: int = settings.common.llm.max_model_len
```
dharaneeshvrd:
```suggestion
    max_model_len: int = llm_model_dict.get('max_model_len')
```
this would return you the dynamic max model len queried from the vllm

services/extract/api/v1/jobs.py
```diff
@@ -1,49 +1,261 @@
...
+    except Exception as exc:
+        logger.error(f"Tokenization failed: {exc}", exc_info=True)
+        raise ExtractException(
+            503, "TOKENIZATION_ERROR",
+            "Failed to tokenise the input text. "
+            "Ensure the vLLM /tokenize endpoint is reachable.",
+        )
+
+    # ── 4. Hard context-window guard ─────────────────────────────────
+    #       check_extraction_budget raises SchemaValidationError(413)
```
dharaneeshvrd: incorrect

services/extract/api/v1/jobs.py
```diff
@@ -1,49 +1,261 @@
...
+        choices = vllm_resp.get("choices", [])
+        if not choices:
+            raise ExtractException(500, "LLM_ERROR", "vLLM returned an empty choices list.")
+
+        choice = choices[0]
+        finish_reason: str = choice.get("finish_reason", "")
+
+        # ── 7. Output-budget exceeded — fail fast, do NOT burn the retry ─
+        if finish_reason == "length":
+            raise ExtractException(
```
dharaneeshvrd: are we not going to retry if we have room for increasing the max tokens ? I think we can give it one more try.

services/extract/api/v1/jobs.py
manalilatkar: I think it will work even for optional fields. If it doesn't work out, I'll remove in that PR.

services/extract/utils/vllm.py
manalilatkar: yes, it is working. We have a story for testing on watsonX. Have added testing on CPU to it as well.

services/extract/api/v1/jobs.py
manalilatkar: retry added

services/extract/api/v1/jobs.py
manalilatkar: removed

services/extract/api/v1/jobs.py
manalilatkar: removed

services/extract/api/v1/jobs.py
manalilatkar: fixed

services/extract/api/v1/jobs.py
manalilatkar: removed

services/extract/api/v1/jobs.py
dharaneeshvrd: can we please replace this with asyncio.create_task?
Context regarding the change https://github.com/IBM/project-ai-services/pull/1232#discussion_r3764210794

services/extract/api/v1/jobs.py
```diff
@@ -1,49 +1,263 @@
...
+        404: http_error_responses[404],
+        413: http_error_responses[413],
+        422: {"description": "Extraction output failed schema validation after retry"},
+        429: http_error_responses[429],
+        500: http_error_responses[500],
+        503: http_error_responses[503],
+    },
+    include_in_schema=True,
+)
+async def extract_sync(request: Request) -> JSONResponse:
```
dharaneeshvrd: don't we need to expect `ExtractionRequest` directly?
```suggestion
async def extract_sync(body: ExtractionRequest) -> JSONResponse:
```

dharaneeshvrd (commented):
Lgtm
Post addressing the last 2 comments, can be merged.

services/extract/api/v1/jobs.py
manalilatkar: @dharaneeshvrd I don't think this will work, as the input variable contains body as well as headers, and ExtractionRequest corresponds to the body(text, schema_id). We later cast the body into ExtractionRequest in line no 176. (Maybe I can rename ExtractionRequest to ExtractionRequestBody to make it clearer) Also, we are not doing this anywhere else. For the summarize sync request input, we expect (request: Request) same as here. Could you explain your thought process for this change?

dharaneeshvrd: Please check other services like chatbot and digitize, we should have followed this for summarize as well(please create a JIRA to track later) if we don't have this information documented, for the user who consumes swagger, how would they know the payload structure to be used?

dharaneeshvrd (approved):

PR #1238, opened 2026-08-11
Core translation service logic and async endpoints.

Niharika0306:
This PR implements the async document translation pipeline.

Core utilities added: token-based chunker, context-window guard, lingua-based language detection, prompt builder, async vLLM client, and file staging helpers.

Endpoints:

POST /v1/translate/jobs — upload a .txt / .md file, returns 202 {job_id}
GET /v1/translate/jobs — paginated job list with optional status filter
GET /v1/translate/jobs/{job_id} — job detail and phase metadata
GET /v1/translate/jobs/{job_id}/result — full translation result as JSON (409 if running, 410 if failed)
GET /v1/translate/jobs/{job_id}/result/download — translated file download with matching MIME type and Content-Disposition


services/translate/api/v1/jobs.py
```diff
@@ -1,17 +1,455 @@
...
+)
+from translate.workers.concurrency import concurrency_manager
+from translate.workers.translation_worker import run_translation_job
+
+logger = get_logger("jobs_api")
 
 router = APIRouter()
 
+# ---------------------------------------------------------------------------
+# Translation-specific HTTP error helpers
```
dharaneeshvrd: can we please move these to a different file?

dharaneeshvrd: let's keep only the API handlers here

services/translate/api/v1/jobs.py
```diff
@@ -1,17 +1,455 @@
...
+    if norm_source != "auto" and norm_source == norm_target:
+        _raise_same_language(
+            f"source_language and target_language must differ "
+            f"(both are '{norm_source}')."
+        )
+
+    return norm_source, norm_target
+
+
+def _format_dt(dt: Optional[datetime]) -> Optional[str]:
```
dharaneeshvrd: there is a get_utc_timestamp() in misc_utils, please see whether that can be used here

services/translate/api/v1/jobs.py
```diff
@@ -1,17 +1,455 @@
...
+    )
+
+
+# ---------------------------------------------------------------------------
+# POST /v1/translate/jobs
+# ---------------------------------------------------------------------------
+
+@router.post(
+    "",
+    status_code=status.HTTP_202_ACCEPTED,
```
dharaneeshvrd, line 149: let's add error responses as well here

dharaneeshvrd: response_description also missing

services/translate/api/v1/jobs.py
```diff
@@ -1,17 +1,455 @@
...
+    job_name: Optional[str] = Form(
+        default=None,
+        description="Optional human-readable label for the job",
+    ),
+) -> JobCreatedResponse:
+    filename = file.filename or "upload"
+
+    # 1. Validate file extension.
+    try:
+        input_type = validate_file_extension(filename)
```
dharaneeshvrd, line 182: please take a ref from extract file validation and follow here

services/translate/api/v1/jobs.py
```diff
@@ -1,17 +1,455 @@
...
+    except ValueError as exc:
+        _raise_unsupported_file_type(str(exc))
+
+    # 2. Validate languages.
+    norm_source, norm_target = _validate_languages(source_language, target_language)
+
+    # 3. Check file size before reading content.
+    content = await file.read()
+    max_bytes = settings.translate.max_upload_size_mb * 1024 * 1024
```
dharaneeshvrd, line 196: maybe missed in proposal, but do we really need this?
we are not doing this in any of the other services

services/translate/workers/translation_worker.py
```diff
@@ -0,0 +1,352 @@
...
+                status=JobStatus.IN_PROGRESS,
+                metadata={"phase": "chunking"},
+            )
+            logger.info(f"[{job_id}] Worker started — chunking")
+
+            # ------------------------------------------------------------------
+            # Step 2 — Read staged file as UTF-8
+            # ------------------------------------------------------------------
+            try:
+                text = staged_file_path.read_text(encoding="utf-8")
```
dharaneeshvrd, line 158: dont we need to parse md file using md parsers?
I thought it would help in processing it better

services/translate/utils/chunking.py
```diff
@@ -0,0 +1,225 @@
...
+        text: Full document text (may be markdown).
+        llm_endpoint: vLLM endpoint used for ``/tokenize`` calls.
+        source_language_code: Resolved ISO-639-1 code (e.g. ``"DE"``) used to pick
+            the correct ``SentenceSplitter`` language.  Defaults to ``"en"`` if None
+            or unresolvable (lingua fell below confidence threshold).
+
+    Returns:
+        Ordered list of ``TranslationChunk`` objects (index 0 … N-1).
+    """
+    budget = settings.translate.chunk_token_budget
```
dharaneeshvrd: need to calculate this value dynamically, can't use static value like this to accommodate dynamic max model len of different models

services/translate/workers/translation_worker.py
```diff
@@ -0,0 +1,352 @@
...
+
+            chunking_secs = round(time.perf_counter() - start_chunking, 3)
+            logger.info(
+                f"[{job_id}] Chunked into {len(chunks)} chunk(s) in {chunking_secs}s"
+            )
+
+            # Guard check: run eagerly before dispatching any LLM calls.
+            # A breach here means CHUNK_TOKEN_BUDGET is misconfigured.
+            for chunk in chunks:
+                max_tokens, diag = check_chunk_guard(chunk.token_count, settings)
```
dharaneeshvrd: this is not required at all if you compute the chunk itself with original max model Len's token budget.

services/translate/utils/guard.py
```diff
@@ -0,0 +1,149 @@
...
+        "max_model_len": max_model_len,
+        "input_tokens": input_tokens,
+        "prompt_overhead_tokens": prompt_overhead_tokens,
+        "min_output_buffer_tokens": min_output_tokens,
+        "total_required_tokens": total_required,
+        "excess_tokens": max(0, total_required - max_model_len),
+    }
+
+
+def check_sync_guard(
```
dharaneeshvrd: unused

services/translate/workers/translation_worker.py
```diff
@@ -0,0 +1,352 @@
...
+    Translate a single chunk, respecting both the per-job and global semaphores.
+
+    Returns:
+        ``(chunk, translated_text, input_tokens, output_tokens)``
+
+    Raises:
+        httpx.HTTPStatusError: Non-2xx response from vLLM (propagates to gather).
+        RuntimeError:          Per-chunk context guard breach.
+    """
+    max_tokens, diag = check_chunk_guard(chunk.token_count, settings)
```
dharaneeshvrd: I dont think this is required again

services/translate/workers/translation_worker.py
```diff
@@ -0,0 +1,352 @@
...
+
+    ``"paragraph"`` boundaries use ``"\\n\\n"``.
+    ``"sentence"`` boundaries (sentence-fallback sub-chunks) use ``" "``.
+
+    Args:
+        results: List of ``(chunk, translation, in_tok, out_tok)`` in chunk
+                 index order (as returned by ``asyncio.gather``).
+
+    Returns:
+        ``(translated_markdown, total_input_tokens, total_output_tokens)``
```
dharaneeshvrd: it needs to process txt file as well
Please see if we need to use markdown libraries to assemble better

services/translate/api/v1/jobs.py
manalilatkar: @dharaneeshvrd may be we can move the extract file validation method to common and use the same thing. WDYT?

services/translate/utils/chunking.py
```diff
@@ -0,0 +1,225 @@
...
+
+def _count_tokens_sync(text: str, llm_endpoint: str) -> int:
+    """Synchronous token count — wraps tokenize_with_llm (returns a list)."""
+    return len(tokenize_with_llm(text, llm_endpoint))
+
+
+async def _count_tokens(text: str, llm_endpoint: str) -> int:
+    """Async wrapper: runs the blocking tokenise call in the default thread-pool."""
+    loop = asyncio.get_running_loop()
+    return await loop.run_in_executor(
```
dharaneeshvrd: let's use await asyncio.to_thread

services/translate/api/v1/jobs.py
```diff
@@ -1,17 +1,455 @@
...
+    try:
+        input_type = validate_file_extension(filename)
+    except ValueError as exc:
+        _raise_unsupported_file_type(str(exc))
+
+    # 2. Validate languages.
+    norm_source, norm_target = _validate_languages(source_language, target_language)
+
+    # 3. Check file size before reading content.
+    content = await file.read()
```
manalilatkar: This line makes the code read the file first and compare size later. A 2GB text file will allocate 2GB of memory before the guard fires. Fix: declare max_bytes before and then read like this:
`content = await file.read(max_bytes + 1)`

services/translate/api/v1/jobs.py
manalilatkar: @dharaneeshvrd I think this is good, saves the processing time in case of really large files which are not processable anyway. We should include this everywhere else.

services/translate/api/v1/jobs.py
```diff
@@ -1,17 +1,455 @@
...
+    # 5. Generate job ID; stage the file.
+    job_id = generate_uuid()
+    staged_path = await stage_uploaded_file(
+        job_id=job_id,
+        filename=filename,
+        content=content,
+    )
+
+    # 6. Insert DB row with status=accepted.
+    db_manager.create_job(
```
manalilatkar: What if a job with job_name already exists? That exception will get caught in the create_job method and a new background task will still get added. Please handle the case where `create_job` returns None.

services/translate/api/v1/jobs.py
```diff
@@ -1,17 +1,455 @@
...
+    response_model=JobsListResponse,
+    summary="List translation jobs",
+    description="Paginated list of jobs with optional status filter.",
+)
+async def list_jobs(
+    limit: int = 20,
+    offset: int = 0,
+    status_filter: Optional[str] = None,
+) -> JobsListResponse:
+    if limit < 1 or limit > 100:
```
manalilatkar: instead of manually checking these values use query parameters like:
```
limit: int = Query(default=20, ge=1, le=100, description="Records per page"),
offset: int = Query(default=0, ge=0, description="Records to skip"),
```

services/translate/utils/chunking.py
```diff
@@ -0,0 +1,225 @@
...
+                        index=chunk_index,
+                        text=block,
+                        join_after="paragraph",
+                        token_count=block_tokens,
+                    )
+                )
+                chunk_index += 1
+            else:
+                # Step 3 — sentence-level fallback for oversized prose blocks.
+                splitter = SentenceSplitter(language=splitter_lang)
```
manalilatkar: Can we initialize SentenceSplitter once before the for block begins instead of initializing it in every iteration?

services/translate/utils/chunking.py
```diff
@@ -0,0 +1,225 @@
...
+                    start_index=chunk_index,
+                    parent_join_after="paragraph",
+                )
+                for sc in sentence_chunks:
+                    sc.index = chunk_index
+                    chunk_index += 1
+                chunks.extend(sentence_chunks)
+        else:
+            # Normal packing: would this block overflow the running chunk?
+            if running_tokens + block_tokens > budget and current_blocks:
```
manalilatkar: We are not counting seperator ("\n\n") tokens and over big texts with multiple paragraphs, it will add up. May be something like this:
```
SEPARATOR_TOKEN_ESTIMATE = 2  # "\n\n" ≈ 2 tokens
if running_tokens + block_tokens + SEPARATOR_TOKEN_ESTIMATE > budget and current_blocks:
```
@dharaneeshvrd WDYT?

services/translate/utils/prompt.py
```diff
@@ -0,0 +1,214 @@
...
+    for pos in positions:
+        sample = text[pos : pos + sample_size]
+        code = detect_language(sample, min_confidence=min_confidence)
+        if code:
+            votes[code] = votes.get(code, 0) + 1
+
+    if not votes:
+        return None
+
+    best_code = max(votes, key=lambda c: votes[c])
```
manalilatkar, line 169: This line has a bug.
If position 0 detects English, position 1 returns None (below confidence), and position 2 detects German, the function returns whichever of English/German appears first in max() iteration — essentially random.

A document with an English letterhead, an untranslatable middle section, and a German body would produce exactly this scenario.
Fix:
```
best_code = max(votes, key=lambda c: votes[c])
if votes[best_code] >= 2:
    return best_code
return None  # no majority — fall back to LLM auto-detect
```

services/translate/utils/llm.py
```diff
@@ -0,0 +1,84 @@
...
+        headers={"Content-Type": "application/json"},
+    )
+    response.raise_for_status()
+    data = response.json()
+
+    choices = data.get("choices", [])
+    if not choices:
+        raise ValueError("vLLM returned an empty 'choices' array")
+
+    logger.info(f"vLLM choices: {choices}")
```
manalilatkar: this log line should be removed or moved to debug otherwise it will overpopulate logs for large documents.

services/translate/workers/translation_worker.py
manalilatkar: @dharaneeshvrd IMO it should not be required as LLM understands markdown natively.

manalilatkar (commented):
good job on the chunking math.

services/translate/api/v1/jobs.py
Niharika0306: sure thing.

services/translate/api/v1/jobs.py
Niharika0306: noted.

services/translate/api/v1/jobs.py
Niharika0306: yes,  that was the intent. Better to avoid the tokenize calls too.

services/translate/api/v1/jobs.py
```diff
@@ -1,17 +1,450 @@
...
+        source_language=norm_source,
+        target_language=norm_target,
+        input_type=input_type,
+        job_name=job_name,
+        document_name=filename,
+        submitted_at=datetime.now(timezone.utc),
+    )
+
+    # 8. Launch background worker.
+    background_tasks.add_task(
```
dharaneeshvrd: can we please replace this with asyncio.create_task?
Context regarding the change https://github.com/IBM/project-ai-services/pull/1232#discussion_r3764210794

services/translate/utils/llm.py
```diff
@@ -0,0 +1,84 @@
...
+        "max_tokens": max_tokens,
+        "temperature": temperature,
+    }
+
+    logger.debug(
+        f"Calling vLLM translate: model={model}, max_tokens={max_tokens}, "
+        f"temperature={temperature}, messages_count={len(messages)}"
+    )
+
+    response = await client.post(
```
dharaneeshvrd, line 90: let's use misc_utils.SESSION to follow the behaviour from other services.
Please take a reference from https://github.com/IBM/project-ai-services/pull/1237

services/translate/api/v1/jobs.py
Niharika0306: @dharaneeshvrd , I just rechecked to be sure before changing. Bob gave me this analysis for the scenario.

----------------------------------------------------------

The short answer is: no, and BackgroundTasks is the right choice here. Here's the precise reasoning:

asyncio.create_task vs BackgroundTasks in this context
asyncio.create_task schedules a coroutine on the event loop immediately — before the HTTP response has been sent. The task runs concurrently with response serialisation and sending. More critically:

If the task raises an unhandled exception, it becomes a silent "lost exception" — asyncio logs it to sys.stderr as Task exception was never retrieved, but the caller gets no signal and no structured error log from your logger
There is no lifecycle hook — nothing cancels orphaned tasks on shutdown, and uvicorn/starlette has no visibility into them
You'd have to manually hold a reference to the task (e.g. in a set) or it gets garbage-collected before it finishes
BackgroundTasks.add_task is Starlette's purpose-built mechanism for exactly this pattern:

The task runs after the response has been fully sent to the client — the 202 is guaranteed to land before the worker starts
Starlette integrates it with the ASGI lifespan — on shutdown, in-flight background tasks are allowed to drain
Exceptions surface through the normal ASGI exception handler chain and your logger catches them (the try/except in run_translation_job already handles this)
It requires no manual task-set bookkeeping
The only scenario where asyncio.create_task would be preferred is if you needed the task to outlive the request beyond Starlette's background task drain window, or to be cancellable from elsewhere — neither applies here.

services/translate/utils/llm.py
Niharika0306: I'm still analysing Bob's explanation for this:

====================

Look at chatbot/app.py:565 and :587 — the chatbot, which is fully async, wraps every SESSION-based call in asyncio.to_thread:

vllm_stream = await asyncio.to_thread(query_vllm_stream, ...)
vllm_non_stream = await asyncio.to_thread(query_vllm_models, ...)

So the other services do not call SESSION.post() directly from async context — they always bridge it through asyncio.to_thread. That's consistent with what our chunking.py does for tokenize_with_llm.

This actually reinforces why SESSION is wrong for query_vllm_translate: you'd need to rewrite it as a sync function and wrap it in asyncio.to_thread at every call site in the worker. Doing that for the LLM translation calls specifically would mean:

Each chunk translation runs in the default thread pool, not on the event loop
asyncio.gather still fires all tasks concurrently, but they block thread-pool workers rather than yielding the event loop — the semaphore-based concurrency control still works, but you lose the efficiency benefit of true async I/O for what can be multi-second LLM calls
For a service whose primary I/O is long-lived LLM requests (vs. the brief tokenize calls in chunking), that's a meaningful throughput regression
The architecture is correct as-is: SESSION + asyncio.to_thread for short, high-frequency blocking calls (tokenize); httpx.AsyncClient + native await for long-lived, latency-sensitive LLM completion calls.

--------------------------------------------------------------

I mean it should be okay to switch to misc_utils.SESSION for consistency ( just some minor setback during long calls probably). Let me know if you have any thoughts on this?

services/translate/workers/translation_worker.py
Niharika0306: Yes, its working well for now, and didn't find any discrepancies so far.

services/translate/utils/chunking.py
Niharika0306: Addressed.

services/translate/workers/translation_worker.py
Niharika0306: The current libraries work fine for both the txt and md files. Will experiment this if any issues seen.  
Have changed the variable name though for better representation. ( translated_text)

services/translate/utils/chunking.py
Niharika0306: Done.

services/translate/api/v1/jobs.py
Niharika0306: Done.

services/translate/api/v1/jobs.py
Niharika0306: Handled the case where it can return None.

services/translate/api/v1/jobs.py
Niharika0306: Done.

services/translate/utils/chunking.py
Niharika0306: Addressed.

services/translate/utils/chunking.py
Niharika0306: Have addressed this.

services/translate/utils/prompt.py
Niharika0306: Handled.

services/translate/api/v1/jobs.py
dharaneeshvrd: Yes @manalilatkar we should do that, I see its just duplicated in translate.
@Niharika0306 can you please move to common utils and consume it in both the services?

services/translate/api/v1/jobs.py
dharaneeshvrd: Ok it might work for txt and md content, but when we start supporting pdf and word files, this might not be applicable, please add a note here that it needs to be revisited when pdf and word are supported(of course with digitize integration).

services/translate/workers/translation_worker.py
dharaneeshvrd: Agree, LLM would nt have any issue with it, I was thinking md parsers would help in processing(splitting and reorganising post translation) md files better, If its working well for now, we can keep it. But let's create a JIRA story to do some experimentation with this in future.

services/translate/api/v1/jobs.py
dharaneeshvrd: @Niharika0306 
We are not expecting to receive exception or any communication from this triggered task back to the API handler.
Bakcground task is something that needs to be used for tasks that live within the lifecycle of the API request & response, in our case we are spawning the task and let it finish the even though the request ends. And any error or failure of the spawned task is updated to the db.
Hence we decided to move to asyncio's create_task, you can compare between the two.

services/translate/utils/llm.py
dharaneeshvrd: Ok maybe we can go ahead with this for now, but I m worried running these calls in event loop may block if we receive 8 translation jobs at once. Maybe you can validate this scenario and see how it behaves. If it's blocking then we need to move to run via asyncio.to_thread.
Can you please create a story for this to do it later?

services/translate/workers/translation_worker.py
```diff
@@ -0,0 +1,354 @@
...
+            for chunk in chunks:
+                available = max_model_len - chunk.token_count - prompt_overhead
+                if available <= 0:
+                    raise RuntimeError(
+                        f"CONTEXT_LIMIT_EXCEEDED: chunk {chunk.index} "
+                        f"({chunk.token_count} tokens) + prompt overhead ({prompt_overhead}) "
+                        f"leaves no room for output in a {max_model_len}-token context window. "
+                        f"Reduce CHUNK_TOKEN_BUDGET."
+                    )
+                chunk_max_tokens[chunk.index] = min(available, chunk.token_count * 2)
```
dharaneeshvrd: Is there any chance for this to raise the CONTEXT_LIMIT_EXCEEDED?
Since we are building the chunks with in the budge of max model len, I dont think this would occur ever
Please remove if not required

dharaneeshvrd (commented):
Overall lgtm
I m just bit hesitant on running all the vLLM calls on the event loop, please create the story and validate it like I asked in one of the comment

manalilatkar:
LGTM too.

services/translate/api/v1/jobs.py
Niharika0306: I have now moved it to common utility, and imported from there for translate service. 
@manalilatkar , can you import these changes into extract service in the next PR? ( I did try, but had some issues and also felt to keep this PR translate specific).

services/translate/api/v1/jobs.py
Niharika0306: Added note.

services/translate/workers/translation_worker.py
Niharika0306: Created story to track - [JIRA](https://jsw.ibm.com/browse/AISERVICES-1639)

services/translate/api/v1/jobs.py
Niharika0306: Agree. Have done the respective changes.

services/translate/utils/llm.py
Niharika0306: Created story - [JIRA](https://jsw.ibm.com/browse/AISERVICES-1641)

services/translate/workers/translation_worker.py
Niharika0306: Right, usually this will be taken care during the chunking itself. 
But there is one edge case worth noting: oversized GFM tables are allowed to exceed budget and are emitted as-is. So a table chunk could have token_count > budget. So just to handle some of these edge cases and as a precaution, planning to keep this block.

Niharika0306:
@dharaneeshvrd @manalilatkar , updated all review comments and bumped the versions too. Let me know if its in good state to merge.

manalilatkar (approved):
LGTM

PR #1241, opened 2026-08-11
fix: set nginx worker_processes to fixed value to prevent OOMKill

pranithraoibm:
nginx worker_processes auto reads /proc/cpuinfo (node CPU count) instead of the pod's cpu limit, causing pods to spawn 48 worker processes on a 48-vCPU node and get OOMKilled.

Set worker_processes to a fixed value aligned with each pod's cpu limit:
- catalog (500m limit): 2
- chatbot (50m limit):  1
- digitize (50m limit): 1

Also reduced worker_connections from 1024 to 256 to match the reduced worker footprint.

### DIFF
--- a/ui/digitize/nginx.conf.tmpl
+++ b/ui/digitize/nginx.conf.tmpl
@@ -1,10 +1,10 @@
-worker_processes auto;
+worker_processes 1;
 
 error_log  /var/log/nginx/error.log warn;
 pid        /tmp/nginx.pid;
 
 events {
-    worker_connections 1024;
+    worker_connections 256;
 }
 
 http {
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.228
+TAG?=v0.0.229
 CONTAINER_BUILDER?=podman
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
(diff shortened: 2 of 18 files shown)


Shubhamag12 (approved):
lgtm
one question: are these configurations tested?

ui/digitize/nginx.conf.tmpl
```diff
@@ -1,10 +1,10 @@
-worker_processes auto;
+worker_processes 1;
 
 error_log  /var/log/nginx/error.log warn;
 pid        /tmp/nginx.pid;
 
 events {
-    worker_connections 1024;
+    worker_connections 256;
 }
```
mayuka-c, line 8: How did we come up with this number?
How exactly will this be impacted?. Did we do thorough validations and ensure no issues?

mayuka-c: cc. @yussufsh @dharaneeshvrd @manju956

dharaneeshvrd: @pranithraoibm can you please explain what is this for?

yussufsh: @pranithraoibm please put a comment to change the numbers when someone changes the limits. Better add comment at both the places.

pranithraoibm: So this change was something that Bob had suggested

worker_connections defines the maximum number of simultaneous connections per worker process.
Total max connections = worker_processes × worker_connections
In our case, it is 512 for catalog and 256 for chatbot and digitize

256 was chosen because it is proportionate to what the pod can actually handle given its CPU limit (500m and 50m). 
So this right-sizes the connection pool to match the worker count and CPU budget, reducing unnecessary memory reservation. 

Just wanted to highlight this, we can revert back if this seems as a unnecessary improvement.

PR #1243, opened 2026-08-11
docs: add jira 3-in-a-box workflow with data connectors worked example

tsambari2:
- 3-in-a-box model overview (PM / Design / Engineering)
- Issue hierarchy, lifecycle, phased delivery, link types
- Components, labels and fix version reference
- Saved JQL filters and status transitions
- Section 10: worked example using AISERVICES-1411 epic

### DIFF
--- a/docs/product-tracking/jira-3-in-a-box-workflow.md
+++ b/docs/product-tracking/jira-3-in-a-box-workflow.md
@@ -1,0 +1,402 @@
+# Jira 3-in-a-Box Enhancement Workflow — AI Services
+
+> A **3-in-a-box model** (Product Management · Design · Engineering) for managing product enhancements from idea to delivery, grounded in the **AISERVICES** Jira project's actual issue types, components, fix versions and custom fields.
+
+---
+
+## 1. What is 3-in-a-Box?
+
+3-in-a-box means all three disciplines — **Product Management, Design, and Engineering** — are equal co-owners of an enhancement. No discipline throws work over the wall to the next. Each has a defined role at every stage, and no stage begins without the previous owner completing their gate.
+
+```mermaid
+graph LR
+    PM["📋 Product Management\n─────────────────\n• Defines the WHY\n• Writes requirements\n• Accepts delivery"]
+    DES["🎨 Design\n─────────────────\n• Defines the HOW (UX)\n• Produces Figma frames\n• Approves before dev starts"]
+    ENG["⚙️ Engineering\n─────────────────\n• Defines the WHAT (impl)\n• Breaks into stories\n• Delivers & demos"]
+
+    PM <-->|"continuous\ncollaboration"| DES
+    DES <-->|"continuous\ncollaboration"| ENG
+    ENG <-->|"continuous\ncollaboration"| PM
+```
+
+---
+
+## 2. Issue Hierarchy
+
+Every enhancement flows through three levels. Epic is owned by PM; Design Story by Design; Dev Story by Engineering.
+
+```mermaid
+graph TD
+    E["🟣 Epic\nEnhancement Initiative\nOwner: PM"]
+    D["🟠 Story — Design\nFigma deliverable\nOwner: Design"]
+    S["🟢 Story — Engineering\nDev implementation unit\nOwner: Eng"]
+    ST["⚪ Sub-Task\nPR split / BE vs FE\nOwner: Eng"]
+
+    E -->|"Epic Link\ncustomfield_10100"| D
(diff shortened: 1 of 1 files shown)


docs/product-tracking/jira-3-in-a-box-workflow.md
```diff
@@ -0,0 +1,402 @@
...
+
+### Design inbox responsibilities
+- If PM has not created the Design Story: creates a `Story` linked to the Epic; sets component `Design`; `label: design`
+- Produces Figma frames; attaches Figma URL to description
+- Moves story → **In Review** when frames are ready; iterates on PM feedback
+- **Done** = design gate passed — unblocks Engineering
+
+### Engineering inbox responsibilities
+
+- Creates Dev Stories per component linked to the Epic; sets `Fix Version`, `Component`, Sprint; `label: development`
```
dharaneeshvrd, line 79: IMO, it would be better if the PM created the high-level development stories as well. I am not clear on when the development team needs to create these stories.

docs/product-tracking/jira-3-in-a-box-workflow.md
```diff
@@ -0,0 +1,402 @@
...
+| `Performance` | Theresa Xu | Benchmark Epics / Stories |
+| `Documentation` | Ira Pandey | Doc Stories / Tasks / Sub-Tasks |
+
+### Label conventions
+
+| Label | Applied to | Inbox filter |
+|---|---|---|
+| `design` | Design Story | Design inbox |
+| `development` | Dev Story | Eng sprint board |
+| `no-design-needed` | Epic | Bypasses design gate (back-end / infra only) |
```
dharaneeshvrd, line 187: If you consider these new services, for current release it does not have UI. But next release it might have. How do we handle this situation?

docs/product-tracking/jira-3-in-a-box-workflow.md
mkumatag: Well, I prefer that the development team owns the creation of dev stories, retaining full control over their creation and content based on our needs.

docs/product-tracking/jira-3-in-a-box-workflow.md
mkumatag: Perhaps you could create separate stories with clear release mappings to show target delivery dates. Since an epic can span multiple releases, we can develop and deliver non-UI stories first and UI/design stories later. What do you think?

mkumatag:
All: Please review this PR. To keep things moving, we'll merge this today by lazy consensus if there are no further comments, and tackle any follow-ups later.

docs/product-tracking/jira-3-in-a-box-workflow.md
```diff
@@ -0,0 +1,402 @@
...
+| `Automation+QE+DevOps` | Manjunath AC | CI/CD, test-coverage Stories, Automation Stories, Bugs |
+| `Performance` | Theresa Xu | Benchmark Epics / Stories |
+| `Documentation` | Ira Pandey | Doc Stories / Tasks / Sub-Tasks |
+
+### Label conventions
+
+| Label | Applied to | Inbox filter |
+|---|---|---|
+| `design` | Design Story | Design inbox |
+| `development` | Dev Story | Eng sprint board |
```
yussufsh, line 186: Is it ok to assign this label just for the stories? If I am not wrong (as discussed in the scrum yesterday) even sub-tasks were asked to be labeled.
We can have this rule applied only for a story. If needed for sub-tasks as well, some kind of automation would work since I assume all tasks under a story will belong to the same category.

docs/product-tracking/jira-3-in-a-box-workflow.md
dharaneeshvrd: Agree, so we would follow more or less similar approach for development stories what is followed today, If I understand correctly this process would bridge the gap between PM, Design & UI mainly.

docs/product-tracking/jira-3-in-a-box-workflow.md
tsambari2: I agree. Epic can span multiple releases. We can create the design story under same epic and corresponding engineering stories with affects version mapping to the next release.

docs/product-tracking/jira-3-in-a-box-workflow.md
tsambari2: Yes. you can apply the label to the stories. I will see how I can automate the sub tasks to have same label as the parent

mkumatag (approved):

PR #1244, opened 2026-08-11
Vendor ibmcloud-api-mcp into mcp/, remove IBM Cloud region and hostname logic

noahmasoud:
- **vendor source code and renamed module for ai-services**
- **remove hostname validation, rename the binary to fit project**
- **remove region server logic from interface, aggregator, and provider**
- **remove unused RegionServer type**


mkumatag:
This sounds like an entire copy from a remote project, is there a scope of reusing the remote code by referring that wherever needed?

ryanmarc:
> This sounds like an entire copy from a remote project, is there a scope of reusing the remote code by referring that wherever needed?

The MCP server is code that I wrote, but the repo is private. It was supposed to be open-sourced, but ultimately never was because I changed teams and the focus of that team ended up shifting to other things. That repo is no longer maintained and I cannot make it public. This PR repurposes that code to remove the specific IBM Cloud API bits. Even if the repo were public, it can not be used as-is because it was originally written with some constraints around the IBM Cloud API that are not relevant to AI Services.

mkumatag:
> The MCP server is code that I wrote, but the repo is private. It was supposed to be open-sourced, but ultimately never was because I changed teams and the focus of that team ended up shifting to other things. That repo is no longer maintained and I cannot make it public. This PR repurposes that code to remove the specific IBM Cloud API bits. Even if the repo were public, it can not be used as-is because it was originally written with some constraints around the IBM Cloud API that are not relevant to AI Services.

thanks for the clarification.. I'm adding @mayuka-c to review the golang code..

ryanmarc (approved):

PR #1248, opened 2026-08-12
refactor : migrate DigitalAssistants to shared table infrastructure

Niranjan-J-Joshi:
### PR 4 — Migrate DigitalAssistants to shared infrastructure

Replace duplicated implementations in `DigitalAssistants.tsx` and its supporting files with the shared hooks, utilities, and components from PRs 1–3. 

**Key changes:**

- Replace existing toolbar JSX with `<TableToolbarActions>`
- Replace inline delete modal JSX with `<DeleteModal>`
- Replace inline export modal JSX with `<ExportModal>`
- Replace inline delete-error and export toast JSX with `<TableToasts>`
- Replace inline empty state JSX with `<TableEmptyStates entityName="digital assistant">` 
- Remove `StatusCell`, `MessageCell`,`ActionCell` and `NameCell` from `DigitalAssistants/CellRenderers.tsx` — import from shared
- Remove inline `downloadCSV` function — adopt `useCSVExport`
- Remove fetch lifecycle `useEffect`s — adopt `useAutoRefresh`
- Remove export toast auto-dismiss `useEffect` — adopt `useExportToastAutoDismiss`
- Replace inline search filter logic with `tableUtils.filterRowsBySearch`
- Replace inline visible-header calculation with `tableUtils.getVisibleHeaders`
- Replace column reset hardcoded object with `reducerUtils.resetColumnVisibility(DEFAULT_VISIBLE_COLUMNS)`
- Rename `isLoadingApplications` → `isLoading` 
- Update `DigitalAssistantRow["status"]` union: replace `"Deploying..." | "Deleting..."` with `"Deploying" | "Deleting"`;
- Add `UPDATE_ROW_STATUS` action to DA reducer — dispatched immediately on delete confirm to set the row to `"Deleting"` while the API call is in flight; matches the existing DS pattern
- Extend `DigitalAssistants.module.scss` to compose from `table.shared.module.scss`; remove the now-duplicated rules


ui/catalog/src/pages/DigitalAssistants/types.ts
```diff
@@ -1,119 +1,67 @@
...
+  handleSharedTableAction,
+  isSharedTableAction,
+  setLoading,
+} from "@/components/Table/utils/reducerUtils";
 
 export interface DigitalAssistantRow {
   id: string;
   name: string;
-  status: "Deploying..." | "Deleting..." | "Error" | "Stopped" | "Running";
+  status:
```
ryanmarc, line 16: Stopped has no producer. ApplicationStatus in `internal/pkg/catalog/db/models/application.go` is exactly `Downloading | Deploying | Running | Deleting | Error`.

`Initializing` is missing, and it is real. This row type is also used for the child rows built in `applications.api.ts`, where the value is `service.status`, and `ServiceStatus` at `application.go` is `Initializing | Running | Error`.

Nothing renders wrong right now because the shared `STATUS_CONFIG` has entries for both, so an Initializing child gets the right tag anyway. But the as casts at `applications.api.ts` mean the compiler cannot tell you when this drifts, so the union is only worth what we put in it. Either:

`status: "Downloading" | "Deploying" | "Running" | "Deleting" | "Error" | "Initializing";`

or, if you would rather keep parent and child honest separately, give the children their own row type with the three service statuses. Second option is more work than this PR signed up for, so the one-liner is fine by me.

ryanmarc (changes requested):

ui/catalog/src/pages/DigitalAssistants/types.ts
Niranjan-J-Joshi: Added "Initializing" in row status and removed "Stopped"

ryanmarc (dismissed):

pranithraoibm (approved):

PR #1249, opened 2026-08-13
Add gRPC gateway for the workers

yussufsh:
Adds a gRPC worker gateway to the catalog API server so remote workers can connect and receive deployment commands.

- A protobuf service (WorkerGateway) with Register and bidirectional CommandStream RPCs, an in-memory Registry that tracks worker state and heartbeats, and a Gateway gRPC server that authenticates workers with single-use bootstrap tokens.
- DB migration (20260707000001_create_workers_table.sql) adds a workers table with typed enums for status and runtime type, a metadata JSONB column, and the standard updated_at trigger.
- Catalog apiserver starts the gRPC gateway on boot (default port 9090, configurable via --workergateway-port). The same flag is added to catalog configure so it is written into the Podman deployment at setup time.


ai-services/internal/pkg/worker/registry/registry.go
```diff
@@ -0,0 +1,266 @@
...
+	"github.com/google/uuid"
+	"github.com/project-ai-services/ai-services/internal/pkg/catalog/db/models"
+	"github.com/project-ai-services/ai-services/internal/pkg/catalog/db/repository"
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+	workerpb "github.com/project-ai-services/ai-services/internal/pkg/worker/proto"
+)
+
+const (
+	// commandChannelSize is the buffer size for the per-worker command channel.
+	commandChannelSize = 32
```
mayuka-c, line 28: A query, what is this commandChannelSize?

ai-services/internal/pkg/worker/proto/worker.proto
```diff
@@ -0,0 +1,96 @@
...
+  COMMAND_TYPE_LIST_SECRETS       = 12;
+  COMMAND_TYPE_DELETE_SECRET      = 13;
+  COMMAND_TYPE_SECRET_EXISTS      = 14;
+  COMMAND_TYPE_DELETE_VOLUME      = 15;
+  COMMAND_TYPE_VOLUME_EXISTS      = 16;
+  COMMAND_TYPE_INSPECT_CONTAINER  = 17;
+  COMMAND_TYPE_CONTAINER_EXISTS   = 18;
+  COMMAND_TYPE_CONTAINER_LOGS     = 19;
+  COMMAND_TYPE_LIST_ROUTES        = 20;
+  COMMAND_TYPE_DELETE_PVCS        = 21;
```
mayuka-c, line 81: Should we segregate based on runtime and different enums for podman and openshift and then use composition here?

ai-services/internal/pkg/worker/proto/worker.proto
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// ──────────────────────────────────────────────────────────────────────────────
+// Registration messages
+// ──────────────────────────────────────────────────────────────────────────────
+
+message RegisterRequest {
+  string worker_name      = 1;
+  string pre_shared_token = 2;
+  map<string, string> metadata = 3;
```
mayuka-c, line 30: What will be present in metadata?
Is it runtime and domain_suffix for now?

ai-services/internal/pkg/worker/gateway/gateway.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	g.grpcServer = grpc.NewServer()
+	workerpb.RegisterWorkerGatewayServer(g.grpcServer, g)
+
+	go func() {
+		logger.InfofCtx(ctx, "WorkerGateway gRPC server listening on %s", addr)
+		if err := g.grpcServer.Serve(lis); err != nil {
+			logger.ErrorfCtx(ctx, "WorkerGateway gRPC server exited: %v", err)
+		}
```
mayuka-c, line 73: Dont we need to panic if we are not able to open grpc server so that the pod restarts

mayuka-c: Maybe returning error and let the caller panic if anything fails in starting grpc serever

ai-services/internal/pkg/worker/gateway/gateway.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	workers, err := g.repo.GetAll(ctx)
+	if err != nil {
+		logger.WarningfCtx(ctx, "WorkerGateway sweeper: failed to fetch workers: %v", err)
+
+		return
+	}
+
+	status := models.WorkerStatusDisconnected
```
mayuka-c: I see this variable being used in only one place. Can we just not directly assign the state there?

ai-services/internal/pkg/worker/gateway/gateway.go
```diff
@@ -0,0 +1,260 @@
...
+	now := time.Now()
+
+	for _, w := range workers {
+		if w.Status == models.WorkerStatusDisconnected {
+			continue
+		}
```
mayuka-c, line 121: I see if we didnt receive heartbeat we are marking as disconnected. What if the LPAR was restarted, dont we need to mark it connected in next cycle?

Is it being checked somewhere else?

ai-services/internal/pkg/worker/gateway/gateway.go
```diff
@@ -0,0 +1,260 @@
...
+// identifyWorker reads the first message from the stream, validates the worker is known,
+// and returns the worker name and registry entry.
+func (g *Gateway) identifyWorker(ctx context.Context, stream grpc.BidiStreamingServer[workerpb.CommandResult, workerpb.Command]) (string, *registry.WorkerEntry, error) {
+	firstMsg, err := stream.Recv()
+	if err != nil {
+		return "", nil, fmt.Errorf("CommandStream: failed to receive first message: %w", err)
+	}
+	workerName := firstMsg.GetWorkerName()
+	if workerName == "" {
+		return "", nil, fmt.Errorf("CommandStream: first message missing worker name")
```
mayuka-c: What if the control plane couldnt receive first msg or there was something missing. I see we are returning error. Will there be a retry or do we need to again do registration process in worker?

ai-services/internal/pkg/worker/registry/registry.go
yussufsh: This is the channel size, ie. how many commands can sit in line waiting to be processed before the gRPC server starts blocking new incoming requests.

ai-services/internal/pkg/worker/proto/worker.proto
yussufsh: Yes, planning to do this in the future when OpenShift comes in the picture.

ai-services/internal/pkg/worker/proto/worker.proto
yussufsh: Yes, agent will send this information in future. Few I can think of is models path, domain suffix, runtime, etc.

ai-services/internal/pkg/worker/gateway/gateway.go
yussufsh: Did not think of it in that way since it is running async in a goroutine.

ai-services/internal/pkg/worker/gateway/gateway.go
yussufsh: It needs a pointer, if I am not wrong?

ai-services/internal/pkg/worker/gateway/gateway.go
yussufsh: Will try to fix this.

yussufsh: Please check the latest commit.

ai-services/internal/pkg/worker/gateway/gateway.go
yussufsh: This needs to be registered manually or auto on restart. That is why the register is using the upsert method.

ai-services/internal/pkg/worker/gateway/gateway.go
yussufsh: Yes if worker is trying to open the stream connection without registeration first this will fail. Or if the stream connection drop, the worker needs to re-register itself.

yussufsh: Now improved this to use the grpc error codes.

ai-services/internal/pkg/worker/registry/registry.go
mayuka-c: Will this not be be a blocker with no of current users?

ai-services/internal/pkg/worker/gateway/gateway.go
mayuka-c: So after every system reboot, user needs to run the command again?

ai-services/internal/pkg/worker/gateway/gateway.go
mayuka-c: I feel we use something like StringPtr similar to BoolPtr method present in utils

ai-services/internal/pkg/worker/registry/registry.go
yussufsh: So this is per worker and every deployment will send the command one at a time, that means we can perform ~32 deployments at a time for a specific worker. I feel this is more than enough.

ai-services/internal/pkg/worker/gateway/gateway.go
yussufsh: Will add a generic Ptr func.

ai-services/internal/pkg/worker/gateway/gateway.go
yussufsh: Yes needs to be run again (or auto-restarted in a container or `Restart=on-failure` may be via systemd).
Worker will call `Register` (which is an upsert — idempotent), then open `CommandStream` which should work if mTLS is already set up.

ai-services/internal/pkg/worker/registry/registry.go
yussufsh: Added more comments to the constant.

mayuka-c: It might not just be deployment right?
We may have sync service even that would consume, also if users use application ps even that will consume.

I guess this should be fine.

Also is this queing mechanism? If yes, then it may also add some delays. Maybe we need to have goroutine pools and do some concurrent processing too.

Maybe can be taken later

mayuka-c (dismissed):
LGTM

mayuka-c (approved):

PR #1250, opened 2026-08-13
refactor: extract shared useDeployFlowReducer hook and unify initial state

pranithraoibm:
- Extracts the shared reducer cases and four callbacks into `Shared/hooks/useDeployFlowReducer.ts`
- Moves RESET_STATE out of SharedDeployFlowAction and into each flow's own ACTION_TYPES and DeployFlowAction union — the shared reducer never handled it, each flow resets to its own initial state
- Extracts BASE_INITIAL_STATE into `Shared/utils/formData.ts` — both getInitialState functions spread it and only declare their own flow-specific fields
- Moves SHOW_DEPLOY_TOAST / HIDE_DEPLOY_TOAST into the shared reducer — they previously only existed in the DA flow; Services gains them here so DeployTearsheetShell (PR 7) can read deployToastOpen from both flows.


ui/catalog/src/components/DeployFlow/Shared/types.ts
```diff
@@ -84,6 +86,7 @@ export interface BaseDeployFlowState {
   isEditing: boolean;
   hasInsufficientResources: boolean;
   deployError: string | null;
+  deployToastOpen: boolean;
```
ryanmarc, line 89: Adding `deployToastOpen` to the base state is the right call, but the Services flow never sets it. `SHOW_DEPLOY_TOAST` is only dispatched from `DigitalAssistantDeployFlow.tsx`, so on Services this field is permanently false

ryanmarc (changes requested):

ui/catalog/src/components/DeployFlow/Shared/types.ts
pranithraoibm: Added it to services as well

ryanmarc (approved):

PR #1252, opened 2026-08-13
Swap IAM JWT validation for Bearer token passthrough

noahmasoud:


mcp/internal/server/http.go
```diff
@@ -309,14 +262,11 @@ func (s *HTTPServer) rateLimitMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		r.Header.Add("Accept", "text/event-stream")
 
-		claims, ok := r.Context().Value(claimsContextKey).(*IAMAccessTokenClaims)
-		if !ok || claims.IAMID == "" {
-			http.Error(w, "Invalid or missing claims", http.StatusUnauthorized)
-			return
+		host, _, err := net.SplitHostPort(r.RemoteAddr)
+		if err != nil {
+			host = r.RemoteAddr
```
ryanmarc: because of Caddy, the `RemoteAddr` here is the Caddy IP. This ends up making the host the same value for everyone and hitting the rate limit would end up causing a 429 for all users.

This should look for something specific to the actual caller like `X-Forwarded-For` from `r.Header`

noahmasoud: added some changes

ryanmarc (approved):

PR #1257, opened 2026-08-14
Fix the lint rule issues regarding D400 and D103

Bleama02:
This change is necessary for a future addition of lint rules D400 and D103 to help with MCP work. 

Please see below links for more info:
[AISERVICES-1601 ](https://jsw.ibm.com/browse/AISERVICES-1601)

[AISERVICES-1598](https://jsw.ibm.com/browse/AISERVICES-1598)

PR #1258, opened 2026-08-14
Added a ruff pydocstyle rules (D103, D400) 

Bleama02:
Added a ruff pydocstyle rules (D103, D400) to /services/common/pyproject.toml. Added lint check step in .gith
ub/workflows/service-unit-tests.yml which adds an extra step to the CI.

Please look at ticket below for more info:
[AISERVICES-1598](https://jsw.ibm.com/browse/AISERVICES-1598)


.github/workflows/service-unit-tests.yml
```diff
@@ -31,7 +31,14 @@ jobs:
           pip install -r services/requirements-test.txt \
             --extra-index-url https://wheels.developerfirst.ibm.com/ppc64le/linux \
             --prefer-binary
-      
+
+      - name: Install ruff
+        run: pip install ruff
+
+      - name: Lint docstrings (common)
+        run: ruff check common/ --exit-zero
```
adarshagrawal38: We should be falling job when there is lint error.

why are we doing `--exit-zero`?

.github/workflows/service-unit-tests.yml
adarshagrawal38: How about naming action to `Services Lint & Unit Tests`, as now we have lint check as well

.github/workflows/service-unit-tests.yml
```diff
@@ -31,7 +31,14 @@ jobs:
           pip install -r services/requirements-test.txt \
             --extra-index-url https://wheels.developerfirst.ibm.com/ppc64le/linux \
             --prefer-binary
-      
+
+      - name: Install ruff
+        run: pip install ruff
+
+      - name: Lint docstrings (common)
```
adarshagrawal38: is there a reason why only in common folder we are running check, instead of entire service folder?

.github/workflows/service-unit-tests.yml
Bleama02: @ryanmarc suggested that it be like this for now until we fix the actual lint errors that the project currently has. Just so it doesn't brake all the current PR's. It will be removed later.

.github/workflows/service-unit-tests.yml
Bleama02: Yeah, I can do that.

.github/workflows/service-unit-tests.yml
Bleama02: You know what. That is a good question haha. When originally thinking about doing this, I thought that common had all the things it needed in order for the MCP work to be as efficient as possible. But looking into it further it probably should be in each service. Ill talk to Ryan about how to go about this since checking all the service folders will cause it to have 300+ issues that need to be fixed.

.github/workflows/service-unit-tests.yml
Bleama02: I made changes to to fix all the issues with the lint rules in this PR(https://github.com/IBM/project-ai-services/pull/1257). 

So I took out the --exit-zero because it is no longer needed.

.github/workflows/service-unit-tests.yml
Bleama02: Done

.github/workflows/service-unit-tests.yml
Bleama02: Changes made so that it hits all the services

adarshagrawal38 (approved):
LGTM

ryanmarc (approved):

PR #1261, opened 2026-08-17
Implement routes for workers

yussufsh:
Worker gateway infrastructure — gRPC control plane for remote workers

- worker/registry — Registry (in-memory worker map + lifecycle: Register, Preregister, Disconnect, Deregister, SweepStale, UpdateHeartbeat, WaitForResult, DeliverResult) and TokenStore (single-use 24-hour bootstrap tokens).
- worker/gateway — gRPC server (Register RPC + CommandStream bidi-stream) that authenticates workers via token, routes commands to them, and sweeps stale heartbeats. Listens on :9090 by default, always active.
- catalog/apiserver — WorkerGatewayPort + WorkerRegistry wired into APIServerOptions; gateway started in Start(ctx); worker REST routes (POST/GET/DELETE /api/v1/workers) added to the router via WorkerHandler.
- catalog/db/repository/worker_repo.go — WorkerRepository interface with Upsert, Update, Delete, GetAll; full pgx implementation.
- cmd/catalog/apiserver.go — --workergateway-port flag (default 9090), signal.NotifyContext, workerRepo → workerReg wiring.

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/handlers/worker_handler.go
+++ b/ai-services/internal/pkg/catalog/apiserver/handlers/worker_handler.go
@@ -1,0 +1,128 @@
+package handlers
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
+	"github.com/project-ai-services/ai-services/internal/pkg/worker/registry"
+)
+
+// WorkerHandler handles worker management endpoints.
+type WorkerHandler struct {
+	reg *registry.Registry
+}
+
+// NewWorkerHandler creates a new WorkerHandler.
+func NewWorkerHandler(reg *registry.Registry) *WorkerHandler {
+	return &WorkerHandler{reg: reg}
+}
+
+// createWorkerReq is the request body for registering a new worker.
+type createWorkerReq struct {
+	WorkerName string `json:"worker_name" binding:"required,min=1,max=100"`
+}
+
+// createWorkerResp is the response body for a newly registered worker.
+type createWorkerResp struct {
+	WorkerName string `json:"worker_name"`
+	Token      string `json:"token"`
+}
+
+// CreateWorker godoc
+//
+//	@Summary		Register a new worker
+//	@Description	Pre-registers a worker by name, creates a pending DB row, and returns a single-use bootstrap token.
+//	@Description	The operator passes this token when starting the worker daemon (`worker start --token <token>`).
+//	@Tags			Workers
+//	@Accept			json
+//	@Produce		json
+//	@Param			worker	body		createWorkerReq		true	"Worker registration request"
+//	@Success		201		{object}	createWorkerResp	"Worker registered; token valid for 24 hours"
+//	@Failure		400		{object}	map[string]interface{}	"Invalid payload"
+//	@Failure		500		{object}	map[string]interface{}	"Internal error"
+//	@Security		BearerAuth
+//	@Router			/workers [post]
+func (h *WorkerHandler) CreateWorker(c *gin.Context) {
+	var req createWorkerReq
+	if err := c.ShouldBindJSON(&req); err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "invalid payload"})
+
+		return
+	}
+
+	ctx := c.Request.Context()
+
+	token, err := h.reg.Preregister(ctx, req.WorkerName)
+	if err != nil {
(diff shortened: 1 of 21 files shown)


ai-services/internal/pkg/catalog/apiserver/handlers/worker_handler.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// NewWorkerHandler creates a new WorkerHandler.
+func NewWorkerHandler(reg *registry.Registry) *WorkerHandler {
+	return &WorkerHandler{reg: reg}
+}
+
+// createWorkerReq is the request body for registering a new worker.
+type createWorkerReq struct {
+	WorkerName string `json:"worker_name" binding:"required,min=1,max=100"`
```
mayuka-c, line 28: Can we also add sanitizations to WorkerName and also feel like better to be case insensitive

ai-services/internal/pkg/worker/gateway/gateway.go
```diff
@@ -133,19 +95,21 @@ func (g *Gateway) sweepStaleWorkers(ctx context.Context) {
 // ──────────────────────────────────────────────────────────────────────────────
 
 // Register implements WorkerGatewayServer. Workers call this once at bootstrap.
-// Metadata supplied in the request is persisted to the
-// DB metadata JSON column by registry.Register — no separate handling needed here.
+// The worker name is taken from the token, not from the request — the worker
+// cannot self-assign a name different from what was pre-registered by an admin.
```
mayuka-c, line 99: Isnt worker name got from the Create Worker (POST worker call). Didnt quite get about the name got from token

ai-services/internal/pkg/worker/gateway/gateway.go
```diff
@@ -133,19 +95,21 @@ func (g *Gateway) sweepStaleWorkers(ctx context.Context) {
...
+// Metadata supplied in the request is persisted to the DB metadata JSON column.
 func (g *Gateway) Register(ctx context.Context, req *workerpb.RegisterRequest) (*workerpb.RegisterResponse, error) {
-	workerName := req.GetWorkerName()
-	logger.InfofCtx(ctx, "WorkerGateway: Register request from worker_name=%s", workerName)
+	logger.InfofCtx(ctx, "WorkerGateway: Register request (token=%s)", req.GetPreSharedToken())
```
mayuka-c: Is it safe to log the token?

ai-services/internal/pkg/worker/registry/registry.go
```diff
@@ -129,6 +126,37 @@ func (r *Registry) Register(ctx context.Context, req *workerpb.RegisterRequest)
...
+// supersedes the old one. The registry's in-memory map is not touched — the worker
+// is not "connected" until it calls Register via gRPC.
+func (r *Registry) Preregister(ctx context.Context, workerName string) (string, error) {
+	if r.repo == nil {
+		return "", fmt.Errorf("worker registry: no repository configured")
+	}
+
+	w := &models.Worker{
+		Name:        workerName,
+		RuntimeType: models.WorkerRuntimeTypePodman,
```
mayuka-c: This has to be received during register process from worker agent. They need to pass the runtime

ai-services/internal/pkg/worker/gateway/gateway.go
yussufsh: The worker name is stored as part of the token during Pre-registration (create worker flow). That is what it means.

ai-services/internal/pkg/worker/gateway/gateway.go
yussufsh: Not allowed. Sorry missed this, will remove.

ai-services/internal/pkg/worker/registry/registry.go
yussufsh: I was avoiding it now since initially we do not know the runtime and it is a NOT NULL field in the db. I will implement an unknown runtime in the enum to fix this.
We can change it later if needed.

ai-services/internal/pkg/worker/gateway/gateway.go
mayuka-c: Ok so this means the Token object and not actual bootstrap token containing it (like hashed or something)?

mayuka-c (dismissed):
LGTM

mayuka-c (approved):

PR #1262, opened 2026-08-17
[BYOS] DB Schema + Handlers Implementation

mayuka-c:
- New DB table for storing custom asset bundles
- DB Handlers implementation for the same.


ai-services/internal/pkg/catalog/db/repository/bundle_repo.go
```diff
@@ -0,0 +1,241 @@
...
+	// GetByID retrieves a single bundle row by its UUID primary key.
+	// Returns (nil, nil) when not found.
+	GetByID(ctx context.Context, id uuid.UUID) (*models.CatalogBundle, error)
+
+	// GetActiveByCatalogID returns the single active row for a (catalog_type, catalog_id) pair.
+	// Returns (nil, nil) when no active row exists.
+	GetActiveByCatalogID(ctx context.Context, catalogType, catalogID string) (*models.CatalogBundle, error)
+
+	// Activate sets status='active', version, name, and size_bytes for an existing row.
+	Activate(ctx context.Context, id uuid.UUID, version, name string, sizeBytes int64) error
```
yussufsh: What is the reason for this method? Do we have similar Deactive support?

mayuka-c: This is used within POST and PUT call where we first insert the DB record -> Untar the bundle tar.gz -> calculate the size, other info and then activate once all are done.

Deactive we dont really need (But we can add it later, if we allow user to deactive explicitly using API). It is kind of used within Delete call (so that a user will not be able deploy using this bundle while delete is proceeding) but it will just make the bundle status to `Deleting`. It reuses the `UpdateStatus` method itself.

Shubhamag12: shall we rather rename it to `UpdateBundle`?

ai-services/internal/pkg/catalog/db/repository/bundle_repo.go
```diff
@@ -0,0 +1,241 @@
...
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf("error iterating catalog bundles: %w", err)
+	}
+
+	return bundles, nil
+}
+
+// isNotFound returns true when the error represents a "no rows" result.
+func isNotFound(err error) bool {
+	return err != nil && err.Error() == "no rows in result set"
```
Shubhamag12: shall we do this rather than hardcoding string and directly use it at caller site?

```suggestion
	return errors.Is(err, pgx.ErrNoRows)
```

Shubhamag12 (commented):
minor comment

ai-services/internal/pkg/catalog/db/repository/bundle_repo.go
mayuka-c: Yeah prob we can do that, Let me update

mayuka-c: Done updated

Shubhamag12 (dismissed):
lgtm

ai-services/internal/pkg/catalog/db/repository/bundle_repo.go
```diff
@@ -0,0 +1,250 @@
...
+		"UPDATE catalog_bundles SET %s WHERE id = $%d",
+		strings.Join(setClauses, ", "),
+		i,
+	)
+	args = append(args, id)
+
+	_, err := r.pool.Exec(ctx, query, args...)
+	if err != nil {
+		return fmt.Errorf("failed to update catalog bundle: %w", err)
+	}
```
yussufsh, line 208: Update silently ignores missing rows:
`bundle_repo.go` — Update: r.pool.Exec returns a CommandTag containing RowsAffected(). If the id doesn't exist, the query succeeds (no error) but zero rows are updated — the caller gets no signal. This should check tag.RowsAffected() == 0 and return a sentinel error (e.g. ErrNotFound) so callers don't silently no-op on stale/wrong IDs.
```
tag, err := r.pool.Exec(ctx, query, args...)
if err != nil {
    return fmt.Errorf("failed to update catalog bundle: %w", err)
}
if tag.RowsAffected() == 0 {
    return fmt.Errorf("catalog bundle not found: %s", id)
}
```

ai-services/internal/pkg/catalog/db/migrations/assets/20260430094507_create_catalog_bundles_table.sql
```diff
@@ -0,0 +1,65 @@
...
+    -- The catalog item type: "service" or "component".
+    catalog_type     VARCHAR(50)    NOT NULL,
+
+    -- Unique identity of the catalog item within its type.
+    -- Services:   bare id                               e.g. "my-service"
+    -- Components: composite <component_type>--<id>      e.g. "llm--my-provider"
+    catalog_id       VARCHAR(200)   NOT NULL,
+
+    -- Semantic version of this bundle: e.g. "1.0.0", "2.1.0"
+    version          VARCHAR(50)    NOT NULL DEFAULT '',
```
yussufsh: IMO version should be required. Remove default='' in db.

ai-services/internal/pkg/catalog/db/repository/bundle_repo.go
```diff
@@ -0,0 +1,250 @@
...
+	if createdBy.Valid {
+		b.CreatedBy = createdBy.String
+	}
+
+	return &b, nil
+}
+
+const selectCols = `
+	id, name, status, size_bytes, catalog_type, catalog_id, version, error, created_by, created_at, updated_at
+`
```
yussufsh: This adds `\n\t` to the select statement. Can we have single-line string?

yussufsh (changes requested):

ai-services/internal/pkg/catalog/db/migrations/assets/20260430094507_create_catalog_bundles_table.sql
mayuka-c: Make sense, removed default since it is sync. Maybe when we make async (in future), then it could be empty initially when we insert and later it gets populated (Can be modified later)

ai-services/internal/pkg/catalog/db/repository/bundle_repo.go
mayuka-c: Addressed, ptal

yussufsh (approved):
/lgtm

PR #1266, opened 2026-08-17
Data Source Connectors: APScheduler & Startup Recovery 

sats-23:
-APScheduler logic 
--Schedules and fires a job immediately when a connector is added
--Same job is removed during connector tear down 
--Same job is recovered on restart (without live trigger)

-Startup Recovery 
--When digitize service crashes and comes back up, handles 
1. Marking status of Connectors as OUT OF SYNC 
2. Marking status of ConnectorSyncLog as FAILED
3. Add all jobs back to APScheduler

### DIFF
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -95,9 +95,10 @@ async def create_connector(body: ConnectorCreateRequest):
                 connector_id, sync_interval, fire_immediately=True
             )
         except Exception as sched_exc:
-            logger.warning(
+            logger.error(
                 f"Scheduler registration failed for {connector_id!r}: {sched_exc}"
             )
+            raise
 
         db_ops.insert_connector(
             connector_id=connector_id,
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -100,8 +100,19 @@ async def create_connector(body: ConnectorCreateRequest):
             f"Connector {connector_id!r} ({body.connector_name!r}) attached "
             f"(type={body.type}, interval={sync_interval}s)"
         )
-        # Worker start is a no-op stub in PR3 — the worker manager (PR7) will hook
-        # in here once implemented.
+
+        # Register the connector with the scheduler so it starts ticking
+        # immediately (fire_immediately=True for the first-ever sync).
+        try:
+            import digitize.connectors.scheduler as _sched
+            await _sched.register_connector_job(
+                connector_id, sync_interval, fire_immediately=True
+            )
+        except Exception as sched_exc:
+            logger.warning(
+                f"Scheduler registration failed for {connector_id!r}: {sched_exc}"
+            )
+
         return Response(
             content=f'{{"connector_id": "{connector_id}"}}',
             status_code=201,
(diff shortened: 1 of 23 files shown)


services/digitize/app.py
```diff
@@ -105,7 +108,63 @@ async def lifespan(app: FastAPI):
...
+    # ------------------------------------------------------------------ #
+    # Connector scheduler                                                  #
+    # ------------------------------------------------------------------ #
+    import digitize.connectors.scheduler as scheduler_module
+    from digitize.utils.db import list_connectors
+    from apscheduler import AsyncScheduler
+    from apscheduler.datastores.sqlalchemy import SQLAlchemyDataStore
+    from digitize.db.connection import engine as db_engine
+
+    try:
```
dharaneeshvrd, line 143: can we move this logic to a method and call from here?
lifespan has become really huge.

services/digitize/db/manager.py
```diff
@@ -1494,6 +1494,77 @@ def merge_document_metadata(doc_id: str, metadata: dict) -> bool:
...
+        """
+        try:
+            with get_db_session() as session:
+                result = session.execute(
+                    update(Connector)
+                    .where(Connector.sync_status == ConnectorStatus.SYNCING)
+                    .values(
+                        sync_status=ConnectorStatus.OUT_OF_SYNC,
+                        last_sync_error=error,
+                        error=error,
```
dharaneeshvrd, line 1518: error field is still not introduced right?

dharaneeshvrd (commented):
mostly lgtm

services/digitize/api/v1/connectors.py
dharaneeshvrd, line 103: can we move this after we successfully register the connector with apscheduler?
so that in case of a failure, user won't need to manage this failed connector

services/digitize/db/manager.py
sats-23: It is
Connectors table has an error field from last PR and that is being populated here

dharaneeshvrd: Oh ok, then do we need last_sync_error as well?
can we remove it?

services/digitize/app.py
sats-23: Sure, done

services/digitize/api/v1/connectors.py
sats-23: Done

services/digitize/db/manager.py
sats-23: We don't need 2 unless we plan to use the generic error for things like auth failure and so on.
If not, let's remove it

dharaneeshvrd: yeah not required IMO, maybe you can remove in the next PR, let's not touch it part of this PR

sats-23: Sure

services/digitize/api/v1/connectors.py
```diff
@@ -87,6 +87,18 @@ async def create_connector(body: ConnectorCreateRequest):
...
 
+        # Register the connector with the scheduler so it starts ticking
+        # immediately (fire_immediately=True for the first-ever sync).
+        try:
+            import digitize.connectors.scheduler as _sched
+            await _sched.register_connector_job(
+                connector_id, sync_interval, fire_immediately=True
+            )
+        except Exception as sched_exc:
+            logger.warning(
```
dharaneeshvrd: you need to reject the request if you are not able to schedule it

dharaneeshvrd (approved):

services/digitize/api/v1/connectors.py
```diff
@@ -87,6 +87,19 @@ async def create_connector(body: ConnectorCreateRequest):
...
+        try:
+            import digitize.connectors.scheduler as _sched
+            await _sched.register_connector_job(
+                connector_id, sync_interval, fire_immediately=True
+            )
+        except Exception as sched_exc:
+            logger.error(
+                f"Scheduler registration failed for {connector_id!r}: {sched_exc}"
+            )
+            raise
```
dharaneeshvrd, line 101: Plain raise is enough?
What would be the status code?

sats-23: Yes, that will do. It will return a 500

PR #1247, opened 2026-08-12
AIS-1582 - Changes for Encryption Key for connectors

prachikurade1:
- Add encryption key secret for podman and openshift
- Mount the key for the catalog backend pod
- Load the key in the API server

PR #1268, opened 2026-08-17
AIS-1582 - Introduce DB Encryption Secret key to encrypt data at rest

prachikurade1:
- Closed https://github.com/IBM/project-ai-services/pull/1247 due to change in approach.
- Verified that the secret is created and mounted on the catalog backend pod
- Verified that re-run of the catalog configure command does not re-generate the secret
- Verified that uninstall of catalog deletes the secret.

Note: Change was needed in password generator since it was not working as expected when length is passed. Existing code uses default values.


ai-services/assets/catalog/openshift/templates/catalog-backend.yaml
```diff
@@ -92,6 +92,11 @@ spec:
                 secretKeyRef:
                   name: catalog-db-secret
                   key: db-password
+            - name: CONNECTOR_ENCRYPTION_KEY
```
mayuka-c: Let this be DB_ENCRYPTION_KEY

ai-services/assets/catalog/openshift/templates/catalog-connector-secret.yaml
```diff
@@ -0,0 +1,13 @@
+{{- $existingSecret := lookup "v1" "Secret" .Release.Namespace "catalog-connector-secret" }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: "catalog-connector-secret"
```
mayuka-c: Same here `catalog-db-encryption-key`

ai-services/internal/pkg/utils/password_generator.go
```diff
@@ -383,12 +383,13 @@ func applyOption(key, value string, opts *passwordOptions) error {
...
 func generateValue(annotation string) (string, error) {
-	parts := strings.Split(annotation, ":")
+	parts := strings.SplitN(annotation, ":", minAnnotationParts)
 	if len(parts) < minAnnotationParts {
 		return "", fmt.Errorf("invalid annotation format: %s", annotation)
 	}
 
-	annotationType := parts[1]
+	// Extract just the type keyword before any options (e.g. "password" from "password length=32")
+	annotationType := strings.Fields(parts[1])[0]
```
mayuka-c, line 392: What exactly is this change for?
What would be the parts and annotationType values looks like?

ai-services/internal/pkg/cli/templates/embed.go
```diff
@@ -234,18 +234,37 @@ func (e *embedTemplateProvider) LoadPodTemplateWithValues(app, file, appName str
...
+	return values, nil
+}
+
+func (e *embedTemplateProvider) LoadValues(app string, valuesFileOverrides []string, cliOverrides map[string]string) (map[string]interface{}, error) {
+	// Load the default values.yaml (including @generate annotation processing).
+	valuesPath := e.buildPath(app, getRuntime(), "values.yaml")
+	values, err := e.loadDefaultValues(valuesPath)
+	if err != nil {
+		return nil, err
+	}
```
mayuka-c, line 266: I hope you have validated the application flow with legacy flag and ensure no regressions

ai-services/internal/pkg/utils/password_generator.go
prachikurade1: When we have an annotation with params (@generate:password length=20), the split would cause the annotation type to be `password length=20` instead of just "password" (would work fine when there are params - which is how it has been used so far). Hence this fix was needed.

ai-services/internal/pkg/cli/templates/embed.go
prachikurade1: Verified for openshift. For podman I am getting errors that are not related to this change. Have reported the same in the dev-channel.

ai-services/assets/catalog/openshift/templates/catalog-backend.yaml
prachikurade1: Done

ai-services/assets/catalog/openshift/templates/catalog-connector-secret.yaml
prachikurade1: Done

ai-services/assets/catalog/openshift/templates/catalog-db-encryption-secret.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: "catalog-db-encryption-secret"
+  labels:
+    ai-services.io/application: {{ .Release.Name }}
+    ai-services.io/template: {{ .Chart.Name }}
+    ai-services.io/version: {{ default .Chart.AppVersion | quote }}
+    ai-services.io/component: catalog
+type: Opaque
+stringData:
+  db-encryption-key: {{ if $existingSecret }}{{ index $existingSecret.data "db-encryption-key" | b64dec | quote }}{{ else }}{{ .Values.backend.dbEncryptionKey | quote }}{{ end }}
```
yussufsh, line 13: Can you explain what will happen here? Otherwise, the changes lgtm.

prachikurade1: This checks if a secret already exists and reads that value and re-writes it. If not then new value is generated.
I have added this for re-runs of catalog configure.

mayuka-c: @Shubhamag12 @adarshagrawal38  Please create a task and let us have this for other secrets like DB, admin pasword and others too during catalog configure to ensure idempotency is working for Openshift too!

mayuka-c (approved):
LGTM,

@prachikurade1 Pls merge once all the job passes

PR #1270, opened 2026-08-18
Support for catalog worker cli

yussufsh:
1.  Implemented catalog worker control-plane lifecycle cli
2. Introduced worker types (`pkg/catalog/types`) for cli and api

### DIFF
--- a/ai-services/cmd/ai-services/cmd/catalog/worker.go
+++ b/ai-services/cmd/ai-services/cmd/catalog/worker.go
@@ -46,7 +46,7 @@ bootstrap token.
 Pass the token to the worker daemon at startup so it can authenticate with the
 catalog gRPC gateway:
 
-  worker start --token <token> --gateway <catalog-host>:9090`,
+  worker join --token <token> --gateway <catalog-host>:9090`,
 		Example: `  ai-services catalog worker register node-1`,
 		Args:    cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
--- a/ai-services/cmd/ai-services/cmd/catalog/worker.go
+++ b/ai-services/cmd/ai-services/cmd/catalog/worker.go
@@ -124,13 +124,12 @@ func newWorkerDeregisterCmd() *cobra.Command {
 	var runtimeType string
 
 	cmd := &cobra.Command{
-		Use:   "deregister <id>",
+		Use:   "deregister <name>",
 		Short: "Permanently deregister a worker",
-		Long: `Permanently removes a worker from the catalog by its UUID.
+		Long: `Permanently removes a worker from the catalog by name.
 
-If the worker is currently connected its gRPC stream is also cleaned up.
-Use 'ai-services catalog worker list' to find the worker's ID.`,
-		Example: `  ai-services catalog worker deregister 4b3e1f2a-8c7d-4e5b-9f6a-1d2e3f4a5b6c --runtime podman`,
+If the worker is currently connected its gRPC stream is also cleaned up.`,
+		Example: `  ai-services catalog worker deregister node-1 --runtime podman`,
 		Args:    cobra.ExactArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			return common.InitAndValidateRuntimeFlag(runtimeType)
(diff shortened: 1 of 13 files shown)


ai-services/cmd/ai-services/cmd/catalog/worker.go
```diff
@@ -0,0 +1,168 @@
...
+	if len(workers) == 0 {
+		logger.Infoln("No workers registered.")
+
+		return nil
+	}
+
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, workerTablePadding, ' ', 0)
+	if _, err := fmt.Fprintln(w, "ID\tNAME\tRUNTIME\tSTATUS\tLAST HEARTBEAT"); err != nil {
+		return err
+	}
```
mayuka-c, line 153: Maybe let us follow structure similar to UI. Instead of Runtime it would be Type and also the values would be different. Also please check other columns

mayuka-c (commented):
Rest LGTM

ai-services/cmd/ai-services/cmd/catalog/worker.go
yussufsh: I would like to keep it the same since we have runtime as a parameter during `worker join` and would be proper to display the same value. UI will need to make the runtime options as per their choice.

ai-services/cmd/ai-services/cmd/catalog/worker.go
```diff
@@ -0,0 +1,168 @@
...
+		Use:   "deregister <name>",
+		Short: "Permanently deregister a worker",
+		Long: `Permanently removes a worker from the catalog by name.
+
+If the worker is currently connected its gRPC stream is also cleaned up.`,
+		Example: `  ai-services catalog worker deregister node-1`,
+		Args:    cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.SilenceUsage = true
+
```
mayuka-c, line 119: I feel printing "Performing Deregistering..." log msg would help here, as we are doing external call and would let user know that there is process ongoing

ai-services/cmd/ai-services/cmd/catalog/worker.go
```diff
@@ -0,0 +1,168 @@
...
+Pass the token to the worker daemon at startup so it can authenticate with the
+catalog gRPC gateway:
+
+  worker start --token <token> --gateway <catalog-host>:9090`,
+		Example: `  ai-services catalog worker register node-1`,
+		Args:    cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.SilenceUsage = true
+
```
mayuka-c, line 54: Same here

mayuka-c (commented):
Minor logs, Rest LGTM

ai-services/cmd/ai-services/cmd/catalog/worker.go
mayuka-c: Yeah but I feel we might need to standardize as we show Type in UI and here its runtime and also other columns. Might need to be consistent for users to follow it easily.

Maybe we can see later

ai-services/cmd/ai-services/cmd/catalog/worker.go
yussufsh: It is pretty quick and success/error message is displayed.
```
[root@onprem133721154 ai-services]# ./bin/ai-services-linux-ppc64le catalog worker deregister abc
Worker "abc" deregistered.
```

ai-services/cmd/ai-services/cmd/catalog/worker.go
yussufsh: Same it is just token generation and db entry task.
```
[root@onprem133721154 ai-services]# ./bin/ai-services-linux-ppc64le catalog worker register d
Worker registered successfully.
  Name:  d
  Token: 6f0c124d-f22c-4761-bb1a-30749189ce5b

Pass this token to the worker daemon with --token.
The token is single-use and expires after 24 hours.
```

ai-services/cmd/ai-services/cmd/catalog/worker.go
mayuka-c: Since its an external API call, never know if there is latency issue or any other. But I'm fine for this if its quick

mayuka-c (approved):

PR #1271, opened 2026-08-18
refactor: standardise eager schema fetching across both deploy flows

pranithraoibm:
- `useDeployOptions` extended to fetch all component provider schemas upfront on mount via Promise.allSettled; deleted `useProviderParams` and `useServiceParams` which fetched lazily on provider selection
- useDeployOptions now returns isLoading and error as plain values instead of dispatching them into the reducer, matching how useServiceDeployOptions already works
- useServiceDeployOptions upgraded from Promise.all to Promise.allSettled so a single schema failure does not block the rest
- StepOne (DA) adds InlineNotification on failed schema loads listing affected component types; gates the Next button via onSchemaError callback (Same will be added to services in the next PR)*

image (1)


ui/catalog/src/components/DeployFlow/DigitalAssistant/steps/StepOne.tsx
```diff
@@ -1,81 +1,81 @@
...
+        const cached = providerParams[`${component.type}:${provider.id}`];
+        if (cached) result[component.type][provider.id] = cached.data;
+      });
     });
     return result;
-  }, [paramsByType, errorsByType]);
+  }, [deployOptions.global_components, providerParams]);
+
+  // Detect component types whose schemas failed to load (at least one provider missing).
+  const failedComponentTypes = useMemo(() => {
```
ryanmarc, line 48: This treats "not in the store" as "failed", but that is also exactly what the store looks like while the fetch is still running, so I think this fires a false error on every cold open.

Worth noting you already have the right pattern on the Services side in #1279, where `ServicesStepOne` keys off `componentModelsError` rather than off absence. Could this do the same, or track the in-flight keys so "loading" and "failed" are separable?

ui/catalog/src/components/DeployFlow/DigitalAssistant/steps/StepOne.tsx
```diff
@@ -1,81 +1,81 @@
...
+    return deployOptions.global_components
+      .filter((component) =>
+        component.providers.some(
```
ryanmarc: Separate from the loading-vs-failed thing above: `.some()` here means one bad provider fails the whole component type. If llm offers watsonx, vllm and openai, the user has watsonx selected, and only openai's schema 500s, this still lists LLM as failed and disables Next.

That is a bigger deal now than before this PR, because the fetch is eager across every provider rather than lazy on the selected one, so we are touching more endpoints and any of them can take the flow down.

ui/catalog/src/components/DeployFlow/DigitalAssistant/hooks/useDeployOptions.ts
```diff
@@ -1,57 +1,140 @@
...
+    // Collect service IDs that need their service-level schema fetched
+    const serviceIds = deployOptions.services.map((s) => s.id);
+    const serviceIdsToFetch = serviceIds.filter((serviceId) => {
+      const cached = getServiceParams(serviceId);
+      return !cached || isServiceParamsStale(serviceId);
+    });
+
+    if (pairsToFetch.length === 0 && serviceIdsToFetch.length === 0) return;
+
+    Promise.allSettled([
```
ryanmarc: The Promise.allSettled result is not assigned, so every rejection here is dropped on the floor. The three hooks this replaces all surfaced failures somehow (useBatchProviderParams returned an errors map plus console.warn, useMultiTypeProviderParams returned errorsByType, useProviderParams returned error), and the only thing replacing that is the absence check in StepOne.

That check does not cover service-level schemas, and ServiceConfigCard.tsx changed at the same time from useServiceParams(serviceId) (which fetched on demand) to a plain store read. So if a serviceParams fetch rejects, the card renders with serviceFields empty: no fields, no error, no retry, nothing in the console, and no lazy refetch to save it. From the user's side the card just looks like it has fewer inputs than it should.

ryanmarc (changes requested):

ui/catalog/src/components/DeployFlow/DigitalAssistant/steps/StepOne.tsx
pranithraoibm: Done, added providerParamsError map to do the same

ui/catalog/src/components/DeployFlow/DigitalAssistant/steps/StepOne.tsx
pranithraoibm: Scoped to the selected provider only. An error in an unselected provider no longer blocks Next. If you switch to a failed provider, Next is disabled; switching back to a working provider re-enables it.

ui/catalog/src/components/DeployFlow/DigitalAssistant/hooks/useDeployOptions.ts
pranithraoibm: Added .catch() to service params fetch, failures now write to serviceParamsError in the store instead of being silently dropped. ServiceConfigCard reads serviceParamsError[serviceId] and shows an error notification if the service schema failed, so the user sees an explicit message instead of silently missing fields. Retry on reopen is also now explicit, errored IDs are included in the refetch filter the same way provider params are.

pranithraoibm: Also added changes to disable the deploy button in step two if any fetch fails to both Digital assistant and services step two.

image

ui/catalog/src/components/DeployFlow/DigitalAssistant/DigitalAssistantDeployFlow.tsx
```diff
@@ -216,16 +208,20 @@ export const DeployFlow = ({
       currentStep={state.currentStep}
       isLastStep={isLastStep}
       isDeploying={state.isDeploying}
-      isPrimaryDisabled={state.isLoading || (isLastStep && state.isEditing)}
+      isPrimaryDisabled={
+        isLoading ||
+        (!isLastStep && (isProviderParamsLoading || hasStep1SchemaError)) ||
```
ryanmarc: The error-map rework is great and it fixes the false banner properly, thank you. One thing came in with it that I think is worth a follow-up.

`isProviderParamsLoading` is `inflightCount > 0`, and the in-flight set is seeded from the full pair list in `useDeployOptions.ts`, which is every provider of every global component plus every provider of every service component. So on a cold cache, step one's Next is disabled until the llm and reranker schemas load too, and those are only read in StepTwo. On the current rag catalog that is 7 provider schemas gating a step that needs 2 of them.

`isProviderParamsLoading` only reaches `isPrimaryDisabled`, not the shell's `isLoading`, so the user sees a fully populated step one with a greyed-out Next, no spinner, no text.

`ServicesDeployFlow.tsx` already does this the way I would suggest: it folds `isStep1ComponentsLoading` into `shellIsLoading` so the spinner shows. Could DA do the same, and narrow the in-flight set to the global-component providers step one actually reads?

ryanmarc (changes requested):

ui/catalog/src/components/DeployFlow/DigitalAssistant/DigitalAssistantDeployFlow.tsx
pranithraoibm: Done

ryanmarc (approved):

PR #1274, opened 2026-08-18
AISERVICES-1623: Fix translate image & rag-dev legacy application 

manalilatkar:
- added github workflow to build and push translate service image
- added translate template to podTemplateExecutions of rag-dev metadata

### DIFF
--- a/.github/workflows/translate-service-image.yml
+++ b/.github/workflows/translate-service-image.yml
@@ -1,0 +1,32 @@
+name: Translate Service
+
+on:
+  push:
+    paths:
+      - 'services/translate/**'
+    branches:
+      - main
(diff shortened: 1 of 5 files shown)


dharaneeshvrd (commented):
Dont think we need to bump the ai-services version, since we would deal with legacy templates directly without catalog.

manalilatkar:
> Dont think we need to bump the ai-services version, since we would deal with legacy templates directly without catalog.

@dharaneeshvrd without it, a check was failing so I added. Should I remove?

dharaneeshvrd:
Ok let it be.

dharaneeshvrd (approved):

PR #1278, opened 2026-08-18
AISERVICES-1383 added async extraction background worker implementation

manalilatkar:
In this PR:
- async endpoint background task for extraction

### DIFF
--- a/services/extract/api/v1/jobs.py
+++ b/services/extract/api/v1/jobs.py
@@ -438,16 +438,311 @@ async def create_extract_job(
 
 
 async def _process_extract_job(job_id: str) -> None:
-    """
-    Background worker stub.
-
-    A full worker (tokenization, vLLM call, schema validation) will be added
-    in a follow-up iteration.  For now it acquires the job_limiter slot so
-    semaphore accounting is correct and immediately releases it.
+    """Background worker: tokenize → guard → extract → validate → persist.
     """
     from extract import state
+
     async with state.job_limiter:
-        logger.info(f"Background worker invoked for job {job_id} (stub — no processing yet)")
+        t_start = time.monotonic()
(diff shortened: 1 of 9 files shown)


services/extract/api/v1/jobs.py
```diff
@@ -438,16 +438,311 @@ async def create_extract_job(
...
+
+        llm_model_dict = get_llm_endpoint()
+        llm_endpoint: str = llm_model_dict.get("llm_endpoint", "")
+        llm_model: str = llm_model_dict.get("llm_model", "")
+        max_model_len: int = llm_model_dict.get("max_model_len")
+
+        # Fetch the job row so we know the schema_id and document name.
+        job_row = db_repo.get_job_by_id(job_id)
+        if job_row is None:
+            logger.error(f"Job {job_id} not found in DB at worker start; aborting.")
```
dharaneeshvrd, line 462: dont you need to update the failure on db as well?

dharaneeshvrd (commented):
I see cleanup_staging_directory getting called at multiple occasions, why not call once via finally?

Otherwise overall lgtm since its same as the previous implementation.
Let's ensure to reuse the functionalities via a common method when you implement the batch processing.

services/extract/api/v1/jobs.py
manalilatkar: @dharaneeshvrd not sure what you mean. If we end up not finding any row with the given job_id, what would we update in the db? Technically this scenario shouldn't happen but we have to handle errors in case db pod goes down or network failure occurs.

dharaneeshvrd: Yeah agree, never mind :)

dharaneeshvrd (approved):

PR #1279, opened 2026-08-18
refactor: extract shared StepOne component across both deploy flows

pranithraoibm:
- Added `SharedStepOne` to `Shared/steps/`which receives all data as props, renders name, version, and component dropdowns with model/provider toggle and error banner
- Replaced DigitalAssistant/steps/StepOne.tsx → `DAStepOne.tsx`, delegates rendering to SharedStepOne, retains DA-specific provider deduplication and default model seeding logic
- Replaced Services/steps/StepOne.tsx → `ServicesStepOne.tsx`, delegates rendering to SharedStepOne, moves default model seeding useEffect here from the root (was ServicesDeployFlow)
- Added `onComponentError` to BaseStepProps in Shared/types.ts; removed onSchemaError from DA's local StepProps
- Added `hasComponentError` state + gate to ServicesDeployFlow which was already present in DA, now both flows block Next on component load failure


pranithraoibm:
@ryanmarc FYI this PR has changes from PR7 and PR 8a as well.

ui/catalog/src/components/DeployFlow/DigitalAssistant/steps/DAStepOne.tsx
```diff
@@ -0,0 +1,162 @@
...
+      component.providers.forEach((provider) => {
+        const cached = providerParams[`${component.type}:${provider.id}`];
+        if (cached) result[component.type][provider.id] = cached.data;
+      });
+    });
+    return result;
+  }, [deployOptions.global_components, providerParams]);
+
+  // Component types whose schemas failed to load (at least one provider missing).
+  const failedComponentTypes = useMemo(() => {
```
ryanmarc, line 38: Same finding I left on #1271, carried up with the rename, so no need to answer it twice. Short version: !providerParams[key] is also what the store looks like while the eager fetch is in flight, so on a cold cache this renders the red banner and disables Next on first open until the schemas land. And .some() over all providers means one unrelated provider failing takes down the whole component type.

What I would add here is that ServicesStepOne.tsx in this same PR already does it the right way, keying off componentModelsError rather than off absence. Whatever you do to the DA side, that is the shape to copy.

ryanmarc (changes requested):

ui/catalog/src/components/DeployFlow/DigitalAssistant/steps/DAStepOne.tsx
pranithraoibm: Pulled the changes made in https://github.com/IBM/project-ai-services/pull/1271 and updated

ui/catalog/src/components/DeployFlow/DigitalAssistant/DigitalAssistantDeployFlow.tsx
```diff
@@ -216,16 +210,20 @@ export const DeployFlow = ({
       currentStep={state.currentStep}
       isLastStep={isLastStep}
       isDeploying={state.isDeploying}
-      isPrimaryDisabled={state.isLoading || (isLastStep && state.isEditing)}
+      isPrimaryDisabled={
+        isLoading ||
+        (!isLastStep && (isProviderParamsLoading || hasStep1SchemaError)) ||
```
ryanmarc: Same comment from #1271. This should be fixed in #1271 and that should be merged to main to update this before this PR is merged.

ui/catalog/src/components/DeployFlow/Services/ServicesDeployFlow.tsx
```diff
@@ -385,6 +312,7 @@ export const ServicesDeployFlow = ({
           deployOptions={deployOptions}
           selectedServiceId={state.selectedServiceId}
           showNameError={state.showStepOneNameError}
+          onComponentError={setHasSchemaError}
```
ryanmarc: 1. This wire does nothing on step one. hasSchemaError is only read under isLastStep && at :272. And ServicesStepOne's failedComponentTypes (:23-33) computes the same set as hasStep1ComponentsError (:125-131) from the same componentModelsError map with the same llm/reranker filter, so we now compute it twice and consult one of them.
This wire does nothing on step one. hasSchemaError is only read under isLastStep && at :272. And ServicesStepOne's failedComponentTypes (:23-33) computes the same set as hasStep1ComponentsError (:125-131) from the same componentModelsError map with the same llm/reranker filter, so we now compute it twice and consult one of them.

2. :330 hands the same setter to StepTwo under the name onSchemaError, where it is driven by llmModelsError. One boolean, two writers, two steps, last mount wins. Fail on step two, hit Back, and step one's effect clears the flag that was gating Deploy. Step two re-sets it when it remounts so I could not land a click inside that window, but I would rather not leave it as a race we are relying on losing.

DA already solved this in this same series: two separate flags, each scoped to its step (DigitalAssistantDeployFlow.tsx:210-223). Could Services get the same shape? A hasStep1SchemaError for this wire, and then either it or hasStep1ComponentsError becomes the single owner of the step one gate. Happy either way, just not both.

ui/catalog/src/components/DeployFlow/Services/steps/ServicesStepOne.tsx
```diff
@@ -0,0 +1,208 @@
...
+    );
+  }, [
+    deployOptions.components,
+    formData.services,
+    selectedServiceId,
+    getComponentModels,
+    providerSchemas,
+  ]);
+
+  // Set default model param for each component when its models arrive from the store.
```
ryanmarc: This effect iterates every deployOptions.components with no llm / reranker filter (unlike the components memo right above it), so it is the only thing that seeds params.model for the llm component, which lives on step two. In ServicesDeployFlow it ran on every step. Here it only runs while step one is mounted.

That matters because of the fetch ordering in useServiceDeployOptions.ts stage 2 (llm models) only starts after stage 1 (step one components) has settled, and stage 1 settling is exactly what drops isStep1ComponentsLoading and lets the user click Next. So the gap where llm models have not landed yet opens at the same moment Next becomes clickable.

If the user clicks Next in that gap, this effect unmounts and nothing else seeds the default. StepTwo.tsx has one useEffect and it only calls onSchemaError; the only writer of llm.params.model there is the Dropdown's own onChange. So the LLM dropdown comes up empty and the deploy payload goes out without params for llm, same shape as the thing we were looking at on #1271.

Two options that both look fine to me: keep the llm/reranker part of the seeding in ServicesDeployFlow where it can see every step, or filter this effect to step one components and give step two its own seeding effect. The second is probably tidier given where this series is heading.

ryanmarc (changes requested):

ui/catalog/src/components/DeployFlow/Services/ServicesDeployFlow.tsx
pranithraoibm: Added two separate flags same as DA

ui/catalog/src/components/DeployFlow/Services/steps/ServicesStepOne.tsx
pranithraoibm: Done, implemented the second option as suggested

pranithraoibm:
@ryanmarc along with the review changes I have also added changes to fix the error message display on failed schema fetches and re fetch of the same on close and reopen of the tearsheet. Happy path was tested but this I came across when testing the 1st review comment so thought of adding it here, PTAL.

https://github.com/user-attachments/assets/5aba2df0-f8c2-4db6-a7f0-7d2c17bfec54

ryanmarc (approved):

PR #1280, opened 2026-08-18
Implementation of Global Weighted Queue to process Digitize/Ingestion requests

manju956:
- Implements the design proposal https://github.com/IBM/project-ai-services/blob/main/docs/proposals/docling_conversion_queue.md
- Covers various scenarios like baseline, mixed normal and large files, Queue overflow, round robin fairness, HOL blocking etc.


services/digitize/api/v1/jobs.py
```diff
@@ -181,27 +184,24 @@ async def create_job(
...
+        #    The gate requires at least one free slot — it does NOT require all N
+        #    new tasks to fit.
         op_key = operation.value  # "ingestion" | "digitization"
-        if concurrency_manager.is_locked(op_key):
+        if op_key == "ingestion":
+            quota = settings.digitize.ingestion_queue_quota
+        else:
+            quota = settings.digitize.digitization_queue_quota
+
+        quota_ok, queued_for_op = db_manager.check_quota_atomic(op_key, quota)
```
sats-23, line 196: Will this be the entry point for a connector job as well? Are there plans to reject connector jobs as well with 429?

services/digitize/api/v1/jobs.py
```diff
@@ -12,82 +12,85 @@
...
+def _run_digitize(
     job_id: str,
     doc_id_dict: dict,
     output_format: models.OutputFormat,
     file_checksum_dict: Optional[dict] = None,  # filename -> md5 hex
 ) -> None:
-    """Run the digitization pipeline and release the semaphore slot."""
-    status_mgr = get_status_manager(job_id)
-    job_staging_path = settings.digitize.staging_dir / job_id
+    """
```
dharaneeshvrd, line 37: We need to use the quotes inline with the comment as per [ruff rules](https://github.com/IBM/project-ai-services/pull/1258) added for MCP work
Please be aware of this @sats-23 @manalilatkar @Niharika0306

services/digitize/api/v1/jobs.py
```diff
@@ -12,82 +12,85 @@
...
 
+    Plain (non-async) so Starlette's BackgroundTasks runs it in a thread-pool
+    worker rather than on the event loop.
+    """
+    job_staging_path = settings.digitize.staging_dir / job_id
     try:
-        logger.info(f"🚀 Digitization started for job: {job_id}")
-        await asyncio.to_thread(digitize, job_staging_path, job_id, doc_id_dict, output_format, file_checksum_dict)
-        logger.info(f"Digitization for job {job_id} completed successfully")
+        logger.info(f"🚀 Digitization pipeline started for job: {job_id}")
```
dharaneeshvrd: @manju956 Some of these log messages and earlier structures were heavily modified by bob even though when its not required. Can we please revert those and keep the earlier changes and allow only the changes which are absolutely needed?
I know some of these are from my commit, apologies for doing this.

services/digitize/api/v1/jobs.py
dharaneeshvrd: We have to design how the connector needs to be integrated into this queuing logic. I was thinking of introducing a new op type for connector jobs so that it would round robin between these op types. Let's discuss more on this separately.

services/digitize/utils/recovery.py
```diff
@@ -194,3 +201,83 @@ def recover_connector_sync_state() -> int:
         logger.debug("No stuck connector syncs found on startup")
 
     return len(affected_ids)
+
+
+def recover_conversion_tasks() -> int:
```
dharaneeshvrd, line 207: I think now we need to redesign the recovery a bit different since we would have track of what are the documents yet to be converted. Maybe we need to give them a try if necessary staged files are present and continue the pipeline.
Maybe for future, I ll create a JIRA story for this.

services/digitize/app.py
```diff
@@ -110,17 +112,16 @@ def _recover_zombie_jobs():
     except Exception as exc:
         logger.error(f"Error during zombie job recovery: {exc}", exc_info=True)
 
-
-def _shutdown():
```
dharaneeshvrd, line 114: No need to remove this method
Just extend this method to call cancel on the dispatcher

services/digitize/db/scripts/init_schema.sql
```diff
@@ -91,6 +91,41 @@ CREATE TABLE IF NOT EXISTS connector_sync_logs (
...
+    page_count      INTEGER,
+    is_large        BOOLEAN         NOT NULL DEFAULT FALSE,
+    -- lifecycle
+    status          VARCHAR(50)     NOT NULL,
+    result_path     TEXT,                                   -- written on completion
+    error           TEXT,
+    queued_at       TIMESTAMPTZ     NOT NULL DEFAULT NOW(),
+    started_at      TIMESTAMPTZ,
+    completed_at    TIMESTAMPTZ,
+    updated_at      TIMESTAMPTZ     NOT NULL DEFAULT NOW(),
```
dharaneeshvrd: dont think this field is needed since we would have only one operation, it won't be updated regularly to be that useful

dharaneeshvrd (commented):
@manju956 Need to rebase with main to review effectively.

services/digitize/api/v1/jobs.py
manju956: sure, @dharaneeshvrd I will revert the log messages.

manju956:
> @manju956 Need to rebase with main to review effectively.

rebased the PR with latest to help reviewers.

services/digitize/api/v1/jobs.py
```diff
@@ -159,6 +161,7 @@ async def create_job(
         description="Output format for digitization: 'json', 'md', or 'txt'",
     ),
     job_name: Optional[str] = Query(None, description="Optional human-readable name for the job"),
+    background_tasks: BackgroundTasks = BackgroundTasks(),
```
dharaneeshvrd: We got rid of this and moved to `asyncio.create_task` recently, can you revert this change?

services/digitize/api/v1/jobs.py
```diff
@@ -185,27 +188,24 @@ async def create_job(
...
+        #    The gate requires at least one free slot — it does NOT require all N
+        #    new tasks to fit.
         op_key = operation.value  # "ingestion" | "digitization"
-        if concurrency_manager.is_locked(op_key):
+        if op_key == "ingestion":
+            quota = settings.digitize.ingestion_queue_quota
+        else:
+            quota = settings.digitize.digitization_queue_quota
+
+        quota_ok, queued_for_op = db_manager.check_quota_atomic(op_key, quota)
```
manalilatkar, line 194: What happens when 2 concurrent requests read queued=0 and both stage files worth a full quota?

services/digitize/api/v1/jobs.py
```diff
@@ -287,39 +287,49 @@ async def create_job(
         filenames     = novel_filenames
         file_contents = novel_contents
 
-        # 6. Acquire semaphore slot.
-        await concurrency_manager.acquire(op_key)
+        # 6. Stage files to disk.
+        await dg_util.stage_upload_files(
```
manalilatkar, line 289: We are staging files before initialize_job_state and enqueue_conversion_tasks. If either of those fails we return a 500 with N files orphaned in the staging dir. We should add some try:except block with cleanup in the except.

services/digitize/api/v1/jobs.py
dharaneeshvrd: Context regarding the change https://github.com/IBM/project-ai-services/pull/1232#discussion_r3764210794

services/digitize/db/manager.py
```diff
@@ -1482,6 +1485,436 @@ def merge_document_metadata(doc_id: str, metadata: dict) -> bool:
...
+                    .order_by(ConversionTask.queued_at)
+                    .limit(1)
+                )
+                task = session.scalar(stmt)
+                if task:
+                    _ = (task.task_id, task.job_id, task.doc_id, task.operation,
+                         task.cached_file, task.output_format, task.page_count,
+                         task.is_large, task.status, task.result_path, task.error,
+                         task.queued_at, task.started_at, task.completed_at)
+                    session.expunge(task)
```
manalilatkar: These lines are getting duplicated multiple times. Can be extracted into a method.

services/digitize/app.py
manju956: added back

services/digitize/db/manager.py
```diff
@@ -1482,6 +1485,436 @@ def merge_document_metadata(doc_id: str, metadata: dict) -> bool:
...
+
+                if not candidates:
+                    return 0
+
+                candidate_ids = [row.task_id for row in candidates]
+                now = datetime.now(timezone.utc)
+                stmt = (
+                    update(ConversionTask)
+                    .where(ConversionTask.task_id.in_(candidate_ids))
+                    .values(status="queued", queued_at=now)
```
manalilatkar: The queued_count scalar statement, the select candidates, and update status to queued are three separate statements. Maybe we should have an advisory lock on the operation key to prevent interleaving.

services/digitize/api/v1/jobs.py
```diff
@@ -185,27 +185,24 @@ async def create_job(
...
+        #    new tasks to fit.
         op_key = operation.value  # "ingestion" | "digitization"
-        if concurrency_manager.is_locked(op_key):
+        if op_key == "ingestion":
+            quota = settings.digitize.ingestion_queue_quota
+        else:
+            quota = settings.digitize.digitization_queue_quota
+
+        quota_ok, queued_for_op = db_manager.check_quota_atomic(op_key, quota)
+        if not quota_ok:
```
dharaneeshvrd, line 195: let's add log message here

services/digitize/parsing/pdf.py
```diff
@@ -140,10 +140,9 @@ def load_pdf_pages(pdf_path):
         with pdfplumber.open(pdf_path) as pdf:
             for page in pdf.pages:
                 pdf_pages.append(page.extract_words(extra_attrs=["size", "fontname"]))
-    except Exception as e:
-        logger.warning(f"Failed to load PDF pages from {pdf_path}: {e}")
-        return []
```
dharaneeshvrd, line 145: Any idea why this is removed?

services/digitize/db/models.py
```diff
@@ -283,4 +284,86 @@ class ConnectorSyncLog(Base):
...
 
+
+class ConversionTask(Base):
+    """
+    Conversion task model representing a single Docling conversion work item.
+
+    Maps to the 'conversion_tasks' table in PostgreSQL.
+    Each task is linked to a job and drives the dispatcher-based conversion queue.
+    """
+    __tablename__ = "conversion_tasks"
```
manalilatkar, line 304: updated_at is missing which was present in the proposal if I remember correctly.

services/digitize/pipeline/digitize.py
```diff
@@ -2,104 +2,76 @@
...
-                "pages": page_count,
-                "completed_at": get_utc_timestamp(),
-                "timing_in_secs": {"digitizing": round(conversion_time, 2)},
-                "file_hash": file_hash,
-            })
-            status_mgr.update_job_progress(doc_id, DocStatus.COMPLETED, JobStatus.COMPLETED)
+    if task is None or task.status == "failed":
+        error = (task.error or "Conversion failed") if task else "Task row missing"
+        logger.error(f"Digitization task for job {job_id} failed: {error}")
+        # The dispatcher already updated job/doc status on failure; nothing more to do.
```
dharaneeshvrd: Dispatcher should not touch job and doc status
It should be updated from here.
Can we please move from there to here?

services/digitize/workers/conversion_dispatcher.py
```diff
@@ -0,0 +1,222 @@
...
+            )
+            logger.warning(f"Task {task.task_id}: cached file missing — marked failed")
+            return
+
+        # Mark task and associated job/doc as running / in_progress
+        db_manager.update_task_status(task.task_id, "running")
+        if task.job_id and task.doc_id:
+            status_mgr = get_status_manager(task.job_id)
+            status_mgr.update_doc_metadata(task.doc_id, {"status": DocStatus.IN_PROGRESS})
+            status_mgr.update_job_progress(task.doc_id, DocStatus.IN_PROGRESS, JobStatus.IN_PROGRESS)
```
dharaneeshvrd: let's not do this
let the respective owners update their status based on task status
this method and dispatcher in general should update only its task status

services/digitize/processing/orchestrator.py
```diff
@@ -559,173 +571,204 @@ def _run_batch(batch_paths, convert_worker, max_worker, doc_id_dict, indexing_ca
...
-                    batch_stats.pop(path, {})
+                    process_futures[p_future] = path
+
+                elif task.status == "failed":
+                    pending_task_ids.discard(task_id)
+                    converted_pdf_stats.pop(path, None)
                     if doc_id is not None:
-                        status_mgr.update_doc_metadata(doc_id, {"status": DocStatus.FAILED}, error=f"failed to convert document: {str(e)}")
-                        status_mgr.update_job_progress(doc_id, DocStatus.FAILED, JobStatus.IN_PROGRESS)
+                        status_mgr.update_doc_metadata(
```
dharaneeshvrd, line 612: see here also we are trying to update the doc and job status
let's avoid doing it in dispatcher logic

services/digitize/pipeline/digitize.py
```diff
@@ -2,104 +2,76 @@
...
+        return
 
-        # Collect metadata (page count may be 0 for DOCX)
-        page_count = get_document_page_count(str(file_path))
+    # Poll until the dispatcher marks the task terminal.
+    while task.status not in ("completed", "failed"):
+        time.sleep(settings.digitize.conversion_poll_interval)
+        task = db_manager.get_conversion_task_by_job_id(job_id)
+        if task is None:
+            logger.warning(f"Task for job {job_id} disappeared during polling")
```
manalilatkar: @manju956 what is the scenariob in which this could happen? i.e task disappearing during polling. or is it just standard error handling?

services/digitize/pipeline/digitize.py
```diff
@@ -2,104 +2,76 @@
...
+        doc_id = next(iter(doc_id_dict.values()), "")
+        if doc_id:
+            status_mgr.update_doc_metadata(doc_id, {"status": DocStatus.FAILED}, error=error)
+        status_mgr.update_job_progress("", DocStatus.FAILED, JobStatus.FAILED, error=error)
+        return
 
-        # Collect metadata (page count may be 0 for DOCX)
-        page_count = get_document_page_count(str(file_path))
+    # Poll until the dispatcher marks the task terminal.
+    while task.status not in ("completed", "failed"):
```
manalilatkar: May be we should add a max_poll condition or timeout here in case the dispatcher gets stuck for some reasona nd never completes or fails the task.

services/digitize/workers/conversion_dispatcher.py
```diff
@@ -0,0 +1,222 @@
...
+logger = get_logger("conversion_dispatcher")
+
+# Round-robin state — alternates each tick when a task was successfully claimed.
+_rr_turn: str = "ingestion"
+
+# One shared process pool for all conversions.
+# max_workers == semaphore capacity so a worker process is always immediately
+# available when the semaphore grants a slot — no internal queuing in the pool.
+_process_pool = ProcessPoolExecutor(
+    max_workers=conversion_semaphore.capacity  # default 4
```
manalilatkar: This will spawn worker processes as soon as the module is imported, including during test collection, CLI tools, or any code that transitively imports this module. We should move the pool creation into dispatch_loop() (or a lazy initializer) so processes are only spawned when the dispatcher actually starts.

services/digitize/workers/conversion_dispatcher.py
```diff
@@ -0,0 +1,222 @@
...
+
+    Started once in app.py's lifespan and cancelled on shutdown.
+    """
+    global _rr_turn
+    logger.info("Conversion dispatcher loop started")
+
+    while True:
+        try:
+            available = conversion_semaphore.available
+            if available > 0:
```
manalilatkar: When available == 0, the dispatcher silently sleeps. Adding a periodic logger.debug (e.g., every 10th skip) would help us diagnose capacity starvation without flooding logs.

services/digitize/utils/jobs.py
```diff
@@ -178,6 +180,60 @@ def initialize_job_state(
...
+    slots_free = max(0, quota - queued_for_op)
+
+    for idx, filename in enumerate(filenames):
+        task_id = generate_uuid()
+        doc_id = doc_id_dict[filename]
+        file_path = staging_dir / filename
+        page_count = await asyncio.to_thread(get_document_page_count, str(file_path))
+        is_large = page_count >= settings.digitize.heavy_doc_page_threshold
+        task_status = "queued" if idx < slots_free else "pending"
+        db_manager.create_conversion_task(
```
manalilatkar: In this line, each file triggers a separate db_manager.create_conversion_task call, each in its own transaction. For a 30-file ingestion job, that's 30 DB round-trips. Consider batching inserts into a single transaction using session.add_all() in a new create_conversion_tasks_batch method. This also makes the set of inserts atomic — either all succeed or none do.

services/digitize/utils/recovery.py
```diff
@@ -194,3 +201,83 @@ def recover_connector_sync_state() -> int:
...
+    - ``pending``  → keep if cached file present; else ``failed``
+
+    Returns:
+        Number of tasks that were recovered (status changed).
+    """
+    from digitize.db.manager import db_manager
+
+    recovered = 0
+
+    try:
```
manalilatkar, line 229: This outer try-except block means a failure recovering running tasks (step 1) silently skips recovery of queued and pending tasks (steps 2–3). Consider individual try/except blocks per step so a failure in one category doesn't prevent recovery of others

services/digitize/tests/conftest.py
```diff
@@ -321,4 +321,61 @@ def mock_db_job():
...
+    task.output_format = "json"
+    task.page_count = 10
+    task.is_large = False
+    task.status = "queued"
+    task.result_path = None
+    task.error = None
+    task.queued_at = datetime(2024, 1, 1, 0, 0, 0, tzinfo=timezone.utc)
+    task.started_at = None
+    task.completed_at = None
+    task.updated_at = datetime(2024, 1, 1, 0, 0, 0, tzinfo=timezone.utc)
```
manalilatkar: updated_at is not defined on the ConversionTask ORM model or in init_schema.sql. Either add the column to the model (the proposal includes it) or remove this line from the mock to avoid tests passing against a field that doesn't exist in production.

manalilatkar (commented):
rest LGTM

services/digitize/tests/conftest.py
manju956: `updated_at` field is removed from the schema, somehow traces left out in the test files. i will do the cleanup.

services/digitize/workers/conversion_dispatcher.py
manju956: makes sense, will add the debug log

services/digitize/db/models.py
manju956: its removed as per suggestion by @dharaneeshvrd above.

services/digitize/pipeline/digitize.py
manju956: could be when we disconnect the connector, we intend to remove the task entries corresponding to the connector as well from the coversion_tasks table.

services/digitize/api/v1/jobs.py
manju956: `check_quota_atomic ` uses advisory lock to prevent this case I guess.

services/digitize/parsing/pdf.py
manju956: accidentally removed, will revert it back.

services/digitize/pipeline/digitize.py
manju956: sure, will move around the logic

services/digitize/workers/conversion_dispatcher.py
manju956: makes sense, will take care

services/digitize/api/v1/jobs.py
manju956: makes sense

services/digitize/db/manager.py
manju956: yep, would be neat

services/digitize/utils/recovery.py
manju956: there is todo for redesigning recovery and @dharaneeshvrd will be creating separate jira for this.
`I think now we need to redesign the recovery a bit different since we would have track of what are the documents yet to be converted. Maybe we need to give them a try if necessary staged files are present and continue the pipeline.
Maybe for future, I ll create a JIRA story for this.`

services/digitize/utils/jobs.py
manju956: addressed

services/digitize/workers/conversion_dispatcher.py
manju956: done

services/digitize/workers/conversion_dispatcher.py
manju956: addressed

services/digitize/pipeline/digitize.py
manju956: done

services/digitize/db/manager.py
manju956: done

services/digitize/api/v1/jobs.py
manju956: done

services/digitize/processing/orchestrator.py
dharaneeshvrd: converter_executor logic needs to be removed

dharaneeshvrd (commented):
Need to cleanup old services/digitize/workers/concurrency.py

services/digitize/pipeline/digitize.py
```diff
@@ -2,104 +2,122 @@
...
-        # Mark COMPLETED
-        if status_mgr:
-            logger.debug(f"Conversion Done: updating doc & job metadata for document: {doc_id}")
-            file_hash = (
-                file_checksum_dict.get(filename)
-                if file_checksum_dict
-                else generate_file_checksum(file_path.read_bytes())
+    # Poll until the dispatcher advances the task past queued state.
+    while task.status not in ("running", "completed", "failed"):
+        if time.monotonic() >= deadline:
```
dharaneeshvrd: this logic is good but 1h seems to be on the shorter side, let's bump to 3hrs

dharaneeshvrd: can we use similar logic in ingestion pipeline as well to time out the conversion tasks?

services/digitize/pipeline/digitize.py
```diff
@@ -2,104 +2,122 @@
...
+from digitize.db.manager import db_manager
+from digitize.models import JobStatus, DocStatus
+from digitize.settings import settings
 from digitize.utils.db import get_status_manager
-from concurrent.futures import ProcessPoolExecutor
 
 logger = get_logger("digitize")
 
+
 def digitize(
```
dharaneeshvrd, line 59: only job_id and doc_id_dict are used with new logic

dharaneeshvrd, line 59: overall this method can be simplified by having an internal method to reuse the deadline exceed logic

services/digitize/processing/orchestrator.py
```diff
@@ -559,173 +571,204 @@ def _run_batch(batch_paths, convert_worker, max_worker, doc_id_dict, indexing_ca
...
+                        )
+                        status_mgr.update_job_progress(
+                            doc_id, DocStatus.FAILED, JobStatus.IN_PROGRESS
+                        )
 
-            # C. Handle Processing -> Submit Chunking
-            for fut in as_completed(process_futures):
-                path = process_futures[fut]
+            # --- B. Drain completed process futures → submit chunking ---
+            for fut in list(process_futures):
```
dharaneeshvrd, line 621: why did we move away from as_completed logic?

manju956: `as_completed` is a blocking iterator. It would stall the loop at section B until every process_future finished before it could proceed to drain chunk_futures. However with the fut.done() pattern does a non-blocking drain on each tick: pick up whatever is already finished, submit it downstream immediately, and continue the loop.

services/digitize/pipeline/digitize.py
manju956: done

services/digitize/processing/orchestrator.py
dharaneeshvrd: yes agree, but why are we making as part of this PR?
I had this change planned for cancel PR ;)

services/digitize/processing/orchestrator.py
```diff
@@ -482,250 +477,296 @@ def clean_intermediate_files(doc_id, out_path):
...
+                    continue
+                if task.status not in ("completed", "failed"):
+                    # Start the deadline clock the first time we see this task.
+                    if task_id not in task_deadlines:
+                        task_deadlines[task_id] = time.monotonic() + timeout_s
+                    if time.monotonic() >= task_deadlines[task_id]:
+                        path = task_id_to_path[task_id]
+                        doc_id = doc_id_dict.get(Path(path).name)
+                        error = (
+                            f"Conversion task {task_id} did not complete within "
```
dharaneeshvrd: let's print doc path to easily associate the doc name with the task id

services/digitize/utils/recovery.py
dharaneeshvrd: Till we implement the resuming the job for processing the queued and pending tasks, lets fail them. 
There is no point in letting them stuck in these states, since underlying ingestion and digitization would have been marked as failed.
The redesign I m talking about is a major story involves all of these changes, let's see in future.

services/digitize/db/models.py
```diff
@@ -282,4 +283,86 @@ class ConnectorSyncLog(Base):
...
+    output_format: Mapped[str] = mapped_column(String(10), nullable=False)
+
+    # Page count; may be 0 for DOCX
+    page_count: Mapped[int | None] = mapped_column(Integer, nullable=True)
+
+    # True when page_count >= heavy_doc_page_threshold
+    is_large: Mapped[bool] = mapped_column(Boolean, nullable=False, default=False)
+
+    # Lifecycle status
+    status: Mapped[str] = mapped_column(String(50), nullable=False)
```
dharaneeshvrd, line 335: can we introduce an enum for possible status values of conversion tasks and refer it every where?

dharaneeshvrd (commented):
overall lgtm
let's address these minor comments

services/digitize/processing/orchestrator.py
manju956: makes sense

services/digitize/db/models.py
manju956: sure, can be done.

services/digitize/processing/orchestrator.py
manju956: shall i revert these changes from here?

services/digitize/db/models.py
manju956: done

services/digitize/processing/orchestrator.py
manju956: done

services/digitize/utils/recovery.py
```diff
@@ -194,3 +201,83 @@ def recover_connector_sync_state() -> int:
...
+        # 1. running → failed
+        running_tasks = db_manager.get_conversion_tasks(status="running")
```
dharaneeshvrd: why do we have three steps?
can it not be done in a single loop?

services/digitize/utils/recovery.py
```diff
@@ -194,3 +201,83 @@ def recover_connector_sync_state() -> int:
...
+            db_manager.update_task_status(
+                task.task_id, "failed",
+                error="Service restarted during conversion",
+            )
+            logger.warning(f"Recovery: task {task.task_id} running→failed")
+            recovered += 1
+
+        # 2. queued — verify cached file
+        queued_tasks = db_manager.get_conversion_tasks(status="queued")
+        for task in queued_tasks:
```
dharaneeshvrd: as per last comment let's merge the three steps
dont we need to clean up the staged or processed things?

services/digitize/utils/recovery.py
manju956: Yes, I need to alter the `get_conversion_tasks` method to accept various status filters to merge these steps into one loop. Will be doing it.

services/digitize/utils/recovery.py
manju956: I guess `recover_zombie_jobs ()` already cleans up staging directory before calling `recover_conversion_tasks `, so we are good.

services/digitize/utils/recovery.py
manju956: done

services/digitize/db/scripts/init_db.py
```diff
@@ -14,7 +14,7 @@
 SCHEMA_FILE = SCRIPT_DIR / 'init_schema.sql'
 
 # Expected tables for digitize service
-EXPECTED_TABLES = {'jobs', 'documents'}
+EXPECTED_TABLES = {'jobs', 'documents', 'conversion_tasks'}
```
sats-23: Can we add document_checksum, connectors, connector_document_checksum & connector_sync_logs as well?

services/digitize/tests/conftest.py
manju956: removed.

services/digitize/db/scripts/init_db.py
manju956: done

services/digitize/utils/recovery.py
```diff
@@ -219,58 +219,41 @@ def recover_conversion_tasks() -> int:
...
+                    error="Service restarted during conversion",
                 )
+                logger.warning(f"Recovery: task {task.task_id} running→failed")
                 recovered += 1
             else:
-                logger.info(
-                    f"Recovery: task {task.task_id} stays pending (file intact)"
-                )
+                # queued / pending — keep only if the cached file is still present
+                if not Path(task.cached_file).exists():
```
dharaneeshvrd: we need to clean even if it has the cache file, since it cannot be used anyway right?

manju956: we need to retain the cached_file(input doc files in case of Queued, Pending tasks) exists to proceess during the next restart rt?

dharaneeshvrd: But the underlying jobs won't be restarted, that's need to be taken care first.

services/digitize/db/manager.py
```diff
@@ -1495,6 +1498,500 @@ def merge_document_metadata(doc_id: str, metadata: dict) -> bool:
...
+                    cached_file=t["cached_file"],
+                    output_format=t["output_format"],
+                    page_count=t["page_count"],
+                    is_large=t["is_large"],
+                    status=t["status"],
+                    queued_at=now,
+                )
+                for t in tasks
+            ])
+            session.flush()
```
sats-23: We can drop this. No exception handling is happening within method, so it will get committed either way at end of method

services/digitize/utils/recovery.py
manju956: yep, right. will take care of this

services/digitize/db/manager.py
manju956: done

services/digitize/utils/recovery.py
manju956: addressed

services/digitize/pipeline/digitize.py
```diff
@@ -2,104 +2,123 @@
...
 
     try:
-        # Mark document/job as IN_PROGRESS
-        if status_mgr:
-            logger.debug(f"Submitted for conversion: updating job & doc metadata to IN_PROGRESS for document: {doc_id}")
+        # Phase 1: wait until the dispatcher picks up the task (queued → running).
+        task = _poll_until(job_id, {"running", "completed", "failed"}, deadline, timeout_s, "start")
+
+        # Mark IN_PROGRESS as soon as the dispatcher starts running.
+        if task is not None and task.status == "running" and doc_id:
```
dharaneeshvrd: here also please use enum values

services/digitize/processing/orchestrator.py
```diff
@@ -482,250 +477,297 @@ def clean_intermediate_files(doc_id, out_path):
...
+
+            # --- A. React to completed/failed conversion tasks ---
+            timeout_s = settings.digitize.conversion_timeout_s
+            for task_id in list(pending_task_ids):
+                task = db_manager.get_conversion_task(task_id)
+                if task is None:
+                    pending_task_ids.discard(task_id)
+                    task_deadlines.pop(task_id, None)
+                    continue
+                if task.status not in ("completed", "failed"):
```
dharaneeshvrd: let's use enum here

dharaneeshvrd (commented):
let's make sure all task status are referred using enum vals

services/digitize/processing/orchestrator.py
dharaneeshvrd: no its ok.

services/digitize/processing/orchestrator.py
manju956: done

services/digitize/pipeline/digitize.py
manju956: done

manju956:
> let's make sure all task status are referred using enum vals

done

dharaneeshvrd (approved):

PR #1281, opened 2026-08-18
docs: add proposal for image versioning

ryanmarc:


docs/proposals/image-versioning-automation.md
```diff
@@ -0,0 +1,133 @@
...
+
+To keep the sequences from ever meeting, the first bump on a release branch appends the branch name and starts a counter (`v0.0.26` → `v0.0.26-release-0.2.1`), and later bumps on that branch just increment the counter (`.2`, `.3`, ...). Main never produces suffixed tags, so collisions are impossible, and the tag becomes self-describing: `v0.0.26-release-0.2.3` is the third fix on `release-0.2`, which forked while chatbot-service was at `v0.0.26`. Signing and promotion don't care about the format, though we should double-check nothing in the Jenkins pipeline assumes tags look exactly like `vX.Y.Z` before the first release-branch bump.
+
+**The bot needs permission to push to main.** The plan is a machine-user account with a fine-grained PAT (contents: write, this repo only) added to the branch protection bypass list.
+
+## How we'd roll it out
+
+Four steps, each its own PR, each safe on its own:
+
+1. **Close the existing gaps.** Add the missing translate and extract image build workflows and scanner entries, copying the pattern the other services already use.
```
manalilatkar, line 121: We have already done this: [extract](https://github.com/IBM/project-ai-services/blob/main/.github/workflows/extract-service-image.yml) and [translate](https://github.com/IBM/project-ai-services/blob/main/.github/workflows/translate-service-image.yml)

docs/proposals/image-versioning-automation.md
```diff
@@ -0,0 +1,133 @@
...
+
+To keep the sequences from ever meeting, the first bump on a release branch appends the branch name and starts a counter (`v0.0.26` → `v0.0.26-release-0.2.1`), and later bumps on that branch just increment the counter (`.2`, `.3`, ...). Main never produces suffixed tags, so collisions are impossible, and the tag becomes self-describing: `v0.0.26-release-0.2.3` is the third fix on `release-0.2`, which forked while chatbot-service was at `v0.0.26`. Signing and promotion don't care about the format, though we should double-check nothing in the Jenkins pipeline assumes tags look exactly like `vX.Y.Z` before the first release-branch bump.
```
mkumatag, line 113: may be we can consider using `v0.0.27-x` versions for the release branches

mkumatag (changes requested):
this looks promising, have a minor comment otherwise LGTM from my side..

manju956 (approved):

docs/proposals/image-versioning-automation.md
```diff
@@ -0,0 +1,133 @@
...
+
+There are also a couple of quieter problems with the current setup:
+
+- The CI check that verifies values files match the Makefiles (`check_image_names.py`) works off a hardcoded list of file/key pairs. That list is out of date. Around nine values files with pinned image versions aren't checked at all (all of `catalog/openshift`, several of the single-service charts, the watsonx litellm files, and a few individual keys elsewhere). If one of those drifts to a stale version, nothing catches it.
+- The translate and extract services have no image build workflows in CI at all. When their code changes, no image gets built or published unless someone does it by hand. They're also missing from the image scanner matrix.
+
+## What we'd change
+
+The short version: **versions move out of PRs entirely.** You merge code, CI figures out what changed, builds and publishes the image with the next version number, and commits the bump itself. Nobody edits a version string by hand for normal day-to-day work.
```
mayuka-c, line 22: Dont we need to give permissions to github bot to be able to directly merge to main branch or is it safe for it to have permissions to only raise a PR and we merge it to be safe?

docs/proposals/image-versioning-automation.md
```diff
@@ -0,0 +1,133 @@
...
+images:
+  chatbot-service:
+    tag: v0.0.25
+    source: [services/chatbot, services/common]
+  chatbot-ui:
+    tag: v0.0.50
+    source: [ui/chatbot]
+  digitize-service:
+    tag: v0.0.43
+    source: [services/digitize, services/common]
```
mayuka-c, line 41: This looks good :)

docs/proposals/image-versioning-automation.md
ryanmarc: I'll remove this part from the proposal.

docs/proposals/image-versioning-automation.md
ryanmarc: I went with this format to make it descriptive at a glance, but sure, this could be changed if you prefer.

Adding the branch also ensures there is absolutely never any tag naming collision. Using this other format is likely ok, but it will not guarantee collision is impossible like it is when including the branch name.

docs/proposals/image-versioning-automation.md
ryanmarc: This is specifically mentioned near the end of the proposal.

The reason I decided on this is to make it completely automatic, not have a trail of additional PRs that have to be manually merged every time anything else is merged.

docs/proposals/image-versioning-automation.md
mkumatag: Our versions are not 100% perfect. However, that is a separate topic for discussion and, I believe, out of scope for this document. We can brainstorm it further and decide on the right approach separately. CC: @adarshagrawal38

dharaneeshvrd (commented):
Overall lgtm
Can we please design the bump workflow to also make changes in rag-dev template's values.yaml as well? since we are using those legacy templates to share the machine for development which is really easy for us to plug and play individual services which is not the case in catalog way of deployment. 
If we don't do this, then those templates will become unusable.

ryanmarc:
> Overall lgtm Can we please design the bump workflow to also make changes in rag-dev template's values.yaml as well? since we are using those legacy templates to share the machine for development which is really easy for us to plug and play individual services which is not the case in catalog way of deployment. If we don't do this, then those templates will become unusable.

With this proposal, every values.yaml file would update when the version bump workflow runs. Is that what you mean here, or are you referring to a specific change about that template file?

dharaneeshvrd:
Ok, if it updates every values.yaml, then we are good. I am looking at this [values.yaml](https://github.com/IBM/project-ai-services/blob/main/ai-services/assets/applications/rag-dev/podman/values.yaml) updated whenever there is a service version bump.

docs/proposals/image-versioning-automation.md
```diff
@@ -0,0 +1,133 @@
...
+  # ... and so on for all 15 of our images, plus third-party
+  # images (opensearch, vllm) which are listed but never auto-bumped
+```
+
+The `source` paths tell CI which image a changed file belongs to. The image lines in all 26 values.yaml files become generated output: a small sync tool reads `versions.yaml` and rewrites the tag on every matching `image:` line. Want to know what version of anything we're currently shipping? It's one file, and `git log` on that file is a complete release history.
```
adarshagrawal38, line 46: We will have sync tool to update tags in `values.yaml` file.

But what about the `TAG` value in the `Makefile`? are we not going to update them as part of sync tool.

PR #1284, opened 2026-08-19
Data Source Connectors: Propagate error and misc

sats-23:
1. Cleanup removed failed_files from init_schema
2. Removed last_sync_error from connectors table 
3. Async auth validation on every PUT/POST call on connectors which updates the connectors table 
--This error is never overwritten by a sync tick failure error message (highlighting root cause being wrong creds)
--Only correct creds via another PUT request can clear the previous auth failure message 
4. Updated proposal 
5. Increased UT coverage 
6. Used the secret as per catalog deployment 
7. Rebrand connectors response into simple ID and name

### DIFF
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -97,8 +97,9 @@ async def _probe_connector_credentials(
         if current_error == _CREDENTIAL_ERROR_MSG:
             db_ops.set_connector_error(connector_id, None)
     except Exception as exc:
-        logger.warning(
-            f"Credential probe failed for connector {connector_id!r}: {exc}"
+        logger.error(
+            f"Credential probe failed for connector {connector_id!r}: {exc}",
+            exc_info=True,
         )
         db_ops.set_connector_error(connector_id, _CREDENTIAL_ERROR_MSG)
         try:
@@ -149,7 +150,8 @@ async def create_connector(body: ConnectorCreateRequest):
             )
         except Exception as sched_exc:
             logger.error(
-                f"Scheduler registration failed for {connector_id!r}: {sched_exc}"
+                f"Scheduler registration failed for {connector_id!r}: {sched_exc}",
+                exc_info=True,
             )
             raise
 
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -36,6 +36,7 @@ from digitize.connectors.models import (
     ConnectorStatus,
     SyncLogStatus,
     SyncTriggerResponse,
+    ConnectorError,
 )
 from digitize.connectors.encryption import (
     encrypt_secrets,
@@ -54,7 +55,7 @@ logger = get_logger("connectors_router")
 
 _KEY_PATH = None  # resolved lazily via _get_key_path()
 
-_CREDENTIAL_ERROR_MSG = "Authentication failed: unable to connect with the provided credentials"
+_CREDENTIAL_ERROR_MSG = ConnectorError.CREDENTIAL_ERROR_MSG  # canonical value lives in connectors.models
 
 
 def _get_key_path() -> str:
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -145,7 +145,10 @@ async def create_connector(body: ConnectorCreateRequest):
                 f"Scheduler registration failed for {connector_id!r}: {sched_exc}",
                 exc_info=True,
             )
-            raise
+            raise RuntimeError(
+                f"Failed to register scheduler job for connector {connector_id!r} "
+                f"during connector creation: {sched_exc}"
+            ) from sched_exc
 
         db_ops.insert_connector(
             connector_id=connector_id,
@@ -186,7 +189,10 @@ async def create_connector(body: ConnectorCreateRequest):
         # encryption key not found
         logger.error(f"Encryption key error: {exc}")
         APIError.raise_error(ErrorCode.INTERNAL_SERVER_ERROR, str(exc))
-    except HTTPException:
+    except HTTPException as exc:
+        logger.error(
+            f"Failed to create connector {connector_id!r}: HTTP {exc.status_code} - {exc.detail}"
+        )
         raise
     except Exception as exc:
         logger.error(f"Unexpected error creating connector: {exc}", exc_info=True)
@@ -283,7 +289,10 @@ async def update_connector(connector_id: str, body: ConnectorUpdateRequest):
     except RuntimeError as exc:
         logger.error(f"Encryption key error: {exc}")
         APIError.raise_error(ErrorCode.INTERNAL_SERVER_ERROR, str(exc))
-    except HTTPException:
+    except HTTPException as exc:
+        logger.error(
+            f"Failed to update connector {connector_id!r}: HTTP {exc.status_code} - {exc.detail}"
+        )
         raise
     except Exception as exc:
         logger.error(f"Unexpected error updating connector {connector_id}: {exc}", exc_info=True)
@@ -337,7 +346,10 @@ async def delete_connector(connector_id: str):
 
         return Response(status_code=204)
 
-    except HTTPException:
+    except HTTPException as exc:
+        logger.error(
+            f"Failed to delete connector {connector_id!r}: HTTP {exc.status_code} - {exc.detail}"
+        )
         raise
     except Exception as exc:
         logger.error(f"Unexpected error deleting connector {connector_id}: {exc}", exc_info=True)
@@ -508,7 +520,10 @@ async def list_connectors():
             )
             for c in connectors
         ]
-    except HTTPException:
+    except HTTPException as exc:
+        logger.error(
+            f"Failed to list connectors: HTTP {exc.status_code} - {exc.detail}"
+        )
         raise
     except Exception as exc:
         logger.error(f"Unexpected error listing connectors: {exc}", exc_info=True)
@@ -559,7 +574,10 @@ async def get_connector(connector_id: str):
             connection_details=strip_secrets(connector.type, connector.connection_details or {}),
             total_files=connector.total_files,
         )
-    except HTTPException:
+    except HTTPException as exc:
+        logger.error(
+            f"Failed to get connector {connector_id!r}: HTTP {exc.status_code} - {exc.detail}"
+        )
         raise
     except Exception as exc:
         logger.error(f"Unexpected error fetching connector {connector_id}: {exc}", exc_info=True)
@@ -683,7 +701,10 @@ async def trigger_sync(connector_id: str):
         APIError.raise_error(ErrorCode.RESOURCE_NOT_FOUND, str(exc))
     except SyncLocked as exc:
         APIError.raise_error(ErrorCode.RESOURCE_LOCKED, str(exc))
-    except HTTPException:
+    except HTTPException as exc:
+        logger.error(
+            f"Failed to trigger sync for connector {connector_id!r}: HTTP {exc.status_code} - {exc.detail}"
+        )
         raise
     except Exception as exc:
         logger.error(f"Unexpected error triggering sync for {connector_id}: {exc}", exc_info=True)
@@ -749,7 +770,11 @@ async def cancel_sync(connector_id: str, sync_seq: int):
         logger.info(f"Cancel-sync signal sent for connector {connector_id!r} (seq={sync_seq})")
         return Response(status_code=204)
 
-    except HTTPException:
+    except HTTPException as exc:
+        logger.error(
+            f"Failed to cancel sync for connector {connector_id!r} (seq={sync_seq}): "
+            f"HTTP {exc.status_code} - {exc.detail}"
+        )
         raise
     except Exception as exc:
         logger.error(f"Unexpected error cancelling sync for {connector_id}: {exc}", exc_info=True)
@@ -803,7 +828,11 @@ async def get_sync_history(
 
         return SyncLogResponse(total=total, limit=limit, offset=offset, items=items)
 
-    except HTTPException:
+    except HTTPException as exc:
+        logger.error(
+            f"Failed to get sync history for connector {connector_id!r}: "
+            f"HTTP {exc.status_code} - {exc.detail}"
+        )
         raise
     except Exception as exc:
         logger.error(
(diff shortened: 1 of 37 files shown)


services/digitize/api/v1/connectors.py
```diff
@@ -53,12 +55,58 @@
...
 
+_CREDENTIAL_ERROR_MSG = ConnectorError.CREDENTIAL_ERROR_MSG  # canonical value lives in connectors.models
+
 
 def _get_key_path() -> str:
     """Return the encryption key path from settings."""
     return settings.digitize.connector.encryption_key_path
 
 
+async def _probe_connector_credentials(
```
sats-23, line 59: This is a makeshift solution to async update the DB (for failed creds). 
This still does not provide direct validation to the end-user when a PUT/POST on connector is done

dharaneeshvrd: I am not fully convinced this is really needed, because in case of POST the immediate sync operation will do this and report this failure anyway.
And for PUT this would be done in next sync which would be 5 mins later.
But somehow after seeing this, I m thinking why not trigger the sync immediately after PUT as well like POST, if someone updates the connector, I feel they would like the sync to happen immediately wdyt?
If we do that, this probing is not needed.

services/digitize/api/v1/connectors.py
```diff
@@ -101,26 +142,36 @@ async def create_connector(body: ConnectorCreateRequest):
             )
         except Exception as sched_exc:
             logger.error(
-                f"Scheduler registration failed for {connector_id!r}: {sched_exc}"
+                f"Scheduler registration failed for {connector_id!r}: {sched_exc}",
+                exc_info=True,
             )
             raise
```
dharaneeshvrd: These kind of empty raise statements needs to be updated with a prefix message saying on what operation and situation, this specific exception is raised. Please take care of all the plain raise statements wrt to the connector changes added.

services/digitize/connectors/models.py
```diff
@@ -66,34 +66,51 @@ class SyncLogStatus(str, Enum):
...
+# Error message constants
+# ---------------------------------------------------------------------------
+
+class ConnectorError(str, Enum):
+    """String enum of well-known error messages stored on the Connector row.
+
+    Inherits from str so values can be compared directly against DB strings.
+    """
+
+    CREDENTIAL_ERROR_MSG = "Authentication failed: unable to connect with the provided credentials"
```
dharaneeshvrd, line 79: this should be used on auth issues faced during sync operation as well

services/digitize/db/scripts/init_schema.sql
```diff
@@ -82,7 +81,6 @@ CREATE TABLE IF NOT EXISTS connector_sync_logs (
     total_files      INTEGER     NOT NULL DEFAULT 0,
     new_files        INTEGER     NOT NULL DEFAULT 0,
     removed_files    INTEGER     NOT NULL DEFAULT 0,
-    failed_files     INTEGER     NOT NULL DEFAULT 0,
```
dharaneeshvrd, line 85: can you help me remember why are we removing failed_files?

services/digitize/db/manager.py
```diff
@@ -820,7 +824,7 @@ def update_connector(
                     values["allowed_extensions"] = allowed_extensions
                 if total_files is not None:
                     values["total_files"] = total_files
-                if error is not None:
+                if error is not _UNSET:
```
dharaneeshvrd, line 816: can you not simply do this by removing this if?
what ever is passed for error arg needs to be set as is.
So when an error exists currently, if its cleared in next run, the error passed here would be none, which can be set as is. wdyt?

services/digitize/api/v1/connectors.py
```diff
@@ -129,7 +180,7 @@ async def create_connector(body: ConnectorCreateRequest):
         # id or name already exists
```
dharaneeshvrd, line 120: I think we also discussed to log the exceptions before raising, can you please take care in all the occurrences?

services/digitize/utils/db.py
```diff
@@ -1412,7 +1412,9 @@ def finalize_sync_log_and_update_connector(
     )
     if not found:
         return False
-    db_manager.update_connector_after_sync(connector_id, status=status, last_sync_at=now)
+    db_manager.update_connector_after_sync(
```
dharaneeshvrd, line 1409: It's better to rephrase the error message of a last sync when we update it on connector to depict that it s from the last sync.

services/digitize/connectors/sync_tick.py
dharaneeshvrd: This is going to be a user facing error message, can we please rephrase accordingly?
Something like: One or more documents failed to sync. See more details on digitize jobs `<f"Connector-{connector_name}-{sync_seq}-*>`

services/digitize/api/v1/connectors.py
dharaneeshvrd: This applies to other endpoint handlers as well. 
Also please ensure to log them before raising

dharaneeshvrd (commented):
It would be great, if you can produce possible error cases and how the connector and sync log's error fields looks like.

services/digitize/api/v1/connectors.py
dharaneeshvrd: @manalilatkar @Niharika0306 
I think similar behaviour need to be taken care in extract and translate service as well.
Please see whether you can address them in your upcoming PRs

services/digitize/connectors/models.py
sats-23: Done, added

services/digitize/db/scripts/init_schema.sql
sats-23: We had already removed it because we needed more complex counting logic with failed due to what reason, stats related to that and how we update it for good UX. So we decided to remove for time being. 

I had removed all its references throughout, just the schema was leftover

services/digitize/connectors/sync_tick.py

services/digitize/utils/db.py
sats-23: Done, added

services/digitize/api/v1/connectors.py
sats-23: Done

services/digitize/db/manager.py
sats-23: This is not really recommended coz in the other instances we are also using update_connector for updating total file count and on PUT request to update connection details (both cases where we do not pass any error). 
In such cases, we are using this UNSET to distinguish between a None error (means reset error status) vs UNSET (hold onto previous error, keep it unchanged)

In db_ops.set_connector_error, we are passing None in case we want to clear the error (after creds are fine). We went up losing this distinction.

services/digitize/api/v1/connectors.py
sats-23: Fixed

services/digitize/api/v1/connectors.py
```diff
@@ -129,13 +183,16 @@ async def create_connector(body: ConnectorCreateRequest):
...
```
dharaneeshvrd, line 120: Add logger.error here

services/digitize/api/v1/connectors.py
```diff
@@ -129,13 +183,16 @@ async def create_connector(body: ConnectorCreateRequest):
...
     except RuntimeError as exc:
         # encryption key not found
         logger.error(f"Encryption key error: {exc}")
         APIError.raise_error(ErrorCode.INTERNAL_SERVER_ERROR, str(exc))
-    except HTTPException:
+    except HTTPException as exc:
+        logger.error(
+            f"Failed to create connector {connector_id!r}: HTTP {exc.status_code} - {exc.detail}"
+        )
         raise
```
dharaneeshvrd: Need to fix this plain raise.

services/digitize/api/v1/connectors.py
dharaneeshvrd: Please add some context here too

services/digitize/db/manager.py
dharaneeshvrd: Ok understood.

services/digitize/db/scripts/init_schema.sql
dharaneeshvrd: ok

services/digitize/utils/db.py
```diff
@@ -1403,7 +1405,10 @@ def finalize_sync_log_and_update_connector(
     )
     if not found:
         return False
-    db_manager.update_connector_after_sync(connector_id, status=status, last_sync_at=now)
+    connector_error = f"error coming from sync seq {seq}: {error}" if error else error
```
dharaneeshvrd:
```suggestion
    connector_error = f"Error from last sync: {error}" if error else error
```

services/digitize/api/v1/connectors.py
sats-23: Done, removed. 
Connection validation and error propagation to tables happens via sync tick

services/digitize/api/v1/connectors.py
dharaneeshvrd: this should be
```suggestion
            db_ops.set_connector_error(connector_id, "Failed to delete connector, error: {some error msg}")
```

dharaneeshvrd: Please revisit and check all the user facing error messages have proper context or not.

services/digitize/api/v1/connectors.py

services/digitize/api/v1/connectors.py
sats-23: Done, added

services/digitize/api/v1/connectors.py
sats-23: Done

services/digitize/utils/db.py
sats-23: Done

services/digitize/api/v1/connectors.py
sats-23: Done, PTAL

services/digitize/connectors/encryption.py
```diff
@@ -33,25 +32,24 @@
 
 
 @lru_cache(maxsize=1)
-def _load_key(key_path: str) -> AESGCM:
-    """Load and cache the AES-256-GCM cipher from the key file."""
-    path = Path(key_path)
-    if not path.exists():
+def _load_key() -> AESGCM:
+    """Load and cache the AES-256-GCM cipher from the DB_ENCRYPTION_KEY env var."""
+    raw = os.environ.get("DB_ENCRYPTION_KEY", "")
```
dharaneeshvrd, line 37: Is the catalog side changes are ready to add this in assets?

services/digitize/api/v1/connectors.py
```diff
@@ -298,14 +315,14 @@ async def _run_teardown(connector_id: str) -> None:
...
-    deletion_failed = False
+    deletion_errors: list[str] = []
     try:
         # Step 1: Remove the scheduled job so no new ticks fire after this point.
         try:
             import digitize.connectors.scheduler as _sched
             await _sched.remove_connector_job(connector_id)
         except Exception as sched_exc:
-            deletion_failed = True
+            deletion_errors.append(f"scheduler job removal failed: {sched_exc}")
```
dharaneeshvrd:
```suggestion
            deletion_errors.append(f"Failed to remove scheduler job: {sched_exc}")
```

services/digitize/api/v1/connectors.py
```diff
@@ -317,9 +334,9 @@ async def _run_teardown(connector_id: str) -> None:
...
                 if remaining == 0 and doc_id:
                     if not _best_effort_delete_document(doc_id):
-                        deletion_failed = True
+                        deletion_errors.append(f"document deletion failed for checksum {checksum!r}")
             except Exception as exc:
-                deletion_failed = True
+                deletion_errors.append(f"checksum removal failed for {checksum!r}: {exc}")
                 logger.error(
                     f"Error removing checksum {checksum!r} for connector "
                     f"{connector_id!r}: {exc}",
```
dharaneeshvrd:
```suggestion
        # Steps 2+3: remove checksum ownership; delete orphaned documents
        owned_checksums = db_ops.list_connector_checksums(connector_id)
        checksum_removal_failures: list[str] = []
        doc_deletion_failures: list[str] = []
        for checksum in owned_checksums:
            try:
                remaining, doc_id = db_ops.remove_connector_checksum_entry(connector_id, checksum)
                if remaining == 0 and doc_id:
                    if not _best_effort_delete_document(doc_id):
                        doc_deletion_failures.append(checksum)
            except Exception as exc:
                checksum_removal_failures.append(checksum)
                logger.error(
                    f"Error removing checksum {checksum!r} for connector "
                    f"{connector_id!r}: {exc}",
                    exc_info=True,
                )
        if checksum_removal_failures:
            deletion_errors.append(
                f"checksum removal failed for {len(checksum_removal_failures)} checksum(s)"
            )
        if doc_deletion_failures:
            deletion_errors.append(
                f"document deletion failed for {len(doc_deletion_failures)} checksum(s)"
            )
```

dharaneeshvrd: See whether this can be used. Earlier version would have lot of content if there is a failure on the document deletion, since we tried to add all the doc's error messages.

dharaneeshvrd: Seems this logic is used in _delete_orphans method as well see whether we can reuse here with the error messages returned the way we want

dharaneeshvrd:
Remaining mostly lgtm
> It would be great, if you can produce possible error cases and how the connector and sync log's error fields looks like.

Any update on this?

services/digitize/connectors/encryption.py
sats-23: They have already added it in their recent PR's
The DB_ENCRYPTION_KEY is exposed just like other creds (generate and place in secrets)
So we are just consuming it now

services/digitize/api/v1/connectors.py

services/digitize/api/v1/connectors.py
sats-23: Done, I have extracted the logic to individual log during error and also extract and send back to caller to show len like suggested above. PTAL

services/digitize/connectors/encryption.py
dharaneeshvrd: Ok I saw its updated only in catalog digitize service asset, can we update on the legacy templates as well?

sats-23: Sure

sats-23: Done

dharaneeshvrd (approved):

PR #1285, opened 2026-08-19
fix: preserve existing secrets on OpenShift re-configure

Shubhamag12:
## Problem
On every re-run of catalog configure on OpenShift, the Helm upgrade was silently overwriting the db-password in catalog-db-secret and admin-password in catalog-secret with empty strings.

## Fix
Use lookup in both secret templates to consider existing secrets data

### DIFF
--- a/ai-services/assets/catalog/openshift/templates/catalog-db-encryption-secret.yaml
+++ b/ai-services/assets/catalog/openshift/templates/catalog-db-encryption-secret.yaml
@@ -1,4 +1,4 @@
-{{- if not (lookup "v1" "Secret" .Release.Namespace "catalog-db-encryption-secret") }}
+{{- $existingSecret := lookup "v1" "Secret" .Release.Namespace "catalog-db-encryption-secret" }}
 apiVersion: v1
 kind: Secret
 metadata:
@@ -10,5 +10,4 @@ metadata:
     ai-services.io/component: catalog
 type: Opaque
 stringData:
-  db-encryption-key: {{ .Values.backend.dbEncryptionKey | quote }}
-{{- end }}  
+  db-encryption-key: {{ if $existingSecret }}{{ index $existingSecret.data "db-encryption-key" | b64dec | quote }}{{ else }}{{ .Values.backend.dbEncryptionKey | quote }}{{ end }}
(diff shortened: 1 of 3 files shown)


prachikurade1 (dismissed):
Approved

mayuka-c (dismissed):
LGTM,

Pls ensure its thoroughly validated with multiple reruns.

yussufsh:
Did you test reset password after this?
Please hold merging this pr until regression testing is done.
I had put similar comment in connectors pr for @prachikurade1 asking what it will do.

yussufsh:
Another thing is instead of creating the secret again with same value can we just not skip it wrapping around if condition? This way we are not changing the create time of the resources.

Shubhamag12:
> Did you test reset password after this? Please hold merging this pr until regression testing is done. I had put similar comment in connectors pr for @prachikurade1 asking what it will do.

reset-password operation is independent of helm operation. reset-password updates the secret directly and restarts the catalog pod. That said, rest of testing is still pending, waiting for setup to be available

mayuka-c:
> Another thing is instead of creating the secret again with same value can we just not skip it wrapping around if condition? This way we are not changing the create time of the resources.

I think this is because on multiple reruns in case of openshift `helm upgrade` is done and these changes are for it to tell not to override.

@Shubhamag12 @prachikurade1 Correct if I'm wrong

prachikurade1:
> > Another thing is instead of creating the secret again with same value can we just not skip it wrapping around if condition? This way we are not changing the create time of the resources.
> 
> I think this is because on multiple reruns in case of openshift `helm upgrade` is done and these changes are for it to tell not to override.
> 
> @Shubhamag12 @prachikurade1 Correct if I'm wrong

yes, this is to handle helm upgrade during catalog configure re-runs.

Shubhamag12:
> Another thing is instead of creating the secret again with same value can we just not skip it wrapping around if condition? This way we are not changing the create time of the resources.

so you mean skip the resource entirely if it already exists? 
If yes, then that also makes sense.

yussufsh (changes requested):
Please do same for `ai-services/assets/catalog/openshift/templates/catalog-db-encryption-secret.yaml`

Shubhamag12:
> Another thing is instead of creating the secret again with same value can we just not skip it wrapping around if condition? This way we are not changing the create time of the resources.

I tried this and added condition `{{- if not (lookup "v1" "Secret" .Release.Namespace "catalog-db-secret") }}`:
What happened is: 
When we re-ran configure cmd, secrets exists and above condition evaluates to false and template renders nothing. Now helm compares old release manifests (17 resources) to new manifest (14 resources, secrets missing), and then helm deletes the 3 secrets.

So we have to go with current approach that is already done for `catalog-db-encryption-secret` by @prachikurade1

yussufsh (approved):

PR #1286, opened 2026-08-19
AIS-1583 - Changes for DB Encryption Key for Digitize Service

prachikurade1:
Verified that the secret is created and mounted on the digitize backend pod (standalone/rag for both runtimes)
Verified that deletion of app/service deletes the secret.


ai-services/assets/services/digitize/openshift/templates/digitize-api-deployment.yaml
```diff
@@ -71,6 +71,11 @@ spec:
                 secretKeyRef:
                   name: "digitize-db-secret"
                   key: password
+            - name: DB_ENCRYPTION_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: "digitize-db-encryption-secret"
+                  key: db-encryption-key
```
mayuka-c, line 78: @dharaneeshvrd Any key name that you want to keep/prefer the env?
We are using DB_ENCRYPTION_KEY even for our catalog backend as this key is not only meant for encrypting private key for connectors but can be used to ensure data is encrypted at rest (if required)

mayuka-c (dismissed):
LGTM

ai-services/assets/services/digitize/openshift/templates/digitize-api-deployment.yaml
dharaneeshvrd: @sats-23 Hope we can follow the similar env name.

Shubhamag12 (dismissed):
lgtm

ai-services/assets/services/digitize/openshift/templates/digitize-api-deployment.yaml
sats-23: Yes, this works 
Currently we have it as connector_encryption_key, but will follow up with a PR to change the naming as this sounds better

mayuka-c (approved):

PR #1287, opened 2026-08-19
[BYOS] Create Bundle API

mayuka-c:
- POST API for supporting upload of a bundle
- Update necessary status in DB
- Create volumes for storing bundles in both podman and openshift
- Follows the process as documented in proposal here: https://github.com/IBM/project-ai-services/blob/main/docs/proposals/catalog/custom-service-templates-proposal.md#62-new-api-endpoints
- This is part of my POC PR but with lot more structured and fixes: https://github.com/IBM/project-ai-services/compare/main...mayuka-c:project-ai-services:poc-custom-assets

Note:
- Validation of bundle as part of POST call is not covered here (will be doing as part of Validate API and integrating).
- Reload of Catalog Items cache to reflect the bundle is pending (Taken up later when we hook this bundle with ListServices)

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/services/bundle/archive.go
+++ b/ai-services/internal/pkg/catalog/apiserver/services/bundle/archive.go
@@ -29,6 +29,15 @@ const (
 	// (50 MB). Any single file — or the aggregate of all files written — whose declared
 	// tar header Size exceeds this value is rejected before any bytes are written to disk.
 	maxExtractedFileSize int64 = 50 * 1024 * 1024
+
+	// splitTwo is the n argument for strings.SplitN when splitting on the first slash.
+	splitTwo = 2
+
+	// dirPerm is the permission bits used when creating directories during extraction.
+	// 0o750: owner=rwx, group=rx, other=none. The bundle volume is a container-private
+	// mount; world access is blocked while group read/execute is retained for any
+	// supplementary GID the container runtime assigns to the apiserver process.
+	dirPerm = 0o750
 )
 
 // rawMetadataYAML holds the minimal set of fields decoded from root metadata.yaml.
@@ -52,6 +61,7 @@ type rawMetadataYAML struct {
 //	component llm--my-prov    1.0.0  → /data/catalog-bundles/components/llm--my-prov-1.0.0
 func bundleDirPath(catalogType, catalogID, version string) string {
 	subdir := catalogTypeToDir(catalogType)
+
 	return filepath.Join(bundleStorageRoot, subdir, catalogID+"-"+version)
 }
 
@@ -104,10 +114,16 @@ func parseMetadataFromBytes(data []byte) (BundleMetadata, error) {
 			Message: fmt.Sprintf("invalid gzip archive: %s", err),
 		}
 	}
-	defer gr.Close()
+	defer func() { _ = gr.Close() }()
 
 	tr := tar.NewReader(gr)
 
+	return scanArchiveForMetadata(tr)
+}
+
+// scanArchiveForMetadata walks tar entries looking for a root metadata.yaml
+// and delegates parsing to parseMetadataYAML.
+func scanArchiveForMetadata(tr *tar.Reader) (BundleMetadata, error) {
 	// topDir is inferred from the first archive entry that contains a slash.
 	// The directory name is never validated — it is only used to locate
 	// <topDir>/metadata.yaml. A flat archive (no top-level directory) is also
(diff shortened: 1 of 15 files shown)


ai-services/internal/pkg/catalog/apiserver/services/bundle/archive.go
```diff
@@ -0,0 +1,408 @@
...
+//   - Per-file size guard: regular files larger than maxExtractedFileSize are rejected.
+//
+// Only regular files and directories are extracted; symlinks and other special
+// entry types are silently skipped.
+func extractAndMeasure(data []byte, destDir string) (int64, error) {
+	gr, err := gzip.NewReader(newBytesReader(data))
+	if err != nil {
+		return 0, &validators.ValidationError{
+			Code:    http.StatusBadRequest,
+			Message: fmt.Sprintf("invalid gzip archive: %s", err),
```
adarshagrawal38, line 249:
```suggestion
			Message: fmt.Sprintf("invalid gzip archive: %w", err),
```

adarshagrawal38 (dismissed):
LGTM

adarshagrawal38 (dismissed):
lgtm

ai-services/internal/pkg/catalog/apiserver/services/bundle/archive.go
mayuka-c: %w is used for wrapping the errors. Since here we are printing we would have to use %s

ai-services/internal/pkg/catalog/apiserver/services/bundle/archive.go
```diff
@@ -0,0 +1,408 @@
...
+		}
+		if err != nil {
+			return 0, fmt.Errorf("error reading archive entry: %w", err)
+		}
+
+		n, err := processEntry(tr, hdr, destDir, &topDir)
+		if err != nil {
+			return 0, err
+		}
+		totalSize += n
```
Shubhamag12, line 278: I think we should check for maxExtractedFileSize after each call to process entry and fail early

ai-services/internal/pkg/catalog/apiserver/services/bundle/archive.go
```diff
@@ -0,0 +1,408 @@
...
+	}
+
+	return data, meta, nil
+}
+
+// parseMetadataFromBytes walks the gzip-compressed tar archive stored in data,
+// finds the root metadata.yaml (either at the top level or one directory deep),
+// and delegates to parseMetadataYAML.
+func parseMetadataFromBytes(data []byte) (BundleMetadata, error) {
+	gr, err := gzip.NewReader(newBytesReader(data))
```
Shubhamag12: why is `newBytesReader` required? 
can we not use `bytes.NewReader(data)`?

ai-services/internal/pkg/catalog/apiserver/services/bundle/service.go
```diff
@@ -2,63 +2,124 @@ package bundle
...
+	// Step 5: insert DB row with status=processing.
+	row := &models.CatalogBundle{
+		Name:        meta.DisplayName(),
+		CatalogType: meta.CatalogType(),
+		CatalogID:   meta.CatalogID(),
+		Version:     meta.Version(),
+		CreatedBy:   userID,
+	}
+	if err := s.repo.Insert(ctx, row); err != nil {
+		_ = os.RemoveAll(destDir) // best-effort cleanup of the extracted directory
```
Shubhamag12, line 101: I think we should delete even if extraction fails, not just here. As if extraction fails mid way, we are still left with stale files.
Maybe we can add defer func and do `os.RemoveAll` if the entire operation is not successful?

ai-services/internal/pkg/catalog/apiserver/services/bundle/archive.go
```diff
@@ -0,0 +1,408 @@
...
+func catalogTypeToDir(catalogType string) string {
+	switch catalogType {
+	case CatalogTypeService:
+		return bundleDirServices
+	case CatalogTypeComponent:
+		return bundleDirComponents
+	default:
+		// Fallback: append 's' so unknown future types degrade gracefully.
+		return catalogType + "s"
+	}
```
Shubhamag12: I did not understand the default case. Can you elaborate on this?

ai-services/internal/pkg/catalog/apiserver/services/bundle/archive.go
mayuka-c: Yeah we could use the bytes.NewReader instead of simply implementing the READER interface. Let me make changes and validate

ai-services/internal/pkg/catalog/apiserver/services/bundle/archive.go
mayuka-c: Its being done inside processEntry when we do preprocessing within the loop

ai-services/internal/pkg/catalog/apiserver/services/bundle/archive.go
mayuka-c: The default would be when its not of either of the type, but I guess it will be simpler to just have the catalogType and suffix with 's' for everything. Should be simpler. Will update

ai-services/internal/pkg/catalog/apiserver/services/bundle/service.go
mayuka-c: The defer would run even for success and also I was thinking we could maintain a new var bool and check for isError and then based on the flag we can check to clean up or not inside defer.

Had this mentioned to Bob and it says go with cleanups if failed. So have added os.RemoveAll even if extract measure fails now.


BOB Suggestion
```
The defer+bool pattern pays off when you have many early-return error paths all needing the same cleanup — it avoids repetition. Here there are exactly two sites, both already written. Adding a defer just to replace two identical lines with a closure, a bool variable, and a flag-set on each error branch would be more code, not less, and harder to follow.

But it has a real problem here: destDir isn't known at the point where defer is registered — it's computed on line 84, after peekMetadata. You'd need to use a pointer or closure over a variable declared earlier, which adds noise.
```

ai-services/internal/pkg/catalog/apiserver/services/bundle/archive.go
Shubhamag12: but that is only for a single file right? are we expecting single file to be <50MB or aggregated?
```
	// maxExtractedFileSize is the total uncompressed size limit enforced during extraction
	// (50 MB). Any single file — or the aggregate of all files written — whose declared
	// tar header Size exceeds this value is rejected before any bytes are written to disk.
```

as per this comment, it says both. However the above existing check is only per file. 

Please correct me if I am wrong.

mayuka-c: Yeah I was thinking its aggregation of all files :)
I have updated and we will check if totalSize exceeds then do early return

mayuka-c: Added a UT to cover the test case and works fine. PTAL

ai-services/internal/pkg/catalog/apiserver/services/bundle/archive.go
Shubhamag12: so what if they specify "Service" or "architecture"? This would still proceed right?

mayuka-c: No during validation phase we will check and support only service and component as catalog types!. This is used post validation during extraction for top level dir name.

Shubhamag12: ah okay, makes sense.

Shubhamag12 (approved):
lgtm

PR #1288, opened 2026-08-19
feat: propagate context through Runtime interface for cancellation support

shwetha-s-poojary:
Added `ctx context.Context` to all methods on the `Runtime` interface so that cancellation signals propagate correctly through the entire call chain.

**Why each group changed:**

- **`runtime/interface.go` + implementations (Podman, OpenShift)** — core change; every method now accepts `ctx` so callers can cancel in-flight operations
- **`application/podman` & `openshift` (`create.go`, `delete.go`)** — deployment path: `ctx` now cancels image pulls, pod creation, and readiness waits mid-flight
- **`cli/podman/deploy.go`, `cli/helpers/helper.go`** — readiness wait loops; cancellation stops waiting immediately
- **`image/`** — image pull respects cancellation
- **`catalog/apiserver/services/deployment/`, `deletion/`** — API-triggered create/delete path
- **`application/{podman,openshift}` (`info.go`, `logs.go`, `ps.go`, `start.go`, `stop.go`) + `catalog/cli/configure/podman` reset functions + `catalog/apiserver/services/sync`** — accept and propagate `ctx` from the cmd layer (`cmd.Context()`) so the full call chain is consistent

**Validation:**
- Applications deployed and deleted via catalog — no regressions
- Applications created and deleted via CLI — no regressions

**Note:** `cmd.Context()` currently behaves the same as `context.Background()` since OS signal wiring at the root command is not yet in place. This change establishes the correct plumbing so that when signal handling is added (tracked separately), cancellation will propagate end-to-end with no further changes required in the internal packages.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/delete.go
+++ b/ai-services/cmd/ai-services/cmd/application/delete.go
@@ -182,6 +182,9 @@ func waitForApplicationDeletion(ctx context.Context, appClient *catalogClient.Ap
 		maxAttempts  = 12
 	)
 
+	timer := time.NewTimer(pollInterval)
+	defer timer.Stop()
+
 	for range maxAttempts {
 		// Check if application still exists via API
 		app, err := appClient.GetApplication(ctx, appID)
@@ -204,9 +207,10 @@ func waitForApplicationDeletion(ctx context.Context, appClient *catalogClient.Ap
 
 		// Wait before next poll, but honour context cancellation
 		select {
-		case <-time.After(pollInterval):
 		case <-ctx.Done():
 			return ctx.Err()
+		case <-timer.C:
+			timer.Reset(pollInterval)
 		}
 	}
 
--- a/ai-services/cmd/ai-services/cmd/application/delete.go
+++ b/ai-services/cmd/ai-services/cmd/application/delete.go
@@ -202,8 +202,12 @@ func waitForApplicationDeletion(ctx context.Context, appClient *catalogClient.Ap
 			// Application still exists, continue polling
 		}
 
-		// Wait before next poll
-		time.Sleep(pollInterval)
+		// Wait before next poll, but honour context cancellation
+		select {
+		case <-time.After(pollInterval):
+		case <-ctx.Done():
+			return ctx.Err()
+		}
 	}
 
 	return fmt.Errorf("timeout waiting for application deletion after %v", maxAttempts*pollInterval)
--- a/ai-services/cmd/ai-services/cmd/application/delete.go
+++ b/ai-services/cmd/ai-services/cmd/application/delete.go
@@ -82,7 +82,7 @@ Arguments:
 		}
 
 		// Default: use new implementation via catalog
-		return deleteApplication(applicationName)
+		return deleteApplication(cmd.Context(), applicationName)
 	},
 }
 
@@ -126,12 +126,12 @@ func buildDeleteFlagValidator() *flagvalidator.FlagValidator {
 	return builder.Build()
 }
 
-func deleteApplication(appName string) error {
-	appClient, err := catalogClient.NewApplicationClient()
+func deleteApplication(ctx context.Context, appName string) error {
+	appClient, err := catalogClient.NewApplicationClient(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to create application client: %w", err)
 	}
-	app, err := cliUtils.GetAppByName(appClient, appName)
+	app, err := cliUtils.GetAppByName(ctx, appClient, appName)
 	if err != nil {
 		return err
 	}
@@ -157,8 +157,8 @@ func deleteApplication(appName string) error {
 
 	// Retry deletion if it fails
 	logger.Infof("Deleting application %s...\n", appName)
-	err = utils.Retry(context.Background(), vars.RetryCount, vars.RetryInterval, nil, func() error {
-		return appClient.DeleteApplication(app.ID, &deleteParams)
+	err = utils.Retry(ctx, vars.RetryCount, vars.RetryInterval, nil, func() error {
+		return appClient.DeleteApplication(ctx, app.ID, &deleteParams)
 	})
 	if err != nil {
 		return fmt.Errorf("failed to delete application after %d retries: %w", vars.RetryCount, err)
@@ -166,7 +166,7 @@ func deleteApplication(appName string) error {
 
 	// Poll to verify deletion is complete
 	logger.Infof("Waiting for application %s to be deleted...\n", appName)
-	if err := waitForApplicationDeletion(appClient, app.ID); err != nil {
+	if err := waitForApplicationDeletion(ctx, appClient, app.ID); err != nil {
 		return fmt.Errorf("failed to verify application deletion: %w", err)
 	}
 
@@ -176,7 +176,7 @@ func deleteApplication(appName string) error {
 }
 
 // waitForApplicationDeletion polls the application status until it's fully deleted.
-func waitForApplicationDeletion(appClient *catalogClient.ApplicationClient, appID string) error {
+func waitForApplicationDeletion(ctx context.Context, appClient *catalogClient.ApplicationClient, appID string) error {
 	const (
 		pollInterval = 5 * time.Second
 		maxAttempts  = 12
@@ -184,7 +184,7 @@ func waitForApplicationDeletion(appClient *catalogClient.ApplicationClient, appI
 
 	for range maxAttempts {
 		// Check if application still exists via API
-		app, err := appClient.GetApplication(appID)
+		app, err := appClient.GetApplication(ctx, appID)
 		if err != nil {
 			// Check if it's an HTTPError with 404 status code
 			var httpErr *catalogClient.HTTPError
--- a/ai-services/internal/pkg/application/openshift/info.go
+++ b/ai-services/internal/pkg/application/openshift/info.go
@@ -46,7 +46,7 @@ func (o *OpenshiftApplication) Info(ctx context.Context, opts types.InfoOptions)
 	// Step3: Read and print the info.md file
 	tp := templates.NewEmbedTemplateProvider(&assets.ApplicationFS)
 
-	if err := helpers.PrintInfo(tp, o.runtime, opts.Name, appTemplate); err != nil {
+	if err := helpers.PrintInfo(ctx, tp, o.runtime, opts.Name, appTemplate); err != nil {
 		// not failing if overall info command, if we cannot display Info
 		logger.Errorf("failed to display info: %v\n", err)
 
--- a/ai-services/internal/pkg/application/openshift/info.go
+++ b/ai-services/internal/pkg/application/openshift/info.go
@@ -13,8 +13,7 @@ import (
 )
 
 // Info displays detailed information about an application.
-func (o *OpenshiftApplication) Info(opts types.InfoOptions) error {
-	ctx := context.Background()
+func (o *OpenshiftApplication) Info(ctx context.Context, opts types.InfoOptions) error {
 	// Step1: Do List pods and filter for given application name
 
 	listFilters := map[string][]string{}
--- a/ai-services/internal/pkg/application/openshift/info.go
+++ b/ai-services/internal/pkg/application/openshift/info.go
@@ -1,6 +1,7 @@
 package openshift
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/project-ai-services/ai-services/assets"
@@ -13,6 +14,7 @@ import (
 
 // Info displays detailed information about an application.
 func (o *OpenshiftApplication) Info(opts types.InfoOptions) error {
+	ctx := context.Background()
 	// Step1: Do List pods and filter for given application name
 
 	listFilters := map[string][]string{}
@@ -20,7 +22,7 @@ func (o *OpenshiftApplication) Info(opts types.InfoOptions) error {
 		listFilters["label"] = []string{fmt.Sprintf("ai-services.io/application=%s", opts.Name)}
 	}
 
-	pods, err := o.runtime.ListPods(listFilters)
+	pods, err := o.runtime.ListPods(ctx, listFilters)
 	if err != nil {
 		return fmt.Errorf("failed to list pods: %w", err)
 	}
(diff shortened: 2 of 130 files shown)


ai-services/internal/pkg/application/openshift/info.go
```diff
@@ -13,14 +14,15 @@ import (
 
 // Info displays detailed information about an application.
 func (o *OpenshiftApplication) Info(opts types.InfoOptions) error {
+	ctx := context.Background()
```
mayuka-c, line 17: Let us propogate from cmd

ai-services/internal/pkg/application/openshift/logs.go
```diff
@@ -9,19 +10,20 @@ import (
 
 // Logs displays logs from an application pod.
 func (o *OpenshiftApplication) Logs(opts types.LogsOptions) error {
+	ctx := context.Background()
```
mayuka-c, line 13: Same here

ai-services/internal/pkg/application/openshift/ps.go
```diff
@@ -14,8 +15,9 @@ func (o *OpenshiftApplication) List(opts appTypes.ListOptions) ([]appTypes.Appli
 		return nil, fmt.Errorf("application name is required for openshift runtime")
 	}
 
+	ctx := context.Background()
```
mayuka-c, line 18: Here too

ai-services/internal/pkg/application/podman/info.go
```diff
@@ -13,14 +14,15 @@ import (
 
 // Info displays detailed information about an application.
 func (p *PodmanApplication) Info(opts types.InfoOptions) error {
+	ctx := context.Background()
```
mayuka-c, line 17: Here

ai-services/internal/pkg/application/podman/logs.go
```diff
@@ -9,19 +10,20 @@ import (
 
 // Logs displays logs from an application pod.
 func (p *PodmanApplication) Logs(opts types.LogsOptions) error {
+	ctx := context.Background()
```
mayuka-c, line 13: Here

ai-services/internal/pkg/application/podman/ps.go
```diff
@@ -9,8 +11,9 @@ import (
 
 // List returns information about running applications.
 func (p *PodmanApplication) List(opts appTypes.ListOptions) ([]appTypes.ApplicationInfo, error) {
+	ctx := context.Background()
```
mayuka-c, line 14: Here

ai-services/internal/pkg/application/podman/start.go
```diff
@@ -14,11 +15,12 @@ import (
 
 // Start starts a stopped application.
 func (p *PodmanApplication) Start(opts appTypes.StartOptions) error {
+	ctx := context.Background()
```
mayuka-c, line 18: Here too should be propogated from cmd

ai-services/internal/pkg/application/podman/stop.go
```diff
@@ -13,7 +14,8 @@ import (
 
 // Stop stops a running application.
 func (p *PodmanApplication) Stop(opts appTypes.StopOptions) error {
-	pods, err := p.listApplicationPods(opts)
+	ctx := context.Background()
```
mayuka-c, line 17: Here

ai-services/internal/pkg/catalog/apiserver/services/sync/openshift.go
```diff
@@ -31,7 +31,7 @@ func newOpenShiftSync() *openShiftSync {
 
 // FetchPodStatuses fetches all pods labelled with the given templateID using the OpenShift runtime.
 func (s *openShiftSync) FetchPodStatuses(rt runtime.Runtime, templateID string) ([]*PodStatus, error) {
-	filteredPods, err := common.FetchFilteredPods(rt, templateID)
+	filteredPods, err := common.FetchFilteredPods(context.Background(), rt, templateID)
```
mayuka-c, line 34: ctx needs to be propogated from caller

ai-services/internal/pkg/catalog/apiserver/services/sync/podman.go
```diff
@@ -32,7 +32,7 @@ func newPodmanSync(catalogProvider *catalogpkg.CatalogProvider) *podmanSync {
 // FetchPodStatuses fetches all pods labelled with the given templateID and returns their statuses.
 // It uses InspectPod and InspectContainer (via common.ProcessPod) to derive state and health.
 func (s *podmanSync) FetchPodStatuses(rt runtime.Runtime, templateID string) ([]*PodStatus, error) {
-	filteredPods, err := common.FetchFilteredPods(rt, templateID)
+	filteredPods, err := common.FetchFilteredPods(context.Background(), rt, templateID)
```
mayuka-c, line 35: Here also

ai-services/internal/pkg/catalog/apiserver/services/sync/podman.go
```diff
@@ -43,7 +43,7 @@ func (s *podmanSync) FetchPodStatuses(rt runtime.Runtime, templateID string) ([]
 
 	var podStatuses []*PodStatus
 	for _, pod := range filteredPods {
-		processedPod, err := common.ProcessPod(rt, pod)
+		processedPod, err := common.ProcessPod(context.Background(), rt, pod)
```
mayuka-c, line 46: Here

ai-services/internal/pkg/catalog/cli/configure/podman/reset_certificate.go
```diff
@@ -23,7 +23,7 @@ func ResetCatalogCertificate(sslCertPath, sslKeyPath string) error {
 	}
 
 	// Validate catalog service is running
-	isCatalogRunning, err := IsCatalogServiceRunning(deployCtx.Runtime)
+	isCatalogRunning, err := IsCatalogServiceRunning(context.Background(), deployCtx.Runtime)
```
mayuka-c, line 26: Need to define one context variable at the beginning and same should be used across

ai-services/internal/pkg/catalog/cli/configure/podman/reset_password.go
```diff
@@ -19,7 +19,7 @@ func ResetCatalogPassword() error {
 	}
 
 	// Validate catalog service and confirm reset action
-	shouldProceed, err := validateCatalogServiceAndConfirmReset(deployCtx.Runtime, "password")
+	shouldProceed, err := validateCatalogServiceAndConfirmReset(context.Background(), deployCtx.Runtime, "password")
```
mayuka-c, line 22: Here also define one at the beginning and reuse

mayuka-c: Pls ensure its followed everywhere in this cli package

mayuka-c (commented):
@shwetha-s-poojary Please ensure its thoroughly validated and there are no regressions

yussufsh (changes requested):
Wondering if we already have the ctx in the podman and openshift client why are we not using it? If we are passing in each func call then please remove it from the client struct.

shwetha-s-poojary:
> Wondering if we already have the ctx in the podman and openshift client why are we not using it? If we are passing in each func call then please remove it from the client struct

If you are referring to pc.Context and ctx, pc.Context on the Podman client is not a cancellation context - it carries the Podman connection handle, we have podmanCtx()  to merge it with the caller's ctx to get both the connection value and cancellation.We can't remove it IMO.

ai-services/internal/pkg/catalog/apiserver/services/sync/podman.go
```diff
@@ -91,11 +91,14 @@ func validateResourceExistenceChecks(ctx context.Context, expectedSecretNames, e
 
 	var errorMessages []string
 
-	if msg := validateResourceExistence(ctx, expectedSecretNames, "secret", rt.SecretExists); msg != "" {
+	secretExistsFn := func(nameOrID string) (bool, error) { return rt.SecretExists(ctx, nameOrID) }
+	volumeExistsFn := func(nameOrID string) (bool, error) { return rt.VolumeExists(ctx, nameOrID) }
+
+	if msg := validateResourceExistence(ctx, expectedSecretNames, "secret", secretExistsFn); msg != "" {
```
yussufsh: Why is this need for func vars?
You can change the signature of validateResourceExistence() itself.

ai-services/cmd/ai-services/cmd/application/delete.go
yussufsh: Minor concern: This waits for 5 seconds even if ctx is cancelled in future. If a signal arrives mid-sleep, the process will hang for up to 5 seconds before the next iteration. Let's add a TODO if it's complex, or fix it now.

ai-services/internal/pkg/catalog/cli/info/podman/info.go
```diff
@@ -76,7 +77,7 @@ func DisplayCatalogInfo() error {
 // GetCatalogRouteInfo retrieves route domains and HTTPS port for the catalog service.
 // This orchestrates: deployContext gets pod name and route info from templates,
 // caddy.Context queries Caddy for route domains and HTTPS port.
-func GetCatalogRouteInfo(rt *rt.PodmanClient) (map[string]string, string, error) {
+func GetCatalogRouteInfo(ctx context.Context, rt *rt.PodmanClient) (map[string]string, string, error) { //nolint:revive
```
yussufsh: Why `nolint`?
Something related to previous code?

ai-services/internal/pkg/proxy/caddy.go
```diff
@@ -58,12 +58,12 @@ func GetCaddyProxyManager() (ProxyManager, error) {
 }
 
 // HealthCheck verifies Caddy is running and accessible.
-func (c *caddyManager) HealthCheck() error {
+func (c *caddyManager) HealthCheck(ctx context.Context) error {
```
yussufsh, line 61: You are missing adding client ctx for `GetRouteByID` and `UnregisterRoute`.

yussufsh (changes requested):

ai-services/internal/pkg/catalog/apiserver/services/sync/podman.go
shwetha-s-poojary: Yeah we can do that.

ai-services/internal/pkg/catalog/cli/info/podman/info.go
shwetha-s-poojary: When I added ctx context.Context ,revive linter's context-as-argument rule kicked in and caught the import-shadowing on rt.
We can also change the parameter name from rt to something else.What do we do keep this nolint or change the naming and fix it?

ai-services/internal/pkg/proxy/caddy.go
shwetha-s-poojary: Thanks for catching ! Will address that.

ai-services/cmd/ai-services/cmd/application/delete.go
shwetha-s-poojary: Right, that should be straightforward. I’ll fix it by making the sleep context-aware.

ai-services/internal/pkg/catalog/cli/info/podman/info.go
shwetha-s-poojary: I've renamed it to runtime

ai-services/cmd/ai-services/cmd/application/delete.go
```diff
@@ -202,8 +202,12 @@ func waitForApplicationDeletion(appClient *catalogClient.ApplicationClient, appI
 			// Application still exists, continue polling
 		}
 
-		// Wait before next poll
-		time.Sleep(pollInterval)
+		// Wait before next poll, but honour context cancellation
+		select {
+		case <-time.After(pollInterval):
```
yussufsh: Bob suggested not to use time.After():
```
// Before (current PR — leaks a timer per iteration)
select {
case <-time.After(pollInterval):
case <-ctx.Done():
    return ctx.Err()
}

// After (correct)
timer := time.NewTimer(pollInterval)
defer timer.Stop()

for range maxAttempts {
    app, err := appClient.GetApplication(ctx, appID)
    // ... handle response ...

    select {
    case <-ctx.Done():
        return ctx.Err()
    case <-timer.C:
        timer.Reset(pollInterval)
    }
}
```

yussufsh: Here's why:

time.After in a loop leaks timers. Each call to time.After creates a new time.Timer internally that is not garbage collected until it fires. If you call it on every iteration of a loop, you create N timers — all but the last one leak until their duration expires.
```
// ❌ leaks a timer on every iteration
for range maxAttempts {
    // ... API call ...
    select {
    case <-ctx.Done():
        return ctx.Err()
    case <-time.After(pollInterval): // new timer created, old one leaks
    }
}
```
time.NewTimer + Reset is the correct pattern for a loop because you reuse a single timer object:
```
// ✅ single timer, no leaks
timer := time.NewTimer(pollInterval)
defer timer.Stop()

for range maxAttempts {
    // ... API call ...
    select {
    case <-ctx.Done():
        timer.Stop()
        return ctx.Err()
    case <-timer.C:
        timer.Reset(pollInterval)
    }
}
```
The leak from time.After is bounded here (at most maxAttempts = 12 timers of 5s each, all short-lived), so it wouldn't be catastrophic in practice — but it's still incorrect idiom. time.NewTimer + Reset is the right call.

ai-services/internal/pkg/catalog/cli/info/podman/info.go
yussufsh: Then better approach would have been to remove the runtime/podman alias in the import.
It should have been... 
```
rt *podman.PodmanClient
```

ai-services/cmd/ai-services/cmd/application/delete.go
shwetha-s-poojary: I can make the change.Any idea how to simulate or validate this?

yussufsh: Just testing the main delete path is enough.

yussufsh (approved):
lgtm

PR #1292, opened 2026-08-20
[BYOS] PUT API to support Replace Bundle

mayuka-c:
- Follows the process as documented in proposal here: https://github.com/IBM/project-ai-services/blob/main/docs/proposals/catalog/custom-service-templates-proposal.md#62-new-api-endpoints
- This is part of my POC PR but with lot more structured and fixes: https://github.com/IBM/project-ai-services/compare/main...mayuka-c:project-ai-services:poc-custom-assets

Note:

- Validation of bundle as part of PUT call is not covered here (will be done as part of Validate API and integrating).
- Reload of Catalog Items cache to reflect the bundle is pending (Taken up later when we hook this bundle with ListServices)

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/handlers/bundle_handler.go
+++ b/ai-services/internal/pkg/catalog/apiserver/handlers/bundle_handler.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog/apiserver/middleware"
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog/apiserver/repository"
 	bundlesvc "github.com/project-ai-services/ai-services/internal/pkg/catalog/apiserver/services/bundle"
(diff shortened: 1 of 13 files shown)


ai-services/internal/pkg/catalog/apiserver/handlers/bundle_handler.go
```diff
@@ -119,15 +120,57 @@ func (h *BundleHandler) ValidateBundle(c *gin.Context) {
...
+
+	file, header, err := c.Request.FormFile("file")
+	if err != nil {
+		c.JSON(http.StatusBadRequest, ErrorResponse{Error: "missing or unreadable 'file' field: " + err.Error()})
+
+		return
+	}
+	defer func() { _ = file.Close() }()
+
+	if !strings.HasSuffix(strings.ToLower(header.Filename), ".tar.gz") {
```
adarshagrawal38: How about using constant for file extension `.tar.gz`,  as string literal is used in multiple places.

ai-services/internal/pkg/catalog/apiserver/services/bundle/service.go
```diff
@@ -127,28 +127,119 @@ func (s *bundleService) ProcessBundle(ctx context.Context, file io.Reader, userI
...
+		return nil, fmt.Errorf("invalid existing bundle id %q: %w", existing.ID, err)
+	}
+
+	// Step 4: mark existing row processing.
+	statusProcessing := models.BundleStatusProcessing
+	if updateErr := s.repo.Update(ctx, existingID, models.BundleUpdate{Status: &statusProcessing}); updateErr != nil {
+		return nil, fmt.Errorf("failed to mark bundle as processing: %w", updateErr)
+	}
+
+	// Steps 5–9: extract, rename, activate, cleanup. Any failure after step 4
```
adarshagrawal38:
```suggestion
	// Steps 5–10: extract, rename, activate, cleanup. Any failure after step 4
```

adarshagrawal38 (approved):
LGTM

ai-services/internal/pkg/catalog/apiserver/services/bundle/service.go
```diff
@@ -127,28 +130,124 @@ func (s *bundleService) ProcessBundle(ctx context.Context, file io.Reader, userI
...
-//  8. Reload CatalogProvider.
+//  7. UPDATE existing row in-place (status=active, version, name, size_bytes) via BundleRepository.Update.
+//  8. TODO — CatalogProvider.Reload() once the provider reference is wired in.
 //  9. Delete old on-disk directory when it differs from the new final path.
 //  10. Re-fetch via BundleRepository.GetByID and return as *BundleResponse.
 //     On failure after step 4: mark row failed, store error message.
-func (s *bundleService) ReplaceBundle(_ context.Context, _ *BundleRecord, _ io.Reader, _ string) (*BundleResponse, error) {
-	// TODO: implement
-	panic("not implemented")
+func (s *bundleService) ReplaceBundle(ctx context.Context, existing *BundleResponse, file io.Reader, _ string) (*BundleResponse, error) {
```
Shubhamag12, line 145: wondering if we should use userID here and have a column stating who updated it?

ai-services/internal/pkg/catalog/apiserver/services/bundle/service.go
```diff
@@ -237,6 +336,57 @@ func (s *bundleService) ListBundles(ctx context.Context, req BundleListRequest)
...
+			return &validators.ValidationError{
+				Code: http.StatusConflict,
+				Message: fmt.Sprintf(
+					"cannot replace bundle: one or more components are currently running with type %q and provider %q",
+					componentType, provider,
+				),
+			}
+		}
+	}
+
```
Shubhamag12, line 386: shall we have a default case for unknown types?

ai-services/internal/pkg/catalog/apiserver/services/bundle/service.go
mayuka-c: That would be kind of like an audit trial which requires separate implementation everywhere (Would be required for Update/Delete Application too).

For now I prefer just to have userID of who created it (also it is a single user now) but definitely we would need when we implement audit trial

ai-services/internal/pkg/catalog/apiserver/services/bundle/service.go
mayuka-c: Before this we would have a validation check to error out for unknown types during validation flow (not added yet will be taken along with Validate API and integrated here).

Even then I feel we could have a default condition and error out saying unsupported. Will it take it up in delete flow as I will be reusing this method.

ai-services/internal/pkg/catalog/apiserver/services/bundle/service.go
Shubhamag12: Sure, I was thinking to add updatedby here since we already had createdby, but yes we can take it as part of audit implementation

Shubhamag12 (approved):
lgtm

ai-services/internal/pkg/catalog/apiserver/services/bundle/service.go
mayuka-c: Yeah but I dont think we can simply have updated_by as it could be updated multiple times by different users. We may need to record everything with timestamps

PR #1294, opened 2026-08-20
AIS-1630 - Create Datasource API

prachikurade1:
- Create Datasource API:
- Validates the request body - datasource name, provider id, metadata 
- Metadata is validated as per the schema.json set for the provider
- Validation for duplicate datasource name
- Test the connection - return an error if it fails - network error, auth error or access error
- Insert if all the above passes and return the id of the datasource created.


ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -0,0 +1,161 @@
...
+
+const (
+	// ErrMsgDatasourceNameExists is returned when a connector with the given name already exists.
+	ErrMsgDatasourceNameExists = "Datasource with name %q already exists"
+
+	// ProviderObjectStorage is the provider ID for S3-compatible object storage connectors.
+	ProviderObjectStorage = "object_storage"
+	// ProviderFileSystem is the provider ID for SSH/SFTP file system connectors.
+	ProviderFileSystem = "file_system"
+)
```
Shubhamag12, line 20: are these constants used anywhere else other than this file? If no, then we should not be exporting these, or move it to common constants if they are being used.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/file_system.go
```diff
@@ -0,0 +1,137 @@
...
+	remotePath, _ := params["remote_path"].(string)
+
+	// Port is not a required schema field; default to 22 when absent.
+	port := "22"
+	if p, ok := params["port"].(string); ok && p != "" {
+		port = p
+	}
+
+	// ── 1. Network ────────────────────────────────────────────────────────────
+	addr := net.JoinHostPort(host, port)
```
Shubhamag12: this will panic if somehow host is empty from line 41.
we should be checking for empty fields from params imo.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/file_system.go
```diff
@@ -0,0 +1,137 @@
...
+// automatically cancelled when the deadline expires.
+func (t *fileSystemTester) TestConnection(ctx context.Context, params map[string]any) error {
+	host, _ := params["host"].(string)
+	username, _ := params["username"].(string)
+	privateKeyPEM, _ := params["private_key"].(string)
+	remotePath, _ := params["remote_path"].(string)
+
+	// Port is not a required schema field; default to 22 when absent.
+	port := "22"
+	if p, ok := params["port"].(string); ok && p != "" {
```
Shubhamag12, line 41: I dont see port in schema. Do we want to include it there?

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/file_system.go
```diff
@@ -0,0 +1,137 @@
...
+	info, err := sftpClient.Stat(remotePath)
+	if err != nil {
+		return &ConnectionCheckError{
+			CheckType: ConnectionCheckAccess,
+			Message:   fmt.Sprintf("sftp.Stat(%q) failed: %v", remotePath, err),
+		}
+	}
+
+	if info.IsDir() {
+		if _, err := sftpClient.ReadDir(remotePath); err != nil {
```
Shubhamag12: should we be reading entire dir? is there something like open supported?

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/file_system.go
prachikurade1: Port is not included in the schema as it is not there in the UI fields.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/file_system.go
prachikurade1: We are reading the entire directory since we need to check access to that particular directory.

Shubhamag12: my only concern is what if that directory has a lot of files

prachikurade1: Let me check if there is any other quicker mechanism.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
prachikurade1: They are not used currently outside currently, so have updated them now.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/file_system.go
prachikurade1: I checked this, stat is enough to check the access permission. ReadDir is not required. So have changed the check to just check stat and if it is a directory.

ai-services/internal/pkg/catalog/apiserver/models/create_datasource.go
```diff
@@ -0,0 +1,21 @@
...
+
+// CreateDatasourceRequest is the request body for creating a new datasource connector.
+type CreateDatasourceRequest struct {
+	// Name is the unique human-readable label for this connector.
+	Name string `json:"name" binding:"required,min=3,max=100"`
+	// ProviderID identifies the provider implementation (e.g. "object_storage", "file_system").
+	ProviderID string `json:"provider_id" binding:"required"`
+	// Metadata holds the provider-specific configuration. Sensitive fields (format: "password"
+	// in the JSON schema) are encrypted at rest; all other fields are stored in plain text.
+	Metadata map[string]any `json:"metadata" binding:"required"`
```
mayuka-c: Lets make this params similar to CreateApplication

ai-services/internal/pkg/catalog/apiserver/models/create_datasource.go
```diff
@@ -0,0 +1,21 @@
+package models
+
+// CreateDatasourceRequest is the request body for creating a new datasource connector.
+type CreateDatasourceRequest struct {
+	// Name is the unique human-readable label for this connector.
+	Name string `json:"name" binding:"required,min=3,max=100"`
+	// ProviderID identifies the provider implementation (e.g. "object_storage", "file_system").
```
mayuka-c, line 9: Can we also do other validations like supporting chars, case insensitive validations etc

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -0,0 +1,161 @@
...
+
+const (
+	// ErrMsgDatasourceNameExists is returned when a connector with the given name already exists.
+	ErrMsgDatasourceNameExists = "Datasource with name %q already exists"
+
+	// providerObjectStorage is the provider ID for S3-compatible object storage connectors.
+	providerObjectStorage = "object_storage"
+	// providerFileSystem is the provider ID for SSH/SFTP file system connectors.
+	providerFileSystem = "file_system"
```
mayuka-c: Let us move this to constant package as it might be used in multiple places

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -0,0 +1,161 @@
...
+	// Phase 1: validate request (provider existence + param schema).
+	if err := s.validator.ValidateCreateDatasourceRequest(ctx, req); err != nil {
+		return nil, err
+	}
+
+	// Phase 2: duplicate-name guard.
+	existing, err := s.connectorRepo.GetByName(ctx, req.Name)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check for existing connector: %w", err)
+	}
```
mayuka-c, line 78: As mentioned above let us make this case insensitive

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/file_system.go
```diff
@@ -0,0 +1,139 @@
...
+func NewFileSystemTester() ConnectionTester {
+	return &fileSystemTester{}
+}
+
+// SensitiveFields returns the credential fields that must be encrypted at rest.
+func (t *fileSystemTester) SensitiveFields() map[string]bool {
+	return map[string]bool{
+		"private_key": true,
+	}
+}
```
mayuka-c: Let us not hardcode this. Let it be generic and be computed from values.schema.json whether its sensitive or not

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/file_system.go
```diff
@@ -0,0 +1,139 @@
...
+			CheckType: ConnectionCheckAuth,
+			Message:   "SSH authentication failed: verify the username and private key are correct for the target host",
+		}
+	}
+
+	sshClient := ssh.NewClient(sshConn, chans, reqs)
+	defer func() { _ = sshClient.Close() }()
+
+	// ── 3. Access — SFTP stat ────────────────────────────────────────────────
+	return checkSFTPAccess(sshClient, remotePath)
```
mayuka-c, line 94: Is this for a permission check?

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/interface.go
```diff
@@ -0,0 +1,49 @@
...
+// CreateDatasource flow (validation, encryption, DB insert) lives in DatasourceService
+// and delegates to this interface for the parts that vary.
+type ConnectionTester interface {
+	// TestConnection runs provider-specific connectivity checks (network → auth → access).
+	// Returns nil when all checks pass, or a *ConnectionCheckError on the first failure.
+	TestConnection(ctx context.Context, params map[string]any) error
+
+	// SensitiveFields returns the set of param keys whose values are credentials and
+	// must be encrypted before being stored in the database.
+	SensitiveFields() map[string]bool
```
mayuka-c: Better would be to have a different interface for this as this is not part of ConnectionTester Ideally

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/object_storage.go
```diff
@@ -0,0 +1,181 @@
...
+		HTTPClient:  awshttp.NewBuildableClient().WithTimeout(s3ConnectTimeout),
+	}
+
+	client := s3.NewFromConfig(cfg, func(o *s3.Options) {
+		if !isAWS {
+			// IBM COS, MinIO, and other S3-compatible stores: supply the custom
+			// endpoint and enable path-style addressing.
+			o.BaseEndpoint = aws.String(endpointURL)
+			o.UsePathStyle = true
+		}
```
mayuka-c: Let us make this inverse,
For AWS do the `if` check as it is specific and rest flow can be generic

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/object_storage.go
```diff
@@ -0,0 +1,181 @@
...
+	"fmt"
+	"net"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
+	"github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/smithy-go"
```
mayuka-c, line 16: Is there not a common package which could provide common s3 methods?
or else we need to use aws-sdk is it?

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/object_storage.go
```diff
@@ -0,0 +1,181 @@
...
+// regionFromEndpoint extracts the SigV4 region from the endpoint URL.
+// Falls back to "us-east-1" when the pattern is not recognised.
+//
+// Supported patterns:
+//
+//	s3.<region>.amazonaws.com                              (AWS S3)
+//	s3.<region>.cloud-object-storage.appdomain.cloud      (IBM COS direct-region)
+//	s3.us.cloud-object-storage.appdomain.cloud            (IBM COS cross-region alias → us-south)
+func regionFromEndpoint(endpointURL string) string {
+	cosAliases := map[string]string{"us": "us-south", "eu": "eu-de", "ap": "jp-tok"}
```
mayuka-c: Again here is there not a generic way?
What if a region is removed/modified/deleted?

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service.go
```diff
@@ -0,0 +1,25 @@
...
+// NewDatasourceService creates the DatasourceService wired with all known provider testers.
+// The DB_ENCRYPTION_KEY environment variable is read once at startup and injected into the
+// service so that business logic is decoupled from the environment at call time.
+func NewDatasourceService(
+	connectorRepo dbrepo.ConnectorRepository,
+	provider *catalog.CatalogProvider,
+) DatasourceServiceInterface {
+	validator := validators.NewConnectorValidator(provider)
+	encryptionKey := os.Getenv("DB_ENCRYPTION_KEY")
+
```
mayuka-c: Let us have constant for DB_ENCRYPTION_KEY in constants pkg

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/object_storage.go
```diff
@@ -0,0 +1,181 @@
...
+// single ListObjectsV2(MaxKeys=0) call — a zero-cost probe that transfers no object data.
+// The SDK response (or error) is classified into network / auth / access failure categories.
+func (t *objectStorageTester) TestConnection(ctx context.Context, params map[string]any) error {
+	endpointURL, _ := params["endpoint_url"].(string)
+	bucket, _ := params["bucket_name"].(string)
+	accessKeyID, _ := params["access_key_id"].(string)
+	secretKey, _ := params["secret_access_key"].(string)
+	prefix, _ := params["prefix"].(string)
+
+	isAWS := strings.Contains(endpointURL, awsEndpointSuffix)
```
mayuka-c: What if user wants to use GCS or ABS, will this work?
Do we need to again check and add specific logics?

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -0,0 +1,161 @@
...
+	if encryptionKey == "" {
+		return nil, fmt.Errorf("encryption key is not configured (DB_ENCRYPTION_KEY must be set)")
+	}
+
+	result := make(map[string]any, len(params))
+	for k, v := range params {
+		if sensitiveKeys[k] {
+			plaintext, ok := v.(string)
+			if !ok {
+				return nil, fmt.Errorf("sensitive field %q must be a string", k)
```
yussufsh: This should be StatusBadRequest instead of server error (500).

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service.go
yussufsh: Also, there should be a check for an empty string. I see you are checking during runtime, which isn't ideal. It should fail fast.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/object_storage.go
yussufsh: Also I see default as "us-east-1". Was this discussed with the design team?

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/file_system.go
prachikurade1: Yes, this does a permission check.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/object_storage.go
prachikurade1: For GCS it will work, will need minor tweaks to handle error codes for access/auth errors.
For Azure , it will not work because Azure Blob storage is not s3 compatible.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/object_storage.go
prachikurade1: This is the way even Digitize Service has implemented. A mapping needs to be maintained and updated if regions are changed.
Digitize mapping - 
https://github.com/IBM/project-ai-services/blob/main/services/digitize/connectors/scanners/config.py#L48

"us-east-1" is the standard default region used for AWS. Checked Digitize implementation as well and they have done the same. 
Default region - https://github.com/IBM/project-ai-services/blob/main/services/digitize/connectors/scanners/config.py#L36

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/object_storage.go
prachikurade1: This package is compatible with s3 based storage systems and is the most recommended one.

ai-services/internal/pkg/catalog/apiserver/models/create_datasource.go
prachikurade1: Done

ai-services/internal/pkg/catalog/apiserver/models/create_datasource.go
prachikurade1: Done

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
prachikurade1: Done

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/file_system.go
prachikurade1: Done

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/object_storage.go
prachikurade1: Done

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service.go
prachikurade1: Done

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/interface.go
prachikurade1: Removed as not needed after making this dynamic based on schema.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
prachikurade1: image

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/object_storage.go
mayuka-c: So currently we support AWS and IBM COS is it?
Are we telling user that only those 2 is supported anywhere?

prachikurade1: No, we are supporting any S3 compatible service. GCS is partially S3 compatible and Azure is not.

mayuka-c: Is it been tested out with all S3 like local and cloud based?. I still feel only AWS and IBM COS are only supported based on looking the code.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/object_storage.go
```diff
@@ -0,0 +1,192 @@
...
+		CheckType: ConnectionCheckNetwork,
+		Message:   fmt.Sprintf("S3 connectivity error for %q: %v", endpointURL, err),
+	}
+}
+
+// regionFromEndpoint extracts the SigV4 region from an S3-compatible endpoint URL using
+// the same regex-based approach as the digitize Python service (_REGION_FROM_URL_RE in config.py).
+//
+// It matches AWS S3 and IBM COS hostname patterns and resolves IBM COS cross-region aliases
+// to their canonical SigV4 region values. Falls back to s3DefaultRegion ("us-east-1") for
```
mayuka-c, line 175: I was searching online and this is what it recommended. They mention use defaults based on each providers: 

# 1. Maintain a map of strict provider rules
PROVIDER_REGION_DEFAULTS = {
    "cloudflare": "auto",
    "r2": "auto",
    "gcs": "auto",
    "google": "auto",
    "localstack": "us-east-1",
    "minio": "us-east-1"
}

I'm not much sure if this is correct. PTAL

mayuka-c (dismissed):
LGTM,
Have few concerns on making generic for object storage, but rest is good

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/object_storage.go
prachikurade1: Tried to check more on this:
Why the provider-name map is the wrong abstraction for our scenario:
The recommended approach tries to detect the provider identity from the endpoint URL string (e.g. "cloudflare", "minio", "r2"). This has three problems for our codebase:

1. We do not have a provider name — We would have to infer the provider from the URL string, which is fragile (a MinIO instance at storage.mycompany.com would fall through every check).

2. The map adds a maintenance burden for zero gain.

3. Every new S3-compatible store a user points at (Wasabi, Backblaze B2, Ceph, Dell ECS, NetApp StorageGRID…) would need an entry. And for all of them, "us-east-1" already works perfectly.

The fallback "us-east-1" for non-AWS, non-IBM stores is not a guess — it is the correct AWS SDK v2 convention for custom endpoints. The SDK itself defaults to "us-east-1" when no region is configured. The region value is included in the SigV4 Authorization header but these stores do not enforce it.
The only case where region matters precisely is AWS S3 (extracted from the URL) and IBM COS (extracted + alias-resolved) — both of which are already handled. For everything else "us-east-1" is the correct no-op default and matches what the AWS SDK itself would use if you didn't set the field at all.

mayuka-c: If it works across, then fine :)

mayuka-c (approved):

PR #1297, opened 2026-08-21
upgrade vllm-spyre image to 3.5.0-ea.2 and increase memory to 200Gi

Shubhamag12:


mayuka-c (changes requested):
Need to bump the Catalog backend image for it to be able to pick and deploy latest assets

mkumatag (changes requested):
avoid caddy and postgres change part of this PR for now, let us look at only the ai-services binary and image only..

mkumatag (approved):

PR #1298, opened 2026-08-21
Fix datasource params order for UI

prachikurade1:
Re-ordered the params as per the UI requirements.


ai-services/assets/connectors/datasource/file_system/schema.json
```diff
@@ -16,6 +16,12 @@
       "description": "Hostname or IP address of the remote file system server.",
       "ui:section": "Location"
     },
+    "remote_path": {
+      "type": "string",
+      "title": "Remote folder path",
+      "description": "Absolute path on the remote file system server to use as the content root (e.g. /data/documents).",
+      "ui:section": "Location"
```
mayuka-c, line 23: Is this a standard practice. I hope you have checked online and from multiple sources

prachikurade1: Yes, I had checked online and this was one of the recommended method.

prachikurade1: I  also confirmed with @pranithraoibm if they are ok with this before the first PR.

yussufsh (dismissed):

mayuka-c (approved):
LGTM

PR #1299, opened 2026-08-21
chore: reduce debug log noise across digitize services

dharaneeshvrd:
Remove or consolidate low-value debug/info log statements that were generating excessive noise in normal operation:

- db/manager.py: drop per-operation debug logs for create/get/update job and document rows; logging at the caller boundary is sufficient
- utils/db.py: remove redundant "Created job/doc" info logs and the per-item "Updated document" debug log; collapse update_job result check that spuriously warned on no-op updates
- utils/jobs.py: remove per-document-id debug log; emit a single summary info log after all documents are created for the job
- utils/storage.py: replace per-file "Staged file" debug with a single "Staged N file(s)" summary after the loop
- parsing/docx.py: strip trace-level debug logs from internal table-caption helpers (_parse_ref_index, _get_body_children_refs, _get_text_value_by_ref, _looks_like_table_caption, _find_matching_caption_near_refs, _get_enclosing_section_header); condense surviving logs in recover_table_caption_from_body_context to compact "ref -> result" format
- parsing/converter.py: drop docling-models-path debug log
- common/opensearch.py: drop init banner and redundant shard-count log
- processing/orchestrator.py: add doc_path to language-detection info log; shorten conversion-submission debug to filename only

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.260
+TAG?=v0.0.261
 CONTAINER_BUILDER?=podman
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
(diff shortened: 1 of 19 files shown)


sats-23 (dismissed):
LGTM

manju956 (dismissed):
lgtm

sats-23 (approved):

PR #1300, opened 2026-08-21
feat(digitize): add cancel endpoint and cooperative pipeline cancellation

dharaneeshvrd:
Introduces a new POST /{job_id}/cancel endpoint and the supporting pipeline machinery required to cleanly stop an in-flight digitize or ingest job at the next safe checkpoint.

--- API & status model ---
* New JobStatus values: CANCEL_PENDING (set immediately by the endpoint) and CANCELLED (set by the background task after draining).
* New DocStatus value: CANCELLED for non-terminal documents at job cancellation time.
* DB check constraints and init_schema.sql updated to accept the new status strings.
* POST /{job_id}/cancel (202 Accepted):
  - Returns 404 if the job does not exist.
  - Returns 409 if the job is already in a terminal/pending-cancel state.
  - Persists CANCEL_PENDING + optional clean_files flag in job stats.
* _run_digitize / _run_ingest catch JobCancelledError and mark all non-terminal documents and the job itself as CANCELLED. Ingest additionally removes already-indexed VDB chunks when clean_files=true.

--- Pipeline checkpoints ---
* New digitize/exceptions.py: JobCancelledError — a clean signal that propagates up the call stack without being treated as an error.
* db_manager.is_job_cancelled(): lightweight single-column SELECT used at every pipeline checkpoint without loading the full job row.
* ingest.py: pre-pipeline CHECK 1 (before process_documents) and post-pipeline CHECK 5 (before writing final status).
* digitize.py: pre-flight check before the ProcessPoolExecutor starts.
* orchestrator.process_documents: all four as_completed loops replaced with cancellation-aware while-pending polling loops:
  - Conversion stage: pending futures cancelled; running ones drained; cancel_event (multiprocessing.Manager proxy) propagated to worker processes so long multi-chunk conversions abort between page chunks.
  - Processing stage: pending futures cancelled; running ones drained; _process_stop_event (threading.Event) propagated to process_converted_document / process_table so in-flight LLM table calls are aborted after the current call completes.
  - Chunking stage: same pattern.
  - Indexing stage: pending futures cancelled; already-running index futures are let to complete to avoid leaving stale VDB entries.
  - is_cancelled flag is latched True across stages; if all pending work has been cancelled, JobCancelledError is raised immediately to skip remaining stages.

--- LLM / processing layer ---
* summarize_and_classify_tables accepts stop_event; checked after each individual table LLM call; queued futures cancelled on signal.
* process_table / process_converted_document: stop_event threaded through so table summarization can be cut short mid-document.
* convert_doc / convert_document accept cancel_event; checked between 100-page chunks so a multi-chunk conversion can be aborted early without any DB access from the worker process.

--- Correctness fixes ---
* DatabaseStatusManager.update_job_progress: protected-status guard prevents any IN_PROGRESS update from overwriting cancel_pending, cancelled, completed, or failed.
* Extra job-stat keys (e.g. clean_files) are preserved during stats merge so the flag is readable by the background cleanup code.
* completed_at is now stamped for CANCELLED jobs as well as COMPLETED/FAILED ones.
* recover_zombie_jobs: CANCEL_PENDING added to orphan statuses so interrupted cancel operations are recovered on restart.

--- Tests ---
* services/digitize/exceptions.py: new module.
* services/digitize/tests/test_cancel_job.py: 1 344-line test suite covering the cancel endpoint (404/409/202/500), _run_digitize and _run_ingest cancellation paths (including VDB cleanup), and all four orchestrator pipeline stages with cancellation scenarios.
* services/digitize/tests/test_types_models.py: extended for the two new JobStatus and one new DocStatus values.

### DIFF
--- a/services/common/llm_utils.py
+++ b/services/common/llm_utils.py
@@ -1,5 +1,6 @@
 import os
 import requests
+import threading
 import time
 import json
 from concurrent.futures import ThreadPoolExecutor, as_completed
@@ -108,12 +109,23 @@ def summarize_and_classify_single_table(prompt, gen_model, llm_endpoint, max_tok
         logger.error(f"Error summarizing/classifying table: {e}")
         raise
 
-def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, doc_path, prompt_template: str, max_tokens: int = 1024, max_workers=32):
+def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, doc_path, prompt_template: str, max_tokens: int = 1024, max_workers=32, stop_event: threading.Event | None = None):
     """Combined function to summarize and classify tables using a single prompt.
 
-    Returns tuple: (summaries, decisions, failures).
-    failures is a dict mapping table index to error message for failed tables.
-    Table failures are non-fatal: failed tables use fallback values and processing continues.
+    Args:
+        table_mds: List of table markdown strings.
+        gen_model: LLM model name.
+        llm_endpoint: LLM endpoint URL.
+        doc_path: Document path (used for logging).
+        prompt_template: Prompt template string with a {content} placeholder.
+        max_tokens: Maximum tokens for the LLM response.
+        max_workers: Maximum parallel LLM workers.
+        stop_event: Optional threading.Event. When set, the function stops
+                    collecting further LLM results between table calls and
+                    cancels any queued futures that have not yet started.
+
+    Returns tuple: (summaries, decisions)
+    Raises JobCancelledError if stop_event is set mid-processing.
     """
     all_prompts = [prompt_template.format(content=md) for md in table_mds]
 
@@ -131,6 +143,16 @@ def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, doc_path, 
             for idx, prompt in enumerate(all_prompts)
         }
         for future in as_completed(futures):
+            # Check for job cancellation after each table's LLM call completes.
+            if stop_event is not None and stop_event.is_set():
+                # Cancel any futures that haven't started yet.
+                for pending_fut in futures:
+                    if not pending_fut.running() and not pending_fut.done():
+                        pending_fut.cancel()
+                from digitize.exceptions import JobCancelledError
+                raise JobCancelledError(
+                    f"Job cancelled during table summarization for '{doc_path}'"
+                )
             idx = futures[future]
             try:
                 results[idx] = future.result()
--- a/services/common/llm_utils.py
+++ b/services/common/llm_utils.py
@@ -3,7 +3,7 @@ import requests
 import threading
 import time
 import json
-from concurrent.futures import ThreadPoolExecutor, as_completed
+from concurrent.futures import ThreadPoolExecutor
 from common.misc_utils import get_logger, resolve_model_max_len
 from common.settings import settings
 from common.retry_utils import retry_on_transient_error
@@ -142,26 +142,33 @@ def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, doc_path, 
             executor.submit(summarize_and_classify_single_table, prompt, gen_model, llm_endpoint, max_tokens): idx
             for idx, prompt in enumerate(all_prompts)
         }
-        for future in as_completed(futures):
-            # Check for job cancellation after each table's LLM call completes.
-            if stop_event is not None and stop_event.is_set():
-                # Cancel any futures that haven't started yet.
-                for pending_fut in futures:
-                    if not pending_fut.running() and not pending_fut.done():
-                        pending_fut.cancel()
+        pending = dict(futures)  # future -> idx, shrinks as futures complete
+        while pending:
+            for fut in list(pending):
+                if not fut.done():
+                    # Cancel queued (not-yet-running) futures if stop_event is set.
+                    if stop_event is not None and stop_event.is_set() and not fut.running():
+                        fut.cancel()
+                        del pending[fut]
+                    continue
+                idx = pending.pop(fut)
+                try:
+                    results[idx] = fut.result()
+                    logger.debug(f"Summarized table {idx + 1}/{len(all_prompts)} from '{doc_path}'")
+                except Exception as e:
+                    error_msg = f"Failed to process table {idx}: {str(e)}"
+                    logger.error(error_msg)
+                    results[idx] = ("No summary.", False)
+                    failures[idx] = error_msg
+
+            if stop_event is not None and stop_event.is_set() and not pending:
                 from digitize.exceptions import JobCancelledError
                 raise JobCancelledError(
                     f"Job cancelled during table summarization for '{doc_path}'"
                 )
-            idx = futures[future]
-            try:
-                results[idx] = future.result()
-                logger.debug(f"Summarized table {idx + 1}/{len(all_prompts)} from '{doc_path}'")
-            except Exception as e:
-                error_msg = f"Failed to process table {idx}: {str(e)}"
-                logger.error(error_msg)
-                results[idx] = ("No summary.", False)
-                failures[idx] = error_msg
+
+            if pending:
+                time.sleep(0.05)
 
     # Unpack results in index order. Every slot starts as the fallback
     # and is overwritten either by a successful future or left as-is on failure.
--- a/services/common/llm_utils.py
+++ b/services/common/llm_utils.py
@@ -168,7 +168,7 @@ def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, doc_path, 
                 )
 
             if pending:
-                time.sleep(0.05)
+                time.sleep(0.5)
 
     # Unpack results in index order. Every slot starts as the fallback
     # and is overwritten either by a successful future or left as-is on failure.
--- a/services/common/llm_utils.py
+++ b/services/common/llm_utils.py
@@ -124,7 +124,12 @@ def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, doc_path, 
                     collecting further LLM results between table calls and
                     cancels any queued futures that have not yet started.
 
-    Returns tuple: (summaries, decisions)
+    Returns tuple: (summaries, decisions, failures).
+        summaries:  List of summary strings in table-index order.
+        decisions:  List of booleans (keep/discard) in table-index order.
+        failures:   Dict mapping table index to error message for tables that
+                    could not be processed; those tables use fallback values
+                    and processing continues.
     Raises JobCancelledError if stop_event is set mid-processing.
     """
     all_prompts = [prompt_template.format(content=md) for md in table_mds]
@@ -167,6 +172,9 @@ def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, doc_path, 
                     f"Job cancelled during table summarization for '{doc_path}'"
                 )
 
+            # Sleep once per outer-loop cycle so the loop never spins when all
+            # remaining futures are still running.  Skipped only when pending is
+            # empty (the while condition will end the loop on the next check).
             if pending:
                 time.sleep(0.5)
 
--- a/services/common/opensearch.py
+++ b/services/common/opensearch.py
@@ -159,7 +159,7 @@ class OpensearchVectorStore(VectorStore):
             raise
 
     @retry_on_transient_error(max_retries=3, initial_delay=5.0, backoff_multiplier=2.0)
-    def insert_chunks(self, chunks, vectors=None, embedding=None, batch_size=10):
+    def insert_chunks(self, chunks, vectors=None, embedding=None, batch_size=10, cancel_event=None):
         """Supports 2 modes of insertion with retry logic for transient failures.
 
         1. Pure embedding: pass 'chunks' and 'vectors'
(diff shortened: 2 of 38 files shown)


services/common/llm_utils.py
```diff
@@ -131,10 +145,26 @@ def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, doc_path,
...
+                results[idx] = fut.result()
+
+            if stop_event is not None and stop_event.is_set() and not pending:
+                from digitize.exceptions import JobCancelledError
+                raise JobCancelledError(
+                    f"Job cancelled during table summarization for '{doc_path}'"
+                )
+
+            if pending:
+                time.sleep(0.05)
```
manju956: 0.05 sec looks very small interval. intentional?

services/digitize/api/v1/jobs.py
```diff
@@ -40,14 +42,30 @@ async def _run_digitize(
...
+                models.DocStatus.COMPLETED.value,
+                models.DocStatus.FAILED.value,
+                models.DocStatus.CANCELLED.value,
+                models.DocStatus.ALREADY_EXISTS.value,
+            ):
+                status_mgr.update_doc_metadata(
+                    doc.doc_id,
+                    {"status": models.DocStatus.CANCELLED, "completed_at": datetime.now(timezone.utc).isoformat()},
+                )
+        status_mgr.update_job_progress("", models.DocStatus.CANCELLED, models.JobStatus.CANCELLED)
```
manju956: we can have a method to perform the cancellation status updates.

services/digitize/exceptions.py
```diff
@@ -0,0 +1,13 @@
+"""
```
sats-23, line 1: We should probably aggregate all our custom exceptions in this file later. 
Currently we have them scattered around digitize.py and connectors.py

services/digitize/workers/conversion_dispatcher.py
```diff
@@ -128,18 +162,41 @@ async def _run_conversion(task: ConversionTask, weight: int) -> None:
...
+        # Check 2: if the job was cancelled while we were converting, write
+        # CANCELLED so the pipeline's poll loop reaches the correct terminal state.
+        fresh = db_manager.get_conversion_task(task.task_id)
+        if fresh is not None and fresh.status == ConversionTaskStatus.CANCEL_PENDING:
+            db_manager.update_task_status(
+                task.task_id, ConversionTaskStatus.CANCELLED,
+                error="Job cancelled during conversion",
+            )
+            logger.info(f"Task {task.task_id} cancelled after conversion completed")
+            return
```
sats-23: Can we extract this into a separate method? It is called thrice in same class

services/digitize/utils/recovery.py
sats-23, line 56: Can we modify the SQL query to make one single query instead of 3?
Like accept a list of JobStatus and use IN query.

sats-23:
Can we add connector support in the same PR?
I believe it is just one extra line to be added in sync_tick#_wait_for_job

dharaneeshvrd:
> Can we add connector support in the same PR?
I believe it is just one extra line to be added in sync_tick#_wait_for_job

Yeah, let me add if it is a simple change.

services/common/llm_utils.py
dharaneeshvrd: I wanted to use 0.5 only, bob changed during refactoring maybe

services/digitize/api/v1/jobs.py
```diff
@@ -70,14 +88,56 @@ async def _run_ingest(
...
+
+        # Sub-Task 6: clean vector DB if requested
+        try:
+            job_row = db_manager.get_job_by_id(job_id)
+            if job_row and job_row.stats.get("clean_files"):
+                import common.db_utils as db_utils
+                indexed_statuses = {
```
manalilatkar: Where is this variable `indexed_statuses` being used?

services/digitize/api/v1/jobs.py
```diff
@@ -473,3 +533,78 @@ async def delete_job(job_id: str):
...
+        "the background pipeline will observe this at its next checkpoint, stop, "
+        "and then mark the job as 'cancelled'. "
+        "Pass clean_files=true to also delete any vector-DB chunks already indexed "
+        "for this job's documents (ingestion jobs only)."
+    ),
+    response_description="Cancellation accepted",
+)
+async def cancel_job(
+    job_id: str,
+    clean_files: bool = Query(False, description="Delete already-indexed vector DB chunks for this job"),
```
manalilatkar, line 424: What kind of usecase would it be where user wants to stop ingestion but not clean the files already ingested from this batch? Shouldn't this be a default choice instead of a user driven flag?

services/digitize/db/manager.py
```diff
@@ -1953,6 +1990,50 @@ def update_task_status(
...
+        ]
+        try:
+            with get_db_session() as session:
+                stmt = (
+                    update(ConversionTask)
+                    .where(
+                        ConversionTask.job_id == job_id,
+                        ConversionTask.status.in_(active_statuses),
+                    )
+                    .values(status=ConversionTaskStatus.CANCEL_PENDING)
```
manalilatkar, line 2042: This code flips PENDING / QUEUED / RUNNING → `CANCEL_PENDING`, but `claim_head`, `peek_head` (manager.py) and `promote_pending` all select strictly on `QUEUED` / `PENDING`. A task moved to `cancel_pending` while queued is never picked up again, so the row is stuck at `cancel_pending` permanently.
So, for PENDING/QUEUED rows, we should directly write CANCELLED so they never get picked up.

services/digitize/processing/orchestrator.py
```diff
@@ -171,6 +176,9 @@ def chunk_text(input_path, out_path, emb_endpoint, max_tokens=512, doc_id=None,
             current_subsubsection = None
 
             for idx, block in enumerate(data):
+                if cancel_event is not None and cancel_event.is_set():
+                    logger.info(f"Cancellation requested — aborting text chunking for '{input_path}'")
+                    raise JobCancelledError("Text chunking cancelled")
```
manalilatkar: This JobCancelledError will be caught by the except block on line no 242 and never reach the except block on line no 333 where its supposed to unless you reraise it.

services/digitize/api/v1/jobs.py
```diff
@@ -70,14 +88,56 @@ async def _run_ingest(
...
+                    models.DocStatus.COMPLETED.value,
+                }
+                # Re-fetch docs after marking cancelled to get accurate statuses
+                all_docs = db_manager.get_documents_by_job_id(job_id)
+                # Use original doc_id_dict values since docs are now CANCELLED;
+                # any doc that was CHUNKED or COMPLETED before cancellation may have vector data
+                doc_ids_to_clean = [
+                    d.doc_id for d in all_docs
+                    if d.doc_id in doc_id_dict.values()
+                ]
```
manalilatkar: Documents that finished indexing *before* the cancel arrived keep `status=completed` and their `file_hash` in `document_checksum`, but lose their chunks. They become unsearchable, and re-uploading the same file returns `already_exists` — there is no way to recover them.

`indexed_statuses` is built two lines above and never used, which seems like the intended filter was dropped during editing.

services/digitize/exceptions.py
dharaneeshvrd: Let's do it in a future PR

services/common/opensearch.py
```diff
@@ -200,6 +203,15 @@ def insert_chunks(self, chunks, vectors=None, embedding=None, batch_size=10):
...
         for i in tqdm(range(0, len(chunks), batch_size)):
+            # Check for cancellation before each batch. cancel_event is only set
+            # when clean_files=True, so any partial inserts will be removed by the
+            # post-cancellation VDB cleanup.
+            if cancel_event is not None and cancel_event.is_set():
+                logger.info(
+                    f"Cancellation requested — aborting insert_chunks after {i} of "
+                    f"{len(chunks)} chunks for index '{self.index_name}'"
+                )
+                return False
```
manalilatkar: Why not raise JobCancelled Error instead of returning False? The caller `index_document_chunks` treats `False` as failure: doc -> `FAILED` with error `"Failed to index document chunks into vector database"`. That status is terminal, so `_run_ingest` won't relabel it, and the user sees a fabricated infrastructure error in response to their own cancel request.

services/digitize/utils/recovery.py
manalilatkar: For CANCEL_PENDING jobs, we should be updating the status as CANCELLED, not FAILED.

services/digitize/api/v1/jobs.py
dharaneeshvrd: Just wanted to give this flexibility to the user, so if they want to they can retain the completed docs.
Maybe out of 10 docs triggered, 5 might be completed and remaining still in progress, if they want to cancel because of those in progress docs, this would be helpful.

services/digitize/db/manager.py
dharaneeshvrd: Nice catch, fixing it.

services/digitize/utils/recovery.py
dharaneeshvrd: Handled this.

services/common/opensearch.py
dharaneeshvrd: You are right, handled this properly by propagating the job cancelled error to its main orchestrator.

services/digitize/api/v1/jobs.py
dharaneeshvrd: yeah missed this in the editing.
Handled this now.

services/digitize/processing/orchestrator.py
dharaneeshvrd: Removed this logic, dont want to check in between each chunk, now checking only between chunking text and table.

services/digitize/api/v1/jobs.py
dharaneeshvrd: Corrected the usage now.

services/digitize/utils/recovery.py
dharaneeshvrd: Let me add this in a follow up PR.

services/digitize/workers/conversion_dispatcher.py
dharaneeshvrd: Agree, taken care.

services/digitize/api/v1/jobs.py
dharaneeshvrd: Added a method and reused it

services/common/llm_utils.py
dharaneeshvrd: Changed to 0.5

services/digitize/connectors/sync_tick.py
```diff
@@ -515,15 +527,6 @@ async def _handle_interrupt(
...
         _cancel_tick(sync_seq, connector_id)
-        # Purge conversion tasks the tick may have enqueued before being interrupted.
-        # Must happen here (Case A only) — _run_teardown is shared with Case B where
-        # no tick was running so no in-flight tasks exist to purge.
-        from digitize.db.manager import db_manager
-        deleted_tasks = db_manager.delete_conversion_tasks_for_connector(connector_id)
-        if deleted_tasks:
-            logger.info(
-                f"Purged {deleted_tasks} queued conversion task(s) for connector {connector_id!r}"
-            )
```
dharaneeshvrd, line 526: @manju956 no need to purge conversion tasks, since job_cancellation will take care of queued tasks cleanup

manju956: sure @dharaneeshvrd , this is good.

dharaneeshvrd:
@manalilatkar @manju956 Can you please review this PR? Addressed all the comments

services/digitize/utils/db.py
```diff
@@ -1131,6 +1131,27 @@ def _update_job(
...
+        # (e.g. CANCELLED written by _run_ingest/_run_digitize) with a
+        # superseding status coming from a late update_job_progress call.
+        # CANCEL_PENDING is also protected: the job has been requested for
+        # cancellation and must not be flipped back to IN_PROGRESS by any
+        # in-flight pipeline step that hasn't yet observed the cancellation flag.
+        protected_statuses = {
+            JobStatus.COMPLETED,
+            JobStatus.FAILED,
+            JobStatus.CANCELLED,
+            JobStatus.CANCEL_PENDING,
```
manalilatkar, line 1144: Shouldn't this list contain COMPLETED_WITH_ERRORS as well?

services/digitize/utils/jobs.py
```diff
@@ -275,6 +278,59 @@ async def enqueue_conversion_tasks(
...
+    models.DocStatus.FAILED.value,
+    models.DocStatus.CANCELLED.value,
+    models.DocStatus.ALREADY_EXISTS.value,
+)
+
+NON_CANCELLABLE_JOB_STATUSES = (
+    models.JobStatus.COMPLETED.value,
+    models.JobStatus.FAILED.value,
+    models.JobStatus.CANCEL_PENDING.value,
+    models.JobStatus.CANCELLED.value,
```
manalilatkar, line 293: COMPLETED_WITH_ERRORS missing from here as well?

services/digitize/exceptions.py
manalilatkar: @dharaneeshvrd do you want to create a task to track this?

services/digitize/utils/jobs.py
dharaneeshvrd: Added.

services/digitize/utils/db.py
dharaneeshvrd: Added

services/digitize/exceptions.py
dharaneeshvrd: Created a JIRA story for this and the one related to improving recovery.

manju956 (dismissed):

manju956 (approved):

PR #1301, opened 2026-08-21
[BYOS] Reload Catalog Items and integrate with APIs

mayuka-c:
- This is required to reload the Catalog Items cache which holds both embedded and custom assets info.
- Changes to ServiceParams, ComponentParams, ListServices to work with custom assets.

Note:
- There is nil assignment in cmd package (`catalog.NewCatalogProvider(nil)`) which would be fixed when we move to using APIs as CLI cannot directly access the assets present inside volumes (Will be taken up later)

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/services/bundle/service_test.go
+++ b/ai-services/internal/pkg/catalog/apiserver/services/bundle/service_test.go
@@ -475,10 +475,10 @@ func TestReplaceBundle_HappyPath(t *testing.T) {
 	require.Error(t, err) // expected: extraction fails without real storage root
 }
 
-// TestReplaceBundle_ActivateFailureMarksRowFailed uses the internal
+// TestActivateFailureMarksRowFailed uses the internal
 // replaceBundleFiles helper to drive the post-processing path directly,
 // using a real temp dir so that extraction succeeds.
-func TestReplaceBundle_ActivateFailureMarksRowFailed(t *testing.T) {
+func TestActivateFailureMarksRowFailed(t *testing.T) {
 	fixedID := uuid.MustParse("550e8400-e29b-41d4-a716-446655440000")
 	tmp := t.TempDir()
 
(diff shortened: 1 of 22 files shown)


ai-services/internal/pkg/catalog/apiserver/services/bundle/service_test.go
```diff
@@ -1335,3 +1336,336 @@ func TestDeleteBundle_NoRunningInstances_Proceeds(t *testing.T) {
...
+		repo:            repo,
+		catalogReloader: failsOnReload(reloadErr),
+	}
+
+	// Drive the reload + markFailed path directly (extraction would fail against
+	// bundleStorageRoot, so we replicate just the post-insert steps).
+	ctx := context.Background()
+
+	if err := svc.catalogReloader.Reload(ctx); err != nil {
+		svc.markFailed(ctx, fixedID, err.Error())
```
yussufsh, line 1480: Why is it marking it as failed in the test?

mayuka-c: This is just a DB mocking. svc.markFailed calls the method defined at line 1463 which does the update of capturedFailUpdate Status to failed

yussufsh: I am confused. On looking closely, I think this should have been handled in the service itself.
`markFailed` is getting called in ProcessBundle() but not getting called in ReplaceBundle() and DeleteBundle(). 
Don't we need to update the row.ID during the replace flow?

mayuka-c: If you look carefully on these tests methods (Reload tests). This is not calling ProcessBundle.
These tests are for testing Reload method and it is doing mocking on the error types and checking if its able to assert the error.

yussufsh: Ok please rename the test name. Otherwise lgtm.

mayuka-c: Done renaming the Test method names

yussufsh (approved):

PR #1305, opened 2026-08-21
refactor: switch global components of DA to model-first selection

pranithraoibm:
- `providerParamsByType` is now built once in the root component of DA and passed down to both StepOne and StepTwo, avoiding code duplication
- Global component dropdowns now show model names and users pick the model they want, the provider is resolved automatically in the background. Aligns the digital assistant flow with how the services flow already works


ui/catalog/src/components/DeployFlow/DigitalAssistant/steps/DAStepOne.tsx
```diff
@@ -0,0 +1,182 @@
...
+  onChange,
+  deployOptions,
+  providerParamsByType,
+  showNameError = false,
+  onComponentError,
+}) => {
+  const providerParamsError = useDeployStore(
+    (state) => state.providerParamsError,
+  );
+  const { getGlobalComponentModels } = useDeployStore();
```
ryanmarc: I think this line quietly takes the whole feature out.

getGlobalComponentModels is a store function, so zustand hands back the same reference forever. It's the only models-related dep on the modelNames memo, the default-seeding effect and the components memo, so none of the three recompute when setGlobalComponentModels lands. The component does re-render, since useDeployStore() with no selector subscribes to everything, but every memo returns its cached value.

Rendered it against the real store to be sure: mount empty, call setGlobalComponentModels('llm', [m1, m2]), render count goes 1 → 2, store reports ["m1","m2"], row stays hasModels: false, modelOptions: [].

The dropdown is the visible half. The half I'd worry about more is the default-seeding effect, because digitalAssistantDeploymentTransform spreads globalComponents[type].params into the deploy payload and the default model no longer gets written there. I ran the same scenario against #1279's head and it seeds {"model":"ibm-granite/granite-embedding-278m-multilingual"}; on this head embedding.params.model is never set. So a cold-open deploy goes out missing a param that used to be in it.

One thing that'll trip you up reproducing it: only happens on a cold cache. If the schemas are already cached, useDeployOptions takes its globalPairsToFetch.length === 0 branch and fills the store before StepOne first renders, and everything works. First open broken, second open fine.

Fix should just be swapping the getter for a subscription:

```const globalComponentModels = useDeployStore((state) => state.globalComponentModels);```

then `globalComponentModels[component.type] ?? EMPTY` at the three read sites, globalComponentModels in the dep arrays, and a module-level `const EMPTY: LLMOption[] = []` so the fallback doesn't churn identity. Gets you off the whole-store subscription as a bonus.

ryanmarc (dismissed):

ryanmarc (approved):

PR #1306, opened 2026-08-21
Updating digitize service workflow

adarshagrawal38:


.github/workflows/digitize-service.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     paths:
       - 'services/digitize/**'
+      - '.github/workflows/build-and-publish-image.yml'
```
mkumatag: but this will trigger the build for any change into that file, is that good?

mkumatag: instead you can do a dummy change in the digitize makefile itself to trigger the build for now.

mkumatag (changes requested):

.github/workflows/digitize-service.yml
adarshagrawal38: Mafefile for digitize-service is now modified, so this will trigger the workflow.

mkumatag (approved):

PR #1307, opened 2026-08-21
Updating namespace in release branch

adarshagrawal38:


ai-services/assets/catalog/podman/values.yaml
```diff
@@ -4,7 +4,7 @@ ui:
 
 backend:
   port: ""
-  image: icr.io/ai-services-cicd/ai-services:v0.0.159-4
+  image: icr.io/ai-services/ai-services:v0.0.159-5
```
yussufsh, line 7: Should not this be `icr.io/ai-services/ai-services:v0.0.159-4`?

adarshagrawal38: as we are updating assets file, so we have to also bump ai-services images.
from `v0.0.159-4 ` to `v0.0.159-5`

mkumatag (approved):
/lgtm

ai-services/assets/catalog/podman/values.yaml
yussufsh: But again, you will have to promote to `ai-services/ai-services:v0.0.159-5` right? Otherwise, the installation won't work.

adarshagrawal38: we have not promoted `v0.0.159-4`, we will only promote `v0.0.159-5` tag

PR #1308, opened 2026-08-23
cve-fix: Bump up pillow and cryptography python pkg.

adarshagrawal38:
Bum up pillow to 12.3.0 and cryptography to 50.0.0 to address high cves in pkgs


adarshagrawal38:
Successfully able to parse PDF document with docling - 2.120.3
image

images/service-base/download_docling_models.py
```diff
@@ -23,3 +23,4 @@
 )
 
 print("Download complete.")
+os._exit(0)
```
yussufsh: Why was this needed? If the previous process is getting hung, we need to figure out what is going on.

adarshagrawal38: Once the python script is completed, we are seeing _PyImport_Init error

```
Download complete.
Fatal Python error: _PyImport_Init: global import state already initialized
Python runtime state: preinitialized
```

```
The Fatal Python error: _PyImport_Init: global import state already initialized is a 
crash-on-exit  caused by a C extension (likely torch or transformers) trying to re-initialize the 
Python interpreter during shutdown.
```

Issue started coming after bump up of  torch pkg

BOB says
`Fix in the Python script itself by catching the crash before the interpreter shuts down, using os._exit(0) to hard-exit before the C extension teardown fires:`

yussufsh: This could be a compatibility issue that needs to be fixed.

images/service-base/prebuilder/pre-requirements.txt
```diff
@@ -1 +1,2 @@
-ujson==5.13.0
+cryptography==50.0.0
+grpcio==1.80.0
```
yussufsh: Why are we building this? It is available in devpi.

adarshagrawal38: Added it here as pulling pkg from devpi was falling for the pkg.

Now I am able to pull pkg from devpi, so removed from here

yussufsh (dismissed):

yussufsh:
@adarshagrawal38 as discussed, let's remove the unused pip, setuptools and other files that are not used by us. This will avoid the scan issue we are having.

images/service-base/prebuilder/pre-requirements.txt
```diff
@@ -1,2 +1,3 @@
 sudachipy==0.6.11
 ujson==5.13.0
```
yussufsh: Not need to build right? available on devpi.

1
yussufsh, line 1: What is this file? Please delete.

images/service-base/Containerfile
```diff
@@ -20,10 +22,11 @@ COPY download_docling_models.py .
 
 RUN --mount=type=bind,from=prebuilder,source=/wheels,target=/wheels,ro \
     python -m venv /var/venv && source /var/venv/bin/activate && \
-    pip install --upgrade pip && \
+    pip install --upgrade pip setuptools wheel && \
     pip install /wheels/*.whl -r requirements.txt \
         --extra-index-url https://wheels.developerfirst.ibm.com/ppc64le/linux \
         --prefer-binary && \
     pip cache purge && \
+    rm -f /var/venv/lib/python3.12/site-packages/pip/_vendor/bom.cdx.json && \
```
yussufsh, line 30: I don't think just deleting the bom file is good solution. Let's discuss this more offline.

yussufsh (changes requested):

images/service-base/prebuilder/pre-requirements.txt
adarshagrawal38: done

1
adarshagrawal38: removed

images/service-base/Containerfile
yussufsh: This is fixed in https://github.com/IBM/project-ai-services/pull/1463
You can use this PR for docling upgrade.

adarshagrawal38:
fixed as part of different PR.

PR #1309, opened 2026-08-24
feat(translate): Add synchronous endpoint for translate service

Niharika0306:
- Implements the stateless POST /v1/translate synchronous text translation endpoint (no DB writes, no file staging).
- Adds utils/language.py to consolidate duplicated language validation logic shared by the sync and async job routers.
- Introduces _raise_context_limit_exceeded with a diagnostics payload; tightens the context guard to require available_for_output ≥ input_tokens.
- Covers all expected error codes: 400 (invalid language), 413 (context limit), 429 (concurrency at capacity), 500/503 (vLLM errors).

The PR also includes changes regarding job_not_complete messages in GET /v1/translate/jobs/{job_id}/result  endpoint

### DIFF
--- a/services/translate/api/v1/jobs.py
+++ b/services/translate/api/v1/jobs.py
@@ -10,7 +10,7 @@ GET    /v1/translate/jobs/{job_id}/result/download — download translated file
 
 import asyncio
 from datetime import datetime, timezone
-from typing import Optional
+from typing import Optional, Union
 
 from fastapi import APIRouter, File, Form, Query, UploadFile, status
 from fastapi.responses import JSONResponse, PlainTextResponse
@@ -285,7 +285,7 @@ async def get_job(job_id: str) -> JobDetailResponse:
         500: http_error_responses[500],
     },
 )
-async def get_job_result(job_id: str):
+async def get_job_result(job_id: str) -> Union[JobResultResponse, JSONResponse]:
     """Return the completed translation result for a job."""
     job = db_manager.get_job_by_id(job_id)
     if job is None:
@@ -348,7 +348,7 @@ async def get_job_result(job_id: str):
         500: http_error_responses[500],
     },
 )
-async def download_job_result(job_id: str):
+async def download_job_result(job_id: str) -> Union[PlainTextResponse, JSONResponse]:
     """Download the translated output file for a completed job."""
     job = db_manager.get_job_by_id(job_id)
     if job is None:
--- a/services/translate/api/v1/jobs.py
+++ b/services/translate/api/v1/jobs.py
@@ -20,12 +20,11 @@ from common.misc_utils import get_logger, get_utc_timestamp
 from common.validation_utils import validate_file_content, validate_file_extension
 from translate.utils.errors import (
     _raise_file_too_large,
-    _raise_invalid_language,
     _raise_job_failed,
     _raise_job_not_complete,
-    _raise_same_language,
     _raise_unsupported_file_type,
 )
+from translate.utils.language import validate_languages
 from translate.db.manager import db_manager
 from translate.models import (
     JobCreatedResponse,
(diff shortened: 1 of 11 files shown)


services/translate/api/v1/translate.py
```diff
@@ -3,11 +3,174 @@
...
+                    llm_endpoint=llm_endpoint,
+                    messages=messages,
+                    model=model,
+                    max_tokens=max_tokens,
+                    temperature=settings.translate.translation_temperature,
+                )
+    except httpx.RequestError as exc:
+        logger.error(f"vLLM unreachable during sync translate: {exc}", exc_info=True)
+        APIError.raise_error(
+            ErrorCode.VECTOR_STORE_NOT_READY,
```
manalilatkar: incorrect error code VECTOR_STORE_NOT_READY instead of VLLM_OUT_OF_SERVICE or something like that.

services/translate/utils/language.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Rules:
+    - ``target`` must be an explicit supported language name (not ``"auto"``).
+    - ``source`` may be ``"auto"`` (triggers lingua / LLM detection) or an
+      explicit supported language name.
+    - When both are explicit they must differ.
+    """
+    supported = settings.translate.supported_languages_list
+    supported_display = ", ".join(s.capitalize() for s in sorted(supported))
+
```
manalilatkar, line 39: Before normalising target_language, we should be validating it for non-emptiness. Currently if the user sends empty or whitespace string, we return the response "target language can't be auto" whereas we should be sending "target language cannot be empty"

services/translate/api/v1/jobs.py
```diff
@@ -333,16 +276,16 @@ async def get_job(job_id: str) -> JobDetailResponse:
...
     responses={
+        202: {"description": "Accepted — job is still in progress. Retry after a short delay."},
         404: http_error_responses[404],   # job not found
-        409: http_error_responses[409],   # JOB_NOT_COMPLETE
         410: {"description": "Gone — job failed. Error details on the job resource."},
         500: http_error_responses[500],
     },
 )
-async def get_job_result(job_id: str) -> JobResultResponse:
+async def get_job_result(job_id: str):
```
manalilatkar: missing return type annotation, use Union[JobResultResponse, JSONResponse]

services/translate/api/v1/translate.py
```diff
@@ -3,11 +3,174 @@
...
+    llm_endpoint = settings.common.llm.endpoint
+    input_tokens: int = await asyncio.to_thread(
+        lambda: len(tokenize_with_llm(body.text, llm_endpoint))
+    )
+
+    max_model_len = get_llm_max_model_len()
+    prompt_overhead = settings.translate.prompt_overhead_tokens
+    available_for_output = max_model_len - input_tokens - prompt_overhead
+
+    if available_for_output < input_tokens:
```
manalilatkar, line 88: This assumes that the output text will be equal to input text whereas we know when we translate from English to German, 100 tokens become approx 135 tokens and when we translate German to English, 100 tokens become approx 75 tokens. So we would be letting some unsuitable inputs pass and reject some suitable inputs.

services/translate/api/v1/translate.py
```diff
@@ -3,11 +3,174 @@
...
+                "input_tokens": input_tokens,
+                "prompt_overhead_tokens": prompt_overhead,
+                "max_model_len": max_model_len,
+                "available_for_output": available_for_output,
+            },
+        )
+
+    # Cap output tokens: 2× input provides generous headroom for most language
+    # pairs (en→de expansion is typically ~20–30 %).
+    max_tokens = min(available_for_output, input_tokens * 2)
```
manalilatkar, line 102: According to the proposal, this was the formula for calculatiung max_tokens:
```
available_output  = MAX_MODEL_LEN - input_tokens - prompt_overhead
buffer            = max(20, int(available_output * 0.10))   # 10% breathing room
effective_max_tokens = available_output - buffer
max_tokens (LLM call) = effective_max_tokens
```
Why did we change it?

manalilatkar (changes requested):

services/translate/api/v1/translate.py
Niharika0306: Right, it was a bit difficult to keep that ratios while converting in between languages, and they are not even completely accurate. So since this is only for the synchronous translation, I thought of just keeping a high level check that can be generic. [Yes, we could lose some tokens at the end if its en -> de ]

services/translate/api/v1/translate.py
Niharika0306: Right, so this was introduced to keep the max_tokens a bit realistic. Currently, if a small chunk is sent for translation, max_token value is too huge and doesn't make sense for the usecase. So to keep things in check and generate accurate responses, this min(available_for_output, input_tokens * 2)   logic was added.

services/translate/api/v1/translate.py
Niharika0306: My bad, didn't notice this. Added an appropriate error code.

services/translate/utils/language.py
Niharika0306: Done.

services/translate/api/v1/jobs.py
Niharika0306: Done.

services/translate/api/v1/translate.py
manalilatkar: ok

manalilatkar (approved):
LGTM

PR #1310, opened 2026-08-24
worker side runtime dispatch flow

yussufsh:
Implements the worker-side command dispatcher and moves RemoteRuntime to the correct package:

- RemoteRuntime implementing runtime.Runtime over the gRPC CommandStream; WorkerRegistry interface decoupling it from the worker/registry package
- Dispatch() routes incoming Command messages to the local runtime.Runtime and returns a CommandResult
- recvLoop wired to dispatch.Dispatch; Run signature simplified (no longer takes runtime.Runtime — creates it internally from opts.RuntimeType)
- WorkerEntry stores RuntimeType string; WorkerCommandChannel and WorkerRuntimeType added to satisfy the WorkerRegistry interface; runtimeTypeFromString removed
- CreateRemote looks up worker runtime type from the registry instead of requiring it as an argument


ai-services/internal/pkg/runtime/remote/payloads.go
```diff
@@ -0,0 +1,69 @@
...
+
+type deletePodPayload struct {
+	ID    string `json:"id"`
+	Force *bool  `json:"force,omitempty"`
+}
+
+// podIDPayload is used by StopPod and StartPod.
+type podIDPayload struct {
+	ID string `json:"id"`
+}
```
mayuka-c: Can we use belore nameOrId itself?. I see both start and stop does supprot it

ai-services/internal/pkg/runtime/remote/runtime.go
```diff
@@ -0,0 +1,416 @@
...
+	"github.com/project-ai-services/ai-services/internal/pkg/runtime/types"
+
+	workerpb "github.com/project-ai-services/ai-services/internal/pkg/worker/proto"
+)
+
+const (
+	// commandTimeout is the maximum time to wait for a worker to respond to a
+	// single command. Long-running operations (e.g. PullImage) may need more
+	// time; callers can pass a context with a tighter deadline if needed.
+	commandTimeout = 5 * time.Minute
```
mayuka-c: Doesnt model download also take lot of time. Is 5 mins fine?

ai-services/internal/pkg/runtime/remote/runtime.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	return exists, nil
+}
+
+// UpdateSecret is not yet supported over the remote worker protocol.
+// TODO: implement when required.
+func (r *RemoteRuntime) UpdateSecret(_, _ string, _ map[string][]byte) error {
+	return fmt.Errorf("remote runtime: UpdateSecret not yet supported on remote workers")
+}
```
mayuka-c, line 224: Can we remove and add it later when required?

ai-services/internal/pkg/runtime/remote/runtime.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	return names, nil
+}
+
+func (r *RemoteRuntime) DeleteSecret(name string) error {
+	_, err := r.send(context.Background(), workerpb.CommandType_COMMAND_TYPE_DELETE_SECRET, namePayload{Name: name})
+
+	return err
+}
```
mayuka-c, line 204: Will it be better to take ctx as args for all and not use default context.Background?

ai-services/internal/pkg/worker/dispatch/payloads.go
```diff
@@ -0,0 +1,64 @@
...
+
+// This file mirrors internal/pkg/runtime/remote/payloads.go exactly.
+// Both sides of the gRPC stream must agree on the JSON field names and
+// struct layout. Any change here must be reflected in the remote package.
+
+// ─── Image ────────────────────────────────────────────────────────────────────
+
+type pullImagePayload struct {
+	Image string `json:"image"`
+}
```
mayuka-c: Is it not possible to use same common model package across both?.

ai-services/internal/pkg/runtime/remote/payloads.go
yussufsh: It actually supports name or id. So while generating the code assumed similar approach compared to other runtimes.
I will rename the params even for podman and openshift.

ai-services/internal/pkg/runtime/remote/runtime.go
yussufsh: I don't think it takes more than 5 minutes. Anyways, I can change the timeout to 10 mins.

ai-services/internal/pkg/runtime/remote/runtime.go
yussufsh: Yes that is what is meant here... func is needed to comply with the runtime interface.

ai-services/internal/pkg/runtime/remote/runtime.go
yussufsh: Planning to do this later.. once the main PR gets merged.

ai-services/internal/pkg/worker/dispatch/payloads.go
yussufsh: Which package you mean?

yussufsh: Moved the files to worker/payload package now.

mayuka-c (approved):
LGTM

ai-services/internal/pkg/runtime/remote/runtime.go
yussufsh: @shwetha-s-poojary FYA

PR #1314, opened 2026-08-24
[BYOS] Validate Bundle

mayuka-c:
- Validate the custom bundle
- Separate Validation process for both podman and openshift runtimes
- Integrate with POST and PUT calls

Note:
- Doesnt validate the template yaml files (Planning to take it up next and do min validations like yaml structure, labels, annotations)


mayuka-c (commented):
Overall this is solid work — the package decomposition (`validate/`, `validate/metadata/`), the concurrent runtime validators, the Helm render-time template check, and the expanded `metadata.yaml` field requirements are all good additions. The tests are well-structured and the naming cleanup (`CatalogReloader` → `CatalogProvider`) is an improvement.

There are three **blocking issues** and a handful of smaller ones:

1. **Silent error drop in concurrent validation** — `errCh` is only drained once after `close`; the second validator's error is lost (see comment on `service.go:94`).
2. **`ProcessBundle` and `ReplaceBundle` call `ValidateBundle` which re-runs `checkCatalogCollision`** — for `ReplaceBundle` this will always 422-reject a legitimate update because the catalog already has the entry. The collision check must be skipped when called from a mutation path (see comments on `service.go:196` and `service.go:317`).
3. **No test coverage for `checkCatalogCollision`** — the most important new behaviour has zero tests (see comment on `service.go:116`).

Other findings: dead `CatalogChecker` interface in `bundlemetadata`, fragile `default` branch in `parseMetadataYAML`, redundant archive passes in the OpenShift validator, and two `gofmt` alignment issues.

Please address the three blocking issues before merge.

ai-services/internal/pkg/catalog/apiserver/services/bundle/validate/metadata/common.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/project-ai-services/ai-services/internal/pkg/catalog/validators"
+)
+
+// runtimeResources is the shared resources block present in every runtime metadata.yaml.
+// storage is allowed to be 0 (cloud-based components carry no local model storage).
+type runtimeResources struct {
+	CPU     int   `yaml:"cpu"`
+	Memory  int64 `yaml:"memory"`
+	Storage int64 `yaml:"storage"`
```
yussufsh: Should not this be all string to accomodate units (eg: Mi, Gi)?

mayuka-c: Currently we do not take Mi, Gi. We expect users to provide it in bytes within metadata.yaml file: https://github.com/IBM/project-ai-services/blob/main/ai-services/assets/services/digitize/podman/metadata.yaml#L11.

So we dont take any units now

yussufsh: I see we are missing `accelerators` as well?

mayuka-c: Yeah accelerators is quite tricky. Its not a mandatory field and its applicable to components (but not all components require it). So I didnt specifically wanted to add any check here.

But I feel we could add a check if accelerator field is present, then ensure its only spyre? Any recommendations?

mayuka-c: Added accelerator nd also restriction for resources section in runtime metadata to support only those 4

yussufsh (approved):
overall lgtm
Will keep the validation checks to minimal and add on in future as needed.

PR #1315, opened 2026-08-24
Add baseDir, domainSuffix, httpsPort as part of worker metadata

yussufsh:
1. Added metadata to the worker register flow.
2. Added remaining CLI params for `join` cmd.
3. Refactored common cmd code for `catalog configure` and `worker join`.

### DIFF
--- a/ai-services/internal/pkg/utils/cert.go
+++ b/ai-services/internal/pkg/utils/cert.go
@@ -93,11 +93,11 @@ func validateSSLCertificates(certPath, keyPath string) error {
 		return fmt.Errorf("certificate validation failed: %w", err)
 	}
 
-	if err := ValidateCertificateKeyPair(certPath, keyPath); err != nil {
+	if err := validateCertificateKeyPair(certPath, keyPath); err != nil {
 		return fmt.Errorf("certificate and key validation failed: %w", err)
 	}
 
-	if err := ValidateWildcardCertificate(certPath); err != nil {
+	if err := validateWildcardCertificate(certPath); err != nil {
 		return fmt.Errorf("wildcard certificate validation failed: %w", err)
 	}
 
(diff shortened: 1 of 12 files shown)


ai-services/internal/pkg/worker/deploy/deploy.go
```diff
@@ -47,7 +48,17 @@ type Options struct {
 	BaseDir string
 	// HTTPSPort is the host port Caddy binds for HTTPS traffic, e.g. "443".
 	// Set once at first join; ignored on subsequent runs if already deployed.
-	HTTPSPort string
+	HTTPSPort int
```
mayuka-c, line 51: Any reason for making this int instead of string now?

ai-services/internal/pkg/utils/cert.go
```diff
@@ -21,8 +21,91 @@ const (
...
+func validateSSLCertificates(certPath, keyPath string) error {
+	if err := validateCertificateFiles(certPath, keyPath); err != nil {
+		return fmt.Errorf("certificate validation failed: %w", err)
+	}
+
+	if err := ValidateCertificateKeyPair(certPath, keyPath); err != nil {
+		return fmt.Errorf("certificate and key validation failed: %w", err)
+	}
+
+	if err := ValidateWildcardCertificate(certPath); err != nil {
```
mayuka-c: Can we make ValidateCertificateKeyPair and ValidateWildcardCertificate private if only used by this file?

ai-services/internal/pkg/worker/deploy/deploy.go
yussufsh: No reason, perhaps it should have been int from before.

mayuka-c (approved):
LGTM

PR #1316, opened 2026-08-24
refactor: unify StepTwo and ServiceConfigCard into shared components

pranithraoibm:
- Move `ServiceConfigCard` to `Shared/`and introduce `SharedStepTwo`; both flows delegate to them
- Remove inferenceBackend from ServiceConfig
- Delete inferenceComponentHelper.ts
- Move `paramFilter.ts` to `Shared/utils/` and add `splitServiceParams`; used by both transforms
- Add COMPONENT_TYPES(LLM, RERANKER...) constants; replace all inline component type strings


ui/catalog/src/components/DeployFlow/Shared/utils/paramFilter.ts
```diff
@@ -0,0 +1,82 @@
...
+    providerSchema?.properties ? Object.keys(providerSchema.properties) : [],
+  );
+  const serviceKeys = new Set(
+    serviceSchema?.properties ? Object.keys(serviceSchema.properties) : [],
+  );
+
+  const serviceBackendParams: Record<string, unknown> = {};
+  const inferenceCredentialParams: Record<string, unknown> = {};
+
+  for (const [key, value] of Object.entries(allParams)) {
```
ryanmarc, line 112: This loop splits params but it does not filter them, and the separateParams it replaces did both. The old one ran every value through shouldIncludeParam(value, schemaProperty) first, which dropped empty strings and dropped anything equal to its schema default.

Is that filtering no longer needed?

ui/catalog/src/components/DeployFlow/Shared/utils/paramFilter.ts
```diff
@@ -0,0 +1,82 @@
...
+  inferenceCredentialParams: Record<string, unknown>;
+} {
+  if (!allParams || Object.keys(allParams).length === 0) {
+    return { serviceBackendParams: {}, inferenceCredentialParams: {} };
+  }
+
+  const providerKeys = new Set(
+    providerSchema?.properties ? Object.keys(providerSchema.properties) : [],
+  );
+  const serviceKeys = new Set(
```
ryanmarc: This reads top-level properties, but service schemas in this repo nest a level deeper. `ai-services/assets/services/chat/podman/values.schema.json` is `{properties: {backend: {properties: {editSystemPrompt, systemPrompt}}}}`, so Object.keys(serviceSchema.properties) is ["backend"] and serviceKeys never contains systemPrompt. The separateParams this replaces walked into properties.backend.properties deliberately, and parseSchema in utils/schemaParser.ts:33-42 recurses for the same reason.

The effect is that the documented fallback ("Service schema is a fallback when no provider schema is supplied") classifies every param as a credential. I ran the transform with providerParamsByType empty and got the credentials plus systemPrompt merged into the llm component's params, with params.backend gone from the payload entirely.

ryanmarc (changes requested):

ui/catalog/src/components/DeployFlow/Shared/utils/paramFilter.ts
pranithraoibm: Yes, added `shouldIncludeParam` back though in the current scenario only empty field possible will be the optional API key as rest all are dropdown values and watsonx fields already have non-empty guard on the UI

ryanmarc (approved):

PR #1317, opened 2026-08-25
Data Connectors: Misc Changes

sats-23:
[AISERVICES-1672](https://jsw.ibm.com/browse/AISERVICES-1672)

-Bulk delete of docs now deletes only user-submitted docs (not connector sourced docs)
-Fixed Swagger doc examples
-Added validation on Connection type and allowed extensions 
-Ported S3 and SSH to object_storage and file_system (converted to ENUM) on all user-facing APIs
-Connectors stuck in DELETE_PENDING due to restart is cleaned up as part of recovery
-Scanner init exception is caught to show AuthException message
-Reorder validate_document_files to do primitive checks first 
-Reuse validate_document_files in connector flow (without it would accept and mark failed)
-Fixed tests appropriately 
-All exception TCs are attached in the ticket

### DIFF
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -72,32 +72,26 @@ logger = get_logger("connectors_router")
 async def create_connector(body: ConnectorCreateRequest):
     """Create a new connector and schedule its worker.
 
-    Validates connector settings, encrypts credentials, persists the new connector
-    configuration in the database, and schedules the background worker.
-
-    The DB insert is performed first so that a duplicate id/name is detected
-    (IntegrityError → 409) before any scheduler state is created.  The
-    scheduler job is only registered once the row is safely committed.
+    Checks for duplicate connector id and name via explicit SELECTs before
+    touching the scheduler or writing to the database.  Only once both checks
+    pass is the scheduler job registered and the row inserted.
     """
     connector_id = body.id or str(uuid.uuid4())
     try:
-        encrypted_details = encrypt_secrets(body.type, body.connection_details)
+        # --- Duplicate checks (GET by id, then GET by name) ---
+        if db_ops.get_connector_by_id(connector_id) is not None:
+            msg = f"Connector id {connector_id!r} already exists"
+            logger.error(msg)
+            APIError.raise_error(ErrorCode.RESOURCE_LOCKED, msg)
 
+        if db_ops.get_connector_by_name(body.name) is not None:
+            msg = f"Connector name {body.name!r} already exists"
+            logger.error(msg)
+            APIError.raise_error(ErrorCode.RESOURCE_LOCKED, msg)
+
+        encrypted_details = encrypt_secrets(body.type, body.connection_details)
         sync_interval = settings.digitize.connector.sync_interval_seconds
 
-        # Persist the connector row first — this validates uniqueness.
-        # If a duplicate id or name exists an IntegrityError is raised here
-        # and we return 409 before touching the scheduler.
-        db_ops.insert_connector(
-            connector_id=connector_id,
-            name=body.name,
-            connector_type=body.type,
-            connection_details=encrypted_details,
-            allowed_extensions=body.allowed_extensions,
-            sync_interval_seconds=sync_interval,
-        )
-
-        # Row committed — now it is safe to register the scheduler job.
         try:
             import digitize.connectors.scheduler as _sched
             await _sched.register_connector_job(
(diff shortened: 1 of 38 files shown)


dharaneeshvrd (commented):
I see duplicate connector is throwing the error after we spawned the AP scheduler.
let's check it before and validate for dups 
then proceed with the AP scheduler

services/digitize/api/v1/connectors.py
dharaneeshvrd: No need to return id, let's use name alone

dharaneeshvrd (commented):
```
        if is_connector_sourced_document(doc_id):
            APIError.raise_error(
                ErrorCode.RESOURCE_NOT_FOUND,
                f"Document with ID '{doc_id}' not found",
            )
```
Instead of this can we return 405 Method Not Allowed, I feel that is more appropriate.

services/digitize/api/v1/connectors.py
sats-23: But IntegrityError is thrown in both cases. 
When a dup name or ID is used. 
There could be a case when the same ID is used and name is different, that way, our error message could be misleading

dharaneeshvrd: Is there a way to differentiate the dup is happening because of id or name?

sats-23: Yes, should be possible. I will do it

services/digitize/api/v1/connectors.py
```diff
@@ -72,15 +74,30 @@ async def create_connector(body: ConnectorCreateRequest):
...
         encrypted_details = encrypt_secrets(body.type, body.connection_details)
 
         sync_interval = settings.digitize.connector.sync_interval_seconds
 
-        # Register the connector with the scheduler so it starts ticking
-        # immediately (fire_immediately=True for the first-ever sync).
+        # Persist the connector row first — this validates uniqueness.
+        # If a duplicate id or name exists an IntegrityError is raised here
+        # and we return 409 before touching the scheduler.
+        db_ops.insert_connector(
```
dharaneeshvrd: Let's check the duplicate by GET on connector id & name
If there is no dup, create ap scheduler and insert the record into db.

services/digitize/api/v1/connectors.py
```diff
@@ -96,33 +113,35 @@ async def create_connector(body: ConnectorCreateRequest):
...
-
         logger.info(
             f"Connector {connector_id!r} ({body.name!r}) attached "
             f"(type={body.type}, interval={sync_interval}s)"
         )
 
         return Response(
             content=f'{{"id": "{connector_id}"}}',
-            status_code=201,
+            status_code=202,
```
dharaneeshvrd, line 127: can we update the proposal too?
also @prachikurade1 should be aware of this.

services/digitize/connectors/scanners/config.py
```diff
@@ -190,19 +190,12 @@ def _check_endpoint_url(cls, v: str) -> str:
     @field_validator("allowed_extensions")
     @classmethod
     def _check_extensions(cls, v: list[str]) -> list[str]:
```
dharaneeshvrd:
```suggestion
    def _normalise_extensions(cls, v: list[str]) -> list[str]:
```

services/digitize/connectors/scanners/config.py
```diff
@@ -284,13 +277,12 @@ def _check_private_key(cls, v: str) -> str:
...
```
dharaneeshvrd:
```suggestion
    def _normalise_extensions(cls, v: list[str]) -> list[str]:
```

services/digitize/api/v1/connectors.py

services/digitize/connectors/sync_tick.py
```diff
@@ -154,16 +157,24 @@ async def run_tick(connector_id: str, sync_seq: int) -> None:
...
             new_files=len(ingest_list),
             removed_files=len(orphan_checksums),
         )
-        update_connector_total_files(connector_id, len(scanned_files))
 
-        await _process_new_files(sync_seq, connector_id, config.name, scanner, ingest_list)
+        invalid_count = await _process_new_files(sync_seq, connector_id, config.name, scanner, ingest_list)
+
+        if invalid_count > 0:
+            total_files -= invalid_count
```
dharaneeshvrd: I dont think we should do this
we should mark the current sync log & connector status as out of sync and update the error as these many invalid files detected or not processed.

services/digitize/connectors/scanners/config.py
sats-23: Done

services/digitize/connectors/scanners/config.py
sats-23: Done

services/digitize/connectors/sync_tick.py
sats-23: Done, failed tick for invalid files

services/digitize/api/v1/connectors.py
sats-23: Done

services/digitize/api/v1/connectors.py
```diff
@@ -112,23 +124,20 @@ async def create_connector(body: ConnectorCreateRequest):
...
             media_type="application/json",
         )
 
     except IntegrityError:
```
dharaneeshvrd, line 131: Is it still required?

sats-23: Yes, good to have. 
Happens very rare, only in race conditions

dharaneeshvrd (approved):

services/digitize/api/v1/connectors.py
prachikurade1: @Shubhamag12 , FYI for the connect API.

PR #1318, opened 2026-08-25
worker uninstall cli

yussufsh:
Add `worker uninstall` command

1. Deletes the Caddy proxy pod and cleans up <baseDir>/worker
2. Recovers base dir from AI_SERVICES_BASE_DIR env via pod inspect.
3. Extracts shared worker constants into worker/constants package.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/worker/uninstall.go
+++ b/ai-services/cmd/ai-services/cmd/worker/uninstall.go
@@ -1,0 +1,57 @@
+package worker
+
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+
+	cmdcommon "github.com/project-ai-services/ai-services/cmd/ai-services/cmd/common"
+	"github.com/project-ai-services/ai-services/internal/pkg/vars"
+	workeruninstall "github.com/project-ai-services/ai-services/internal/pkg/worker/uninstall"
+)
+
+// Flag variables for the worker uninstall command.
+var (
+	uninstallRuntimeType string
+	uninstallAutoYes     bool
+)
+
+func newUninstallCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "uninstall",
+		Short: "Remove this node's worker components",
+		Long: `Removes all worker components deployed by 'worker join' on this node.
+
+The uninstall process will:
+  - Delete the Caddy reverse-proxy pod
+  - Remove the worker data directory (<basedir>/worker)
+
+Application pods deployed on this worker by the catalog are not touched.`,
+		Example: `  # Uninstall worker components (prompts for confirmation)
+  ai-services worker uninstall --runtime podman
+
+  # Skip confirmation prompt
+  ai-services worker uninstall --runtime podman --yes
+
+  ai-services worker uninstall --runtime podman --yes`,
+		Args: cobra.NoArgs,
+		PreRunE: func(cmd *cobra.Command, _ []string) error {
+			cmd.SilenceUsage = true
+
+			return cmdcommon.InitAndValidateRuntimeFlag(uninstallRuntimeType)
+		},
+		RunE: func(_ *cobra.Command, _ []string) error {
+			return workeruninstall.Uninstall(context.Background(), workeruninstall.Options{
+				RuntimeType: vars.RuntimeFactory.GetRuntimeType(),
+				AutoYes:     uninstallAutoYes,
+			})
+		},
+	}
+
+	cmdcommon.ConfigureRuntimeFlag(cmd, &uninstallRuntimeType)
+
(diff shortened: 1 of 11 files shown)


ai-services/cmd/ai-services/cmd/worker/uninstall.go
```diff
@@ -0,0 +1,57 @@
...
+
+func newUninstallCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "uninstall",
+		Short: "Remove this node's worker components",
+		Long: `Removes all worker components deployed by 'worker join' on this node.
+
+The uninstall process will:
+  - Delete the Caddy reverse-proxy pod
+  - Remove the worker data directory (<basedir>/worker)
```
mayuka-c, line 27: Dont we also need to remove the worker daemon service too?

ai-services/internal/pkg/worker/constants/constants.go
```diff
@@ -0,0 +1,19 @@
...
+
+const (
+	// WorkerProxyLabel is the pod label set by deploy.Setup; used by deploy (idempotency) and uninstall (lookup).
+	WorkerProxyLabel = "ai-services.io/component=worker-proxy"
+
+	// WorkerDataSubDir is the on-disk subtree written by deploy.Setup; removed by uninstall.
+	WorkerDataSubDir = "worker"
+
+	// BaseDirEnvVar is injected into the Caddy container at deploy time; read back by uninstall.
+	BaseDirEnvVar = "AI_SERVICES_BASE_DIR"
```
mayuka-c, line 13: I think there is already a constant for this

ai-services/internal/pkg/worker/uninstall/uninstall.go
```diff
@@ -0,0 +1,193 @@
...
+// extractConfigFromEnv populates config from Caddy container environment variables.
+func extractConfigFromEnv(env map[string]string, config *WorkerCaddyConfig) {
+	if value, ok := env[workerconstants.BaseDirEnvVar]; ok {
+		config.BaseDir = value
+	}
+}
+
+// confirmUninstall prints the pod list and prompts the user when autoYes is false.
+// Returns (false, nil) when the user declines, (true, nil) when confirmed.
+func confirmUninstall(ctx context.Context, pods []types.Pod, autoYes bool) (bool, error) {
```
mayuka-c, line 130: This looks similar to our catalog uninstall and even app deletion CLI too, Can we reuse?

ai-services/internal/pkg/worker/constants/constants.go
yussufsh: no there is none, hence added in worker pkg.

ai-services/internal/pkg/worker/uninstall/uninstall.go
yussufsh: I had the same thinking and here is what Bob told me:
> They are essentially the same logic. Two differences prevent a straight merge right now:
>
> podsDeletion takes *podman.PodmanClient (concrete type) — it can't be shared without taking runtime.Runtime.
> Log messages differ slightly.
> These are internal functions in their own packages — extracting a shared helper would create a cross-package dependency that doesn't exist today. The right call here is leave them as-is: the duplication is acceptable because the catalog version is tied to *podman.PodmanClient and lives in a completely separate package tree. Consolidating would mean either pulling catalog code into a runtime.Runtime-based helper package or making catalog import the worker package — neither is appropriate.
>
> The observation is valid but the fix isn't worth the coupling. They can be aligned independently if podsDeletion in catalog is ever refactored to take runtime.Runtime instead of the concrete podman client.

ai-services/cmd/ai-services/cmd/worker/uninstall.go
yussufsh: Right now, the worker daemon is not running as a service, it is a shell process.

mayuka-c: Shouldnt we be killing this long running process?

ai-services/internal/pkg/worker/constants/constants.go
mayuka-c: Ok I guess anyway this will not be required once we have caddy to use volumes

ai-services/cmd/ai-services/cmd/worker/uninstall.go
yussufsh: Since it is the CLI based, sessions would be different for uninstall. We cannot make out if the daemon process was run via systemd, background or something else.

Shubhamag12 (approved):
lgtm

mayuka-c (approved):
LGTM,
Once we have the worker daemon running as a container/systemd we would need to add it here for cleanups

PR #1319, opened 2026-08-25
fix: wait for inferenceService to be active before proceeding

Shubhamag12:
## Problem
On OpenShift, component deployment was completing before the KServe predictor pods were actually running.
This caused Phase 1 (deployComponentsConcurrently) to return immediately with predictors still in Init:0/1, Phase 2 services to deploy against unavailable backends, and the sync service to subsequently mark the application as Error because the LLM/embedding/reranker components weren't actually ready.

A deployment should not be considered ready if its underlying workloads aren't running, but Helm has no awareness of the secondary Deployment that KServe asynchronously creates from an InferenceService.

## Solution
After helmInstallOrUpgrade returns for each component, explicitly poll the InferenceService object `.status.conditions[type=Ready].status` until it is "True" before returning from deployComponent.

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.279
+TAG?=v0.0.280
 CONTAINER_BUILDER?=podman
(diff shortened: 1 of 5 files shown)


yussufsh (commented):
@Shubhamag12 this lgtm. Please rebase and bump up the images.

yussufsh (approved):

PR #1320, opened 2026-08-25
upgrade vllm-spyre image to 3.5.0 and increase memory to 200Gi

Shubhamag12:
Since RHAI GA is done, updating image for vllm-spyre to 3.5.0 and increase memory to 200Gi, for both runtimes.

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.296
+TAG?=v0.0.297
 CONTAINER_BUILDER?=podman
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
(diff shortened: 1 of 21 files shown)


yussufsh:
@Shubhamag12 you need to change resources in metadata.yaml as well for spyre.

yussufsh:
Also @Shubhamag12 , You need to use 200Gi for `vllm-cpu` as well? I do not see version bump for upstream vllm image here. Why was this needed?
And, if it is needed why not bump up the memory requirements in rag-cpu template?

Shubhamag12:
> Also @Shubhamag12 , You need to use 200Gi for `vllm-cpu` as well? I do not see version bump for upstream vllm image here. Why was this needed? And, if it is needed why not bump up the memory requirements in rag-cpu template?

ah my bad. Its a result of find and replace, I reverted cpu files from other places and this one was missed. Will fix it

manju956 (dismissed):
lgtm

dharaneeshvrd (changes requested):
Let's keep the cpu changes in a different PR
I don't think CPU template's memory needs to be bumped.

Shubhamag12:
> Let's keep the cpu changes in a different PR I don't think CPU template's memory needs to be bumped.

yes, I missed reverting that file. will change it
https://github.com/IBM/project-ai-services/pull/1320#issuecomment-5475092189

mayuka-c (dismissed):
LGTM

dharaneeshvrd (approved):

yussufsh (approved):

PR #1321, opened 2026-08-25
don not allow active worker to re-register

yussufsh:
DB Status | In-memory entry? | Register possible?
-- | -- | --
pending | ❌ No entry | ✅ Should succeed
ready | ✅ Entry exists | ❌ Should fail
disconnected | ❌ No entry (removed by Disconnect) | ✅ Should succeed

</div>


mayuka-c (approved):
LGTM

PR #1323, opened 2026-08-25
feat: align backup/restore with catalog approach

Shubhamag12:
This PR aligns OpenShift backup/restore with how Podman works:
1. Resolves app details from catalog API instead of using raw app name
2. Derives namespace as ai-services-<first 8 chars of app UUID>
3. Uses opensearch component UUID as ai-services.io/template label selector to find pods
4. Derives Digitize URL from catalog service endpoints instead of oc get route


prachikurade1 (dismissed):
LGTM. Approved

ai-services/internal/pkg/application/openshift/restore.go
```diff
@@ -18,28 +21,47 @@ func (o *OpenshiftApplication) Restore(ctx context.Context, opts types.RestoreOp
...
+	logger.Infof("OpenSearch component ID: %s\n", templateID)
+
+	// Derive namespace using catalog convention: "ai-services-<first 8 chars of app UUID>"
+	appUUID, err := uuid.Parse(appDetails.ID)
+	if err != nil {
+		return fmt.Errorf("failed to parse application UUID: %w", err)
+	}
+	namespace := catalogUtils.AppNamespace(appUUID)
+	logger.Infof("Namespace: %s\n", namespace)
+
```
yussufsh: This is all OpenSearch related, so it should go in the switch section below.

Shubhamag12: done

yussufsh (dismissed):
overall lgtm!
Hope you have tested both openshift and podman B&R for digitize and opensearch.

yussufsh (approved):

PR #1325, opened 2026-08-25
Add openshift manifests for extract and translate service

dharaneeshvrd:
- Adds openshift manifest for extract and translate service
- Adds db-wait init container to ensure db is healthy before spawning the service containers

### DIFF
--- a/ai-services/assets/services/chat/openshift/templates/chat-bot-backend-deployment.yaml
+++ b/ai-services/assets/services/chat/openshift/templates/chat-bot-backend-deployment.yaml
@@ -22,6 +22,31 @@ spec:
         ai-services.io/component-type: api
     spec:
       automountServiceAccountToken: false
+      initContainers:
+        - name: llm-wait
+          image: "registry.access.redhat.com/ubi9/ubi-minimal:latest"
+          command:
+            - "/bin/sh"
+            - "-c"
+            - |
+              echo "Waiting for LLM endpoint to become ready..."
+              until wget -q -O /dev/null http://llm-predictor:8000/v1/models; do
+                echo "LLM not ready, retrying in 10s..."
+                sleep 10
+              done
+              echo "LLM endpoint is ready."
+        - name: emb-wait
+          image: "registry.access.redhat.com/ubi9/ubi-minimal:latest"
+          command:
+            - "/bin/sh"
+            - "-c"
+            - |
+              echo "Waiting for embedding endpoint to become ready..."
+              until wget -q -O /dev/null http://embedding-predictor:8080/v1/models; do
+                echo "Embedding not ready, retrying in 10s..."
+                sleep 10
+              done
+              echo "Embedding endpoint is ready."
       containers:
         - name: backend-server
           image: "{{ .Values.backend.image }}"
--- a/ai-services/assets/services/chat/openshift/templates/chat-bot-backend-deployment.yaml
+++ b/ai-services/assets/services/chat/openshift/templates/chat-bot-backend-deployment.yaml
@@ -22,31 +22,6 @@ spec:
         ai-services.io/component-type: api
     spec:
       automountServiceAccountToken: false
-      initContainers:
-        - name: llm-wait
-          image: "registry.access.redhat.com/ubi9/ubi-minimal:latest"
-          command:
-            - "/bin/sh"
-            - "-c"
-            - |
-              echo "Waiting for LLM endpoint to become ready..."
-              until curl -sf http://llm-predictor:8000/v1/models > /dev/null; do
-                echo "LLM not ready, retrying in 10s..."
-                sleep 10
-              done
-              echo "LLM endpoint is ready."
-        - name: emb-wait
-          image: "registry.access.redhat.com/ubi9/ubi-minimal:latest"
-          command:
-            - "/bin/sh"
-            - "-c"
-            - |
-              echo "Waiting for embedding endpoint to become ready..."
-              until curl -sf http://embedding-predictor:8080/v1/models > /dev/null; do
-                echo "Embedding not ready, retrying in 10s..."
-                sleep 10
-              done
-              echo "Embedding endpoint is ready."
       containers:
         - name: backend-server
           image: "{{ .Values.backend.image }}"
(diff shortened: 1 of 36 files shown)


ai-services/assets/services/translate/openshift/templates/translate-api-deployment.yaml
```diff
@@ -0,0 +1,140 @@
...
+          command:
+            - "/bin/sh"
+            - "-c"
+            - |
+              echo "Waiting for LLM endpoint to become ready..."
+              until curl -sf http://llm-predictor:8000/v1/models > /dev/null; do
+                echo "LLM not ready, retrying in 10s..."
+                sleep 10
+              done
+              echo "LLM endpoint is ready."
```
mayuka-c: Any reason for adding initContainers?
If the llm pod is not ready then the services pods will not be deployed at all. First the components gets deployed and once ready services will get deployed.

@Shubhamag12 I hope this is how its done even for openshift
cc. @yussufsh

ai-services/assets/services/chat/openshift/templates/chat-bot-backend-deployment.yaml
```diff
@@ -22,6 +22,31 @@ spec:
         ai-services.io/component-type: api
     spec:
       automountServiceAccountToken: false
+      initContainers:
+        - name: llm-wait
+          image: "registry.access.redhat.com/ubi9/ubi-minimal:latest"
+          command:
+            - "/bin/sh"
+            - "-c"
+            - |
```
mayuka-c: Same here and other places where initContainers are added for component readiness

ai-services/assets/services/extract/openshift/values.yaml
```diff
@@ -0,0 +1,22 @@
...
+postgres:
+  image: icr.io/ai-services-cicd/postgres:18-4
+  storage: 5Gi
+  # @generate:password
+  password: ""
+
+llm:
+  model:
+    name: ""
+  apiKey: ""
```
mayuka-c: Also is this required?. This should not be
@Shubhamag12 ^^

ai-services/assets/services/translate/openshift/templates/translate-api-deployment.yaml
dharaneeshvrd: I observed both the components and service pods were deployed simultaneously, and the service pods kept crashing until the LLM pod was up and running. Hence, this init container was added to prevent crash loops. It also provides the same layering logic that we have for Podman. Since these services are not useful without the LLM, it makes sense to validate the LLM's health before spawning them.
But if there is a helm way that we can get this dependency check, please let me know, I ll implement the same.

Shubhamag12: Helm was marking components ready as soon as inference service was deployed, since it did not look for underlying deployments to be ready.
This PR fixes it: https://github.com/IBM/project-ai-services/pull/1319
@dharaneeshvrd above initContainer is not required once my changes go in

ai-services/assets/services/extract/openshift/values.yaml
dharaneeshvrd: @Shubhamag12 can you please confirm whether should I remove this or keep it? seems its there in existing services as well.

Shubhamag12: yes, I dont think so this is required. Let me check and confirm

ai-services/assets/services/translate/openshift/templates/translate-api-deployment.yaml
dharaneeshvrd: can we include db also in the waitlist PR before bringing up the services? since services would try to setup db & tables during the startup. It is better to replicate the layering logic implemented for podman here too.

mayuka-c: This will not work for db as db is part of services and its not a component. So each service would need a initContainer to wait until its db is ready

Shubhamag12: dbs are part of service charts and when service is deployed, all pods are deployed as part of it, including db.
The only way we can implement podman layering in helm is by using helm hook and weights. If that is okay, I can work on that.
cc: @yussufsh

mayuka-c: Nah that is not considered good practice, init container is the best practice for that

ai-services/assets/services/extract/openshift/values.yaml
Shubhamag12: this can be removed @dharaneeshvrd

dharaneeshvrd: Remove this on new and existing services.

ai-services/assets/services/chat/openshift/templates/chat-bot-backend-deployment.yaml
dharaneeshvrd: Removed

ai-services/assets/services/translate/openshift/templates/translate-api-deployment.yaml
dharaneeshvrd: Removed llm-wait and added db-wait initContainer ptal.

ai-services/assets/services/extract/openshift/templates/extract-api-deployment.yaml
```diff
@@ -0,0 +1,138 @@
...
+        - name: extract-db-wait
+          image: "{{ .Values.postgres.image }}"
+          command:
+            - "/bin/sh"
+            - "-c"
+            - |
+              until pg_isready -h extract-db -p 5432; do
+                echo "Waiting for extract-db to be ready..."
+                sleep 5
+              done
```
mayuka-c, line 42: Can we add timeouts to this so that if db is not brought up, it doesnt keep waiting indefinitely?

mayuka-c (commented):
Rest LGTM

ai-services/assets/services/extract/openshift/templates/extract-api-deployment.yaml
dharaneeshvrd: Agree, added a timeout. Thanks

dharaneeshvrd:
@mayuka-c can you please approve this if there are no more comments?

manalilatkar (approved):
LGTM

mayuka-c (approved):
LGTM

PR #1329, opened 2026-08-26
AISERVICES-1573: Adding service base lite image support

adarshagrawal38:
Introduce `service-base-lite` a lightweight base image for services that do not require `Docling`.

The existing `service-base` image bundles Docling packages and its associated models, making it unnecessarily heavy for services that never use document parsing. This PR introduces a new `service-base-lite` base image as a leaner alternative.

All direct third-party dependencies were audited across the following services by scanning every Python source file for external imports:
- chatbot
- summarize
- extract
- translate
- similarity

The identified packages have been pinned and added to `images/service-base-lite/requirements.txt`, keeping the dependency footprint minimal (17 packages vs 144 in service-base). Services that do not rely on Docling can now use this lighter base image, reducing build times and image size

### DIFF
--- a/images/service-base-lite/requirements.txt
+++ b/images/service-base-lite/requirements.txt
@@ -1,0 +1,17 @@
+fastapi==0.135.3
+httpx==0.28.1
+jsonschema==4.26.0
+jsonschema-specifications==2025.9.1
+lingua-language-detector==2.2.0
+psycopg2==2.9.11
+pydantic==2.13.4
+pydantic-settings==2.14.2
+pypdfium2==5.9.0
+python-dotenv==1.2.2
+python-multipart==0.0.32
+requests==2.34.2
+sentence-splitter==1.4
+SQLAlchemy==2.0.49
+starlette==1.3.1
+uvicorn==0.44.0
+uvloop==0.22.1
--- a/.github/workflows/service-base-lite.yml
+++ b/.github/workflows/service-base-lite.yml
@@ -1,0 +1,33 @@
+name: Service Base Lite
+on:
+  push:
+    paths:
+      - 'images/service-base-lite/**'
+    branches:
+      - main
+      - 'release-*'
+  pull_request:
+    paths:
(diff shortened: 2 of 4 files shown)


images/service-base-lite/requirements.txt
```diff
@@ -0,0 +1,17 @@
+fastapi==0.135.3
+httpx==0.28.1
+jsonschema==4.26.0
+jsonschema-specifications==2025.9.1
+lingua-language-detector==2.2.0
+psycopg2==2.9.11
+pydantic==2.13.4
+pydantic-settings==2.14.2
+pypdfium2==5.9.0
```
dharaneeshvrd, line 9: this is not required for all the services, let's not include

dharaneeshvrd (commented):
@manalilatkar @sats-23 @Niharika0306 
team let's review this requirements
we need to ensure the common python dependencies across all the services should go in this image.

sats-23:
Can we instead move docling specific requirements to digitize requirements and keep the service base as lean as the proposed service base lite?

images/service-base-lite/requirements.txt
adarshagrawal38: summarize needs `pypdfium2` pkg right.

Then summarize will make use of `service-base` image?

mkumatag:
I think adding one more image will just add the complexity, I vote for just having same image and share across all the services!

adarshagrawal38:
Image size stats

| Image | Size |
| -------- | -------- | 
| icr.io/ai-services-cicd/service-base | 3.45 GB |
| icr.io/ai-services-private/service-base-lite | 630 MB |

Note: Size of model directory
```
[root@30b666f6d646 /]# du -sh /var/docling-models/
506M	/var/docling-models/
```

mkumatag:
> I think adding one more image will just add the complexity, I vote for just having same image and share across all the services!

as per the offline discussion - let us use lite -> service-base and create a image which contains docling and other packages for the digitize-base

PR #1331, opened 2026-08-26
fix: Avoid bundles vol delete during skip-cleanup

mayuka-c:
- Skip deleting catalog bundles volume if skip-cleanup is used during catalog uninstall

### DIFF
--- a/ai-services/assets/catalog/openshift/templates/catalog-bundles-pvc.yaml
+++ b/ai-services/assets/catalog/openshift/templates/catalog-bundles-pvc.yaml
@@ -2,6 +2,8 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: catalog-bundles
+  annotations:
+    "helm.sh/resource-policy": keep
   labels:
     ai-services.io/application: {{ .Release.Name }}
     ai-services.io/template: {{ .Chart.Name }}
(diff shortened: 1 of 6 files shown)


ai-services/assets/catalog/openshift/templates/catalog-bundles-pvc.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: catalog-bundles
+  annotations:
+    "helm.sh/resource-policy": keep
```
adarshagrawal38, line 6: When skip-cleanup flag is not passed, for that scenario we must cleanup `catalog-bundle` pvc right.

I fell we should cleanup pvc [here](https://github.com/IBM/project-ai-services/blob/main/ai-services/internal/pkg/catalog/cli/uninstall/openshift/uninstall.go#L45)

mayuka-c: Its already taken care in DeletePVCs right. Based on the labels it will delete. https://github.com/IBM/project-ai-services/blob/main/ai-services/internal/pkg/catalog/cli/uninstall/openshift/uninstall.go#L47

yussufsh (dismissed):

adarshagrawal38 (dismissed):
LGTM

adarshagrawal38 (approved):

PR #1332, opened 2026-08-26
Implement httpproxy worker grpc command

yussufsh:
`HTTPProxy` command implementation.
This is just for forwarding HTTP requests to the worker node and returning the response.

### DIFF
--- a/ai-services/internal/pkg/runtime/podman/podman.go
+++ b/ai-services/internal/pkg/runtime/podman/podman.go
@@ -1,10 +1,14 @@
 package podman
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"io"
+	"net"
+	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"os/signal"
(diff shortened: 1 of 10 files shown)


ai-services/internal/pkg/runtime/podman/podman.go
```diff
@@ -815,3 +819,131 @@ func (pc *PodmanClient) ExecInContainerWithCmd(_, _ string, _ []string) (string,
...
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("HTTPProxy: execute request: %w", err)
+	}
+	defer func() {
+		if closeErr := resp.Body.Close(); closeErr != nil {
+			logger.Warningf("HTTPProxy: close response body: %v", closeErr)
+		}
+	}()
```
mayuka-c: Can we use resty package for HTTP calls?

ai-services/internal/pkg/runtime/podman/podman.go
```diff
@@ -815,3 +819,131 @@ func (pc *PodmanClient) ExecInContainerWithCmd(_, _ string, _ []string) (string,
...
 }
+
+// ─── HTTP proxy tunnel ────────────────────────────────────────────────────────
+
+// HTTPProxy makes an HTTP request to targetURL from the worker node and returns
+// the response to the control plane. The request executes against a local pod
+// endpoint so the control plane can reach pods that are not externally routed.
+//
+// The targetURL hostname may be a Podman pod name (e.g. "my-app--chat-bot").
+// Pod name DNS resolution only works inside Podman-networked containers, not
```
mayuka-c: Will this be used anywhere?. Is it for Caddy pod?

ai-services/internal/pkg/runtime/podman/podman.go
```diff
@@ -815,3 +819,131 @@ func (pc *PodmanClient) ExecInContainerWithCmd(_, _ string, _ []string) (string,
...
+
+	return u.String(), nil
+}
+
+// podNameToIP inspects the named pod and returns its infra-container IP address.
+func (pc *PodmanClient) podNameToIP(podName string) (string, error) {
+	podReport, err := pods.Inspect(pc.Context, podName, nil)
+	if err != nil {
+		return "", fmt.Errorf("inspect pod %q: %w", podName, err)
+	}
```
mayuka-c, line 958: If I'm not wrong for a non root user, this will not work. Please check it out once.
A daemon/process running on host will not be able to directly use podIP to make calls

ai-services/internal/pkg/runtime/podman/podman.go
yussufsh: These are for podman network ports that are not accessible via Caddy or HostPort.

ai-services/internal/pkg/runtime/podman/podman.go
yussufsh: Yes you are right, this will not work for rootless, hence I wanted to run the worker daemon process on a container instead for this routing to work. I will add a TODO here may be so we can merge this PR.

ai-services/internal/pkg/runtime/podman/podman.go
yussufsh: Sure we can do that.

ai-services/internal/pkg/runtime/podman/podman.go
mayuka-c: Yeah but I dont think the control plane will talk to any of the pods deployed on workers using the podman network right?
Even for data source connectors, it will use the digitize endpoint

mayuka-c (approved):
LGTM

PR #1333, opened 2026-08-26
Fix the issue with order for datasource params (raw data approach)

prachikurade1:
UI needs the params in a fixed order. Currently for json it was getting sorted alphabetically even if the file order was fixed, hence it needed the raw data to be sent.


mayuka-c (dismissed):
LGTM

yussufsh (dismissed):

yussufsh (approved):

PR #1334, opened 2026-08-26
feat: Add DataSource Connector with  dynamic form, cache, and API integration

Niranjan-J-Joshi:
Adds new flow to create Data Sources Connectors from the Connectors page using a dynamic form. 

## Implementation
 
- Added an Add Data Source SidePanel.
- Fetches available connector types from the API.
- Dynamically renders the connector form.
- Added required field validation.
- Added POST /datasources integration.
- Added cache to store params API response 
- Prefetches connector data once the table loads.
- Normalizes private keys before submitting.
- Resets the form after closing.
- Refreshes the table after a successful create.
- Added loading and error states.

## API Endpoints
 
| Method | Endpoint | Purpose |
|---|---|---|
| GET | /connectors?connector_type=datasource | Get connector types |
| GET | /connectors/datasource/providers/:id/params | Get connector params schema |
| POST | /datasources | Create a data source |
 
## Known Gaps
 
- Inline error messages for the fields after form submission


ui/catalog/src/store/connectors.store.ts
```diff
@@ -0,0 +1,198 @@
...
+          cacheVersion: CACHE_VERSION,
+          connectorTypes: [],
+          connectorTypesError: null,
+          connectorTypesFetchedAt: null,
+          paramsCache: {},
+          paramsCacheError: {},
+        }),
+
+      // Initialize store and validate cache version at runtime
+      initialize: () => {
```
ryanmarc, line 163: initialize() has no caller. `grep -rn "useConnectorsStore" src/`gives four hits: this file, the table, the modal, and auth.ts (which calls clearAll() on logout). The deploy store's equivalent is called from DigitalAssistantDeployFlow.tsx, so I think this one just got missed.

ryanmarc (changes requested):
This needs changes based on what the api response shape will actually be for connectors. https://github.com/IBM/project-ai-services/pull/1346

ui/catalog/src/store/connectors.store.ts
Niranjan-J-Joshi: Added caller for initialize()

ryanmarc (dismissed):

pranithraoibm (approved):

PR #1337, opened 2026-08-26
Add summarization ft

vijay-vin:
Add three Summarization Failure testcases on top of the existing ones
1) Invalid level via JSON body
2) Both level and length supplied
3) Connectivity failure — unreachable summarize API

Test results were added in the Jira ticket, 
https://jsw.ibm.com/browse/AISERVICES-1442

### DIFF
--- a/ai-services/tests/e2e/e2e_suite_test.go
+++ b/ai-services/tests/e2e/e2e_suite_test.go
@@ -2629,7 +2629,7 @@ var _ = ginkgo.Describe("AI Services End-to-End Tests", ginkgo.Ordered, func() {
 			logger.Infof("[TEST] ✓ binary TXT rejected (status=%d, msg=%s)", statusCode, errResp.Error.Message)
 		})
 
-		// ── TC-2: Invalid level via JSON body ─────────────────────────────────
+		// ──  Invalid level via JSON body ─────────────────────────────────
 		//
 		// Rationale: The existing test "returns 400 for invalid level parameter
 		// via multipart form" only covers the multipart/form-data code path.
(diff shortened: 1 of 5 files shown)


manju956 (commented):
looks good.

manju956:
> Add three Summarization Failure testcases on top of the existing ones
> 
> 1. Invalid level via JSON body
> 2. Both level and length supplied
> 3. Connectivity failure — unreachable summarize API
> 
> Test results were added in the Jira ticket, https://jsw.ibm.com/browse/AISERVICES-1442

@vijay-vin pls bump up the versions

vijay-vin:
@manju956 I have now updated the Makefile tag.

Abhishek-Kushwaha1:
looks good @vijay-vin @manju956

Abhishek-Kushwaha1 (approved):

ai-services/tests/e2e/e2e_suite_test.go
```diff
@@ -2623,6 +2629,82 @@ var _ = ginkgo.Describe("AI Services End-to-End Tests", ginkgo.Ordered, func() {
 			logger.Infof("[TEST] ✓ binary TXT rejected (status=%d, msg=%s)", statusCode, errResp.Error.Message)
 		})
 
+		// ──  Invalid level via JSON body ─────────────────────────────────
```
manju956, line 2632: I guess summarize is not part of e2e test suite, it runs as separate pipeline. @Abhishek-Kushwaha1 Can we not integrate into e2e suite then?

Abhishek-Kushwaha1: we can but it will run at the end only then because it uses diff template for creation. 
so better to make it independent i think, we can include if you want .

PR #1342, opened 2026-08-27
AIS-1535 Digitize service table summarization failure returns error

anshuman-agarwala:
Modified Digitize service to return error on table summarization failure

### DIFF
--- a/services/common/llm_utils.py
+++ b/services/common/llm_utils.py
@@ -1,7 +1,6 @@
 import logging
 import os
 import requests
-import threading
 import time
 import json
 from concurrent.futures import ThreadPoolExecutor, as_completed
@@ -56,7 +55,7 @@ def tqdm_wrapper(iterable, **kwargs):
         return iterable
 
 @retry_on_transient_error(max_retries=3, initial_delay=1.0, backoff_multiplier=2.0)
-def summarize_and_classify_single_table(prompt, gen_model, llm_endpoint, max_tokens: int = 1024, cancel_event=None):
+def summarize_and_classify_single_table(prompt, gen_model, llm_endpoint, max_tokens: int = 1024):
     """Combined function to summarize and classify a table in a single LLM call.
 
     Returns tuple: (summary, decision).
@@ -64,9 +63,6 @@ def summarize_and_classify_single_table(prompt, gen_model, llm_endpoint, max_tok
     """
     if misc_utils.SESSION is None:
         raise RuntimeError("LLM session not initialized. Call create_llm_session() first.")
-
-    if cancel_event is not None and cancel_event.is_set():
-        return "No summary.", False
 
     payload = {
         "model": gen_model,
@@ -127,22 +123,21 @@ def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, doc_path, 
 
     Returns tuple: (summaries, decisions, failures).
     failures is a dict mapping table index to error message for failed tables.
+    Table failures are non-fatal: failed tables use fallback values and processing continues.
     """
     all_prompts = [prompt_template.format(content=md) for md in table_mds]
 
     if not all_prompts:
         return [], [], {}
 
-    results: list[tuple[str, bool] | None] = [None] * len(all_prompts)
-
-    cancel_event = threading.Event()
+    results: list[tuple[str, bool]] = [("No summary.", False)] * len(all_prompts)
     summaries: list[str] = []
     decisions: list[bool] = []
     failures: dict[int, str] = {}
 
     with ThreadPoolExecutor(max_workers=min(max_workers, len(all_prompts))) as executor:
         futures = {
-            executor.submit(summarize_and_classify_single_table, prompt, gen_model, llm_endpoint, max_tokens, cancel_event): idx
+            executor.submit(summarize_and_classify_single_table, prompt, gen_model, llm_endpoint, max_tokens): idx
             for idx, prompt in enumerate(all_prompts)
         }
         for future in tqdm_wrapper(as_completed(futures), total=len(all_prompts),
@@ -155,20 +150,12 @@ def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, doc_path, 
                 logger.error(error_msg)
                 results[idx] = ("No summary.", False)
                 failures[idx] = error_msg
-                cancel_event.set()
-                break
 
-    # Separate summaries and decisions with proper None handling
-    for idx, result in enumerate(results):
-        if result is not None:
-            summary, decision = result
-            summaries.append(summary)
-            decisions.append(decision)
-        else:
-            # Default values for failed futures
-            summaries.append("No summary.")
-            decisions.append(False)
-            failures[idx] = "Unknown error: result is None"
+    # Unpack results in index order. Every slot starts as the fallback
+    # and is overwritten either by a successful future or left as-is on failure.
+    for summary, decision in results:
+        summaries.append(summary)
+        decisions.append(decision)
 
     return summaries, decisions, failures
 
--- a/services/common/llm_utils.py
+++ b/services/common/llm_utils.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import requests
+import threading
 import time
 import json
 from concurrent.futures import ThreadPoolExecutor, as_completed
@@ -55,13 +56,17 @@ def tqdm_wrapper(iterable, **kwargs):
         return iterable
 
 @retry_on_transient_error(max_retries=3, initial_delay=1.0, backoff_multiplier=2.0)
-def summarize_and_classify_single_table(prompt, gen_model, llm_endpoint, max_tokens: int = 1024):
+def summarize_and_classify_single_table(prompt, gen_model, llm_endpoint, max_tokens: int = 1024, cancel_event=None):
     """Combined function to summarize and classify a table in a single LLM call.
 
     Returns tuple: (summary, decision).
+    Raises on failure so that the retry decorator and the caller can handle it.
     """
     if misc_utils.SESSION is None:
         raise RuntimeError("LLM session not initialized. Call create_llm_session() first.")
+
+    if cancel_event is not None and cancel_event.is_set():
+        return "No summary.", False
 
     payload = {
         "model": gen_model,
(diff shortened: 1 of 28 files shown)


services/common/llm_utils.py
```diff
@@ -134,23 +136,32 @@ def summarize_and_classify_tables(table_mds, gen_model, llm_endpoint, doc_path,
         for future in tqdm_wrapper(as_completed(futures), total=len(all_prompts),
                                    desc=f"Summarizing and classifying tables of '{doc_path}'"):
             idx = futures[future]
-            results[idx] = future.result()
-
-    # Separate summaries and decisions with proper None handling
+            try:
+                results[idx] = future.result()
+            except Exception as e:
+                error_msg = f"Failed to process table {idx}: {str(e)}"
```
dharaneeshvrd, line 149: can we stop summarising remaining tables in case of failure?

also need to consider the `retry_on_transient_error ` wrapper usage as well.
even after the retry if it fails, we should mark the doc as failed and stop summarizing remaining tables.

anshuman-agarwala: I've added support to end summarisation on any table's failure. 
However, this will only stop threads which are just starting or are yet to start. 
From what I can tell, there is no way of stopping already running threads.
I've also modified `summarize_and_classify_single_table` to handle `retry_on_transient_errorretry_on_transient_error` properly.

services/common/tests/test_llm_utils_table_summarize.py
```diff
@@ -0,0 +1,298 @@
...
+        assert decision is True
+
+    @patch("common.llm_utils.get_vllm_headers", return_value={})
+    @patch("common.llm_utils.settings")
+    def test_cancelled_before_call(self, mock_settings, mock_headers):
+        """Returns (No summary., False) immediately if cancel_event is set before the call."""
+        from common.llm_utils import summarize_and_classify_single_table
+        import common.misc_utils as misc_utils
+
+        misc_utils.SESSION = Mock()
```
sats-23: Use `monkeypatch.setattr(misc_utils, "SESSION", mock_session)` instead
The ordering of tests in CI after misc_utils.SESSION override is causing the failure in test_middleware.py::test_middleware_with_different_endpoints

anshuman-agarwala: But the tests should not be order dependent, right?
I can reproduce the same failure locally by running:
```
PYTHONPATH=common:chatbot:./services .venv/bin/pytest \
    services/chatbot/tests/unit/test_middleware.py \
    services/chatbot/tests/unit/test_endpoints.py \
    -v --tb=short
```

sats-23: Tests are not order dependent. 
The instance of session behaves differently in either case. 
With monkeypatch.setattr, it reset after each test, but the current implementation, it does not reset after.

sats-23: All tests throughout the suite are using monkeypatch.setattr, that's why they never came through this issue

anshuman-agarwala: Shifting to monkeypatch fixed the issues with the test failure, thanks.

services/digitize/processing/tables.py
```diff
@@ -346,12 +346,18 @@ def process_table(converted_doc, doc_path, out_path, gen_model, gen_endpoint, do
...
         table_markdowns, gen_model, gen_endpoint, doc_path,
         prompt_template=selected_prompt,
         max_tokens=selected_max_tokens,
     )
 
+    # Log failures for visibility
+    if failures:
+        logger.warning(f"Table summarization failures for '{doc_path}': {len(failures)} table(s) failed")
+        for idx, error in failures.items():
+            logger.error(f"Table {idx} failed: {error}")
```
dharaneeshvrd: let's return from here

services/common/llm_utils.py
dharaneeshvrd: After looking at this, I am thinking maybe we should not fail the entire flow and use a status like partially_ingested.
Not a good design to stop the ingestion of the entire document since there is a problem with table summarising. Because table is not the only entity in a document, it would have a lot useful text as well.  
Can we please remove this cancel_event logic and continue the document to go through the remaining ingestion stages and mark the document as partially_ingested? 
Sry for the design change.

services/requirements-test.txt
```diff
@@ -51,3 +51,4 @@ docling==2.95.0
 sentence-splitter==1.4
 
 apscheduler[sqlalchemy]==4.0.0a6
+grpcio==1.80.0
```
dharaneeshvrd: why are we adding this?

services/extract/models.py
```diff
@@ -182,7 +182,6 @@ class ExtractionRequest(BaseModel):
     text: str = Field(..., min_length=1, description="Raw text to extract from")
     schema_id: str = Field(..., min_length=1, description="ID of a registered schema")
 
-
```
dharaneeshvrd, line 185: revert this

services/digitize/processing/tables.py
dharaneeshvrd: already we are logging in llm_utils, let's keep only one instance

services/digitize/models.py
```diff
@@ -28,6 +29,7 @@ class DocStatus(str, Enum):
     PROCESSED = "processed"
     CHUNKED = "chunked"
     COMPLETED = "completed"
+    PARTIALLY_INGESTED = "partially_ingested"
```
dharaneeshvrd: I feel we can use the same term COMPLETED_WITH_ERRORS here as well to align with peer status of doc as well as with the Job status .

dharaneeshvrd (commented):
overall lgtm

dharaneeshvrd (commented):
@sats-23 
Can you also please take a look once?
We are adding completed_with_errors status in case of table summarization getting failed.
Changes are touching the sync tick as well.

services/digitize/models.py
anshuman-agarwala: Done

services/digitize/processing/tables.py
anshuman-agarwala: I've removed this

services/extract/models.py
anshuman-agarwala: Done

services/requirements-test.txt
anshuman-agarwala: Removed this

services/digitize/utils/jobs.py
```diff
@@ -59,17 +61,27 @@ def get_job_document_stats(job_id: str) -> dict:
...
             doc for doc in documents
-            if doc.get("status") in (DocStatus.COMPLETED.value, DocStatus.ALREADY_EXISTS.value)
+            if doc.get("status") == DocStatus.COMPLETED_WITH_ERRORS.value
         ]
 
         return {
             "failed_docs": failed_docs,
             "completed_docs": completed_docs,
+            "partial_docs": completed_with_errors_docs,
+            "completed_with_errors_docs": completed_with_errors_docs,
```
dharaneeshvrd: seems dup, please take care

services/digitize/utils/db.py
```diff
@@ -1084,11 +1090,12 @@ def _update_document(
         if update_params:
             success = db_manager.update_document(doc_id, **update_params)
             if success:
-                # Register the checksum once the document is marked COMPLETED so
-                # that future uploads of the same content are caught via the
-                # document_checksum table.
+                # Register the checksum once the document reaches any terminal
+                # success state (COMPLETED or COMPLETED_WITH_ERRORS) so that future
+                # uploads of the same content are caught via document_checksum.
+                _terminal_ok_doc = (DocStatus.COMPLETED, DocStatus.COMPLETED_WITH_ERRORS)
```
dharaneeshvrd: I think we need to avoid setting the file hash for the COMPLETED_WITH_ERRORS doc, since it would stop the document from processing again when its submitted, ideally we should process in case of resubmission.
So let's keep the logic as is.
@sats-23 please take a look once here

anshuman-agarwala: I've modified it to only hash the document on Completed status

services/digitize/utils/jobs.py
anshuman-agarwala: Done

services/digitize/utils/db.py
```diff
@@ -1124,42 +1130,46 @@ def _update_job(
...
             total_docs = stats["total_documents"]
             completed_docs = stats["completed"]
             failed_docs = stats["failed"]
 
             if total_docs > 0 and (completed_docs + failed_docs) == total_docs:
                 update_params["completed_at"] = datetime.now(timezone.utc)
 
-        # Set error if provided
+        # Set error only for hard failures
         if error and job_status == JobStatus.FAILED:
```
dharaneeshvrd: let's set the error also for COMPLETED_WITH_ERRORS if error is not none.

anshuman-agarwala: Done

dharaneeshvrd (dismissed):

services/digitize/db/scripts/init_schema.sql
```diff
@@ -162,4 +162,18 @@ BEGIN
 END
 $$;
 
--- Note: Using postgres superuser, no additional grants needed
\ No newline at end of file
+-- Note: Using postgres superuser, no additional grants needed
+
+-- Schema migrations — idempotent ALTER statements to update existing databases.
```
sats-23: Do we need these alter statements? @dharaneeshvrd 
We've always assumed fresh deployment. In case we need, we will have to revisit our diff since last release and have alter queries for all tables

sats-23: It has to be planned in synergy with upgrade and rollback feature

sats-23:
How does the UI look? Does it interpret Completed with Error as Completed? Should we consider making edits there for transparency? 

Update: Just saw the screenshots on the ticket. The UI also needs mapping for the new status. Maybe a yellow check mark similar to already_exists can be reused next to the completed_with_errors status. 
Right now it looks inconsistent for other statuses to have an icon and for this to not.

services/digitize/utils/db.py
```diff
@@ -1084,9 +1090,9 @@ def _update_document(
         if update_params:
             success = db_manager.update_document(doc_id, **update_params)
             if success:
-                # Register the checksum once the document is marked COMPLETED so
-                # that future uploads of the same content are caught via the
-                # document_checksum table.
+                # Register the checksum only when the document completed fully
+                # (not COMPLETED_WITH_ERRORS) so that a re-upload of a document
+                # that had table failures is not incorrectly de-duplicated.
```
sats-23: This is something to consider. 
If we have a table failure and proceed to ingest to VDB but don't track the checksum, we might end up re-ingesting it in the next sync, so we don't really have a track for the old partial ingestion. 
Now in case the LLM instance is still down, we could lead to uncontrollably and untraceable growing of VDB size even if a connector is attached for overnight (every 5 mins it ingests and fails and repeats)

services/digitize/db/scripts/init_schema.sql
dharaneeshvrd: Might not be required, since we don't have inplace upgrade mechanism, what we follow is, use /v1/import API to backup the DB data and use /v1/export API to import it into the new deployment. We would be able to handle deviated fields in export API by assigning them null. 
So we can update the create statement itself for now to add this new field.

services/digitize/utils/jobs.py
```diff
@@ -59,17 +61,25 @@ def get_job_document_stats(job_id: str) -> dict:
 
         documents = job_data.get("documents", [])
         failed_docs = [doc for doc in documents if doc.get("status") == DocStatus.FAILED.value]
-        completed_docs = [
+        _terminal_ok = (
```
dharaneeshvrd, line 64: we need to remove DocStatus.COMPLETED_WITH_ERRORS here or else those docs will get counted on both cases.

dharaneeshvrd:
> Maybe a yellow check mark similar to already_exists can be reused next to the completed_with_errors status.
Right now it looks inconsistent for other statuses to have an icon and for this to not.

let's do all UI changes in a single PR
we can merge this without the UI changes for now.

services/digitize/utils/db.py
dharaneeshvrd: Agree this is worse than missing out on the table data during re-ingestion.
@anshuman-agarwala sry, can you please revert to the pervious version?
When user wants to ingest it properly again, then they can clean up the doc and re-ingest it by seeing the completed_with_error status.

dharaneeshvrd (dismissed):

dharaneeshvrd:
@sats-23 please merge if things are looking good

services/digitize/db/scripts/init_schema.sql
sats-23: @anshuman-agarwala Please remove the alter statements

dharaneeshvrd: yes

anshuman-agarwala: Done

sats-23 (approved):
LGTM

PR #1345, opened 2026-08-27
feat: add Register worker resource

pranithraoibm:
- Added Register worker resource modal with name input, Generate token button, and a run command snippet shown after token is issued
- Modal state managed via reducer at the page level — same pattern as the rest of the codebase
- Wired up `POST /workers` API call with loading, success, and error states

https://github.com/user-attachments/assets/d0806f87-96c7-4441-9c00-519cae6c7e18


ui/catalog/src/components/WorkerResourcesTable/RegisterWorkerModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const isLoading = phase === "loading";
+  const isSuccess = phase === "success";
+
+  const runCommand = useMemo(() => {
+    if (!isSuccess) return "";
+    return [
+      "ai-services worker join \\",
+      "  --server <host>:<port> \\",
+      `  --name "${workerName}" \\`,
+      `  --token "${token}"`,
```
ryanmarc: This command will not run as written. worker join takes the gateway as a positional argument (Use: "join <gateway>" with Args: cobra.ExactArgs(1) in cmd/worker/join.go) and the flags it registers are --token, --runtime, --basedir, --https-port, --domain-name, --ssl-cert, --ssl-key. There is no --server and no --name - I grepped cmd/ai-services/cmd/worker/ for both and neither exists.

Pasting this fails twice: cobra rejects --server as an unknown flag, and because the host got eaten as a flag value there are no positional args left, so ExactArgs(1) fails too.

--name is also not needed even as a concept. registry.go:112 says the worker name comes from the validated token, not from the caller, and Preregister binds it in via IssueToken(workerName) - the daemon already knows who it is.

The command's own Example block has the shape:

`return ["ai-services worker join <host>:<port> \\", `  --token "${token}"`].join("\n");`

ui/catalog/src/pages/WorkerResources/WorkerResources.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const handleGenerateToken = useCallback(async () => {
+    if (!state.workerName.trim()) {
+      dispatch({ type: "SET_PHASE", payload: "invalid" });
+      return;
+    }
+    dispatch({ type: "SET_PHASE", payload: "loading" });
+    try {
+      const result = await registerWorker(state.workerName.trim());
+      dispatch({ type: "REGISTER_SUCCESS", payload: { token: result.token } });
```
ryanmarc: registerWorker returns { worker_name, token } and this only takes token. The server lowercases and trims the name before storing it (worker_handler.go:58, with a comment saying so explicitly) and hands the canonical value back in the response.

So if someone types Worker-A: the modal keeps displaying Worker-A, the run command says Worker-A, and the table underneath refreshes from GET /workers and says worker-a. Two names for the same worker on screen at once.

WorkerRegisterResponse already declares worker_name, so it is just a matter of threading it through:
```
dispatch({
  type: "REGISTER_SUCCESS",
  payload: { token: result.token, workerName: result.worker_name },
});
```
and having REGISTER_SUCCESS overwrite state.workerName with it. That way the readonly field and the run command both show what the worker will actually be called.

ryanmarc (changes requested):

ui/catalog/src/components/WorkerResourcesTable/RegisterWorkerModal.tsx
pranithraoibm: Yes command was updated again post this PR was raised. The final PR of this workers https://github.com/IBM/project-ai-services/pull/1424 had the updated command. Will update it here as well.

ryanmarc (dismissed):

Niranjan-J-Joshi (approved):

PR #1346, opened 2026-08-27
Changes to the connectors API

prachikurade1:
Change the API response for connectors API.
Filtered out redundant fields like the type and name of the asset which was always connectors.
Updated response format:
```
[{
  "type": "datasource",
  "name": "Data sources",
  "provider": {
    "id": "file_system",
    "name": "File system",
    "description": "SSH based connection...",
    "schema": "/api/v1/connectors/datasource/providers/filesystem/params"
  }
}]
```


prachikurade1:
image

ai-services/docs/docs.go
```diff
@@ -2619,31 +2619,41 @@ const docTemplate = `{
...
-                "connector_type": {
-                    "description": "e.g. \"datasource\"",
+                "id": {
                     "type": "string"
                 },
-                "description": {
-                    "description": "short description",
+                "name": {
                     "type": "string"
                 },
```
yussufsh, line 2640: Minor nit: Can we keep the description fields? Seems swagger derives this from the comments in the struct.

prachikurade1: Done

yussufsh (approved):

PR #1350, opened 2026-08-28
Deploy remote application code with create cli

yussufsh:
**Remote worker deployment support**

- **Worker targeting** — `ai-services create --worker <name>` deploys an application to a named remote worker node instead of the local machine. Defaults to `"local"` (existing behaviour unchanged).

- **Worker validation** — before any DB records are written, `ValidateWorker` confirms the named worker is connected and ready (registry cache + DB check), returning HTTP 400 immediately on failure.

- **Remote runtime routing** — `DeploymentExecutor` resolves the worker's runtime type from the registry, constructs a `RemoteRuntime` backed by the gRPC `CommandStream`, and delegates to the standard Podman or OpenShift deployer — no deployer code needed to know it's talking to a remote machine.

- **Worker metadata injection** — for remote Podman workers, `PodmanWorkerConfig` is populated from the worker's registration metadata (`domain-suffix`, `https-port`, `base-dir`) so Caddy configuration and model paths use the worker's values rather than local environment variables.

- **Model download** — `downloadModels` sends `COMMAND_TYPE_DOWNLOAD_MODEL` over the gRPC stream for remote workers; calls `helpers.DownloadModelContainer` directly for local.

- **Proto cleanup** — `COMMAND_TYPE_RUN_EPHEMERAL_CONTAINER` renamed to `COMMAND_TYPE_EXEC_IN_CONTAINER` (wire value `24` unchanged); `COMMAND_TYPE_DOWNLOAD_MODEL = 27` added.

- **`GetByName` on worker repo** — new DB query to look up a worker by name, used by `IsWorkerConnected`.

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/services/deployment/executor.go
+++ b/ai-services/internal/pkg/catalog/apiserver/services/deployment/executor.go
@@ -18,11 +18,9 @@ import (
 	"github.com/project-ai-services/ai-services/internal/pkg/worker/stream"
 )
 
-// DeploymentExecutor orchestrates the complete deployment process.
-// It uses the DeploymentPlanner to create a plan and then executes it
-// using the appropriate runtime-specific deployer.
+// DeploymentExecutor orchestrates the execution of an already-planned
+// deployment, routing to the correct runtime-specific deployer.
 type DeploymentExecutor struct {
-	planner         *DeploymentPlanner
 	catalogProvider *catalog.CatalogProvider
 	appRepo         repository.ApplicationRepository
 	serviceRepo     repository.ServiceRepository
@@ -39,7 +37,6 @@ func NewDeploymentExecutor(
 	componentRepo repository.ComponentRepository,
 ) *DeploymentExecutor {
 	return &DeploymentExecutor{
-		planner:         NewDeploymentPlanner(catalogProvider, componentRepo),
 		catalogProvider: catalogProvider,
 		appRepo:         appRepo,
 		serviceRepo:     serviceRepo,
@@ -48,20 +45,20 @@ func NewDeploymentExecutor(
 }
 
 // WithWorkerRegistry wires the worker registry into the executor so it can
-// validate workers, read their metadata, and create RemoteRuntime instances.
-// Must be called before any deployment that targets a non-local worker.
+// resolve a RemoteRuntime for worker deployments. Must be called before any
+// deployment that targets a non-local worker.
 func (e *DeploymentExecutor) WithWorkerRegistry(reg stream.WorkerRegistry) *DeploymentExecutor {
 	e.workerRegistry = reg
 
 	return e
 }
 
-// ValidateWorker checks that workerName is non-empty and that the named worker
-// is currently connected with status=ready (cache + DB check). Called before
-// any DB records are written so the API can return 400 immediately.
+// ValidateWorker confirms the named worker is ready before any DB records are
+// written so the API can return 400 immediately.
 func (e *DeploymentExecutor) ValidateWorker(ctx context.Context, workerName string) error {
+	// TODO: Remove when remote deployment is by default
 	if workerName == "" {
-		return fmt.Errorf("worker name must not be empty")
+		return nil
 	}
 
 	if e.workerRegistry == nil {
@@ -91,9 +88,7 @@ func (e *DeploymentExecutor) ExecuteWithPlan(
 }
 
 // executeDeployment routes to the correct deployer based on plan.WorkerName.
-// Any name other than LocalWorkerName is treated as a remote worker and
-// dispatched over the gRPC CommandStream. LocalWorkerName falls back to the
-// local runtime determined by runtimeType.
+// WorkerName is always set by PlanDeployment.
 func (e *DeploymentExecutor) executeDeployment(
 	ctx context.Context,
 	plan *DeploymentPlan,
@@ -101,7 +96,9 @@ func (e *DeploymentExecutor) executeDeployment(
 	runtimeType types.RuntimeType,
 ) error {
 	// ── Remote worker deployment ──────────────────────────────────────────────
-	if plan.WorkerName != "" && plan.WorkerName != workerconstants.LocalWorkerName {
+	// TODO Remove the check when remote deployment is by default
+	// and the remaining code will be dead
+	if plan.WorkerName != "" {
 		return e.executeWorkerDeployment(ctx, plan, req)
 	}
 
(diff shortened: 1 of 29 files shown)


ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -218,7 +239,25 @@ func (d *PodmanDeployer) extractModelsFromParams(params map[string]any, modelSet
...
+			_, err := rt.Send(ctx, workerpb.CommandType_COMMAND_TYPE_DOWNLOAD_MODEL, payload.DownloadModel{
+				Model:     modelName,
+				TargetDir: modelsPath,
+			})
+			if err != nil {
+				return fmt.Errorf("failed to download model %s: %w", modelName, err)
+			}
+		}
+		return nil
+	}
```
mayuka-c, line 259: I feel having a helper might help here and same can be used for application download subcommand too

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
```diff
@@ -1102,23 +1141,36 @@ func (d *PodmanDeployer) registerApplicationRoutes(ctx context.Context, plan *De
...
+		}
+
+		if d.workerConfig.HTTPSPort == "" {
+			return "", "", nil, fmt.Errorf("worker metadata missing %q", workerconstants.MetaKeyHTTPSPort)
+		}
+
+		pm := proxy.NewRemoteProxyManager(rt.Sender)
+
+		return d.workerConfig.DomainSuffix, d.workerConfig.HTTPSPort, pm, nil
+	}
```
mayuka-c, line 1153: Also this, having a method known as getCaddyRemotConfig might help

mayuka-c: Also better place for this method: `getCaddyConfiguration` to be in planner.go and then pass it to deployer.go. Its better to do these checks immediately during Create flow and fail early.

ai-services/internal/pkg/catalog/apiserver/services/deployment/types/plan.go
```diff
@@ -17,6 +17,9 @@ type DeploymentPlan struct {
 	Components      map[string]*ComponentPlan // Key: component hash, Value: component plan
 	Services        map[string]*ServicePlan   // Key: service ID, Value: service plan
 	SpyreCardPool   *SpyreCardPool            // Allocated Spyre card pool (set after allocation)
+	// WorkerName is set when the deployment targets a remote worker node.
+	// Empty means local deployment.
+	WorkerName string
```
mayuka-c: I see we use constant `local` for local deployment, why not let it be set here and need have to assume empty to be local here

ai-services/internal/pkg/catalog/apiserver/services/deployment/executor.go
```diff
@@ -42,29 +47,65 @@ func NewDeploymentExecutor(
...
+// validate workers, read their metadata, and create RemoteRuntime instances.
+// Must be called before any deployment that targets a non-local worker.
+func (e *DeploymentExecutor) WithWorkerRegistry(reg stream.WorkerRegistry) *DeploymentExecutor {
+	e.workerRegistry = reg
+
+	return e
+}
+
+// ValidateWorker checks that workerName is non-empty and that the named worker
+// is currently connected with status=ready (cache + DB check). Called before
```
mayuka-c: Isnt the executor getting executed as async?
I believe DB write is already done and after that the async operation is done. Please check once

ai-services/internal/pkg/catalog/apiserver/services/deployment/executor.go
```diff
@@ -42,29 +47,65 @@ func NewDeploymentExecutor(
...
+// dispatched over the gRPC CommandStream. LocalWorkerName falls back to the
+// local runtime determined by runtimeType.
 func (e *DeploymentExecutor) executeDeployment(
 	ctx context.Context,
 	plan *DeploymentPlan,
 	req apimodels.CreateApplicationRequest,
 	runtimeType types.RuntimeType,
 ) error {
+	// ── Remote worker deployment ──────────────────────────────────────────────
+	if plan.WorkerName != "" && plan.WorkerName != workerconstants.LocalWorkerName {
```
mayuka-c: As said I feel we dont need to do empty check and evaluate it to local. Let us set it right at the beginning inside planner to LOCAL

ai-services/internal/pkg/catalog/apiserver/services/deployment/executor.go
```diff
@@ -75,8 +116,73 @@ func (e *DeploymentExecutor) executeDeployment(
 	}
```
mayuka-c, line 94: I had a one query, if its "Local", we will not have a worker agent installed and the catalog backend (gateway) itself takes care of deploying the pods?

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
yussufsh: `application model download` won't be using grpc stream. It will directly use `helpers.DownloadModelContainer`.
This is only during remote application deployment.

ai-services/internal/pkg/catalog/apiserver/services/deployment/executor.go
yussufsh: ValidateWorker is called synchronously in `CreateApplication` before `InsertDeploymentRecords` and before the goroutine is launched.
It is just that I was using the executor object for Validating worker, moving it to planner.

ai-services/internal/pkg/catalog/apiserver/services/deployment/executor.go
yussufsh: This is the plan. Right now, if I set it to Local, it will add regression.

ai-services/internal/pkg/catalog/apiserver/services/deployment/executor.go
yussufsh: No, it will still be gRPC, and we will reuse Caddy. I will discuss with you more offline.

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/podman/deployer.go
yussufsh: Moved the validation to the top layer.

mayuka-c (approved):
LGTM

PR #1352, opened 2026-08-30
Data Source Connectors: Minor Enhancements

sats-23:
-Enhance logging and error response for invalid files 
-Fixed bug where connectors table uses completed status (it is confined to ONLY Syncing, In Sync & Out of Sync)
-Introduced new completed_files metric that updates on ConnectorSyncLog for every successfully ingested file on connector
-Introduced new sync_message field in Connectors that populates the message used by UI
-Fixed bug to exclude connector sourced docs from GET /v1/docs in any stage
-Pagination support for GET /v1/connectors
-Minor rebranding
-Introduced Job Source to help filter jobs easily
-Updated proposal


docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -278,8 +282,30 @@ _run_teardown(connector_id)  (connectors.py — used for BOTH Case A and Case B)
 
 ### 3.4 `GET /v1/connectors` & `GET /v1/connectors/{connector_id}`
 
-- `GET /v1/connectors`: Returns all connectors ordered by `attached_at` descending. Excludes secret fields. Includes `error` field (covers teardown errors, credential errors, and last-sync error messages).
-- `GET /v1/connectors/{id}`: Returns full connector detail including `connection_details` (secrets stripped), `allowed_extensions`, `sync_interval_seconds`, and file processing counters (`total_files`).
+- `GET /v1/connectors`: Returns a paginated list (`ConnectorListResponse`: `total`, `limit`, `offset`, `items`) of connectors ordered by `attached_at` descending (`limit` default 50, max 200, `offset` default 0). Excludes secret fields. Includes `error`, `total_files`, and `sync_message` fields.
+- `GET /v1/connectors/{id}`: Returns full connector detail (`ConnectorDetailResponse`) including `connection_details` (secrets stripped), `allowed_extensions`, `sync_interval_seconds`, `error`, `total_files`, and `sync_message`.
+
+**`GET /v1/connectors` Response shape — `ConnectorListResponse`:**
```
dharaneeshvrd: Yes, I see the sample response is missing for some of the APIs, let's update them properly

services/digitize/api/v1/connectors.py
```diff
@@ -76,7 +76,7 @@ async def create_connector(body: ConnectorCreateRequest):
     touching the scheduler or writing to the database.  Only once both checks
     pass is the scheduler job registered and the row inserted.
     """
-    connector_id = body.id or str(uuid.uuid4())
```
dharaneeshvrd, line 79: let's do this in a different PR
we might probably need to figure to how to make the connectors API internal

services/digitize/connectors/models.py
```diff
@@ -310,6 +312,37 @@ class SyncLogItem(BaseModel):
...
+                "limit": 50,
+                "offset": 0,
+                "items": [
+                    {
+                        "id": "c1d2e3f4-a5b6-7890-abcd-ef1234567890",
+                        "name": "prod-sftp-reports",
+                        "type": "file_system",
+                        "attached_at": "2025-01-15T10:00:00Z",
+                        "last_sync_at": "2025-01-15T10:30:00Z",
+                        "sync_status": "up to date",
```
dharaneeshvrd: sync_message is missing

services/digitize/connectors/sync_tick.py
```diff
@@ -280,6 +293,15 @@ async def _wait_for_job(job_id: str, connector_id: str, sync_seq: int) -> None:
         job_data = get_job(job_id)
         status = (job_data or {}).get("status", "")
         logger.debug(f"Polling job {job_id!r} for connector {connector_id!r}: status={status!r}")
+
+        # Count any docs that newly reached 'completed' since the last poll.
+        job_stats = get_job_document_stats(job_id)
+        completed_count = len(job_stats["completed_docs"])
+        newly_completed = completed_count - prev_completed_count
```
dharaneeshvrd, line 302: can we just update the completed docs as is?
why this calculation is required?

dharaneeshvrd (commented):
how are we resetting the sync_message to none once sync reaches one of the terminal status?

services/digitize/api/v1/connectors.py
sats-23: That should be a simple flag `include_in_schema=False`. 
I suggest we do just that in a final PR before code freeze. But any other thoughts to keep the UUID generating behavior while it is internal?

services/digitize/connectors/sync_tick.py
sats-23: Docs keep updating in random flow. 
Assume our batch of 10 files finish ingestion at 1..2..4..5..7..10 rate 
If we just add completed count, we are adding 1+2+4+5... so on which is wrong calculation. 
Adding this calculation so that we get granular updates (within batch of 10), but at the same time, don't end up adding more (wrong) numbers to the count

services/digitize/connectors/models.py
sats-23: Fixed

services/digitize/connectors/sync_tick.py
dharaneeshvrd: but its just a total count right, we dont need to worry about the order in which it gets updated.
I dont think once its updated as completed, its not going to decrease, its going to keep increase on the job side right?

services/digitize/api/v1/connectors.py
dharaneeshvrd: no, we should remove it.
want to do this in a dedicated PR, so we can look back when required, or else it would be missed if it is part of a large PR with too many changes.

services/digitize/connectors/sync_tick.py
sats-23: Yes, it is going to increase, but we don't know how many increased in between right?
Say 2/10 jobs are completed and the 2 is passed to increment in completed_files and to update the message. 
Next, when say 5/10 jobs are completed, we need to pass 3 to increment to completed_files. If we pass 5, the whole count messes up

dharaneeshvrd: ok this would be required when we have more than one batch and need to update the count across batches.

services/digitize/api/v1/connectors.py
sats-23: Done, I have removed it

services/digitize/connectors/sync_tick.py
dharaneeshvrd, line 165: here we are updating the only connector sync log
need to update connector as well with total files and new files found message

services/digitize/connectors/sync_tick.py
```diff
@@ -167,14 +168,16 @@ async def run_tick(connector_id: str, sync_seq: int) -> None:
             removed_files=len(orphan_checksums),
         )
 
-        invalid_count = await _process_new_files(sync_seq, connector_id, config.name, scanner, ingest_list)
+        invalid_files = await _process_new_files(sync_seq, connector_id, config.name, scanner, ingest_list)
 
         update_connector_total_files(connector_id, total_files)
```
dharaneeshvrd: I see we are updating after processing, lets move one step above and add the message as well

services/digitize/api/v1/connectors.py
dharaneeshvrd, line 229: Is this exception possible here?

services/digitize/api/v1/connectors.py
```diff
@@ -782,19 +787,22 @@ async def cancel_sync(connector_id: str, sync_seq: int):
 
 @router.get(
     "/{connector_id}/syncs",
-    response_model=SyncLogResponse,
```
dharaneeshvrd, line 785: why is this removed?

services/digitize/api/v1/connectors.py
```diff
@@ -805,6 +813,25 @@ async def get_sync_history(
...
             )
 
+        if latest:
+            log = db_ops.get_latest_sync_log(connector_id)
+            if log is None:
+                APIError.raise_error(
+                    ErrorCode.RESOURCE_NOT_FOUND,
+                    f"No sync history found for connector {connector_id!r}",
+                )
+            return SyncLogDetailResponse(
```
dharaneeshvrd: I think we can return in the same list format, no need to break the structure.
we can just keep the only one item in the list

services/digitize/db/manager.py
```diff
@@ -468,11 +471,11 @@ def get_all_documents(
...
-                    # Exclude connector-sourced documents via NOT EXISTS subquery.
-                    filters.append(
-                        ~select(ConnectorDocumentChecksum.doc_id)
-                        .where(ConnectorDocumentChecksum.doc_id == Document.doc_id)
-                        .exists()
+                    # Exclude connector-sourced documents
+                    # Since the connector doc checksum is inserted after successful ingestion
+                    # we need a mechanism to exclude connector sourced docs irrespective of which stage they are in (Accepted, Digtized, In Progress, etc.)
+                    stmt = stmt.join(Job, Job.job_id == Document.job_id).where(
+                        ~Job.job_name.like("Connector-%")
```
dharaneeshvrd: I m bit skeptical about this way of identifying the connector jobs.
We never know the users, they might copy paste the connector job name from UI and submit a new job.
Can we see is there any other way?

dharaneeshvrd (commented):
Remaining lgtm

services/digitize/connectors/sync_tick.py

services/digitize/connectors/sync_tick.py
sats-23: Done

services/digitize/api/v1/connectors.py
sats-23: Got removed around latest flag, I have re-added it and removed latest flag

services/digitize/api/v1/connectors.py
sats-23: Yes, as discussed, only in high load race conditions

services/digitize/api/v1/connectors.py

services/digitize/db/manager.py
sats-23: Other way I can think of is to probably add another field in the Job table like Job.source to distinguish User and connector created jobs. What do you think?

dharaneeshvrd: yeah that seems to be the possible way.

sats-23: Done, added

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -166,17 +169,48 @@ class ConnectorSyncLog(Base):
...
+
+| Field | Required | Description |
+|-------|----------|-------------|
+| `endpoint_url` | **Yes** | Full URL with scheme, e.g. `"https://s3.us-east-1.amazonaws.com"` |
+| `bucket_name` | **Yes** | S3/COS bucket name |
+| `access_key_id` | **Yes** | Access key ID |
+| `secret_access_key` | **Yes** | Secret access key (encrypted at rest) |
+| `prefix` | No | Key prefix filter |
+| `delimiter` | No | Key delimiter |
+| `download_concurrency` | No | Parallel download threads |
```
dharaneeshvrd: dont think we support this, remove it

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -288,44 +358,136 @@ _run_teardown(connector_id)  (connectors.py — used for BOTH Case A and Case B)
 #### `GET /v1/connectors/{connector_id}/syncs`
```
dharaneeshvrd, line 419: move this after post API def

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
```diff
@@ -288,44 +358,136 @@ _run_teardown(connector_id)  (connectors.py — used for BOTH Case A and Case B)
...
+Returns a single sync log entry identified by its sequence number, wrapped in a `SyncLogResponse` envelope. Returns `404` if the connector or the specific `sync_seq` does not exist.
+
+**`200 OK` — `SyncLogResponse`** (`total=1`, `limit=1`, `offset=0`, `items` contains the single entry):
+
+```json
+{
+  "total": 1,
+  "limit": 1,
+  "offset": 0,
+  "items": [
```
dharaneeshvrd: invalid for get on id

services/digitize/connectors/sync_tick.py
```diff
@@ -158,23 +161,27 @@ async def run_tick(connector_id: str, sync_seq: int) -> None:
...
             total_files=total_files,
-            new_files=len(ingest_list),
+            new_files=new_files,
             removed_files=len(orphan_checksums),
         )
 
-        invalid_count = await _process_new_files(sync_seq, connector_id, config.name, scanner, ingest_list)
-
         update_connector_total_files(connector_id, total_files)
+        set_connector_message(connector_id, f"{new_files} new files found")
```
dharaneeshvrd: can't we combine to make a single db call?

services/digitize/db/models.py
```diff
@@ -31,6 +31,12 @@ class Base(DeclarativeBase):
     pass
 
 
+class JobSource(str, enum.Enum):
```
dharaneeshvrd, line 34: why not define this for document as well?
would become easy in filtering out user docs on get docs requests
those join statements can be avoided

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Fixed

docs/proposals/data-source-connectors/digitize-connector-processing-proposal.md
sats-23: Done

services/digitize/connectors/sync_tick.py

services/digitize/db/models.py
sats-23: Done, introduced that too. 
Only one query (delete user jobs) uses the JobSource, so I kept it as is. Rest sources are able to do operations without table join

dharaneeshvrd (dismissed):

dharaneeshvrd (dismissed):

dharaneeshvrd (approved):

PR #1341, opened 2026-08-27
AIS-1667 - Connect datasources to application

prachikurade1:
Adds a PUT /api/v1/applications/:id/datasources endpoint that connects one or more datasource connectors to eligible services in an application.

1. Accepts a list of datasource_ids in the request body; each datasource is processed independently
2. Resolves eligible services via accepts_datasource: true in catalog metadata

PR #1358, opened 2026-08-31
AIS-1669 - List Application Datasources

prachikurade1:
Fetch the list of datasources associated with an application:
- Fields fetched from Digitize API for connectors:
   - Sync status 
   - Files 
   - Message


prachikurade1:
application_datasources
Note: The status "completed" is a bug from the Digitize API which is being fixed.

ai-services/internal/pkg/catalog/client/digitize.go
```diff
@@ -94,3 +94,57 @@ func (c *DigitizeClient) GetConnectorSync(ctx context.Context, connectorID strin
...
+		SetContext(ctx).
+		SetResult(&result).
+		Get("/v1/connectors")
+	if err != nil {
+		return nil, fmt.Errorf("digitize request failed: %w", err)
+	}
+
+	if resp.IsError() {
+		return nil, fmt.Errorf("digitize returned status %d", resp.StatusCode())
+	}
```
mayuka-c: As mentioned here: https://github.com/IBM/project-ai-services/pull/1341#discussion_r3893681515. Lets avoid digitize mentioned in catalog here and let us rename the file too

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -678,4 +688,180 @@ func extractAPIEndpointURL(endpointsJSON json.RawMessage) string {
...
+
+	data := make([]apimodels.ApplicationDatasourceItem, 0, len(connectorIDs))
+	for _, cid := range connectorIDs {
+		item, buildErr := s.buildApplicationDatasourceItem(ctx, cid, baseURL, digitizeConnectors)
+		if buildErr != nil {
+			return nil, buildErr
+		}
+
+		data = append(data, *item)
+	}
```
mayuka-c: I thought we would just make the call to their API and just return back to UI. I see we are doing lot of processing steps which we should not be doing. I feel it should be handled by them whatever is required to be present on UI

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -678,4 +688,180 @@ func extractAPIEndpointURL(endpointsJSON json.RawMessage) string {
...
+			HasPrev:    req.Page > 1,
+		},
+	}, nil
+}
+
+// fetchDigitizeConnectors resolves the Digitize base URL from the first connector in the
+// list (using the existing GetLinkedServiceEndpoints + extractAPIEndpointURL path), then
+// calls GET /v1/connectors once to return a map keyed by connector ID.
+// Returns an empty string and nil map when no endpoint is found or the call fails.
+func (s *DatasourceService) fetchDigitizeConnectors(ctx context.Context, connectorIDs []uuid.UUID) (string, map[string]apimodels.DigitizeConnectorItem) {
```
mayuka-c: Lets make this generic and let us not have any reference of digitize

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
prachikurade1: They are not sending the message that UI expects. The message gets formed using the sync_status from the response "/connectors" API and the the individual sync job /connectors/:id/sync?latest=true.

mayuka-c: Any reason of why it cant be done by them in the GET API itself?
Let us avoid all these processing, this adds lot of latencies.

prachikurade1: @dharaneeshvrd , @sats-23 , can you form the message field also in the GET /connectors API ?

dharaneeshvrd: It's not possible, it's per sync level detail, we can't add those details at connector level. Hence suggested to call sync log API.

mayuka-c: These are paginated calls, if user selects 50-100 items, we cannot make that many network calls and the latency would be really really bad and cannot gaurantee consistency (What if few network calls fails here).

Any reason of why it cant be added?. Since its part of digitize service, it should be able to get list of connectors from DB and also sync details from same DB and return back?. Am I missing something here?

dharaneeshvrd: Sync details are exposed through a separate API because they contain sync-specific attributes that need to be maintained independently. Storing these details at the connector level would be redundant and not a good design IMO.
Also, sync details only need to be fetched when the status is syncing, not all the time.

mayuka-c: You could expose it in separate API, but definitely while returning List response, the message field should have appropriate details based on syncing logic that you mentioned.

As said this will not work and causes lot of inconsistent state. Whatever that you're expecting the catalog backend needs to do processing should be done internally in the GET call and appropriately message field needs to be updated.

mayuka-c: If you still feel we cannot add, then I would request to have a separate API which would give us required fields with list because of below issues (without having to do any processing):

```
These are paginated calls, if user selects 50-100 items, we cannot make that many network calls and the latency would be really really bad and cannot gaurantee consistency (What if few network calls fails here).
```

ai-services/internal/pkg/catalog/client/digitize.go
prachikurade1: Since @Shubhamag12 's PR has most of the changes done, will pull his code once merged and then update my code here.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
dharaneeshvrd: Digitize service is not tailored to serve the UI, it needs to be generic since it can be consumed directly by AI agent via MCP and as well as by user's other applications, we can't have a field that is specifically needed for catalog UI.
And for 100s of data sources, do you really think user would be connecting that many ssh and s3 sources? As I mentioned earlier, this API needs to be called only when the status is sync. Do you still need a dedicated API that provides both?

mayuka-c: Well MCP/ any user shouldnt be directly called to digitize for connector, we will hit the inconsistency issue that we had already discussed. Even if MCP needs to be used for connector, it has to be used via catalog backend and cannot be directly calling digitize.

We would need certain framework, protocol that each service needs to follow and message is one such field that has to be followed by each and every service and it will be its responsibility to display on the UI (Digitize might be imposing this today but a different service may impose different logic to determine message). Having catalog backend do service specific will not make it generic and cannot be extended if customer wants to have their own service use the connectors.

We could discuss more if required but I definitely think every service needs to follow the contract and it would be its respondibility to do it.

cc. @yussufsh

mayuka-c: `And for 100s of data sources, do you really think user would be connecting that many ssh and s3 sources? As I mentioned earlier, this API needs to be called only when the status is sync. Do you still need a dedicated API that provides both?`

We cannot restrict it. Who knows there might be and we should ensure our service should handle it :)
Users may/ may not have it today but we need to ensure our services can scale and handle the load effectively today and also as and when we bring new type of connectors.

prachikurade1:
Will rework on this post the changes to get message field from the Digitize connectors API.
cc: @yussufsh @mayuka-c

prachikurade1:
image
Tested with new changes

ai-services/internal/pkg/catalog/client/digitize.go
prachikurade1: Done

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -1122,4 +1121,153 @@ func extractAPIEndpointURL(endpointsJSON json.RawMessage) string {
...
+
+// buildApplicationDatasourceItem fetches connector metadata from the DB and enriches it
+// with sync state from the pre-fetched serviceConnectors map (status, last_sync, total_files,
+// message). The message field is sourced directly from the service — no separate sync-log
+// call is needed.
+func (s *DatasourceService) buildApplicationDatasourceItem(ctx context.Context, connectorID uuid.UUID, baseURL string, serviceConnectors map[string]apimodels.ConnectorItem) (*apimodels.ApplicationDatasourceItem, error) {
+	connector, err := s.connectorRepo.GetByID(ctx, connectorID, false)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch connector: %w", err)
+	}
```
mayuka-c: Instead of making DB call per connector, can we not take list of connectors for givenAppID and just do computing. This avoids lot of DB network calls

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -1122,4 +1121,153 @@ func extractAPIEndpointURL(endpointsJSON json.RawMessage) string {
...
+	app, err := s.appRepo.GetByID(ctx, appID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to look up application: %w", err)
+	}
+
+	if app == nil {
+		return nil, &ValidationError{Code: http.StatusNotFound, Message: "application not found"}
+	}
+
+	connectorIDs, total, err := s.svcDepRepo.GetConnectorsByAppID(ctx, appID, req.PageSize, (req.Page-1)*req.PageSize)
```
mayuka-c: I see a problem here, what if you fetch a page with contains 20 items and the ListConnectors from service gave you a page with diff 20 items, this will cause a problem.

The pagination support for this API should entirely rely on services response

mayuka-c (changes requested):

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go

prachikurade1:
page1
page2

mayuka-c (dismissed):
LGTM

mayuka-c (approved):

PR #1359, opened 2026-08-31
feat: add deregister worker flow

pranithraoibm:
Added the worker deregister workflow along with error handling

https://github.com/user-attachments/assets/84af02ec-1006-4304-90b3-4bf2f6bc3371


ui/catalog/src/components/WorkerResourcesTable/DeregisterWorkerModal.tsx
```diff
@@ -0,0 +1,88 @@
...
+  isDeregistering: boolean;
+  workerName: string;
+  workerStatus: string;
+  runtimeType: string;
+  onConfirm: () => void;
+  onClose: () => void;
+}
+
+const buildCleanupCommand = (runtimeType: string) =>
+  `ai-services worker uninstall --runtime ${!runtimeType || runtimeType === "unknown" ? "<podman|openshift>" : runtimeType} --yes`;
```
ryanmarc: The command itself is right now, thanks. One thing on the fallback: <podman|openshift> is shell syntax, so if the runtime is unknown the copied snippet does not run.

Bash reads <podman as a redirect and |openshift> as a pipe, so worker uninstall never runs and the errors do not mention the real problem. And unknown is not rare here: it is the runtime a worker has before it first connects, which is a likely thing to be deregistering.

Quoting it ("<podman|openshift>") or using a bare RUNTIME would at least let the CLI produce its own "required flag(s) "runtime" not set" error.

ui/catalog/src/components/WorkerResourcesTable/DeregisterWorkerModal.tsx
```diff
@@ -0,0 +1,88 @@
...
+      size="sm"
+      modalLabel={`Deregister ${workerName}`}
+      modalHeading="Deregister worker resource"
+      primaryButtonText={isDeregistering ? "Deregistering..." : "Deregister"}
+      secondaryButtonText="Cancel"
+      primaryButtonDisabled={isDeregistering}
+      preventCloseOnClickOutside
+      onRequestSubmit={onConfirm}
+      onRequestClose={onClose}
+    >
```
ryanmarc, line 41: Carbon's Modal takes a danger prop that renders the primary button destructively, and this is a removal confirm without it, so "Deregister" renders as an ordinary blue primary.

ui/catalog/src/components/WorkerResourcesTable/RegisterWorkerModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const isSuccess = phase === "success";
+
+  const runCommand = useMemo(() => {
+    if (!isSuccess) return "";
+    return [
+      "ai-services worker join \\",
+      "  --server <host>:<port> \\",
+      `  --name "${workerName}" \\`,
+      `  --token "${token}"`,
+    ].join("\n");
```
ryanmarc, line 45: Please verify that --server and --name are valid flags on this command. It looks like only --token, --basedir, --https-port, --domain-name, --ssl-cert, and --ssl-key are the valid flags.

ryanmarc (changes requested):

ui/catalog/src/components/WorkerResourcesTable/DeregisterWorkerModal.tsx
pranithraoibm: Added "<podman|openshift>" for register modal command as user may not know what values to pass here, so this will give them a idea
Added "RUNTIME" for de-register modal command - but modal won't show the command if the worker is not connected i.e. if the status is pending and runtime type is unknown. 

image
image

ui/catalog/src/components/WorkerResourcesTable/DeregisterWorkerModal.tsx
pranithraoibm: Added the danger prop

image

ui/catalog/src/components/WorkerResourcesTable/RegisterWorkerModal.tsx
pranithraoibm: Yes command is updated as we updated it in the register worker PR post merge and main pull

ryanmarc (approved):

PR #1360, opened 2026-08-31
APIs for Architecture/Services Images + Application Image CLI command to consume the API's

mayuka-c:
- Currently the `application image` sub command fetches the image list by directly looping over the embedded assets.
- This will not work with BYOS, as the custom assets will be stored in volumes/PVC and CLI will not have access to it.
- So exposing 2 different endpoints for fetching the list of images required for architecture/service.
- Make CLI consume these 2 endpoints by default which would work for both embedded and custom assets.


ai-services/internal/pkg/catalog/apiserver/handlers/catalog.go
```diff
@@ -323,6 +323,73 @@ func (h *CatalogHandler) GetServiceParams(c *gin.Context) {
...
+//	@Failure		401	{object}	ErrorResponse	"Unauthorized - Invalid or missing access token"
+//	@Failure		404	{object}	ErrorResponse	"Architecture not found"
+//	@Failure		500	{object}	ErrorResponse	"Internal Server Error"
+//	@Router			/architectures/{id}/images [get]
+func (h *CatalogHandler) GetArchitectureImages(c *gin.Context) {
+	id := c.Param("id")
+
+	images, err := h.provider.GetArchitectureImages(c.Request.Context(), id)
+	if err != nil {
+		c.JSON(http.StatusNotFound, ErrorResponse{
```
yussufsh: This could be 5XX as well no?

ai-services/cmd/ai-services/cmd/application/image/image.go
```diff
@@ -24,19 +25,32 @@ var ImageCmd = &cobra.Command{
...
+	// Only fall through to architecture when the server returned 404.
+	var httpErr *client.HTTPError
+	if !errors.As(err, &httpErr) || httpErr.StatusCode != http.StatusNotFound {
+		return nil, err
 	}
 
-	images, err := provider.GetCatalogImages(ctx, templateID)
+	resp, err = appClient.GetArchitectureImages(ctx, templateID)
 	if err != nil {
 		return nil, err
```
yussufsh, line 68: Here we return architecture not found error.. but this should be catalog item not found or something else related to catalog.

yussufsh (changes requested):

ai-services/internal/pkg/catalog/apiserver/handlers/catalog.go
mayuka-c: Done have handled 404 separately

ai-services/cmd/ai-services/cmd/application/image/image.go
```diff
@@ -24,15 +29,71 @@ var ImageCmd = &cobra.Command{
...
+	if !ok || httpErr.StatusCode != http.StatusNotFound {
+		return nil, err
+	}
+
+	resp, err = appClient.GetArchitectureImages(ctx, templateID)
+	if err != nil {
+		// Both lookups failed — the ID is not in the catalog.
+		if archErr, ok := err.(*client.HTTPError); ok && archErr.StatusCode == http.StatusNotFound {
+			return nil, fmt.Errorf("template '%s' not found as a service or architecture", templateID)
+		}
```
yussufsh: Instead with `err.(*client.HTTPError)` can we use errors.As() as done in other places in the repo.

ai-services/cmd/ai-services/cmd/application/image/image.go
```diff
@@ -24,15 +29,71 @@ var ImageCmd = &cobra.Command{
...
+		return resp.Images, nil
+	}
+
+	// Only retry as architecture when the server returned 404.
+	httpErr, ok := err.(*client.HTTPError)
+	if !ok || httpErr.StatusCode != http.StatusNotFound {
+		return nil, err
+	}
+
+	resp, err = appClient.GetArchitectureImages(ctx, templateID)
```
yussufsh: We should check Architecture first and then Services, as len(arch) < len(services).

yussufsh (changes requested):
Minor comments, else lgtm.

ai-services/cmd/ai-services/cmd/application/image/image.go
mayuka-c: Yes Make sense Thanks will be reversing the order

yussufsh (approved):
lgtm

PR #1361, opened 2026-08-31
AISERVICES-1654 async endpoint batch file handling

manalilatkar:
In this PR:

1. schema changes
2. db manager changes
3. async endpoint now handles a batch of max 64 files
4. result endpoint for each document
5. result download endpoint for each document is also provided
6. unit tests


services/extract/api/v1/jobs.py
```diff
@@ -300,46 +301,6 @@ def _resolve_schema(schema_id: str):
     return row
 
 
-def _stage_and_persist(
```
dharaneeshvrd, line 303: why are we removing this method?
can't we alter to handle multiple files?

services/extract/api/v1/jobs.py
```diff
@@ -403,128 +364,222 @@ async def _validate_file_content(file: UploadFile) -> None:
...
+                schema_id=schema_id,
+                job_name=job_name,
+                submitted_at=datetime.now(timezone.utc),
+                file_count=len(files),
+            )
+        except Exception as exc:
+            logger.error(f"Unexpected DB error creating job {job_id}: {exc}")
+            raise ExtractException(500, "DATABASE_ERROR", "Failed to create job record.")
+
+        if row is None:
```
dharaneeshvrd, line 420: let's log the errors before we raising

services/extract/api/v1/jobs.py
```diff
@@ -403,128 +364,222 @@ async def _validate_file_content(file: UploadFile) -> None:
...
+                "doc_id": str(uuid.uuid4()),
+                "filename": filename,
+                "source_type": source_type,
+            }
+            for filename, source_type in validated
+        ]
+        ok = db_repo.create_documents(job_id, doc_entries)
+        if not ok:
+            db_repo.delete_job(job_id)
+            raise ExtractException(500, "DATABASE_ERROR", "Failed to create document records.")
```
dharaneeshvrd, line 437: log the error

services/extract/api/v1/jobs.py
```diff
@@ -708,37 +754,161 @@ async def _process_extract_job(job_id: str) -> None:
...
+                error=f"Unexpected error: {exc}",
+                completed_at=datetime.now(timezone.utc),
+            )
+
+
+# ---------------------------------------------------------------------------
+# _process_batch_job — batch background worker
+# ---------------------------------------------------------------------------
+
+async def _process_batch_job(job_id: str) -> None:
```
dharaneeshvrd: let's move these methods to utils/job.py

services/extract/api/v1/jobs.py
```diff
@@ -708,37 +754,161 @@ async def _process_extract_job(job_id: str) -> None:
...
+    """Background worker: process each document in the batch sequentially."""
+    from extract import state
+
+    async with state.extract_limiter:
+        t_start = time.monotonic()
+        logger.info(f"Batch worker started for job {job_id}")
+
+        db_repo.update_job(job_id=job_id, status="in_progress")
```
dharaneeshvrd: let's define enum for job status if not done already and update the status usage with enum

services/extract/api/v1/jobs.py
dharaneeshvrd: need to call vllm inside this method with concurrency semaphore locked
seeing this behaviour in sync extraction request as well

services/extract/api/v1/jobs.py
```diff
@@ -708,37 +754,161 @@ async def _process_extract_job(job_id: str) -> None:
                 },
             }
 
-            result_path = settings.extract.results_dir / f"{job_id}_result.json"
+            result_dir = settings.extract.results_dir / job_id
+            result_path = result_dir / f"{doc_id}_result.json"
             try:
-                result_path.parent.mkdir(parents=True, exist_ok=True)
+                result_dir.mkdir(parents=True, exist_ok=True)
                 result_path.write_text(json.dumps(result_payload), encoding="utf-8")
```
dharaneeshvrd: do we need to write the result_payload into local file?
IMO writing only the extracted content is enough, so when user tries to download the result, they can use it directly without parsing the json

services/extract/api/v1/jobs.py
```diff
@@ -922,6 +1120,79 @@ async def get_extract_job_result(job_id: str):
...
+        raise ExtractException(
+            404, "RESOURCE_NOT_FOUND",
+            f"Result file not found for document {doc_id!r}.",
+        )
+
+    filename_stem = os.path.splitext(doc_row.filename)[0]
+    download_filename = f"{filename_stem}_result.json"
+
+    return Response(
+        content=json.dumps(result_data, indent=2),
```
dharaneeshvrd, line 746: I feel its better to return the extracted content

services/extract/api/v1/jobs.py
```diff
@@ -937,7 +1208,7 @@ async def get_extract_job_result(job_id: str):
     },
     summary="Delete extraction job",
     description=(
-        "Delete a job record and its result file.  "
+        "Delete a job record and its result file(s).  "
```
dharaneeshvrd, line 769: Its fine
But can we please elaborate more on all the documents submitted will be deleted by this? 
Also can we add confirm flag?

services/extract/utils/job.py
```diff
@@ -101,11 +135,19 @@ def delete_all_job_files() -> None:
     """Delete all result files and all staging directories."""
     results_dir = settings.extract.results_dir
     if results_dir.exists():
+        # Flat single-file results
         for f in results_dir.glob("*_result.json"):
             try:
                 f.unlink()
             except Exception as exc:
                 logger.error(f"Failed to delete result file {f.name}: {exc}")
+        # Per-job batch results directories
```
dharaneeshvrd, line 173: its just json file right? dont think we have batch specific dirs

services/extract/settings.py
```diff
@@ -131,10 +131,24 @@ class ExtractionConfig(BaseSettings):
...
+    max_concurrent_requests: int = Field(
+        default=8,
         ge=1,
-        description="Maximum number of async extraction jobs running in parallel",
+        description="Maximum number of async + sync extraction requests running in parallel",
+    )
+
+    # Batch submission
+    max_files_per_job: int = Field(
+        default=64,
```
dharaneeshvrd: now I feel 64 is too much, can we keep 32 only?

services/extract/api/v1/jobs.py
manalilatkar: The method was named stage_and_persist and staging files also saving the job in db, which is not needed. So I thought of retiring it and creating a new method more true to its name that just stages and doing the saving to db part separately. 
I don't see a big improvement in functioning if we go back to earlier version.

services/extract/api/v1/jobs.py
manalilatkar: @dharaneeshvrd so then we would be losing out on usage data and precise value of processing_time_ms (but it can be roughly calculated by start_time-end_time). May be we can persist input and output usage tokens in the document table and then this whole payload will be reconstructable. WDYT?

services/extract/api/v1/jobs.py
manalilatkar: @dharaneeshvrd why do we need a confirm flag for deleting a single job? We don't follow that pattern anywhere else.

services/extract/api/v1/jobs.py
dharaneeshvrd: Are we not storing the same data in the db? That's why I commented above.

manalilatkar: @dharaneeshvrd We store only the total tokens that the doc contains. But 
`usage_input_tokens = input prompt+ doc+schema tokens `
and completion_tokens are not stored at all. Shall I add these to the Document model?

services/extract/api/v1/jobs.py
dharaneeshvrd: We follow on digitize jobs and at the only place we have jobs I believe, hence asked this :)

services/extract/api/v1/jobs.py
dharaneeshvrd: If really required we can, also if its simple to compute after retrieving from db, that's also fine.

services/extract/utils/job.py
manalilatkar: yeah just json files, but they are stored as {results_dir}/{job_id}/{doc_id}_result.json so we need to delete all directories. but the lines above that are not required, so I'll remove those.

services/extract/settings.py
manalilatkar: ok

services/extract/api/v1/jobs.py
manalilatkar: Also, what do you mean by `elaborate more on all the documents submitted will be deleted by this`? The documents rows associated with this job id will be automatically deleted. Why should we expose the info to the external user?

manalilatkar: @dharaneeshvrd I just checked for digitize jobs and we don't have confirm flag for deleting digitize job by job_id.

dharaneeshvrd: ok for jobs we dont have since it won't delete the underlying docs, but delete on all docs takes this confirm=true flag. That's why I suggested that here since we would clean up the docs.

dharaneeshvrd: > Also, what do you mean by elaborate more on all the documents submitted will be deleted by this? The documents rows associated with this job id will be automatically deleted. Why should we expose the info to the external user?

Are we not deleting the data as well? I m not worried about the db entry, but all the extracted content also will be deleted right. I want to stress this more.

manalilatkar: The data is being deleted when we call `delete_job_files(job_id)`

services/extract/api/v1/jobs.py
manalilatkar: Now since we are saving only the extracted content, that's what we are providing in the download file, so this should be fine.

services/extract/api/v1/jobs.py
manalilatkar: fixed.

services/extract/api/v1/jobs.py
manalilatkar: fixed.

services/extract/api/v1/jobs.py
manalilatkar: defined enum for Job status and Doc status everywhere.

services/extract/api/v1/jobs.py
manalilatkar: moved all helper methods to utils/job.py

services/extract/api/v1/jobs.py
manalilatkar: done

services/extract/api/v1/jobs.py
manalilatkar: done

dharaneeshvrd (commented):
logging error is missing on all the instances of raising exception on the API handler

services/extract/api/v1/jobs.py
dharaneeshvrd: let's address this please

dharaneeshvrd (approved):

PR #1362, opened 2026-09-01
APIs for Architecture/Services models + Application Models CLI command to consume the API's

mayuka-c:
- This is very much similar to images
- More info on why this is req:  https://github.com/IBM/project-ai-services/pull/1360


ai-services/cmd/ai-services/cmd/application/model/list.go
```diff
@@ -67,8 +67,9 @@ func list(cmd *cobra.Command) error {
 }
 
 // listCatalogModels lists models for services or architectures from the catalog.
+// watsonx models are excluded as they are not locally deployable.
 func listCatalogModels(ctx context.Context, templateID string) error {
-	models, err := getCatalogModels(ctx, templateID)
+	models, err := getCatalogModels(ctx, templateID, "watsonx")
```
yussufsh: Listing watsonx model is fine right? Had implemented in that way in Q2.

mayuka-c: Yeah it should be fine as we are not downloading the watsonx model. I feel in next release watsonx would be moved to connector (model endpoints) and it should work fine

mayuka-c: Just wondering whether I should not change the response now and keep as it is. Maybe I will remove "watsonx" and keep as it is, let it list watsonx model. If people are using model list, dont want to cause confusion on why its missing

mayuka-c: Have reverted this change

ai-services/internal/pkg/catalog/apiserver/handlers/catalog.go
```diff
@@ -396,11 +397,97 @@ func (h *CatalogHandler) GetArchitectureImages(c *gin.Context) {
...
+
+// GetArchitectureModels godoc
+//
+//	@Summary		Get models for an architecture
+//	@Description	Returns all unique model names referenced in the schemas of all component dependencies across every service in the architecture
+//	@Tags			Catalog
+//	@Produce		json
+//	@Security		BearerAuth
+//	@Param			id	path		string				true	"Architecture ID (e.g., 'rag')"
+//	@Success		200	{object}	map[string][]string	"JSON object with a 'models' key containing the list of model names"
```
yussufsh:
```suggestion
//	@Success		200	{object}	[]string	"JSON object with a 'models' key containing the list of model names"
```
If I am not wrong.

yussufsh (changes requested):

ai-services/internal/pkg/catalog/apiserver/handlers/catalog.go
mayuka-c: Its following below structure similar to images actually
{
models: []
}

mayuka-c: This is the reason its a map. I feel it should be fine to go with this, Open for any suggestions

yussufsh: We can do similar to `ImagesResponse`?

mayuka-c: Ok let me change across

mayuka-c: Updated

yussufsh (dismissed):

yussufsh (approved):

PR #1363, opened 2026-09-01
Port apscheduler[sqlalchemy] to latest stable v3.11.3

sats-23:
-Port apscheduler[sqlalchemy] to v3.11.3
-Tested and works fine

### DIFF
--- a/services/digitize/db/scripts/init_schema.sql
+++ b/services/digitize/db/scripts/init_schema.sql
@@ -42,6 +42,9 @@ CREATE TABLE IF NOT EXISTS document_checksum (
     doc_id        TEXT        NOT NULL UNIQUE REFERENCES documents(doc_id) ON DELETE CASCADE
 );
 
+-- Schema for APScheduler job store
+CREATE SCHEMA IF NOT EXISTS scheduler;
+
 -- Connector tables
 CREATE TABLE IF NOT EXISTS connectors (
     id                      TEXT        PRIMARY KEY,
--- a/services/digitize/db/scripts/init_schema.sql
+++ b/services/digitize/db/scripts/init_schema.sql
@@ -43,6 +43,11 @@ CREATE TABLE IF NOT EXISTS document_checksum (
 );
 
 -- Schema for APScheduler job store
+-- APScheduler v3 SQLAlchemyJobStore attempts to auto-create its job store table
+-- (default: 'apscheduler_jobs') under the 'scheduler' schema, but it cannot auto-create
+-- the schema itself, which would cause an error on startup if the schema doesn't exist.
+-- The job store is initialized in services/digitize/app.py via
+-- SQLAlchemyJobStore(engine=db_engine, tableschema="scheduler")
 CREATE SCHEMA IF NOT EXISTS scheduler;
 
 -- Connector tables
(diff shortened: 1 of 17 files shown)


services/digitize/db/scripts/init_schema.sql
```diff
@@ -42,6 +42,9 @@ CREATE TABLE IF NOT EXISTS document_checksum (
     doc_id        TEXT        NOT NULL UNIQUE REFERENCES documents(doc_id) ON DELETE CASCADE
 );
 
+-- Schema for APScheduler job store
+CREATE SCHEMA IF NOT EXISTS scheduler;
```
dharaneeshvrd, line 51: Please add a comment on why this is required

dharaneeshvrd (commented):
I see we have moved from async await on most of the places, please ensure it won't be a blocking call and if there is a possibility for it to be blocking, let's run them via background tasks

services/digitize/db/scripts/init_schema.sql
sats-23: Done

sats-23:
> I see we have moved from async await on most of the places, please ensure it won't be a blocking call and if there is a possibility for it to be blocking, let's run them via background tasks

Yes, they are equivalent in effect — but there's one meaningful difference worth knowing.

### `sched.start()` (v3) vs `await sched.start_in_background()` (v4)

| Aspect | v3: `sched.start()` | v4: `await sched.start_in_background()` |
| :--- | :--- | :--- |
| **Blocking?** | No — returns immediately | No — returns immediately |
| **Job dispatch** | Via `call_later` timer callbacks on the running event loop | Via a background asyncio `Task` on the running event loop |
| **Runs in** | Reuses the caller's event loop; `wakeup` is scheduled with `@run_in_event_loop` → `call_later` | Spawns a dedicated `asyncio.Task` that loops internally |
| **Requires running loop** | Yes — `AsyncIOScheduler.start()` calls `asyncio.get_running_loop()` at call time, so it must be called from within an `async` context | Yes — same requirement |
| **Shutdown** | `sched.shutdown()` — synchronous, cancels pending timers | `await sched.stop()` — async, awaits task cancellation |

Both return control to the caller immediately and run jobs on the same event loop without blocking it. The practical difference is purely mechanical: v3 uses `loop.call_later` to wake up at the next job's fire time, while v4 used a persistent background task that awaited a sleep. For this codebase the behaviour is identical — the scheduler starts, fires jobs on schedule, and keeps running until `sched.shutdown()` is called in the lifespan `finally` block.

> **Note:** Because v3 uses `asyncio.get_running_loop()` at `start()` time, `sched.start()` must be called from inside a running coroutine (which it is, inside `_connector_scheduler_lifespan`). Calling it from synchronous code at module import time would fail — but that was already true of v4 as well.

You don't need to wrap `sched.start()` in a background task. Here's why:

`sched.start()` is not a coroutine and not blocking. Internally, it does three things synchronously and returns:

1. Starts executors and job stores
2. Schedules any pending jobs
3. Calls `loop.call_later(wait_seconds, self.wakeup)` — registers a timer callback on the already-running event loop

After that, it returns immediately. The scheduler then lives entirely as a series of `call_later` callbacks on the same event loop — no thread, no background task, no blocking. The `yield` on line 218 is what keeps the lifespan (and the scheduler) alive for the duration of the app.

dharaneeshvrd (approved):

PR #1364, opened 2026-09-01
Extend the global conversion queue mechanism to connectors

manju956:
- Implements 3 lane fair round robin strategy to process U-ING, U-DIG and C-ING requests with each tick allocating prioirty to only one type in the same order. C-ING requests are further processed in round robin style in each subsequent connector tick turns.

### DIFF
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -317,6 +317,7 @@ async def _run_teardown(connector_id: str) -> None:
     and awaited directly from _handle_interrupt in sync_tick (Case A).
 
     Steps:
+      0. Purge queued conversion tasks so the dispatcher never picks them up
       1. Remove the scheduled job so no new ticks fire
       2. Snapshot checksums owned by this connector
       3. Remove ownership rows; delete documents when last owner
@@ -326,6 +327,15 @@ async def _run_teardown(connector_id: str) -> None:
     logger.info(f"Starting teardown for connector {connector_id!r}")
     deletion_errors: list[str] = []
     try:
+        # Step 0: Purge any queued conversion tasks before anything else so the
+        # dispatcher cannot pick up tasks that belong to a connector being deleted.
+        from digitize.db.manager import db_manager
+        deleted_tasks = db_manager.delete_conversion_tasks_for_connector(connector_id)
+        if deleted_tasks:
+            logger.info(
+                f"Purged {deleted_tasks} queued conversion task(s) for connector {connector_id!r}"
+            )
+
         # Step 1: Remove the scheduled job so no new ticks fire after this point.
         try:
             import digitize.connectors.scheduler as _sched
--- a/services/digitize/api/v1/connectors.py
+++ b/services/digitize/api/v1/connectors.py
@@ -313,11 +313,13 @@ async def _run_teardown(connector_id: str) -> None:
     """
     Teardown for connector deletion.
 
-    Scheduled via asyncio.create_task from the delete endpoint (Case B),
-    and awaited directly from _handle_interrupt in sync_tick (Case A).
+    Scheduled via asyncio.create_task from the delete endpoint (Case B — no
+    tick running). Awaited directly from _handle_interrupt in sync_tick
+    (Case A — tick interrupted). In Case A the caller is responsible for
+    purging in-flight conversion tasks *before* calling this function, since
+    no tick is running when _run_teardown executes in Case B.
 
     Steps:
-      0. Purge queued conversion tasks so the dispatcher never picks them up
       1. Remove the scheduled job so no new ticks fire
       2. Snapshot checksums owned by this connector
       3. Remove ownership rows; delete documents when last owner
@@ -327,15 +329,6 @@ async def _run_teardown(connector_id: str) -> None:
     logger.info(f"Starting teardown for connector {connector_id!r}")
     deletion_errors: list[str] = []
     try:
-        # Step 0: Purge any queued conversion tasks before anything else so the
-        # dispatcher cannot pick up tasks that belong to a connector being deleted.
-        from digitize.db.manager import db_manager
-        deleted_tasks = db_manager.delete_conversion_tasks_for_connector(connector_id)
-        if deleted_tasks:
-            logger.info(
-                f"Purged {deleted_tasks} queued conversion task(s) for connector {connector_id!r}"
-            )
-
         # Step 1: Remove the scheduled job so no new ticks fire after this point.
         try:
             import digitize.connectors.scheduler as _sched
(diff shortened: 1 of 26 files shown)


services/digitize/api/v1/connectors.py
```diff
@@ -326,6 +327,15 @@ async def _run_teardown(connector_id: str) -> None:
     logger.info(f"Starting teardown for connector {connector_id!r}")
     deletion_errors: list[str] = []
     try:
+        # Step 0: Purge any queued conversion tasks before anything else so the
+        # dispatcher cannot pick up tasks that belong to a connector being deleted.
+        from digitize.db.manager import db_manager
+        deleted_tasks = db_manager.delete_conversion_tasks_for_connector(connector_id)
```
dharaneeshvrd: Is this needed here?
Since we would be calling from API only when the status is not syncing.
If its coming from run_tick while its syncing, it should ensure to cleanup all the conversion tasks.

@sats-23 ^^

services/digitize/workers/conversion_dispatcher.py
```diff
@@ -151,82 +150,44 @@ async def dispatch_loop() -> None:
...
+                else:                                    # ── Turn 2: Connector Ingestion ──
+                    cids = db_manager.get_connector_ids_with_queued_tasks()
+                    if cids:
+                        cid = cids[_connector_rr_index % len(cids)]
+                        task = _try_claim_if_fits("ingestion", available, connector_id=cid)
+                        if task:
+                            await _dispatch_one(task)
+                        _connector_rr_index = (_connector_rr_index + 1) % len(cids)
+
+                _op_turn = (_op_turn + 1) % 3            # ALWAYS advance — unconditional
```
dharaneeshvrd: no, should alter the turn only when the semaphore is claimed, during large file, it might need to wait right, till the available becomes two?

services/digitize/workers/conversion_dispatcher.py
```diff
@@ -151,82 +150,44 @@ async def dispatch_loop() -> None:
...
+                        await _dispatch_one(task)
+
+                else:                                    # ── Turn 2: Connector Ingestion ──
+                    cids = db_manager.get_connector_ids_with_queued_tasks()
+                    if cids:
+                        cid = cids[_connector_rr_index % len(cids)]
+                        task = _try_claim_if_fits("ingestion", available, connector_id=cid)
+                        if task:
+                            await _dispatch_one(task)
+                        _connector_rr_index = (_connector_rr_index + 1) % len(cids)
```
dharaneeshvrd: here also, need to rotate the turn only after successful claim

dharaneeshvrd (commented):
Somehow changes seems not that complicated 
I was expecting a large change :)

services/digitize/api/v1/connectors.py
sats-23: That's right. The cancel job should take care of removing all docs from conversion tasks queue

services/digitize/connectors/sync_tick.py
```diff
@@ -382,7 +387,31 @@ async def _process_new_files(
...
+            )
+
+            # Launch the ingestion pipeline as a background task so it drives
+            # process → chunk → index after the dispatcher completes each file.
+            # Pass batch_dir as staging_dir so cleanup targets the correct subtree.
+            asyncio.create_task(
+                launch_ingest_pipeline(
+                    job_id=job_id,
+                    doc_id_dict=doc_id_dict,
+                    file_checksum_dict=filename_to_checksum,
```
sats-23: We should not be passing checksum here. Connectors handles it on its own with add_connector_checksum_entry

sats-23:
As discussed can we normalize steps 4, 7, 8, 9 of jobs#create_job such that it can be reused by both connectors and user flow. 
If its too much work around, can we at least document in comments that these are identical steps followed by both flows?

services/digitize/api/v1/connectors.py
manju956: alright, will remove this.

services/digitize/workers/conversion_dispatcher.py
manju956: right, will change this

services/digitize/workers/conversion_dispatcher.py
manju956: right, will modify accordingly

services/digitize/api/v1/connectors.py
manju956: done

services/digitize/workers/conversion_dispatcher.py
manju956: done

services/digitize/workers/conversion_dispatcher.py
manju956: done

services/digitize/connectors/sync_tick.py
manju956: removed checksum param

services/digitize/api/v1/jobs.py
```diff
@@ -295,39 +266,23 @@ async def create_job(
...
+            queued_for_op=queued_for_op,
+            job_name=job_name,
+            already_exists_files=already_exists_files,
+            file_checksum_dict=file_checksum_dict,
+        )
 
-        # 9. Launch the pipeline as a fire-and-forget asyncio task.
-        #    The pipeline polls conversion_tasks and drives post-conversion work.
+        # Digitization jobs need a separate pipeline task (not launch_ingest_pipeline).
         if operation == models.OperationType.DIGITIZATION:
```
dharaneeshvrd: why are we doing this alone in here?
why not part of dg_util.initialize_and_launch?

services/digitize/connectors/sync_tick.py
```diff
@@ -504,6 +515,15 @@ async def _handle_interrupt(
     elif interrupt_type == InterruptType.DELETE_CONNECTOR:
         logger.info(f"Handling delete connector for {connector_id!r}")
         _cancel_tick(sync_seq, connector_id)
+        # Purge conversion tasks the tick may have enqueued before being interrupted.
+        # Must happen here (Case A only) — _run_teardown is shared with Case B where
+        # no tick was running so no in-flight tasks exist to purge.
+        from digitize.db.manager import db_manager
+        deleted_tasks = db_manager.delete_conversion_tasks_for_connector(connector_id)
```
dharaneeshvrd, line 522: I have to update this logic to call cancel on the job, the job will take care of purging the tasks from the queue.
FYI @manju956

services/digitize/api/v1/jobs.py
manju956: addressed this

dharaneeshvrd (approved):

PR #1365, opened 2026-09-01
feat: disconnect application from datasources

Shubhamag12:
Adds a DELETE /api/v1/applications/:id/datasources endpoint that disconnects one or more datasource connectors in an application.

- Accepts a list of datasource_ids in the request body; each datasource is processed independently

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/handlers/datasource_handler.go
+++ b/ai-services/internal/pkg/catalog/apiserver/handlers/datasource_handler.go
@@ -449,4 +449,57 @@ func (h *DatasourceHandler) GetApplicationDatasource(c *gin.Context) {
 	c.JSON(http.StatusOK, resp)
 }
 
+// DisconnectDatasourcesFromApplication godoc
+//
+//	@Summary		Disconnect a datasource from an application
+//	@Description	Removes a single datasource connector from each eligible service in a running application and deletes the service_dependency record.
+//	@Tags			Datasources
+//	@Produce		json
+//	@Security		BearerAuth
+//	@Param			id				path	string	true	"Application ID (UUID)"
+//	@Param			datasource_id	path	string	true	"Datasource ID (UUID)"
+//	@Success		204				"Datasource disconnected successfully"
+//	@Failure		400				{object}	ErrorResponse	"Invalid application or datasource ID"
+//	@Failure		401				{object}	ErrorResponse	"Unauthorized"
+//	@Failure		404				{object}	ErrorResponse	"Datasource not connected to this application"
+//	@Failure		502				{object}	ErrorResponse	"Downstream Digitize service returned an error"
+//	@Failure		500				{object}	ErrorResponse	"Internal Server Error"
+//	@Router			/applications/{id}/datasources/{datasource_id} [delete]
+func (h *DatasourceHandler) DisconnectDatasourcesFromApplication(c *gin.Context) {
+	applicationID, err := uuid.Parse(c.Param("id"))
+	if err != nil {
+		c.JSON(http.StatusBadRequest, ErrorResponse{
+			Error: fmt.Sprintf("Invalid application ID: %v", err),
+		})
+
+		return
+	}
+
+	datasourceID, err := uuid.Parse(c.Param("datasource_id"))
+	if err != nil {
+		c.JSON(http.StatusBadRequest, ErrorResponse{
+			Error: fmt.Sprintf("Invalid datasource ID: %v", err),
+		})
+
+		return
+	}
+
+	if err := h.datasourceSvc.DisconnectDatasourcesFromApplication(c.Request.Context(), applicationID, datasourceID); err != nil {
+		if valErr, ok := err.(*repository.ValidationError); ok {
+			c.JSON(valErr.Code, ErrorResponse{Error: valErr.Message})
+
+			return
+		}
+
+		logger.ErrorfCtx(c.Request.Context(), "failed to disconnect datasource from application: %v", err)
+		c.JSON(http.StatusInternalServerError, ErrorResponse{
+			Error: "Failed to disconnect datasource from application",
+		})
+
+		return
+	}
+
+	c.Status(http.StatusNoContent)
+}
+
 // Made with Bob
--- a/ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
+++ b/ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
@@ -33,6 +33,8 @@ type ValidationError = validators.ValidationError
 // services that accept datasource connectors.
 type ServiceConnectorClientInterface interface {
 	Connect(ctx context.Context, baseURL string, req apimodels.ConnectDatasourceRequest) error
+	// Disconnect calls DELETE /v1/connectors/{connectorID} on the service.
+	Disconnect(ctx context.Context, baseURL, connectorID string) error
 }
 
 // DatasourceService is the single implementation of the datasource connector business logic.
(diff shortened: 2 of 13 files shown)


ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -708,6 +710,69 @@ func (s *DatasourceService) UpdateDatasource(ctx context.Context, id uuid.UUID,
...
+		return nil, &ValidationError{
+			Code:    http.StatusUnprocessableEntity,
+			Message: "no eligible running service with an API endpoint found in application",
+		}
+	}
+
+	var disconnErrors []apimodels.DatasourceConnectionError
+
+	for _, datasourceID := range datasourceIDs {
+		if disconnectErr := s.disconnectOneDatasource(ctx, datasourceID, linkedServices); disconnectErr != nil {
```
prachikurade1: Should we check if the datasource was connected before initiating disconnect ?

prachikurade1 (changes requested):
The UI disconnects one datasource at a time, so the proposal document would need update since the original API was expecting a single datasource id.

ai-services/internal/pkg/catalog/apiserver/models/datasource.go
```diff
@@ -202,4 +202,16 @@ type ConnectDatasourcesResponse struct {
 	Errors []DatasourceConnectionError `json:"errors"`
 }
 
+// DisconnectDatasourcesRequest is the request body for disconnecting one or more datasources from an application.
+type DisconnectDatasourcesRequest struct {
+	// DatasourceIDs is the list of datasource connector UUIDs to disconnect.
+	DatasourceIDs []string `json:"datasource_ids" binding:"required,min=1"`
+}
```
mayuka-c: Any reason on why this is a list?. I thought we support only 1 connected datasource to be deleted per request

ai-services/internal/pkg/catalog/apiserver/handlers/datasource_handler.go
```diff
@@ -353,4 +353,81 @@ func (h *DatasourceHandler) ConnectDatasourcesToApplication(c *gin.Context) {
...
+	}
+
+	var req models.DisconnectDatasourcesRequest
+	if err := c.ShouldBindJSON(&req); err != nil {
+		c.JSON(http.StatusBadRequest, ErrorResponse{
+			Error: fmt.Sprintf("Invalid request body: %v", err),
+		})
+
+		return
+	}
```
mayuka-c: Its not considered a good practice for DELETE in http request to have request body

ai-services/internal/pkg/catalog/apiserver/handlers/datasource_handler.go
```diff
@@ -353,4 +353,81 @@ func (h *DatasourceHandler) ConnectDatasourcesToApplication(c *gin.Context) {
...
+		if parseErr != nil {
+			c.JSON(http.StatusBadRequest, ErrorResponse{
+				Error: fmt.Sprintf("Invalid datasource ID %q: %v", idStr, parseErr),
+			})
+
+			return
+		}
+
+		datasourceIDs = append(datasourceIDs, id)
+	}
```
mayuka-c: This seems to be same as what would have been done for Connect, Can we not reuse?

ai-services/internal/pkg/catalog/apiserver/router.go
```diff
@@ -105,6 +105,8 @@ func registerApplicationRoutes(v1 *gin.RouterGroup, h *handlers.ApplicationHandl
 		g.GET("/:id/ps", h.ApplicationPS)
 		// PUT /api/v1/applications/:id/datasources — connect one or more datasources to application
 		g.PUT("/:id/datasources", datasourceH.ConnectDatasourcesToApplication)
+		// DELETE /api/v1/applications/:id/datasources — disconnect one or more datasources from application
+		g.DELETE("/:id/datasources", datasourceH.DisconnectDatasourcesFromApplication)
```
mayuka-c: I would highly recommend instead of having request body with 1 item in list which is not a good practice for delete, let us use path param and support by default 1

/:id/datasources/:id

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -708,6 +710,69 @@ func (s *DatasourceService) UpdateDatasource(ctx context.Context, id uuid.UUID,
...
+func (s *DatasourceService) DisconnectDatasourcesFromApplication(ctx context.Context, applicationID uuid.UUID, datasourceIDs []uuid.UUID) (*apimodels.DisconnectDatasourcesResponse, error) {
+	linkedServices, err := s.eligibleServicesForApp(ctx, applicationID)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(linkedServices) == 0 {
+		return nil, &ValidationError{
+			Code:    http.StatusUnprocessableEntity,
+			Message: "no eligible running service with an API endpoint found in application",
```
mayuka-c: The error message doesnt look good for end user.
We should mention appropriate message "There are no connections exist for given id: %s". Please refine accordingly

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -708,6 +710,69 @@ func (s *DatasourceService) UpdateDatasource(ctx context.Context, id uuid.UUID,
...
+			continue
+		}
+
+		if err := s.serviceClient.Disconnect(ctx, svc.URL, datasourceID.String()); err != nil {
+			return &ValidationError{
+				Code:    http.StatusBadGateway,
+				Message: fmt.Sprintf("failed to disconnect datasource from service %s: %v", svc.ServiceCatalogID, err),
+			}
+		}
+
```
mayuka-c: We need to ensure idempotency.
Lets say if we are able to disconnect it from service and the next step failed (Db deletion). We need to ensure the next time user initiates delete we will not throw error but just delete endtry in DB

mayuka-c (changes requested):

ai-services/internal/pkg/catalog/apiserver/models/datasource.go
Shubhamag12: made changes

ai-services/internal/pkg/catalog/apiserver/handlers/datasource_handler.go
Shubhamag12: removed

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
Shubhamag12: updated

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
Shubhamag12: added a check where if service is not found, we then go ahead with db deletion

ai-services/internal/pkg/catalog/client/service.go
```diff
@@ -118,3 +118,26 @@ func (c *ServiceClient) Connect(ctx context.Context, baseURL string, req apimode
...
+	// 404 means the connector was already removed from the downstream service
+	// (e.g. a previous attempt succeeded here but failed at DB cleanup).
+	// Treat it as success so the DB row can still be cleaned up idempotently.
+	if resp.StatusCode() == http.StatusNotFound {
+		return nil
+	}
+
+	if resp.IsError() {
+		return fmt.Errorf("service returned unexpected status %d", resp.StatusCode())
+	}
```
mayuka-c: Can we avoid doing StatusNotFound check here but rather let it error out here wi the appropriate status code and let the caller handle it.

We may want to ignore that in Disconnect flow and in future in other place if used we may want to error out

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -708,6 +710,53 @@ func (s *DatasourceService) UpdateDatasource(ctx context.Context, id uuid.UUID,
...
+//
+// Flow:
+//  1. Resolve eligible services for the application (same filter as connect).
+//  2. Call DELETE /v1/connectors/{id} on each eligible service's Digitize endpoint,
+//     then remove the service_dependency row.
+func (s *DatasourceService) DisconnectDatasourcesFromApplication(ctx context.Context, applicationID uuid.UUID, datasourceID uuid.UUID) error {
+	linkedServices, err := s.eligibleServicesForApp(ctx, applicationID)
+	if err != nil {
+		return err
+	}
```
mayuka-c: Can we first do check to ensure the application exists for a given id that user has passed, if not error out

mayuka-c: Also I feel better to ensure datasourceid also exists check is done after app id existence

mayuka-c (commented):
Also pls update proposal with latest info

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
Shubhamag12: added a check

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
Shubhamag12: added check

prachikurade1 (dismissed):
approved

prachikurade1 (approved):
approved

mayuka-c (approved):
LGTM

PR #1366, opened 2026-09-01
Application templates cli support for BYOS

mayuka-c:
- Application templates commands uses the API to list custom bundles.
- Preserve fallback logic to ensure embedded assets are printed too if not logged into catalog service

### DIFF
--- a/ai-services/internal/pkg/catalog/client/catalog_source.go
+++ b/ai-services/internal/pkg/catalog/client/catalog_source.go
@@ -75,6 +75,7 @@ func (s *apiWithFallback) ListArchitectures(ctx context.Context) ([]types.Archit
 	if err != nil {
 		if isConnectivityError(err) {
 			logger.DebugfCtx(ctx, "API ListArchitectures failed, falling back to embedded: %v", err)
+
 			return toArchitectureSummaries(s.embedded.ListArchitectures())
 		}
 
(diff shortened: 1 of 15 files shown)


ai-services/internal/pkg/catalog/client/catalog_source.go
```diff
@@ -0,0 +1,296 @@
...
+		for j, s := range a.Services {
+			svcIDs[j] = s.ID
+		}
+		summaries[i] = types.ArchitectureSummary{
+			ID:          a.ID,
+			Name:        a.Name,
+			Description: a.Description,
+			CertifiedBy: a.CertifiedBy,
+			Services:    svcIDs,
+		}
```
yussufsh: We can reuse existing catalog.ToArchitectureSummary

ai-services/internal/pkg/catalog/client/catalog_source.go
```diff
@@ -0,0 +1,296 @@
...
+	for i, s := range svcs {
+		summaries[i] = types.ServiceSummary{
+			ID:            s.ID,
+			Name:          s.Name,
+			Description:   s.Description,
+			CertifiedBy:   s.CertifiedBy,
+			Architectures: s.Architectures,
+			Standalone:    s.Standalone,
+			Dependencies:  s.Dependencies,
+		}
```
yussufsh: same

ai-services/internal/pkg/catalog/client/catalog_source.go
mayuka-c: Updated

yussufsh (approved):

PR #1371, opened 2026-09-01
support delete op and applications api response

yussufsh:
1. Supports deletion through the resolved local or remote worker runtime.
2. Includes worker ID, name, and runtime type in application API responses.
3. Uses the worker’s registered runtime when planning deployments.
4. Consolidates Podman/OpenShift application service implementations.
5. Updates Swagger documentation and CLI calls accordingly.


ai-services/internal/pkg/catalog/apiserver/repository/application_service/common.go
```diff
@@ -91,12 +92,70 @@ type ApplicationServiceBase struct {
...
+		rtStr, _ := s.WorkerRegistry.WorkerRuntimeType(workerName)
+		rt, err := runtime.NewRuntimeFactory(runtimeTypes.RuntimeType(rtStr)).CreateRemote(workerName, s.WorkerRegistry)
+		if err != nil {
+			return nil, fmt.Errorf("create remote runtime for worker %q: %w", workerName, err)
+		}
+
+		return rt, nil
+	}
+
+	return vars.RuntimeFactory.Create(catalogutils.AppNamespace(app.ID))
```
mayuka-c, line 135: This means it is openshift?

ai-services/internal/pkg/catalog/apiserver/repository/application_service/openshift.go
```diff
@@ -1,47 +0,0 @@
-package applicationservice
-
-import (
```
mayuka-c, line 3: I dont remember much, is it because of namespace we had 2 diff files with podman and openshift is it?

mayuka-c (commented):
Overall LGTM, have few queries

ai-services/internal/pkg/catalog/apiserver/repository/application_service/common.go
yussufsh: Yes, until we move to auto Local will be used in podman too.

ai-services/internal/pkg/catalog/apiserver/repository/application_service/openshift.go
yussufsh: Not sure; it seems not required, the funcs itself were not consistent.
This code needs lots of refactoring.

ai-services/internal/pkg/catalog/types/application_response.go
```diff
@@ -19,6 +26,7 @@ type Application struct {
 	Services       []ApplicationService `json:"services,omitempty"`
 	CreatedAt      string               `json:"created_at"`
 	UpdatedAt      string               `json:"updated_at"`
+	Worker         *ApplicationWorker   `json:"worker,omitempty"`
 }
```
mayuka-c, line 30: just a query,

Will the GetAppByID response also have Worker information?
I remember it should use same struct, but please check once

mayuka-c (approved):
LGTM

ai-services/internal/pkg/catalog/types/application_response.go
yussufsh: Yes Get/Update reponse will have worker struct.

PR #1372, opened 2026-09-01
AIS-1670 - Get Application Datasource by ID

prachikurade1:
GET `api/v1/applications/:appid/datasources/:datasourceid`

Sample response:
```
{
  "id": "41923864-6eeb-47d9-ad1b-1d1cbedef78e",
  "name": "fs-ds",
  "status": "connected",
  "provider": {
    "id": "file_system",
    "name": "File system"
  },
  "service_details": {
    "sync_status": "completed",
    "total_files": 7,
    "last_sync_at": "2026-09-01T12:29:50.652490Z",
    "last_sync_error": null
  }
}
```
Note: The status "completed" is a bug from the Digitize API which is being fixed.


mayuka-c:
Is this being consumed anywhere?

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -708,6 +708,78 @@ func (s *DatasourceService) UpdateDatasource(ctx context.Context, id uuid.UUID,
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to query linked services for datasource %s: %w", datasourceID, err)
+	}
+
+	var appRows []dbrepo.LinkedServiceRow
+	for _, row := range allRows {
+		if row.ApplicationID == applicationID {
+			appRows = append(appRows, row)
+		}
+	}
```
mayuka-c: I'm thinking maybe we can do this filtering in DB and filter with application ID and let the DB return as anyway DB would be scanning. This avoids us to again loop

Let us not do this now, but just mentioning

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -708,6 +708,78 @@ func (s *DatasourceService) UpdateDatasource(ctx context.Context, id uuid.UUID,
...
+	if err != nil {
+		if err == dbrepo.ErrConnectorNotFound {
+			return nil, &ValidationError{
+				Code:    http.StatusNotFound,
+				Message: fmt.Sprintf("datasource %s not found", datasourceID),
+			}
+		}
+
+		return nil, fmt.Errorf("failed to fetch datasource: %w", err)
+	}
```
mayuka-c: I feel we need reordering
- Check for application exist
- Check for connector exist
- Then do `datasource not connected to this application`

mayuka-c (commented):
REst LGTM

prachikurade1:
> Is this being consumed anywhere?

No, it will not be consumed anywhere currently.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
prachikurade1: Resolving since this will be enhancement later

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -708,6 +708,92 @@ func (s *DatasourceService) UpdateDatasource(ctx context.Context, id uuid.UUID,
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to query linked services for datasource %s: %w", datasourceID, err)
+	}
+
+	var appRows []dbrepo.LinkedServiceRow
+	for _, row := range allRows {
+		if row.ApplicationID == applicationID {
+			appRows = append(appRows, row)
+		}
+	}
```
mayuka-c: allRows would anyway give you list of applications to which a datasource is connected.

For a given applicationID you will always have 1 record, there cannot be same datasource with same application present twice. So why not compute the baseURL in this for loop and just break here.

This avoids unnecessary looping and no need of appRows as list, just need a boolean to say if present or not and use it in line 768

mayuka-c (dismissed):
LGTM

mayuka-c (approved):

PR #1374, opened 2026-09-01
Worker pod correctness

yussufsh:
### Worker pod correctness (the bulk of the work)

1. Image bumped v0.0.285 → v0.0.304 — fixes the unknown flag: --token / missing grpcserver subcommand error
2. Added REGISTRY_AUTH_FILE, AI_SERVICES_BASE_DIR env vars, podman-auth secret mount, ai-services-data basedir mount, and run.oci.keep_original_groups annotation — all missing vs the catalog backend
3. New auth-secret.yaml.tmpl + layer 1 in metadata.yaml — creates podman-auth-secret on the worker machine at join time (same pattern as catalog; reset-podman-auth flag to follow)
4. deploy.go now resolves the actual podman socket at runtime, reads + base64-encodes the auth file, passes AppTemplateName/Version to templates, and forwards --basedir via optionalFlags

### HTTPProxy simplification

- Removed podNameToIP + resolvePodNameInURL — the worker runs in the same podman network so pod-name DNS works natively; no host-side IP lookup needed
### Image version tracking

- Catalog + worker pinned to v0.0.304; CI script now enforces parity between them

### DIFF
--- a/ai-services/internal/pkg/worker/deploy/deploy.go
+++ b/ai-services/internal/pkg/worker/deploy/deploy.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"slices"
 	"strconv"
+	"strings"
 	ttemplate "text/template"
 
 	"github.com/project-ai-services/ai-services/assets"
@@ -21,6 +22,7 @@ import (
 	podmodels "github.com/project-ai-services/ai-services/internal/pkg/models"
 	"github.com/project-ai-services/ai-services/internal/pkg/runtime"
 	"github.com/project-ai-services/ai-services/internal/pkg/specs"
+	"github.com/project-ai-services/ai-services/internal/pkg/utils"
 	workerconstants "github.com/project-ai-services/ai-services/internal/pkg/worker/constants"
 
 	k8syaml "sigs.k8s.io/yaml"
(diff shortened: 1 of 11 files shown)


ai-services/internal/pkg/worker/deploy/deploy.go
```diff
@@ -195,6 +201,55 @@ func deployAll(ctx context.Context, rt runtime.Runtime, tp templates.Template, o
...
+
+	authFileBase64, err := readAuthFileBase64()
+	if err != nil {
+		return nil, err
+	}
+
+	return map[string]string{
+		"caddy.httpsPort":               strconv.Itoa(opts.HTTPSPort),
+		"worker.token":                  token,
+		"worker.gatewayAddr":            gatewayAddr,
```
adarshagrawal38: lets move keys to `workerconstant`, we can reuse keys for openshift deployment.

adarshagrawal38:
Wondering how user will reset podman-auth secret in the worker machine?

Given that: user might not configure catalog in worker machine.

yussufsh:
> Wondering how user will reset podman-auth secret in the worker machine?
> 
> Given that: user might not configure catalog in worker machine.

This is TODO for the worker command.

ai-services/internal/pkg/worker/deploy/deploy.go
yussufsh: done.

mayuka-c (approved):
LGTM

PR #1377, opened 2026-09-02
API for Steps md rendering and integrate with CLIs

mayuka-c:
- Currently the CLI directly uses embedded assets to print the info and next steps.
- This will not work for custom assets as it resides inside volume.
- So a separate API is exposed so that it returns steps md for both embedded and custom assets.

### DIFF
--- a/ai-services/internal/pkg/catalog/catalog.go
+++ b/ai-services/internal/pkg/catalog/catalog.go
@@ -2,6 +2,7 @@ package catalog
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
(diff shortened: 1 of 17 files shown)


mayuka-c:
There is problem in Info command because it expects containers to have certain formats like <name>-ui and <name>-backend. This needs to be made more generic and should not be restricted (requires changes in existing code as its not generic). Will pick it up next

image

ai-services/internal/pkg/catalog/catalog.go
```diff
@@ -1079,4 +1079,52 @@ func (p *CatalogProvider) LoadServicesMD(serviceID string) (map[string]*texttemp
...
+	stepsPath := filepath.Join(servicePath, string(runtime), "steps")
+
+	itemFS, err := p.getItemFS(serviceID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get item filesystem: %w", err)
+	}
+
+	files := make(map[string][]byte)
+
+	err = fs.WalkDir(itemFS, stepsPath, func(path string, d fs.DirEntry, err error) error {
```
yussufsh, line 1112: We should check whether the steps dir exists. Otherwise this will fail I guess.
Since it is an optional dir we can skip it or return an empty map.

mayuka-c: Let me verify how it behaves if its not present

mayuka-c: Thanks looks like it fails if steps dir is not present.
Added a check to do early return with empty map if not present

yussufsh (approved):

PR #1378, opened 2026-09-02
npm CVE fixes

pranithraoibm:
- Fixed a high CVE fix for `browserslist` across all UI folders that was flagged by `npm audit` by running `npm audit fix`
- Fixed the `qs` package CVE by bumping it to `6.16.0` (the latest version of `qs`) using `npm overrides`. Ran `npm run build` and confirmed that there were no build failures post this override.

### DIFF
--- a/ui/chatbot/package.json
+++ b/ui/chatbot/package.json
@@ -15,6 +15,9 @@
         "react-dom": "^18.3.1",
         "sass": "^1.98.0"
     },
+    "overrides": {
+        "qs": "^6.16.0"
+    },
     "devDependencies": {
         "@ianvs/prettier-plugin-sort-imports": "^4.7.1",
         "@vitejs/plugin-react": "^5.2.0",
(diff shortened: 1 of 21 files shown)


ui/chatbot/package.json
```diff
@@ -15,6 +15,9 @@
         "react-dom": "^18.3.1",
         "sass": "^1.98.0"
     },
+    "overrides": {
+        "qs": "^6.16.0"
+    },
```
mayuka-c, line 20: Can we not bump the corresponding direct dependency which is utilizing this pkg or still its not available so going with overrides?

Also I feel it will be helpful if we could add comments, mentioning which pkg depends on it and can be removed once the fix is available there

pranithraoibm: `qs` is coming from `body-parser` package(it's latest version is used in our project) which pins it at `6.15.2` which is coming from `express`(also the latest available version), so `overrides` is the option for us to fix this CVE.

body-parser: https://github.com/expressjs/body-parser/releases#release-v2.3.0
express: https://github.com/expressjs/express/releases/tag/v5.2.1

image

Added an comment for this override in the `readme` file of chatbot folder as `package.json` doesn't support comments

mayuka-c (approved):
LGTM

adarshagrawal38 (approved):

PR #1379, opened 2026-09-02
Ensure Version consistency across Embedded services/components

mayuka-c:
- Currently the helm chart is having diff version but should ideally have the same version number same as what we maintain in our runtimeMetadata.
- Also introduced version in rootMetadata as the version should be same across both runtimes.
- For now maintaining versions in runtimeMetadata, but in future we need to maintain in only rootMetadata and ensure the code while deploying populates appropriate version specified in rootMetadata required for runtime specific.
- Already the above checks are taking care for custom assets to ensure consistency.

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.329
+TAG?=v0.0.330
 CONTAINER_BUILDER?=podman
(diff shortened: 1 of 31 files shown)


mayuka-c:
Have added quotes as helm also prefers the version to have quotes. So now everything in metadata and Chart.yaml have quotes

Shubhamag12 (dismissed):
lgtm

yussufsh (dismissed):
lgtm
Thanks @mayuka-c

Shubhamag12 (approved):

PR #1381, opened 2026-09-02
Generic steps/md files rendering

mayuka-c:
- Currently the next.md and info.md rendering is hardcoded which doesnt work for BYOS.
- Ensuring the above is made generic by making use of vars_file.yaml (Which was not used earlier).
- Updates in vars_file for both podman and openshift to make it generic.
- Ensuring consistency of namings in vars_file and remove redundant ones for all emebedded services.
- Allows user to have their own custom alias defined in vars_file and use in the md.
- Continuation of this PR: https://github.com/IBM/project-ai-services/pull/1377

Note:
- Require this PR: https://github.com/IBM/project-ai-services/pull/1377 to be merged before merging this.
- Also includes a bug fix on Openshift where the Helm chart load was not considering custom bundles.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/create.go
+++ b/ai-services/cmd/ai-services/cmd/application/create.go
@@ -19,14 +19,15 @@ import (
 	apiModels "github.com/project-ai-services/ai-services/internal/pkg/catalog/apiserver/models"
 	catalogClient "github.com/project-ai-services/ai-services/internal/pkg/catalog/client"
 	catalogTypes "github.com/project-ai-services/ai-services/internal/pkg/catalog/types"
+	catalogUtils "github.com/project-ai-services/ai-services/internal/pkg/catalog/utils"
 	appFlags "github.com/project-ai-services/ai-services/internal/pkg/cli/constants/application"
 	"github.com/project-ai-services/ai-services/internal/pkg/cli/flagvalidator"
 	"github.com/project-ai-services/ai-services/internal/pkg/cli/helpers"
 	"github.com/project-ai-services/ai-services/internal/pkg/cli/templates"
-	catalogUtils "github.com/project-ai-services/ai-services/internal/pkg/catalog/utils"
 	cliutils "github.com/project-ai-services/ai-services/internal/pkg/cli/utils"
 	"github.com/project-ai-services/ai-services/internal/pkg/image"
 	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+	runtimeTypes "github.com/project-ai-services/ai-services/internal/pkg/runtime/types"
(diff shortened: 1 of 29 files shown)


mayuka-c:
Openshift Validations
---

image

ai-services/internal/pkg/catalog/utils/common.go
```diff
@@ -112,14 +112,20 @@ func SanitizeFilePath(path string) string {
 
 // LoadChartFromCatalogFS walks assets.CatalogFS at catalogPath and returns a Helm chart.
 func LoadChartFromCatalogFS(catalogPath string) (helmchart.Charter, error) {
+	return LoadChartFromFS(&assets.CatalogFS, catalogPath)
```
Shubhamag12: can we directly implement the `LoadChartFromFS` function here if this is the only step performed?

ai-services/internal/pkg/catalog/catalog.go
```diff
@@ -372,6 +372,13 @@ func (p *CatalogProvider) getItemFS(key string) (fs.FS, error) {
 	return item.itemFS, nil
 }
 
+// GetItemFS returns the filesystem for the catalog item identified by key.
+// Embedded items return &assets.CatalogFS; bundle items return an os.DirFS
+// rooted at their extracted directory on disk.
+func (p *CatalogProvider) GetItemFS(key string) (fs.FS, error) {
```
Shubhamag12: is this wrapper required? can we not directly export getItemFS?

ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -546,46 +548,68 @@ func printNextSteps(ctx context.Context, app *catalogTypes.Application) error {
...
-			continue
-		}
+		return
+	}
 
-		err = printNextStepsMD(tmpls, params)
-		if err != nil {
-			logger.Warningf("Failed to render next steps for service '%s': %v\n", service.CatalogID, err)
+	// Populate status params generically from vars_file.yaml
+	if appPS != nil {
```
Shubhamag12: we should have a similar safe-check at this place as well in info
https://github.com/IBM/project-ai-services/pull/1381/changes#diff-75ff5851a486099ea5436caf75dacbda56a66ee99d13af40e08deffdfc42d05dR139

mayuka-c: Acutally nil check is not required. It can become nil if only there is error and if error we anyway return before proceeding.

It was added in create.go because there it was not failing but made sure it fails early and removed nil check there

yussufsh (approved):

PR #1383, opened 2026-09-02
Changes for remote register routes

yussufsh:
Previously, BuildRoutesFromAnnotation read DOMAIN_SUFFIX from the environment at parse time and stamped the full domain (catalog-api.example.com) directly onto each route. This was correct for the catalog but broken for remote workers: the routes were serialised into the gRPC payload and sent to the worker, which then registered the catalog's domain in its own Caddy instead of its own.

The fix splits responsibility cleanly across two layers:

- BuildRoutesFromAnnotation (utils.go) — now sets Route.Domain to just the subdomain (e.g. "catalog-api"). No env read, no domain construction. The subdomain travels through the gRPC payload as neutral data.

- RegisterRoute (caddy.go) — now owns full domain construction. It reads DOMAIN_SUFFIX from the local environment on whichever machine is running Caddy, then builds domain = route.Domain + "." + domainSuffix. Both the Caddy host-match rule and the returned ExternalURL use this locally-assembled domain.

The effect: the catalog registers catalog-api.<catalog-domain> against its own Caddy, and a remote worker registers catalog-api.<worker-domain> against its own Caddy — each using the DOMAIN_SUFFIX of the machine where Caddy is actually running.

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.315
+TAG?=v0.0.316
 CONTAINER_BUILDER?=podman
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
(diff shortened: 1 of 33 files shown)


adarshagrawal38 (dismissed):
LGTM

adarshagrawal38 (approved):

PR #1336, opened 2026-08-26
feat: add Worker Resources page with table

pranithraoibm:
- New route `/worker-resources` and nav item between Connectors and the divider
- `WorkerResourcesTable` component wired up with search, column toggle, CSV export, pagination, and a Register primary button
- GET `/api/v1/workers` shape aligned with PR https://github.com/IBM/project-ai-services/pull/1261 — WorkerStatus (pending | ready | disconnected), runtime_type (podman | openshift | unknown)
- Fixed OverflowMenuPanel in TableToolbarActions to absorb Carbon-injected closeMenu / handleOverflowMenuItemFocus props which fixes React DOM warnings in all tables

image

PR #1389, opened 2026-09-03
Add worker details to DA & SVC tables and deployment details

pranithraoibm:
- Added `Worker resource` & `Worker type` columns to both the DA and services table.
- Added them to the column filters list as well as to the CSV export functionality.

image (2)
image (3)

- Added worker details to the DA and service deployment details page as well
image (1)
image (4)


yussufsh:
Screenshot shows `podman` as worker type.

pranithraoibm:
@yussufsh Yes, added the mapping will add the updated screenshots soon
Edit: Added the updated screenshot

ui/catalog/src/constants/app.constants.ts
```diff
@@ -1,3 +1,9 @@
+// runtime mapping
+export const RUNTIME_TYPE_LABELS: Record<string, string> = {
```
ryanmarc: This gets the raw podman off the table, which was the main thing, thanks. But it does it with a second map rather than the WORKER_RUNTIME_LABELS fifteen lines down, so the two surfaces still disagree - the table now says RHAIIS and the details card at DeploymentDetails.tsx:528 still says Red Hat Enterprise Linux (RHAIIS).

I bundled the two modules and ran both paths over the real enum values to be sure I wasn't misreading:
```
value      | table cell | details card
podman     | RHAIIS     | Red Hat Enterprise Linux (RHAIIS)
openshift  | RHOAI      | Red Hat OpenShift (RHOAI)
```

If the short form in the table is deliberate (and it is a sensible call for a narrow column), could we get it from one map instead of two? Something like adding short to the existing entries:
```
podman: { short: "RHAIIS", label: "Red Hat Enterprise Linux (RHAIIS)", description: ... },
```
then the cell uses .short and the card uses .label. One place to edit when a product gets renamed, and it also lets powervs and unknown be handled once rather than twice

ui/catalog/src/pages/DigitalAssistants/CellRenderers.tsx
```diff
@@ -74,9 +75,15 @@ interface CellRendererProps {
 
 type CellRendererComponent = (props: CellRendererProps) => ReactElement | null;
 
+export const WorkerTypeCell = ({ value }: CellRendererProps): ReactElement => {
+  const raw = String(value ?? "");
+  return <span>{RUNTIME_TYPE_LABELS[raw] ?? raw}</span>;
```
ryanmarc: Mapping in the renderer changes what is drawn but not what is in the row, and two things read the row instead:

- Search. DigitalAssistants.tsx:279 runs filterRowsBySearch over state.rowsData with workerType in the field list, and that helper compares String(row[field]). I ran it: searching RHAIIS returns 0 rows, searching podman returns 1. So the column says RHAIIS, the user types RHAIIS, and they get the "no results" empty state. That worked before this commit, when the column showed the raw value.
- CSV export. useCSVExport.ts:56 filters rows and passes them straight to the writer without going through CELL_RENDERERS, so the file says podman while the screen said RHAIIS.
DeployedServicesTable has both of the same problems at :268 and :87.

ryanmarc (changes requested):

ui/catalog/src/constants/app.constants.ts
pranithraoibm: Right, added short names as suggested.

ui/catalog/src/pages/DigitalAssistants/CellRenderers.tsx
pranithraoibm: Fixed it in all the 3 tables by pushing the mapping upstream to the transform functions, so the row itself always carries the display value

ui/catalog/src/api/workerResources.api.ts
```diff
@@ -14,7 +15,8 @@ export function transformWorkerToRow(
     id: worker.id,
     name: worker.name,
     status: worker.status,
-    runtime_type: worker.runtime_type,
+    runtime_type:
```
ryanmarc: Moving the label into the transform is the right fix for the DA and SVC tables, but here it changes what row.runtime_type means on the Worker Resources table: it used to be the enum, now it is RHAIIS.

#1359 (and #1424, same modal) build the cleanup command from that row: DeregisterWorkerModal.tsx:15 does ai-services worker uninstall --runtime ${runtimeType} --yes with runtimeType={modalRow?.runtime_type ?? ""}. I ran the merged transform through that function: a podman worker produces --runtime RHAIIS --yes, and the CLI's InitAndValidateRuntimeFlag only accepts podman or openshift, so the copied command fails. No compile error, because WorkerResourceRow.runtime_type is string, and the two branches merge without conflict, so it ships quietly in either order.

Suggest keeping runtime_type: worker.runtime_type as the enum and adding a separate display field (runtime_label?) with the .short lookup, then pointing the column key, searchFields and DEFAULT_VISIBLE_COLUMNS at that. Typing runtime_type as WorkerRuntimeType while you are there would make the next one a type error.

ryanmarc (changes requested):

ui/catalog/src/api/workerResources.api.ts
pranithraoibm: Done, added a separate display field as suggested

pranithraoibm:
@ryanmarc I have added a small extra change to add message column to the workers table, I had added it in https://github.com/IBM/project-ai-services/pull/1336 but looks like while merging main that commit was lost, PTAL at these changes as well

ryanmarc (approved):

PR #1390, opened 2026-09-03
worker api having msg and apps

yussufsh:
- DB schema — added message TEXT column to the workers table to persist status transition reasons.
- Worker messages — new messages.go constants (MsgDisconnected, MsgLostHeartbeat) as the single source of truth; registry's Disconnect and SweepStale write these to the DB.
- Worker repo — WorkerUpdate gained a Message field; Update() persists it; new GetApplicationIDsByWorkerIDs bulk-queries which applications belong to a set of workers.
- API types — catalogtypes.Worker now exposes message and application_ids on the wire.
- Worker handler — NewWorkerHandler takes a WorkerRepository; ListWorkers enriches every worker with its message and app IDs in one query; new GET /workers/:id single-worker endpoint added.
- CLI — worker list table gained MESSAGE and APPS columns (app count, not IDs).

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/handlers/worker_handler.go
+++ b/ai-services/internal/pkg/catalog/apiserver/handlers/worker_handler.go
@@ -1,23 +1,28 @@
 package handlers
 
 import (
+	"context"
 	"net/http"
 	"strings"
 
 	"github.com/gin-gonic/gin"
 	"github.com/google/uuid"
+	"github.com/project-ai-services/ai-services/internal/pkg/catalog/db/models"
+	"github.com/project-ai-services/ai-services/internal/pkg/catalog/db/repository"
 	catalogtypes "github.com/project-ai-services/ai-services/internal/pkg/catalog/types"
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
 	"github.com/project-ai-services/ai-services/internal/pkg/worker/registry"
 )
 
 // WorkerHandler handles worker management endpoints.
 type WorkerHandler struct {
-	reg *registry.Registry
+	reg  *registry.Registry
+	repo repository.WorkerRepository
 }
 
 // NewWorkerHandler creates a new WorkerHandler.
-func NewWorkerHandler(reg *registry.Registry) *WorkerHandler {
-	return &WorkerHandler{reg: reg}
+func NewWorkerHandler(reg *registry.Registry, repo repository.WorkerRepository) *WorkerHandler {
+	return &WorkerHandler{reg: reg, repo: repo}
 }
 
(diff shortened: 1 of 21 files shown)


ai-services/internal/pkg/catalog/apiserver/handlers/worker_handler.go
```diff
@@ -80,38 +84,88 @@ func (h *WorkerHandler) CreateWorker(c *gin.Context) {
...
+
+	for _, w := range workers {
+		if w.ID != workerID {
+			continue
+		}
+
+		appIDMap, err := h.repo.GetApplicationIDsByWorkerIDs(ctx, []uuid.UUID{workerID})
+		if err != nil {
+			c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to fetch application IDs"})
+
```
mayuka-c: Can we print atleast logs with err so that we know what is the exact error as we are not returning it

ai-services/internal/pkg/catalog/apiserver/handlers/worker_handler.go
```diff
@@ -80,38 +84,88 @@ func (h *WorkerHandler) CreateWorker(c *gin.Context) {
...
+		c.JSON(http.StatusBadRequest, gin.H{"error": "invalid worker id"})
+
+		return
+	}
+
+	ctx := c.Request.Context()
+
+	workers, err := h.reg.List(ctx)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to fetch worker"})
```
mayuka-c, line 149: Same here having logs with err object printed would help

ai-services/internal/pkg/catalog/apiserver/handlers/worker_handler.go
```diff
@@ -80,38 +84,88 @@ func (h *WorkerHandler) CreateWorker(c *gin.Context) {
...
+
+		return
+	}
+
+	for _, w := range workers {
+		if w.ID != workerID {
+			continue
+		}
+
+		appIDMap, err := h.repo.GetApplicationIDsByWorkerIDs(ctx, []uuid.UUID{workerID})
```
mayuka-c: Maybe we can simplify it?

We get the workerID from request, query for that from DB in workers table to get worker details. Once we get worker ID, we can directly query for that in applications table and get the list of appIds. This would avoid loopings

yussufsh: I had seen this but thought it is ok 😄 
Will update.

mayuka-c (approved):
LGTM

PR #1392, opened 2026-09-03
helm support for remote deployment

yussufsh:
For remote deployment:
- Added COMMAND_TYPE_HELM_INSTALL/UNINSTALL to proto + wire payload types
- New helm package: HelmManager interface with LocalHelmManager (direct) and RemoteHelmManager (gRPC) implementations
- Single NewOpenShiftDeployer constructor resolves local vs remote HelmManager from runtime type
- Dispatcher handles the two new Helm command types on the worker side

For refactoring:
- Uninstall owns the exists-check internally — callers handle ErrReleaseNotFound instead of pre-checking
- create/delete/uninstall paths simplified to use the new shared operations

### DIFF
--- a/ai-services/assets/applications/rag-dev/openshift/values.yaml
+++ b/ai-services/assets/applications/rag-dev/openshift/values.yaml
@@ -79,7 +79,7 @@ ingest:
 
 opensearch:
   # @hidden
-  image: icr.io/ai-services-cicd/opensearch:3.7.0-ppc64le
+  image: icr.io/ppc64le-oss/opensearch-ppc64le:3.5.0
   # @description Sets the memory limit for the Opensearch service(Default: 8Gi). Override by passing a value with a unit suffix (e.g., Mi, Gi).
   memoryLimit: 8Gi
   # @description Sets the storage limit for the Opensearch service(Default: 10Gi). Override by passing a value with a unit suffix (e.g., Mi, Gi).
--- a/ai-services/assets/applications/rag-dev/openshift/values.yaml
+++ b/ai-services/assets/applications/rag-dev/openshift/values.yaml
@@ -79,7 +79,7 @@ ingest:
 
 opensearch:
   # @hidden
-  image: icr.io/ppc64le-oss/opensearch-ppc64le:3.5.0
+  image: icr.io/ai-services-cicd/opensearch:3.7.0-ppc64le
   # @description Sets the memory limit for the Opensearch service(Default: 8Gi). Override by passing a value with a unit suffix (e.g., Mi, Gi).
   memoryLimit: 8Gi
   # @description Sets the storage limit for the Opensearch service(Default: 10Gi). Override by passing a value with a unit suffix (e.g., Mi, Gi).
--- a/ai-services/assets/worker/openshift/templates/worker-tls-pvc.yaml
+++ b/ai-services/assets/worker/openshift/templates/worker-tls-pvc.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: worker-tls
+  annotations:
+    "helm.sh/resource-policy": keep
+  labels:
+    ai-services.io/application: {{ .Release.Name }}
+    ai-services.io/template: {{ .Chart.Name }}
+    ai-services.io/version: {{ default .Chart.AppVersion | quote }}
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: "{{ .Values.worker.tls.storage }}"
--- a/ai-services/internal/pkg/application/openshift/delete.go
+++ b/ai-services/internal/pkg/application/openshift/delete.go
@@ -2,6 +2,7 @@ package openshift
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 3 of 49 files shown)


ai-services/internal/pkg/worker/dispatch/dispatch.go
```diff
@@ -296,6 +298,34 @@ func handle(ctx context.Context, rt runtime.Runtime, pr *workercaddy.ProxyRouter
...
+		if req.TimeoutSec == 0 {
+			timeout = 10 * time.Minute // worker default
+		}
+
+		chart, err := helmutil.UnmarshalChart(req.ChartFiles)
+		if err != nil {
+			return nil, fmt.Errorf("helm install: reconstruct chart for release %q: %w", req.Release, err)
+		}
+
+		return nil, helmutil.InstallOrUpgrade(ctx, req.Release, req.Namespace, chart, req.Values, req.TemplateID, timeout)
```
mayuka-c, line 352: I see this is doing both internally, will it be better to have 2 diff rpcs (Command Types) helm install and helm upgrade?

mayuka-c (commented):
Rest LGTM

ai-services/assets/worker/openshift/templates/worker-tls-pvc.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: worker-tls
+  annotations:
+    "helm.sh/resource-policy": keep
```
yussufsh, line 6: @adarshagrawal38 you will need to use this in skip-delete during `worker uninstall`.

ai-services/internal/pkg/worker/dispatch/dispatch.go
yussufsh: Should be fine for now, as it is following previous apporach.

mayuka-c (dismissed):
LGTM

ai-services/assets/applications/rag-dev/openshift/values.yaml
```diff
@@ -79,7 +79,7 @@ ingest:
 
 opensearch:
   # @hidden
-  image: icr.io/ai-services-cicd/opensearch:3.7.0-ppc64le
+  image: icr.io/ppc64le-oss/opensearch-ppc64le:3.5.0
```
mayuka-c: Can we avoid downgrading opensearch image everywhere?

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/openshift/deployer.go
```diff
@@ -168,35 +170,34 @@ func (d *OpenShiftDeployer) deployComponentsConcurrently(ctx context.Context, ns
...
+	logger.InfofCtx(ctx, "Waiting for InferenceService '%s' to become ready\n", isvcName)
 
-		waitCtx, cancel := context.WithTimeout(ctx, predictorWaitTimeout)
-		defer cancel()
+	waitCtx, cancel := context.WithTimeout(ctx, constants.PredictorWaitTimeout)
+	defer cancel()
 
-		if err := oc.WaitForInferenceServiceReady(waitCtx, isvcName, predictorPollInterval); err != nil {
-			return fmt.Errorf("InferenceService %q is not ready yet: %w", isvcName, err)
-		}
```
mayuka-c, line 196: I think there seems to be a issue here, if we have a service which doesnt have a component that doesn't use InferenceService, this causes it to wait and will fail eventually.

ai-services/internal/pkg/catalog/utils/common.go
```diff
@@ -141,18 +143,17 @@ func HelmUninstall(ctx context.Context, namespace, release string) error {
 		return fmt.Errorf("failed to create Helm client: %w", err)
 	}
 
-	exists, err := helmClient.IsReleaseExist(release)
-	if err != nil {
-		return fmt.Errorf("failed to check '%s' release existence: %w", release, err)
```
mayuka-c, line 146: Can we still keep this error wrapper msg?

mayuka-c (commented):
Rest LGTM

ai-services/internal/pkg/application/openshift/delete.go
```diff
@@ -47,13 +31,23 @@ func (o *OpenshiftApplication) Delete(ctx context.Context, opts types.DeleteOpti
...
 
 	s.Start(ctx)
 
-	// Perform helm uninstall
 	err = helmClient.Uninstall(app, &helm.UninstallOpts{Timeout: timeout})
 	if err != nil {
+		if errors.Is(err, driver.ErrReleaseNotFound) {
+			s.Stop("Application '" + app + "' does not exist — nothing to delete")
+
+			return nil
```
mayuka-c, line 48: Dont we need to return error here?

ai-services/internal/pkg/catalog/utils/common.go
yussufsh: This is moved to helm.go. The code is same.

ai-services/internal/pkg/application/openshift/delete.go
yussufsh: No right, the application is already deleted, means we are good.

ai-services/internal/pkg/catalog/apiserver/services/deployment/repository/openshift/deployer.go
yussufsh: Also, I see we are considering vector_store as an InferenceService 👎 
We should file a bug and raise a different patch for this.

mayuka-c: Yes not related to this PR, but just wanted to mention as I was seeing the code :)

mayuka-c (approved):
LGTM

PR #1394, opened 2026-09-04
fix: extraction service fixes

anshuman-agarwala:
Made `ExampleItem.text` optional
Fixed `SchemaRegisterRequest` Swagger example body
Added an option to provide `schema_name` instead of `schema_id` for job submission and sync extraction endpoints.

### DIFF
--- a/services/extract/api/v1/jobs.py
+++ b/services/extract/api/v1/jobs.py
@@ -70,6 +70,30 @@ logger = get_logger("jobs_router")
 
 
 # ---------------------------------------------------------------------------
+# Module-level schema resolution helpers
+# ---------------------------------------------------------------------------
+
+def _resolve_schema_id(schema_id: str):
+    """Return the schema row for *schema_id*, raising 404 if not found."""
+    row = db_repo.get_schema_by_id(schema_id)
+    if row is None:
+        msg = f"No schema with id {schema_id!r}."
+        logger.error(msg)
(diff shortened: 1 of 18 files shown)


manalilatkar:
@anshuman-agarwala could you please add a description for this PR?

dharaneeshvrd:
Overall changes lgtm

services/extract/api/v1/jobs.py
manalilatkar, line 464: In the case that only schema_name is passed, it will not be passed to the create_job and the job row will be created with a None for schema_id instead of schema_name. This line should be:
```
row = db_repo.create_job(
    job_id=job_id,
    schema_id=schema_row.schema_id,  
    schema_name=schema_name,
    job_name=job_name,
    submitted_at=datetime.now(timezone.utc),
    file_count=len(files),
)
```

manalilatkar: This line should be :
`"schema_id": schema_row.schema_id,`

So we use the schema_id from resolved schema_row when schema_name was passed.

manalilatkar (changes requested):
rest LGTM.

services/extract/api/v1/jobs.py
anshuman-agarwala: Done

anshuman-agarwala: Done

manalilatkar (approved):
LGTM

PR #1396, opened 2026-09-04
feat: auto-join catalog host as Local worker

Shubhamag12:
During catalog configure --runtime podman, the catalog machine automatically joins itself as the reserved "Local" worker. Caddy pod, podman-auth, and certificates are reused from the catalog deployment.

1. Gateway local-bypass: pre defined token + worker_name == "local" skips token validation
2. Join runs inside handlePostDeployment before next-steps; non-fatal on failure
3. --skip-local-worker flag to opt out

### DIFF
--- a/ai-services/cmd/ai-services/cmd/catalog/apiserver.go
+++ b/ai-services/cmd/ai-services/cmd/catalog/apiserver.go
@@ -149,7 +149,8 @@ func buildAPIServerOptions(ctx context.Context, pool *pgxpool.Pool, secretKey, a
 
 	tokenMgr := auth.NewTokenManager(secretKey, accessTTL, refreshTTL)
 	workerRepo := repository.NewWorkerRepository(pool)
-	workerReg := workerregistry.New(workerRepo)
+	localWorker := os.Getenv("LOCAL_WORKER") == "true"
+	workerReg := workerregistry.New(workerRepo, localWorker)
 
 	stopBackgroundServices, err := startBackgroundServices(ctx, appRepo, svcRepo, compRepo, svcDepRepo, connectorRepo, catalogProvider, workerReg, encryptionKey)
 	if err != nil {
--- a/ai-services/cmd/ai-services/cmd/catalog/apiserver.go
+++ b/ai-services/cmd/ai-services/cmd/catalog/apiserver.go
@@ -149,8 +149,7 @@ func buildAPIServerOptions(ctx context.Context, pool *pgxpool.Pool, secretKey, a
 
 	tokenMgr := auth.NewTokenManager(secretKey, accessTTL, refreshTTL)
 	workerRepo := repository.NewWorkerRepository(pool)
-	localWorker := os.Getenv("LOCAL_WORKER") == "true"
-	workerReg := workerregistry.New(workerRepo, localWorker)
+	workerReg := workerregistry.New(workerRepo)
 
 	stopBackgroundServices, err := startBackgroundServices(ctx, appRepo, svcRepo, compRepo, svcDepRepo, connectorRepo, catalogProvider, workerReg, encryptionKey)
 	if err != nil {
--- a/ai-services/cmd/ai-services/cmd/common/common.go
+++ b/ai-services/cmd/ai-services/cmd/common/common.go
@@ -3,6 +3,7 @@
 package common
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/spf13/cobra"
@@ -13,6 +14,7 @@ import (
 	"github.com/project-ai-services/ai-services/internal/pkg/runtime/types"
 	"github.com/project-ai-services/ai-services/internal/pkg/utils"
 	"github.com/project-ai-services/ai-services/internal/pkg/vars"
+	workerconstants "github.com/project-ai-services/ai-services/internal/pkg/worker/constants"
 )
 
 // InitAndValidateRuntimeFlag validates the runtime flag value, initialises
@@ -49,4 +51,25 @@ func validateRuntimeType(runtimeType types.RuntimeType) error {
 	default:
 		return fmt.Errorf("unsupported runtime type: %s", runtimeType)
 	}
+}
+
+// IsCatalogLocalWorker inspects the running catalog pod and returns true when
+// the LOCAL_WORKER environment variable is set to "true" inside it.
+func IsCatalogLocalWorker(ctx context.Context, rt runtime.Runtime) (bool, error) {
+	pod, err := rt.InspectPod(ctx, workerconstants.PodmanGatewayPodName)
+	if err != nil {
+		return false, fmt.Errorf("inspect catalog pod: %w", err)
+	}
+
+	for _, container := range pod.Containers {
+		cInfo, err := rt.InspectContainer(ctx, container.ID)
+		if err != nil {
+			continue
+		}
+		if cInfo.Env[workerconstants.LocalWorkerEnvVar] == "true" {
+			return true, nil
+		}
+	}
+
+	return false, nil
 }
--- a/ai-services/cmd/ai-services/cmd/common/common.go
+++ b/ai-services/cmd/ai-services/cmd/common/common.go
@@ -64,7 +64,7 @@ func IsCatalogLocalWorker(ctx context.Context, rt runtime.Runtime) (bool, error)
 	for _, container := range pod.Containers {
 		cInfo, err := rt.InspectContainer(ctx, container.ID)
 		if err != nil {
-			continue
+			return false, fmt.Errorf("inspect container %s failed: %w", container.ID, err)
 		}
 		if cInfo.Env[workerconstants.LocalWorkerEnvVar] == "true" {
 			return true, nil
--- a/ai-services/cmd/ai-services/cmd/worker/join.go
+++ b/ai-services/cmd/ai-services/cmd/worker/join.go
@@ -12,7 +12,6 @@ import (
 	cmdcommon "github.com/project-ai-services/ai-services/cmd/ai-services/cmd/common"
 	catalogUtils "github.com/project-ai-services/ai-services/internal/pkg/catalog/utils"
 	"github.com/project-ai-services/ai-services/internal/pkg/constants"
-	"github.com/project-ai-services/ai-services/internal/pkg/logger"
 	"github.com/project-ai-services/ai-services/internal/pkg/runtime"
 	"github.com/project-ai-services/ai-services/internal/pkg/runtime/types"
 	"github.com/project-ai-services/ai-services/internal/pkg/utils"
@@ -102,9 +101,11 @@ func joinPreRunE(cmd *cobra.Command, _ []string) error {
 	if err != nil {
 		return fmt.Errorf("worker join: init runtime: %w", err)
 	}
-	if isLocalWorker, err := cmdcommon.IsCatalogLocalWorker(ctx, rt); err != nil {
-		logger.WarningfCtx(ctx, "worker join: could not determine LOCAL_WORKER from catalog pod: %v\n", err)
-	} else if isLocalWorker {
+	isLocalWorker, err := cmdcommon.IsCatalogLocalWorker(ctx, rt)
+	if err != nil {
+		return fmt.Errorf("could not determine LOCAL_WORKER from catalog pod: %w", err)
+	}
+	if isLocalWorker {
 		return fmt.Errorf("the worker is already co-located with the control plane and cannot be joined independently")
 	}
 
--- a/ai-services/cmd/ai-services/cmd/worker/join.go
+++ b/ai-services/cmd/ai-services/cmd/worker/join.go
@@ -1,6 +1,7 @@
 package worker
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"path/filepath"
@@ -11,9 +12,11 @@ import (
 	cmdcommon "github.com/project-ai-services/ai-services/cmd/ai-services/cmd/common"
 	catalogUtils "github.com/project-ai-services/ai-services/internal/pkg/catalog/utils"
 	"github.com/project-ai-services/ai-services/internal/pkg/constants"
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
 	"github.com/project-ai-services/ai-services/internal/pkg/runtime"
 	"github.com/project-ai-services/ai-services/internal/pkg/runtime/types"
 	"github.com/project-ai-services/ai-services/internal/pkg/utils"
+	"github.com/project-ai-services/ai-services/internal/pkg/vars"
 	workercaddy "github.com/project-ai-services/ai-services/internal/pkg/worker/caddy"
 	workerconstants "github.com/project-ai-services/ai-services/internal/pkg/worker/constants"
 	workeropenshift "github.com/project-ai-services/ai-services/internal/pkg/worker/deploy/openshift"
@@ -91,6 +94,18 @@ func joinPreRunE(cmd *cobra.Command, _ []string) error {
 		if err := validateAddHost(h); err != nil {
 			return err
 		}
+	}
+
+	ctx := context.Background()
+	rtType := vars.RuntimeFactory.GetRuntimeType()
+	rt, err := runtime.CreateRuntime(rtType, "")
+	if err != nil {
+		return fmt.Errorf("worker join: init runtime: %w", err)
+	}
+	if isLocalWorker, err := cmdcommon.IsCatalogLocalWorker(ctx, rt); err != nil {
+		logger.WarningfCtx(ctx, "worker join: could not determine LOCAL_WORKER from catalog pod: %v\n", err)
+	} else if isLocalWorker {
+		return fmt.Errorf("the worker is already co-located with the control plane and cannot be joined independently")
 	}
 
 	return nil
--- a/ai-services/internal/pkg/catalog/apiserver/services/deployment/planner.go
+++ b/ai-services/internal/pkg/catalog/apiserver/services/deployment/planner.go
@@ -335,6 +335,10 @@ func (p *DeploymentPlanner) getRequiredSpyreCardsForComponent(ctx context.Contex
 // WorkerDBID returns the database UUID for the named worker by consulting the
 // in-memory registry.
 func (p *DeploymentPlanner) WorkerDBID(workerName string) (uuid.UUID, bool) {
+	if p.workerRegistry == nil {
+		return uuid.Nil, false
+	}
+
 	return p.workerRegistry.WorkerID(workerName)
 }
 
--- a/ai-services/internal/pkg/catalog/apiserver/services/deployment/planner.go
+++ b/ai-services/internal/pkg/catalog/apiserver/services/deployment/planner.go
@@ -333,17 +333,8 @@ func (p *DeploymentPlanner) getRequiredSpyreCardsForComponent(ctx context.Contex
 }
 
 // WorkerDBID returns the database UUID for the named worker by consulting the
-// in-memory registry. Returns (uuid.Nil, false) when the worker is not
-// connected or the registry is nil (local-only server).
+// in-memory registry.
 func (p *DeploymentPlanner) WorkerDBID(workerName string) (uuid.UUID, bool) {
-	if p.workerRegistry == nil {
-		return uuid.Nil, false
-	}
-
-	if isLocalWorkerName(workerName) {
-		workerName = workerconstants.LocalWorkerName
-	}
-
 	return p.workerRegistry.WorkerID(workerName)
 }
 
@@ -394,7 +385,7 @@ func (p *DeploymentPlanner) ResolveRuntimeType(ctx context.Context, workerName s
 }
 
 func isLocalWorkerName(workerName string) bool {
-	return workerName == "" || strings.EqualFold(workerName, workerconstants.LocalWorkerName)
+	return strings.EqualFold(workerName, workerconstants.LocalWorkerName)
 }
 
 // Made with Bob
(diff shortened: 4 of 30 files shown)


Shubhamag12:
```
[root@lpar-2 ai-services]# podman pod ps
POD ID        NAME                  STATUS      CREATED             INFRA ID      # OF CONTAINERS
84b83384bfca  ai-services--worker   Running     48 seconds ago      73e045f2ffb6  2
ac72c0b2e40b  ai-services--catalog  Running     About a minute ago  6be8d6bfbe4a  3
95ceb02a9c6e  ai-services--db       Running     2 minutes ago       97f83d925628  2
489b52d70aa6  ai-services--caddy    Running     2 minutes ago       f41195827535  2
```
```
[root@lpar-2 ai-services]# podman pod logs ai-services--worker
3d7edfc6b497 {"ts":1788783208777.6624,"caller":"join/join.go:102","msg":"Registering worker with catalog control plane...","v":0,"level":"INFO","caller_fullpath":"/go/src/github.com/project-ai-services/ai-services/internal/pkg/worker/join/join.go:102"}
3d7edfc6b497 {"ts":1788783208778.2485,"caller":"join/join.go:118","msg":"worker join: ca.crt not present, bootstrap connection will use InsecureSkipVerify","v":0,"level":"WARNING","caller_fullpath":"/go/src/github.com/project-ai-services/ai-services/internal/pkg/worker/join/join.go:118"}
3d7edfc6b497 {"ts":1788783208778.3345,"caller":"join/join.go:128","msg":"worker join: registering with catalog control plane...","v":0,"level":"INFO","caller_fullpath":"/go/src/github.com/project-ai-services/ai-services/internal/pkg/worker/join/join.go:128"}
3d7edfc6b497 {"ts":1788783208784.3755,"caller":"join/join.go:146","msg":"worker join: mTLS credentials written to /var/lib/ai-services/worker-tls","v":0,"level":"INFO","caller_fullpath":"/go/src/github.com/project-ai-services/ai-services/internal/pkg/worker/join/join.go:146"}
3d7edfc6b497 {"ts":1788783208784.609,"caller":"join/join.go:94","msg":"Worker \"Local\" registered with control plane.\n","v":0,"level":"INFO","caller_fullpath":"/go/src/github.com/project-ai-services/ai-services/internal/pkg/worker/join/join.go:94"}
3d7edfc6b497 {"ts":1788783208784.8286,"caller":"join/join.go:177","msg":"worker join: connecting as \"Local\" to 10.48.64.154:9090","v":0,"level":"INFO","caller_fullpath":"/go/src/github.com/project-ai-services/ai-services/internal/pkg/worker/join/join.go:177"}
3d7edfc6b497 {"ts":1788783208784.8403,"caller":"join/join.go:194","msg":"Opening CommandStream for worker \"Local\"...\n","v":0,"level":"INFO","caller_fullpath":"/go/src/github.com/project-ai-services/ai-services/internal/pkg/worker/join/join.go:194"}
3d7edfc6b497 {"ts":1788783208787.3948,"caller":"join/join.go:236","msg":"CommandStream open for worker \"Local\" \n","v":0,"level":"INFO","caller_fullpath":"/go/src/github.com/project-ai-services/ai-services/internal/pkg/worker/join/join.go:236"}
```

ai-services/internal/pkg/catalog/cli/configure/podman/local_worker.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if err := utils.CreateDir(filepath.Join(aiServicesDir, "models")); err != nil {
+		return fmt.Errorf("local worker join: failed to create model directory: %w", err)
+	}
+
+	hostIP, err := utils.GetHostIP()
+	if err != nil {
+		return fmt.Errorf("local worker join: resolve host IP: %w", err)
+	}
+
```
shwetha-s-poojary, line 29: We should handle the case where `utils.GetHostIP()` returns "", nil

ai-services/internal/pkg/worker/gateway/gateway.go
```diff
@@ -130,16 +130,40 @@ func (g *Gateway) runSweeper(ctx context.Context) {
...
+
+	if req.GetPreSharedToken() == "" && req.GetWorkerName() == workerconstants.LocalWorkerName {
+		// Local-worker bypass: catalog and worker share the same host.
+		// Self-issue the pre-registration so we can proceed without a token.
+		logger.DebuglnCtx(ctx, "WorkerGateway: local worker detected — bypassing token validation")
+
+		if _, err := g.registry.Preregister(ctx, workerconstants.LocalWorkerName); err != nil {
+			logger.WarningfCtx(ctx, "WorkerGateway: local preregister: %v (continuing)", err)
+		}
+
```
shwetha-s-poojary: Just wondering if it's ok to just Warn here.Or should we error out since it's DB failure.

ai-services/internal/pkg/catalog/cli/configure/podman/local_worker.go
```diff
@@ -0,0 +1,55 @@
...
+	gatewayAddr := fmt.Sprintf("%s:%d", hostIP, opts.WorkerGatewayPort)
+
+	workerOpts := workertypes.PodmanWorkerOptions{
+		WorkerConnectionOptions: workertypes.WorkerConnectionOptions{
+			GatewayAddr: gatewayAddr,
+			Token:       "",
+		},
```
shwetha-s-poojary, line 36: Shouldn't we be setting the RequestedWorkerName here ?Is it being set somewhere else?

shwetha-s-poojary:
@Shubhamag12 when grpcstream command is run manually from a different VM , not the gateway(or the catalog machine), it would set the worker name to Local and bypass token check right.Doesn't this mean anyone can join without token ?Isn't this supposed to just allow token bypass on catalog machine and not other VMs ?
Is my understanding right or did I miss anything here?

ai-services/cmd/ai-services/cmd/worker/join.go
```diff
@@ -238,11 +238,19 @@ func grpcStreamRunE(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("unsupported runtime type: %s", runtimeType)
 	}
 
+	// For the local worker (empty token) set the reserved worker name so
+	// StartGrpcStream can apply the local-bypass registration path.
+	requestedWorkerName := ""
+	if token == "" {
+		requestedWorkerName = workerconstants.LocalWorkerName
+	}
```
yussufsh: This should not be in join cmd... no one else can use this Local mechanism for join.

ai-services/internal/pkg/catalog/cli/configure/podman/local_worker.go
```diff
@@ -0,0 +1,55 @@
...
+	logger.InfolnCtx(ctx, "Joining local machine as worker...")
+
+	aiServicesDir, err := utils.ValidateBaseDir(opts.BaseDir)
+	if err != nil {
+		return fmt.Errorf("local worker join: invalid base directory %q: %w", opts.BaseDir, err)
+	}
+
+	if err := utils.CreateDir(filepath.Join(aiServicesDir, "models")); err != nil {
+		return fmt.Errorf("local worker join: failed to create model directory: %w", err)
+	}
```
yussufsh: Logic is already present in `catalog configure` no need to repeat. Also use the processed basedir.

ai-services/internal/pkg/catalog/cli/configure/podman/local_worker.go
```diff
@@ -0,0 +1,55 @@
...
+	workertypes "github.com/project-ai-services/ai-services/internal/pkg/worker/types"
+)
+
+// JoinAsLocalWorker joins the catalog machine itself as the reserved "Local"
+// worker. It can be skipped with --skip-local-worker.
+//
+// The gateway address uses the host's outbound IP. An empty bootstrap token is passed; the gateway applies the
+// local-bypass registration path when it sees an empty token together with
+// worker_name == LocalWorkerName.
+func JoinAsLocalWorker(ctx context.Context, opts catalogUtils.PodmanConfigureOptions) error {
```
mayuka-c: We would need to take care of openshift too with self join

Shubhamag12:
@shwetha-s-poojary @yussufsh @mayuka-c 
implemented with new approach as discussed. PTAL

mkumatag:
> * oin runs inside handlePostDeployment before next-steps; non-fatal on failure
> * --skip-local-worker flag to opt out

there shouldn't be any difference between registering a local node vs remote, only difference should be running control plane + worker on a single machine with sharing few components like caddy etc..

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -125,6 +124,14 @@ func handlePostDeployment(ctx context.Context, caddyCtx *caddy.Context, deployCt
 		return fmt.Errorf("route registration failed: %w", err)
 	}
 
+	if !opts.SkipLocalWorker {
+		if err := JoinAsLocalWorker(ctx, deployCtx.Runtime, opts); err != nil {
+			// Non-fatal: log and continue so the catalog itself is not
+			// considered failed if the local worker join fails.
+			logger.Warningf("local worker join failed: %v\n", err)
```
yussufsh: This should fail. Without the worker catalog does not make sense, otherwise users will have to uninstall and deploy again.
If we fail users will just retry the same command, and idempotency will kick in and just refresh the worker pod.
cc @adarshagrawal38 had implemented a similar approach in the join cmd.

ai-services/cmd/ai-services/cmd/catalog/apiserver.go
```diff
@@ -149,7 +149,8 @@ func buildAPIServerOptions(ctx context.Context, pool *pgxpool.Pool, secretKey, a
 
 	tokenMgr := auth.NewTokenManager(secretKey, accessTTL, refreshTTL)
 	workerRepo := repository.NewWorkerRepository(pool)
-	workerReg := workerregistry.New(workerRepo)
+	localWorker := os.Getenv("LOCAL_WORKER") == "true"
```
yussufsh: Use existing util func.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -134,97 +141,106 @@ func handlePostDeployment(ctx context.Context, caddyCtx *caddy.Context, deployCt
...
+		return nil, err
 	}
 
-	// Base64 encode the auth file content for Kubernetes secret
-	authFileBase64 := base64.StdEncoding.EncodeToString(authFileContent)
-
-	// Determine the podman URI
-	// Strip unix:// prefix from podmanURI for hostPath volume mount
-	// The CONTAINER_HOST env var needs the full URI, but the hostPath needs just the file path
 	podmanURI, err := utils.ResolvePodmanURI()
```
yussufsh, line 169: Please keep the previous comments. It is really helpful.

ai-services/internal/pkg/catalog/utils/common.go
```diff
@@ -33,7 +33,9 @@ type PodmanConfigureOptions struct {
 	SSLCertPath       string // Path to user-provided SSL certificate
 	SSLKeyPath        string // Path to user-provided SSL private key
 	HttpsPort         int
-	WorkerGatewayPort int // gRPC worker gateway port; always active, default 9090
+	WorkerGatewayPort int  // gRPC worker gateway port; always active, default 9090
+	SkipLocalWorker   bool // When true, skip joining this machine as the Local worker
+	LocalWorker       bool // Reflects LOCAL_WORKER env var from the running catalog-backend container
```
yussufsh: Why do we need both? `SkipLocalWorker` is enough. Either 1 will be true.

ai-services/internal/pkg/worker/registry/registry.go
```diff
@@ -95,22 +95,31 @@ func (w *WorkerEntry) deliverResult(res *workerpb.CommandResult) {
...
+// Set localWorker to true to skip the DB upsert in Preregister (local worker mode).
+func New(repo repository.WorkerRepository, localWorker bool) *Registry {
 	return &Registry{
-		workers:    make(map[string]*WorkerEntry),
-		repo:       repo,
-		tokenStore: NewTokenStore(),
+		workers:     make(map[string]*WorkerEntry),
+		repo:        repo,
+		tokenStore:  NewTokenStore(),
+		localWorker: localWorker,
```
yussufsh: No need to store this extra variable. Just use the env variable where needed.

ai-services/internal/pkg/worker/registry/registry.go
```diff
@@ -228,7 +237,15 @@ func (r *Registry) Restore(ctx context.Context, workerName string) (*WorkerEntry
 // If the worker is currently active in the in-memory map (i.e. its stream is still
 // open), it is evicted first so that the stale connection can no longer update the
 // heartbeat on the now-pending row.
+//
+// When the LOCAL_WORKER environment variable is "true" the catalog-backend is
+// running as the local worker itself. In that case the DB upsert is skipped —
+// only a token is issued so the worker pod can connect via gRPC.
 func (r *Registry) Preregister(ctx context.Context, workerName string) (string, error) {
+	if r.localWorker {
```
yussufsh: DB entry is needed right? How grpc stream will connect if no "Local" worker in DB? Did I miss something?

yussufsh (changes requested):

yussufsh:
> > * oin runs inside handlePostDeployment before next-steps; non-fatal on failure
> > * --skip-local-worker flag to opt out
> 
> there shouldn't be any difference between registering a local node vs remote, only difference should be running control plane + worker on a single machine with sharing few components like caddy etc..

Registering a worker from cli will need to login as admin and  we need to calculate the domain if custom is provided. Instead of this we could think of by default registering the worker at gateway if a flag is set. And then normal join will work with no difference.

ai-services/internal/pkg/worker/registry/registry.go
Shubhamag12: we cannot talk to DB from the host. That was the issue which made us directly register with pre-defined token and not creating one.

ai-services/internal/pkg/catalog/utils/common.go
Shubhamag12: one is flag, and the other one is value we get from env variable

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
Shubhamag12: > If we fail users will just retry the same command, and idempotency will kick in and just refresh the worker pod.

We can fail instead of warn.
But this also behaves the same way. At this step if worker pod is not deployed and failed early, users can re run configure command and then worker will get deployed.

ai-services/internal/pkg/worker/registry/registry.go
Shubhamag12: this makes it easy to check for local worker by r.localWoker

adarshagrawal38:
What happens when catalog gets uninstalled, will local worker also goes off?

Shubhamag12:
> What happens when catalog gets uninstalled, will local worker also goes off?

yes

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -110,7 +109,7 @@ func executeCatalogDeployment(ctx context.Context, deployCtx *deploy.DeployConte
 }
 
 // handlePostDeployment handles route registration and next steps display after catalog deployment.
-func handlePostDeployment(ctx context.Context, caddyCtx *caddy.Context, deployCtx *deploy.DeployContext) error {
+func handlePostDeployment(ctx context.Context, caddyCtx *caddy.Context, deployCtx *deploy.DeployContext, opts catalogUtils.PodmanConfigureOptions) error {
```
adarshagrawal38, line 106: Why are we passing full `PodmanConfigureOptions` here?

only passing `SkipLocalWorker` will work for us.

ai-services/internal/pkg/worker/registry/registry.go
yussufsh: Ok so the register will do `Upsert`. I missed that.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
Shubhamag12: no, it wont work. opts are required in JoinLocalWorker function.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -134,97 +141,106 @@ func handlePostDeployment(ctx context.Context, caddyCtx *caddy.Context, deployCt
...
 	argParams[configure.ArgParamCaddyFileContent] = utils.IndentString(caddyFileContent, caddyFileIndent)
 	argParams[configure.ArgParamSSLCertFileContent] = utils.IndentString(sslCertContent, certContentIndent)
 	argParams[configure.ArgParamSSLKeyFileContent] = utils.IndentString(sslKeyContent, certContentIndent)
 
 	return argParams, nil
 }
 
+// readAuthFileBase64 reads the podman auth file and returns it base64-encoded.
+// If the file does not exist, an encoded empty JSON object is returned.
+func readAuthFileBase64() (string, error) {
```
adarshagrawal38: Duplicate method, please move this to util and update [here](https://github.com/IBM/project-ai-services/blob/6d06c4ab92c9d659c2c242425c886d1a81a882cd/ai-services/internal/pkg/worker/deploy/podman/deploy.go#L237). as well

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
adarshagrawal38: ooh okay got it.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
Shubhamag12: adding a fail here and will work on idempotency once @adarshagrawal38 PR is merged

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
Shubhamag12: done

ai-services/internal/pkg/catalog/apiserver/services/deployment/planner.go
```diff
@@ -336,12 +336,12 @@ func (p *DeploymentPlanner) getRequiredSpyreCardsForComponent(ctx context.Contex
...
 // connected or the registry is nil (local-only server).
 func (p *DeploymentPlanner) WorkerDBID(workerName string) (uuid.UUID, bool) {
-	if isLocalWorkerName(workerName) {
+	if p.workerRegistry == nil {
 		return uuid.Nil, false
 	}
 
-	if p.workerRegistry == nil {
-		return uuid.Nil, false
+	if isLocalWorkerName(workerName) {
```
yussufsh: Not needed to check since workername is already Local.

ai-services/internal/pkg/utils/util.go
yussufsh:
```suggestion
func getAuthFilePath() (string, error) {
```

ai-services/internal/pkg/worker/uninstall/podman/uninstall.go
```diff
@@ -138,6 +144,27 @@ func extractConfigFromEnv(env map[string]string, config *WorkerCaddyConfig) {
 	}
 }
 
+// isCatalogLocalWorker inspects the running catalog pod and returns true when
+// the LOCAL_WORKER environment variable is set to "true" inside it.
+func isCatalogLocalWorker(ctx context.Context, rt runtime.Runtime) (bool, error) {
```
yussufsh: This func should go into `worker.go` parent cmd and should work for both podman/openshift and other worker sub-commands.

yussufsh (changes requested):

ai-services/internal/pkg/catalog/apiserver/services/deployment/planner.go
Shubhamag12: changed

ai-services/internal/pkg/worker/uninstall/podman/uninstall.go
Shubhamag12: moved

ai-services/cmd/ai-services/cmd/common/common.go
```diff
@@ -50,3 +52,24 @@ func validateRuntimeType(runtimeType types.RuntimeType) error {
...
+func IsCatalogLocalWorker(ctx context.Context, rt runtime.Runtime) (bool, error) {
+	pod, err := rt.InspectPod(ctx, workerconstants.PodmanGatewayPodName)
+	if err != nil {
+		return false, fmt.Errorf("inspect catalog pod: %w", err)
+	}
+
+	for _, container := range pod.Containers {
+		cInfo, err := rt.InspectContainer(ctx, container.ID)
+		if err != nil {
+			continue
```
yussufsh: Should return err and not assume if local worker is false if this one was a backend container.

ai-services/internal/pkg/catalog/apiserver/services/deployment/planner.go
```diff
@@ -333,17 +333,8 @@ func (p *DeploymentPlanner) getRequiredSpyreCardsForComponent(ctx context.Contex
...
 // WorkerDBID returns the database UUID for the named worker by consulting the
-// in-memory registry. Returns (uuid.Nil, false) when the worker is not
-// connected or the registry is nil (local-only server).
+// in-memory registry.
 func (p *DeploymentPlanner) WorkerDBID(workerName string) (uuid.UUID, bool) {
-	if isLocalWorkerName(workerName) {
-		return uuid.Nil, false
-	}
-
-	if p.workerRegistry == nil {
```
yussufsh, line 343: Let's keep the nil guard.

ai-services/cmd/ai-services/cmd/common/common.go
Shubhamag12: done

ai-services/internal/pkg/catalog/apiserver/services/deployment/planner.go
Shubhamag12: done

yussufsh (dismissed):

yussufsh (approved):

PR #1397, opened 2026-09-04
Support Openshift as controlplane with remote deployments

mayuka-c:
- Include gateway/tls assets required to support worker gateway.
- mTLS integration and support with openshift as control plane.
- Application deployment, sync service, deletion support for remote deployments.

### DIFF
--- a/ai-services/assets/catalog/openshift/templates/catalog-worker-gateway-route.yaml
+++ b/ai-services/assets/catalog/openshift/templates/catalog-worker-gateway-route.yaml
@@ -7,6 +7,7 @@ metadata:
     ai-services.io/template: {{ .Chart.Name }}
     ai-services.io/version: {{ default .Chart.AppVersion | quote }}
   annotations:
+    ingress.operator.openshift.io/default-to-http2: "true"
     haproxy.router.openshift.io/timeout: 1d 
 spec:
   port:
@@ -14,5 +15,3 @@ spec:
   to:
     kind: Service
     name: "catalog-api"
-  tls:
-    termination: passthrough
--- a/ai-services/assets/catalog/openshift/templates/catalog-worker-gateway-route.yaml
+++ b/ai-services/assets/catalog/openshift/templates/catalog-worker-gateway-route.yaml
@@ -1,0 +1,17 @@
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: "catalog-worker-gateway"
+  labels:
+    ai-services.io/application: {{ .Release.Name }}
+    ai-services.io/template: {{ .Chart.Name }}
+    ai-services.io/version: {{ default .Chart.AppVersion | quote }}
+spec:
+  port:
+    targetPort: worker-gateway
+  to:
+    kind: Service
+    name: "catalog-api"
+  tls:
+    termination: passthrough
+    insecureEdgeTerminationPolicy: None
--- a/ai-services/assets/catalog/openshift/templates/catalog-worker-gateway-route.yaml
+++ b/ai-services/assets/catalog/openshift/templates/catalog-worker-gateway-route.yaml
@@ -6,12 +6,11 @@ metadata:
     ai-services.io/application: {{ .Release.Name }}
     ai-services.io/template: {{ .Chart.Name }}
     ai-services.io/version: {{ default .Chart.AppVersion | quote }}
+  annotations:
+    haproxy.router.openshift.io/h2c: "true"
 spec:
   port:
     targetPort: worker-gateway
   to:
     kind: Service
     name: "catalog-api"
-  tls:
-    termination: passthrough
-    insecureEdgeTerminationPolicy: None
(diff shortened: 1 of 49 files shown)


ai-services/cmd/ai-services/cmd/catalog/configure.go
```diff
@@ -342,14 +343,14 @@ func buildFlagValidator() *flagvalidator.FlagValidator {
 	rt := vars.RuntimeFactory.GetRuntimeType()
 	builder := flagvalidator.NewFlagValidatorBuilder(rt)
 
-	// Common flags, valid for every runtime.
+	// Common flags, valid for all runtimes.
 	builder.AddCommonFlag("reset-password", nil)
+	builder.AddCommonFlag("workergateway-port", nil)
```
yussufsh: Mark as supported by podman only.

ai-services/assets/catalog/openshift/values.yaml
```diff
@@ -9,11 +9,13 @@ ui:
...
 
 backend:
-  image: icr.io/ai-services-cicd/ai-services:v0.0.326
+  image: icr.io/ai-services-cicd/ai-services:v0.0.327
   runtime: "openshift"
   adminPasswordHash: ""
   # @generate:password length=32, special=false
   dbEncryptionKey: ""
+  # workerGatewayPort: port for the gRPC worker gateway. Always active; default is 9090.
+  workerGatewayPort: "9090"
```
hasan4791, line 18: We can hardcode the port to 9090 as it would be exposed via service & routes. It will be local to container. We dont need to customize it.

ai-services/assets/catalog/openshift/templates/catalog-worker-gateway-route.yaml
```diff
@@ -0,0 +1,18 @@
...
+    ai-services.io/template: {{ .Chart.Name }}
+    ai-services.io/version: {{ default .Chart.AppVersion | quote }}
+  annotations:
+    haproxy.router.openshift.io/timeout: 1d
+spec:
+  port:
+    targetPort: worker-gateway
+  to:
+    kind: Service
+    name: "catalog-api"
```
hasan4791, line 16: instead of catalog-api can we use "gateway". Because this pod exposes 2 services, api & gateway.
nvm. got it.

ai-services/cmd/ai-services/cmd/worker/join.go
```diff
@@ -31,6 +36,7 @@ var (
 	domainName  string
 	sslCertPath string
 	sslKeyPath  string
+	addHosts    []string
```
hasan4791, line 39: Do we need this?
What if the ip changes for the hosts, like worker is replaced or something similar to that?

mayuka-c: Yes this would be required especially if gateway is openshift and worker needs to connect using Routes. Then within the pod we need to resolve the host, so for that we need it.

Its optional tho, if user has dns server configured then automatically worker should be able to resolve it, then they dont need to set it.

This doesnt affect if worker is replaced, but affects only if domain of openshift cluster changes (gateway side), then they would have to perform join again with updated addHosts

ai-services/internal/pkg/worker/gateway/pki.go
```diff
@@ -117,11 +119,22 @@ func generateServerCert(caCert *x509.Certificate, caKey *ecdsa.PrivateKey) (*ecd
...
+//
+// For OpenShift the server cert's DNS SAN is populated by fetching the live
+// passthrough route host via GatewayRouteHost rather than using a hardcoded
+// hostname. For all other runtimes the static internal name is used.
+func generateAndPersistPKI(ctx context.Context, pkiDir string, runtimeType types.RuntimeType) (pkiResult, error) {
 	empty := pkiResult{}
 
+	serverName := workerconstants.GatewayServerName
+	if runtimeType == types.RuntimeTypeOpenShift {
+		host, err := GatewayRouteHost(ctx)
```
hasan4791, line 131: We are relying on container restart here for retries. if the route creation takes time, container would be restarted and it will be delayed exponentially right.

ai-services/assets/catalog/openshift/values.yaml
mayuka-c: Yeah still we dont allow it to be overrided, it is just a reference added in values.yaml to be able to reference it in deployment file. I feel we can keep it still.

hasan4791: oh ok, since the comment mentioned default is 9090, thought it would be customizable.

ai-services/internal/pkg/worker/gateway/pki.go
mayuka-c: Yes but I dont see a reason of why route will not get created. We do helm install everything is deployed simultaneously and creating route resource is instant infact pod bring up takes time as it needs to allocate pvc, run db migration script (init container), spawn a main container and then start it

ai-services/cmd/ai-services/cmd/worker/join.go
hasan4791: I meant the openshift node of that ip goes down, it would tear down the connection too.

ai-services/internal/pkg/worker/gateway/pki.go
hasan4791: Route is a different object, so if that controller had issues, it would delay too.

mayuka-c: Yes in that case should be fine to error and let it reconnect, if Route itself is not reachable/created then there is no point of doing join

yussufsh (approved):
lgtm
Reviewed offline and changes were implemented as suggested.

PR #1402, opened 2026-09-04
Rename VLLM_SPYRE_WARMUP envs to SENDNN_INFERENCE_WARMUP - pt2 for rag podman

dharaneeshvrd:


yussufsh:
This is for dev purpose only right? Customers won't install this unless using hidden legacy flag.

dharaneeshvrd:
Agree, but this is the missing piece that we would ship in the bin. Thought of adding this to be on the safer side.

yussufsh (approved):

PR #1405, opened 2026-09-05
support for worker parameter in resources api

yussufsh:


mayuka-c (dismissed):
LGTM

mayuka-c (approved):

ai-services/internal/pkg/catalog/apiserver/handlers/resources.go
```diff
@@ -28,49 +36,79 @@ type ResourcesResponse struct {
 // GetResources godoc
 //
 //	@Summary		Get system resources
-//	@Description	Retrieves system resource information including CPU, memory, and accelerator availability
+//	@Description	Retrieves system resource information including CPU, memory, and accelerator availability.
+//	@Description	When the optional `worker` query parameter is provided, the resources are fetched from
+//	@Description	that remote worker node instead of the local runtime.
```
mkumatag, line 41: But there is nothing called local anymore right? what do you mean local runtime here?

yussufsh: The auto-join PR is still in progress. Once that is done for both podman/openshift (local), we will migrate/remove the local runtime completely.

PR #1408, opened 2026-09-07
Support local service endpoint and podname for remote deployments

mayuka-c:
- This is to allow internal pod-pod communication in Podman and Openshift to be able to establish grpc stream with mtls enabled between gateway and worker pods.

### DIFF
--- a/ai-services/internal/pkg/worker/constants/constants.go
+++ b/ai-services/internal/pkg/worker/constants/constants.go
@@ -52,6 +52,14 @@ const (
 	ArgParamWorkerAuthFile    = "worker.podman.authFileContent"
 
 	// GatewayServerName is the default DNS SAN embedded in the auto-generated
-	// gateway server certificate for the local/podman flow.
+	// gateway server certificate for the local flow.
 	GatewayServerName = "gateway.ai-services.internal"
+
+	// GatewayPodName is the Podman catalog pod DNS name embedded in the
+	// auto-generated gateway server certificate.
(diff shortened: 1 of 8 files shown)


ai-services/internal/pkg/worker/constants/constants.go
```diff
@@ -52,6 +52,14 @@ const (
...
+	// gateway server certificate for the local flow.
 	GatewayServerName = "gateway.ai-services.internal"
+
+	// GatewayPodName is the Podman catalog pod DNS name embedded in the
+	// auto-generated gateway server certificate.
+	GatewayPodName = "ai-services--catalog"
+
+	// GatewayServiceEndpoint is the OpenShift service DNS name embedded in the
+	// auto-generated gateway server certificate.
+	GatewayServiceEndpoint = "catalog-api.ai-services.svc.cluster.local"
```
yussufsh: Can we prefix the variable names with their runtime?
`GatewayServerName` seems to be in use only for podman, but does not mention in the comments.

yussufsh (changes requested):
A minor comment, else lgtm

ai-services/internal/pkg/worker/constants/constants.go
mayuka-c: Addressed.

Shubhamag12 (dismissed):
lgtm

yussufsh (dismissed):

yussufsh (approved):

PR #1409, opened 2026-09-07
feat(deployment-details): add application data sources table

Niranjan-J-Joshi:
## Summary

Adds a **Data sources** tab to DeploymentDetails for Digital Assistants, backed by the application datasources API.

## Changes

- Added `ApplicationDatasourcesTable` with:
  - Server-side pagination
  - Search across datasource fields
  - Column visibility controls
  - CSV export
  - Auto-refresh
  - Loading and empty states
- Added paginated datasource API fetching and full-export support.
- Moved `files` and `last_sync` display formatting into the API transform layer.
- Added datasource sync status types and API response types.
- Added the `GET_APPLICATION_DATASOURCES` endpoint.
- Added the **Data sources** side-nav section for Digital Assistants.
- Extended shared `MessageCell` and `TableEmptyStates` 

## Dependency

Depends on **#1344**, which updates the shared `STATUS_CONFIG` and `ConnectorStatus` casing to match the backend values.

## Follow-up work

- Connect datasource flow
- Delete datasource flow

These will be handled in follow-up PRs.


ryanmarc (approved):

ui/catalog/src/components/Table/components/CellRenderers.tsx
```diff
@@ -110,20 +138,39 @@ export const StatusCell = ({ value }: SharedCellRendererProps) => {
...
+   * Statuses that warrant the error icon instead of the in-progress icon.
+   * Defaults to ["Error", "Offline"] to preserve existing behaviour for the deployments table.
+   */
+  errorStatuses?: string[];
+}
+
+export const MessageCell = ({
+  value,
+  rowData,
+  hideStatuses = ["Running", "up to date"],
```
ryanmarc: Heads up: this default has drifted from main while the PR has been open. #1344 landed and main's MessageCell now reads:
```
if (status === "Running" || status === "connected" || !message) {
```
hideStatuses here has "Running" and "up to date" but not "connected", so merging as-is means a healthy connected connector starts showing its message with an in-progress icon on the Data Source Connectors table.

I think you want hideStatuses = ["Running", "connected", "up to date"]. errorStatuses already keeps "offline", so that half is fine.

ui/catalog/src/components/Table/components/CellRenderers.tsx
```diff
@@ -110,20 +138,39 @@ export const StatusCell = ({ value }: SharedCellRendererProps) => {
...
+  errorStatuses = ["Error", "offline", "out of sync"],
```
ryanmarc, line 160: Typo with a real effect here: this is "offline" lowercase, but the status string the app actually produces is "Offline". ConnectorStatus is "Connected" | "Offline" (api.types.ts:399), the STATUS_CONFIG key is Offline (:54), and the sort map in DataSourceConnectorsTable/types.ts:55 is Offline. Nothing produces lowercase offline.

includes is case sensitive, so on the Data Source Connectors table an Offline row with a message now falls through to the in-progress icon instead of the error icon. The JSDoc two lines up says the default preserves existing behaviour for the deployments table, which is what made me go and check.

Smallest fix is just the capital O. If you want my opinion on the shape though: I'd set the defaults back to exactly ["Running"] and ["Error", "Offline"] and have the new datasources table pass hideStatuses={["up to date"]} and errorStatuses={["out of sync"]} at the call site. Then the shared default is "what it always was", the doc comment is true, and adding a fifth table later can't silently move an existing one.

ui/catalog/src/components/DeploymentDetails/components/ApplicationDatasourcesTable/ApplicationDatasourcesTable.tsx
```diff
@@ -0,0 +1,367 @@
...
+
+      try {
+        const { rows, pagination } = await fetchApplicationDatasources(
+          applicationId,
+          page,
+          pageSize,
+        );
+
+        const totalPages = pagination.total_pages;
+        if (page > totalPages) {
```
ryanmarc: The backend returns total_pages: 0 for an empty list rather than 1:
```
// datasource_service.go:1173
totalPages := 0
if total > 0 { totalPages = (total + req.PageSize - 1) / req.PageSize }
```
so for an application with no data sources connected, page is 1, totalPages is 0, and this branch fires. This creates two HTTP calls (page=1 then page=0), and state.page / pageRef.current both left at 0.

The extra request isn't free, since each one fans out to the Digitize pod backend-side. And the 0 sticks around: once a data source is connected and someone hits refresh, loadDatasources() defaults to pageRef.current = 0, the server normalises it so rows come back fine, the clamp doesn't re-fire, and Carbon Pagination ends up with page={0}.

`if (totalPages > 0 && page > totalPages)` covers both.

ryanmarc (changes requested):

ui/catalog/src/components/Table/components/CellRenderers.tsx
Niranjan-J-Joshi: Addded "connected" status to hideStatues.

ui/catalog/src/components/Table/components/CellRenderers.tsx
Niranjan-J-Joshi: This PR wasn’t up to date with main when you reviewed it. After merging main, ConnectorStatus is "connected" | "offline" (lowercase) everywhere, which matches what the backend sends.

ui/catalog/src/components/DeploymentDetails/components/ApplicationDatasourcesTable/ApplicationDatasourcesTable.tsx
Niranjan-J-Joshi: fixed. Added the totalPages >= 1 guard  to prevent the clamp from firing when the backend returns total_pages: 0 for an empty list. Followed the same pattern which is present for the DA and Services tables.

Niranjan-J-Joshi: fixed. Added the totalPages >= 1 guard to prevent the clamp from firing when the backend returns total_pages: 0 for an empty list. Followed the same pattern as the DA and Services tables.

ryanmarc (dismissed):

pranithraoibm (approved):

PR #1410, opened 2026-09-07
always fetch fresh services list, remove cache

pranithraoibm:
Fix for https://jsw.ibm.com/browse/AISERVICES-1746

Services is now not stored in cache and fetched freshly everytime on the step zero of the services deploy flow and in the catalog tab of the services page

https://github.com/user-attachments/assets/41cabb71-8fc2-470b-8944-964329c5623e


ui/catalog/src/hooks/useServices.ts
```diff
@@ -19,14 +18,10 @@ export const useServices = (autoFetch: boolean = false) => {
...
-
-  // Manual fetch function that can be called by components
+  // Read servicesLoading via getState() to keep it out of deps and prevent
+  // refetch from being recreated on every loading state change.
   const refetch = useCallback(async () => {
-    if (servicesLoading) return; // Don't fetch if already loading
+    if (useServiceDeployStore.getState().servicesLoading) return;
 
     setServicesLoading(true);
     setServicesError(null);
```
ryanmarc: The order of these 2 should be flipped.


```suggestion
    setServicesError(null);
    setServicesLoading(true);
```

ryanmarc (changes requested):

ryanmarc (dismissed):

Niranjan-J-Joshi (approved):

PR #1411, opened 2026-09-07
Podman domain san

shwetha-s-poojary:
## Summary

- Return the worker gateway address from the catalog worker registration API.
- Use the API-provided gateway address in the generated `worker join` command.
- Generate Podman gateway certificates with a SAN matching the domain-derived gateway hostname.
- Support the configured gateway port in the returned address.
- Reject empty, malformed, and IP-based gateway addresses during TLS server-name validation.
- Add unit coverage for domain-derived SAN generation and missing `DOMAIN_SUFFIX` validation.

## Worker registration

The catalog worker registration API now returns the worker name, gateway address, and single-use bootstrap token:

```json
{
  "worker_name": "worker0",
  "gateway_address": "catalog-worker-gateway.example.com:9090",
  "token": "***"
}
```
Join message:
```
$ ./bin/ai-services-linux-ppc64le catalog worker register worker0

Worker registered successfully.
  Name:  worker0
  Token: ***

Run the following command on the worker node:
  ai-services worker join catalog-worker-gateway.example.com:9090 --token *** --runtime <runtime>

The token is single-use and expires after 24 hours.
```

Note: If the configured custom domain is not resolvable from the worker environment, provide an explicit host mapping with --add-host during the worker join.

```
$ ./bin/ai-services-linux-ppc64le worker join \
    catalog-worker-gateway.example.com:9090 \
    --token *** \
    --runtime podman \
    --add-host catalog-worker-gateway.example.com:<gatewayIP>
```

### DIFF
--- a/ai-services/cmd/ai-services/cmd/catalog/worker.go
+++ b/ai-services/cmd/ai-services/cmd/catalog/worker.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"strconv"
 	"strings"
 	"text/tabwriter"
 	"time"
@@ -13,6 +14,7 @@ import (
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog/client"
 	catalogtypes "github.com/project-ai-services/ai-services/internal/pkg/catalog/types"
 	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+	workerconstants "github.com/project-ai-services/ai-services/internal/pkg/worker/constants"
 )
 
 // NewWorkerCmd returns the parent command for worker management.
@@ -149,7 +151,7 @@ If the worker is currently connected its gRPC stream is also cleaned up.`,
 // e.g. "https://catalog-api.10.0.0.1.nip.io" → "gateway.10.0.0.1.nip.io:9090"
 // Falls back to "<catalog-server-host>:9090" if the subdomain cannot be parsed.
 func gatewayAddrFromServerURL(serverURL string) string {
-	const defaultGatewayPort = "9090"
+	defaultGatewayPort := strconv.Itoa(workerconstants.WorkerGatewayPort)
 
 	parsed, err := url.Parse(serverURL)
 	if err != nil || parsed.Host == "" {
--- a/ai-services/cmd/ai-services/cmd/catalog/worker.go
+++ b/ai-services/cmd/ai-services/cmd/catalog/worker.go
@@ -2,10 +2,7 @@ package catalog
 
 import (
 	"fmt"
-	"net/url"
 	"os"
-	"strconv"
-	"strings"
 	"text/tabwriter"
 	"time"
 
@@ -14,7 +11,6 @@ import (
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog/client"
 	catalogtypes "github.com/project-ai-services/ai-services/internal/pkg/catalog/types"
 	"github.com/project-ai-services/ai-services/internal/pkg/logger"
-	workerconstants "github.com/project-ai-services/ai-services/internal/pkg/worker/constants"
 )
 
 // NewWorkerCmd returns the parent command for worker management.
@@ -72,7 +68,7 @@ Pass the token to the worker node and run:
 			logger.Infof("  Token: %s\n", resp.Token)
 			logger.Infoln("\nRun the following command on the worker node:")
 			logger.Infof("  ai-services worker join %s --token %s --runtime <runtime>\n",
-				gatewayAddrFromServerURL(c.ServerURL()), resp.Token)
+				resp.GatewayAddress, resp.Token)
 			logger.Infoln("\nThe token is single-use and expires after 24 hours.")
 
 			return nil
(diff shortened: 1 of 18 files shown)


ai-services/cmd/ai-services/cmd/catalog/worker.go
```diff
@@ -140,6 +144,27 @@ If the worker is currently connected its gRPC stream is also cleaned up.`,
 	return cmd
 }
 
+// gatewayAddrFromServerURL derives the worker gateway dial address from the
+// catalog API server URL stored in credentials.
+// e.g. "https://catalog-api.10.0.0.1.nip.io" → "gateway.10.0.0.1.nip.io:9090"
+// Falls back to "<catalog-server-host>:9090" if the subdomain cannot be parsed.
+func gatewayAddrFromServerURL(serverURL string) string {
+	const defaultGatewayPort = "9090"
```
mayuka-c: Isnt there already a global constant, if not lets add it and utilize here

ai-services/cmd/ai-services/cmd/catalog/worker.go
```diff
@@ -140,6 +144,27 @@ If the worker is currently connected its gRPC stream is also cleaned up.`,
...
+	parsed, err := url.Parse(serverURL)
+	if err != nil || parsed.Host == "" {
+		return serverURL + ":" + defaultGatewayPort
+	}
+
+	host := parsed.Hostname()
+	// Replace the first subdomain (e.g. "catalog-api") with "gateway".
+	if idx := strings.Index(host, "."); idx != -1 {
+		return "gateway" + host[idx:] + ":" + defaultGatewayPort
+	}
```
mayuka-c: Let us follow same naming as openshift, its `catalog-worker-gateway`

mayuka-c: Also please add a global const for this too

ai-services/internal/pkg/worker/gateway/pki.go
```diff
@@ -121,23 +123,35 @@ func generateServerCert(caCert *x509.Certificate, caKey *ecdsa.PrivateKey, serve
...
 		}
-		serverNames = []string{route, workerconstants.OpenShiftGatewayServiceEndpoint}
+		dnsNames = []string{route, workerconstants.OpenShiftGatewayServiceEndpoint}
 	case types.RuntimeTypePodman:
-		serverNames = []string{workerconstants.PodmanGatewayServerName, workerconstants.PodmanGatewayPodName}
+		domainSuffix := utils.GetEnv("DOMAIN_SUFFIX", "")
+		if domainSuffix == "" {
+			return empty, fmt.Errorf("DOMAIN_SUFFIX environment variable not set — cannot generate gateway server cert")
+		}
+		dnsNames = []string{"gateway." + domainSuffix, workerconstants.PodmanGatewayServerName, workerconstants.PodmanGatewayPodName}
```
mayuka-c: Can we remove workerconstants.PodmanGatewayServerName now as we are using domain?

ai-services/internal/pkg/worker/gateway/pki.go
```diff
@@ -121,23 +123,35 @@ func generateServerCert(caCert *x509.Certificate, caKey *ecdsa.PrivateKey, serve
...
 	case types.RuntimeTypeOpenShift:
 		route, err := GatewayRouteHost(ctx)
 		if err != nil {
 			return empty, fmt.Errorf("resolve gateway route host for cert SAN: %w", err)
 		}
-		serverNames = []string{route, workerconstants.OpenShiftGatewayServiceEndpoint}
+		dnsNames = []string{route, workerconstants.OpenShiftGatewayServiceEndpoint}
 	case types.RuntimeTypePodman:
-		serverNames = []string{workerconstants.PodmanGatewayServerName, workerconstants.PodmanGatewayPodName}
+		domainSuffix := utils.GetEnv("DOMAIN_SUFFIX", "")
```
mayuka-c, line 138: Isnt there already a method for this, if not lets add this in utils of doing GetEnv

ai-services/internal/pkg/worker/gateway/pki.go
```diff
@@ -121,23 +123,35 @@ func generateServerCert(caCert *x509.Certificate, caKey *ecdsa.PrivateKey, serve
...
+		if domainSuffix == "" {
+			return empty, fmt.Errorf("DOMAIN_SUFFIX environment variable not set — cannot generate gateway server cert")
+		}
```
mayuka-c, line 141: We should be doing validations right at the beginning and need not have to do here

ai-services/cmd/ai-services/cmd/catalog/worker.go
```diff
@@ -66,8 +68,10 @@ Pass the token to the worker node and run:
 			logger.Infoln("Worker registered successfully.")
 			logger.Infof("  Name:  %s\n", resp.WorkerName)
 			logger.Infof("  Token: %s\n", resp.Token)
-			logger.Infoln("\nPass this token to the worker daemon with --token.")
-			logger.Infoln("The token is single-use and expires after 24 hours.")
+			logger.Infoln("\nRun the following command on the worker node:")
+			logger.Infof("  ai-services worker join %s --token %s --runtime <runtime>\n",
+				gatewayAddrFromServerURL(c.ServerURL()), resp.Token)
```
yussufsh: This whole line should be the response of the register API. So that UI can also make use of the pre-populated join command.

ai-services/internal/pkg/worker/gateway/pki.go
shwetha-s-poojary: GetEnv itself is the utility.

ai-services/cmd/ai-services/cmd/catalog/worker.go
shwetha-s-poojary: There's none yet, will add

ai-services/cmd/ai-services/cmd/catalog/worker.go
shwetha-s-poojary: Renaming it means we need to change it while population SAN and deriving gateway address as well.Will do.

ai-services/internal/pkg/worker/gateway/pki.go
shwetha-s-poojary: This means we reject empty/IP gateway addresses instead of falling back to this default.Is that what is expected?

ai-services/internal/pkg/worker/gateway/pki.go
shwetha-s-poojary: I think you mean validating before injecting injecting during deployment? - the existing flow already does that.This is just  a defensive check in case the env var is missing later.

ai-services/cmd/ai-services/cmd/catalog/worker.go
shwetha-s-poojary: returning the gateway address and token separately for the UI to format would be better right?
Something like this:
```
{
  "worker_name": "node-1",
  "gateway_address": "catalog-worker-gateway.example.com:9090",
  "token": "..."
}
```

yussufsh: That works as well.

ai-services/internal/pkg/worker/gateway/pki.go
hasan4791: @mayuka-c I feel, joining with ip would be simpler for user.  They don't need to add entries in their dns or specify the --add-hosts options as well. Moreover, it works seamlessly with the powervs case.

yussufsh (approved):

PR #1412, opened 2026-09-07
Support hostAliases for Openshift worker

mayuka-c:
- This is to support adding hostAliases for Openshift remote worker to allow worker pod to be able to join to the remote Openshift cluster.
- Also increases Helm Install timeout to 20 mins as 10 mins is quite less as I see isvc storage initilizers sometimes takes time to initialize.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/worker/join.go
+++ b/ai-services/cmd/ai-services/cmd/worker/join.go
@@ -177,13 +177,15 @@ func joinRunE(cmd *cobra.Command, args []string) error {
 				Token:       token,
 			},
 			Setup: workertypes.Options{
-				BaseDir:     aiServicesDir,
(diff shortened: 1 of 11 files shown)


mayuka-c:
It took around more than 10 mins. So increasing to 20 to keep more buffer as sometimes image pull and others might take time

image

yussufsh (dismissed):

ai-services/cmd/ai-services/cmd/worker/join.go
yussufsh: Can we have exact gateway sub-domain here?

mayuka-c: Done

Shubhamag12 (approved):
lgtm

yussufsh (approved):

PR #1415, opened 2026-09-08
feat(worker): Enable encryption for mtls keys on both gateway & worker components

hasan4791:
Enable encryption for the following keys,
1. Root ca key -> Gateway
2. Server key -> Gateway
3. Client key -> Worker


ai-services/internal/pkg/worker/encryption/encryption.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	sum := sha256.Sum256([]byte(raw))
+
+	return sum[:], nil
+}
+
+// EncryptPEM encrypts PEM-encoded private key bytes using AES-256-GCM.
+// The returned bytes are nonce || ciphertext+tag (not base64; written raw to disk).
+func EncryptPEM(keyPEM []byte, aesKey []byte) ([]byte, error) {
```
yussufsh: Can we have a utility for encryption/decryption for both catalog and the worker in a common place?

hasan4791: It is common. need to update the comments though.

yussufsh: I mean reuse code from `ai-services/internal/pkg/catalog/utils/connector_utils.go`

hasan4791: Updated.

yussufsh (commented):
lgtm
update the versions

hasan4791:
> lgtm update the versions

done

yussufsh (approved):
lgtm

PR #1416, opened 2026-09-08
AIS-1684 - switch to internal routes for digitize calls

prachikurade1:


prachikurade1:
Deferred to next quarter.
cc: @yussufsh

mayuka-c:
> Deferred to next quarter. cc: @yussufsh

Dont we need this?. Otherwise we might face issue in PowerVS with custom domain right?

prachikurade1:
> > Deferred to next quarter. cc: @yussufsh
> 
> Dont we need this?. Otherwise we might face issue in PowerVS with custom domain right?

This was specifically to avoid using the "insecure" flag, so in Power VS also with the digitize api url the calls should work right ?

prachikurade1:
Closing this PR - Updated approach to store internal endpoint and use is part of this PR - https://github.com/IBM/project-ai-services/pull/1436

PR #1417, opened 2026-09-08
fix: use proper container path for certs

yussufsh:
The mount paths were moved to non-overlapping container-internal paths:
- worker-tls → /data/worker-tls
- gateway-pki → /data/gateway-pki

This will avoid conflict with the default hostPath mount used by BaseDir.


mayuka-c (approved):
LGTM

PR #1419, opened 2026-09-08
Add get conversion admin endpoint

dharaneeshvrd:
- Adding this as a admin utility to inspect the conversion tasks added to the application

### DIFF
--- a/services/digitize/api/v1/admin.py
+++ b/services/digitize/api/v1/admin.py
@@ -9,13 +9,18 @@ Exposes one router:
 - ``router`` → mounted at ``/v1`` (for /import and /export)
 """
 
+from typing import List, Optional
+
 from fastapi import APIRouter, HTTPException, Query
+from pydantic import BaseModel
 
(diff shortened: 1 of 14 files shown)


services/digitize/api/v1/admin.py
```diff
@@ -155,3 +160,108 @@ async def export_metadata(
...
+
+    class Config:
+        from_attributes = True
+
+
+# ------------------------------------------------------------------ #
+# Conversion tasks endpoint                                           #
+# ------------------------------------------------------------------ #
+
+_VALID_STATUSES = {s.value for s in ConversionTaskStatus}
```
manju956: Can't we directly use ConversionTaskStatus enum values here?

manju956 (commented):
overall looks good.

services/digitize/api/v1/admin.py
dharaneeshvrd: Done, ptal

manju956 (approved):

PR #1420, opened 2026-09-08
Cleanup process pool after every conversion to avoid mem corruption

dharaneeshvrd:
**Problem:**
During a large batch ingestion job, the service crashed with:
```
concurrent.futures.process.BrokenProcessPool: A process in the process pool was
terminated abruptly while the future was running or pending.
```
This caused the entire application to restart, marking all in-progress and queued conversion tasks as failed.

**Analysis:**
The root cause is unbounded memory growth in long-lived worker processes. The ProcessPoolExecutor was created with no max_tasks_per_child limit, so the same 4 worker processes were reused for the entire lifetime of the application. Each Docling DocumentConverter.convert() call loads ML model weights and accumulates internal layout/page state that Python's GC does not fully reclaim between tasks. After processing several hundred-page documents back-to-back (e.g. 1952 pages + 1562 pages + 1494 pages in the same workers without recycling), memory pressure grew until the OS OOM-killer terminated a worker process mid-task, poisoning the entire pool.

A secondary issue is that the default fork start method on Linux causes each worker to inherit the parent process's full memory image (FastAPI, SQLAlchemy, all loaded modules) before Docling even starts, inflating the baseline RSS of every worker.

**Fix:**

Two changes to the ProcessPoolExecutor construction in conversion_dispatcher.py:

- `max_tasks_per_child=1` — recycles each worker process after it completes one conversion task. All Docling model state, ML tensors, and intermediate page data are freed at the OS level between tasks. Since each conversion takes minutes, the process respawn cost is negligible.
- `mp_context=multiprocessing.get_context("spawn")` — workers start with a clean Python interpreter instead of a fork of the parent, so the parent's loaded application state is not duplicated into every worker's address space before the task begins.

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.334
+TAG?=v0.0.335
 CONTAINER_BUILDER?=podman
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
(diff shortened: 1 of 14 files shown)


manju956 (dismissed):
lgtm

manju956 (approved):

PR #1423, opened 2026-09-08
fix: run dispatcher loop in routines

yussufsh:


yussufsh:
Implemented in https://github.com/IBM/project-ai-services/pull/1396

PR #1424, opened 2026-09-08
feat: integrate worker resources and runtime selection into deploy flows

pranithraoibm:
- Added options to pick runtime type and target worker during deployment.
- Automatically filters workers matching the chosen runtime, defaulting to the local worker.
- Users can register a new worker resource directly from the dropdown and refresh the list.
- Updated deploy stores and payloads to include the selected worker and deployment type.
- Updated the resource requirements calculation based on the selected worker

https://github.com/user-attachments/assets/9b8a81af-0b00-42db-a3bd-604962b4c3a3

https://github.com/user-attachments/assets/8abc350a-3860-4763-87d5-f1dfc0093fcf

https://github.com/user-attachments/assets/ce59a962-61da-4d3a-a655-1be94c334fb6


ui/catalog/src/components/DeployFlow/Shared/steps/SharedStepOne.tsx
```diff
@@ -49,16 +113,123 @@ export const SharedStepOne = ({
...
+      onChange({ workerName: next });
+      if (next) onWorkerErrorReset?.();
+    }
+  }, [
+    workerOptions,
+    formData.workerName,
+    onChange,
+    isLoadingWorkers,
+    onWorkerErrorReset,
+  ]);
```
ryanmarc, line 233: This effect can loop forever when the selected runtime has no ready worker: `onWorkerErrorReset` was added to the dep array but is passed as an inline arrow in both flows, so its identity changes every render, and the fallback change from `LOCAL_WORKER_NAME` to `""` removed the one dep change that used to let things settle — `formData.workerName` is already `""`, and `handleUpdateFormData` always returns a fresh object so `useReducer` can't bail out either. A simulation of the dep-comparison and bail-out rules shows it settling in 2 renders before the commit and still looping at 200 renders after; it reproduces by picking a deployment type with no ready worker, or on a `--skip-local-worker` install. Smallest fix is to guard the dispatch with `formData.workerName !== next`, plus `useCallback` on the two `onWorkerErrorReset` props so the effect stops running every render.

ui/catalog/src/hooks/useWorkers.ts
```diff
@@ -0,0 +1,51 @@
...
+      );
+  }, [enabled]);
+  useEffect(() => {
+    if (!enabled) {
+      hasFetched.current = false;
+      return;
+    }
+    if (hasFetched.current) return;
+    hasFetched.current = true;
+    fetchAllWorkerResources()
```
ryanmarc: This is a duplicate of the nine-line fetch in refetch and should collapse into one function both call. Neither copy cancels, and they can overlap — refetch() fires from handleModalClose after a successful registration, potentially before the mount fetch settles, so a slow first response can clobber the list containing the just-registered worker. Also, the .catch sets workers: [], meaning one transient failure empties the dropdown, trips the reconcile effect in SharedStepOne, and resets the user's selection to local; keeping the previous list on error would avoid that.

ryanmarc (changes requested):
This branch was cut from worker-deregister before #1345 landed its review fixes, so it carries an older revision of a file #1345 also modifies — specifically without #1345's dispatch of workerName: result.worker_name, the canonical name from createWorkerResp. git merge-tree shows the two PRs conflicting on exactly this file plus pages/WorkerResources/types.ts, so whoever resolves the conflict silently decides whether that fix survives, and both resolutions type-check. The ask is to rebase onto #1359 once it lands and drop anything #1336/#1345/#1359 already deliver, leaving this PR as just the deploy-flow integration.

ui/catalog/src/components/DeployFlow/Shared/steps/SharedStepOne.tsx
pranithraoibm: Done

ui/catalog/src/hooks/useWorkers.ts
pranithraoibm: Collapsed the duplicate fetch into a single effect called by both the mount path and explicit refetches via a refetchTrigger counter, used dedupe to prevent overlapping requests, and kept the previous worker list on error instead of resetting to [].

ryanmarc (dismissed):

Niranjan-J-Joshi (approved):

PR #1429, opened 2026-09-09
feat(connectors): add ConnectorDetailsPanel with auth key update

Niranjan-J-Joshi:
# Summary

This PR adds a **Connector Details** side panel to the Connectors page.

Users can view connector information and update authentication keys
### New: ConnectorDetailsPanel

Added a slide-in side panel that supports two modes:

#### View Mode

Shows the connector's:

- Connection details
- Location
- Authentication details
- Connected services

The information is displayed based on the connector metadata and provider schema.

#### Update Key Mode

Opens the panel directly on the authentication form.

This mode is opened from the **Update key** option in the connector's overflow menu.

Other features include:

- Uses the existing provider parameter schema when available, so no extra API call is needed.
- Inline validation for authentication fields.
- Loading state while saving.
- Success message that automatically disappears.
- Error banner with errors highlighted on the relevant fields.
- Skeleton loading state while connector details are being fetched.
- Error state when connector details cannot be loaded.
- Uses the Carbon `SidePanel` with slide-in behavior and no overlay.

## Updated: DataSourceConnectorsTable

Updated the connectors table to support the new panel and connector actions.

- Added `OPEN_DETAILS_PANEL` and `CLOSE_DETAILS_PANEL` actions to the table reducer.
- **Update key** now opens the Connector Details panel in `update-key` mode.

## API & Types

Added the following API endpoints:

- `fetchDataSourceById`
  - `GET /datasources/:id`
- `updateDataSourceAuth`
  - `PUT /datasources/:id`

Added these types to `api.types.ts`:

- `DataSourceDetailResponse`
- `ConnectorApplication`
- `UpdateDataSourceAuthRequest`

## Refactor

- Moved `calculateUptime` from `applications.api.ts` to `@/utils/time.ts`.
- `applications.api.ts` still re-exports `calculateUptime` to keep existing imports working.


ui/catalog/src/utils/time.ts
```diff
@@ -0,0 +1,31 @@
...
+  } else if (totalMinutes > 0) {
+    const mins = totalSeconds % 60 > 0 ? totalMinutes + 1 : totalMinutes;
+    return mins === 1 ? "1 minute" : `${mins} minutes`;
+  } else {
+    return totalSeconds === 1
+      ? "1 second"
+      : totalSeconds > 0
+        ? `${totalSeconds} seconds`
+        : "Just now";
+  }
```
ryanmarc: This rounds up rather than to the nearest unit, and because it triggers on any non-zero remainder it fires nearly every time. I ran the function as-is under node:
```
61 seconds ago    -> "2 minutes"
1 hour 30 min ago -> "2 hours"
25 hours ago      -> "2 days"
30 days 1h ago    -> "31 days"
```
It is also inconsistent with itself: 23 hours gives "23 hours" but 25 hours gives "2 days". I think you want the plain floored values here (totalDays, totalHours, totalMinutes without the + 1), which is the usual meaning of uptime.

Separately, an unparseable createdAt gives NaN for every comparison and lands on "Just now", which looks identical to a brand new connector. A Number.isNaN(diffMs) guard returning "Unknown" would make that distinguishable.

ui/catalog/src/components/ConnectorDetailsPanel/ConnectorDetailsPanel.tsx
```diff
@@ -0,0 +1,674 @@
...
+        dispatch({ type: PANEL_ACTION_TYPES.FETCH_SUCCESS, payload: data });
+      })
+      .catch((err: unknown) => {
+        const msg =
+          err instanceof Error
+            ? err.message
+            : "Failed to load connector details";
+        dispatch({ type: PANEL_ACTION_TYPES.FETCH_FAILURE, payload: msg });
+      });
+  }, [open, connectorId]);
```
ryanmarc, line 113: This effect has no cleanup, so responses can land out of order. Open connector A, click connector B before A returns, and if A resolves second the panel renders A's details while connectorId is B. handleSave would then PUT to B's endpoint with fields built from A's schema.

The same-shaped fix that useConnectorDropdownItems (from #1421) already uses:
```
let cancelled = false;
fetchDataSourceById(connectorId)
  .then((data) => { if (!cancelled) dispatch({ type: FETCH_SUCCESS, payload: data }); })
  .catch((err) => { if (!cancelled) dispatch({ type: FETCH_FAILURE, payload: msg }); });
return () => { cancelled = true; };
```

ryanmarc (changes requested):

ui/catalog/src/components/ConnectorDetailsPanel/ConnectorDetailsPanel.tsx
Niranjan-J-Joshi: Fixed. Added the cancelled flag with the same pattern used in useConnectorDropdownItems

ui/catalog/src/utils/time.ts
Niranjan-J-Joshi: Fixed both issues: removed the +1 rounding logic and added the Number.isNaN(diffMs) guard to return "Unknown" for invalid timestamps.

Since we’re not displaying decimal values like 1 hour 4 mins, the output will now be:

61 seconds ago → "1 minute"
1 hour 30 minutes ago → "1 hour"
25 hours ago → "1 day"
30 days 1 hour ago → "30 days"

ui/catalog/src/components/ConnectorDetailsPanel/ConnectorDetailsPanel.tsx
```diff
@@ -0,0 +1,684 @@
...
+            className={styles.inlineNotification}
+          />
+        )}
+
+        {/* View mode — "Update key" button; disabled when connector is offline */}
+        {authSectionState.kind === "button" && (
+          <Button
+            kind="tertiary"
+            size="sm"
+            disabled={state.detail?.status === "offline"}
```
ryanmarc: The backend only has two statuses (connector.go:13-14: connected, offline), and the sync job sets offline on any failed check including [AUTH]. So status !== "connected" at line 370 and status === "offline" here are the same condition: whenever the "Invalid key credentials" notification shows, this button is disabled, and the overflow menu item in CellRenderers.tsx:63 is disabled by the same rule. There is then no way in the UI to fix a rejected key, even though PUT /datasources/:id re-tests and flips the connector back to connected on success.

ryanmarc (changes requested):

ui/catalog/src/components/ConnectorDetailsPanel/ConnectorDetailsPanel.tsx
Niranjan-J-Joshi: I followed the Figma design and the Figma dev notes specified disabling “Update key” when the status is Offline, but this also blocks users from fixing failures. Should I keep “Update key” enabled even when the connector is Offline?

image

ryanmarc (approved):

PR #1430, opened 2026-09-09
correct worker gateway address and server cert SANs

shwetha-s-poojary:
fix: correct worker gateway address and server cert SANs

1. Restore host:443 for OpenShift gateway address (bare hostname broke worker StartGrpcStream validation)
2. Add PodmanGatewayPodName to Podman server cert SANs (was defined but never wired in, causing TLS failures for internal pod DNS connections)
3. Bump image to v0.0.343


mayuka-c (dismissed):
LGTM

yussufsh (approved):

PR #1431, opened 2026-09-09
fix: use pod name instead of hostip during auto join

Shubhamag12:
The worker now connects through the pod name, matching the gateway certificate and avoiding host IP connectivity errors.


yussufsh (approved):

mayuka-c (approved):
LGTM

PR #1432, opened 2026-09-09
Add Namespace and Worker name to Application ps

mayuka-c:
- Currently the app ps doesnt print worker name and namespace which makes it really difficult through CLI to know the worker to which the application is deployed and also the namespace it resides if openshift (in case of podman it is empty)

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/ps.go
+++ b/ai-services/cmd/ai-services/cmd/application/ps.go
@@ -191,27 +191,21 @@ func renderApplicationPS(ctx context.Context, opts appTypes.ListOptions) error {
 	return nil
 }
 
-// PS table column indices (normal output).
+// PS table column indices (shared across normal and wide output).
 const (
-	psColAppName = 0
-	psColWorker  = 1
-	psColRuntime = 5
-)
-
-// PS table column indices (wide output).
-const (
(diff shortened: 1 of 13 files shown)


ai-services/cmd/ai-services/cmd/application/ps.go
```diff
@@ -194,8 +194,8 @@ func renderApplicationPS(ctx context.Context, opts appTypes.ListOptions) error {
 // setApplicationPSTableHeaders sets the table headers based on output format.
 func setApplicationPSTableHeaders(printer *utils.Printer, outputWide bool) {
 	if outputWide {
-		printer.SetHeaders("APPLICATION NAME", "POD ID", "POD NAME", "STATUS", "CREATED", "CONTAINERS")
+		printer.SetHeaders("APPLICATION NAME", "WORKER", "NAMESPACE", "POD ID", "POD NAME", "STATUS", "CREATED", "CONTAINERS")
```
yussufsh: Can we add runtime as well here?

mayuka-c: Done

mayuka-c:
Validation

image

yussufsh (approved):

PR #1434, opened 2026-09-09
Add completed_with_errors as terminal state for the doc

dharaneeshvrd:
During cancellation completed_with_errors docs marked as cancelled instead of its original terminal status. This PR fixes that.


manju956 (approved):
lgtm

PR #1437, opened 2026-09-09
Accept schema json example part of extract sync and async requests

manalilatkar:
Now both sync and async routes for entity extraction will accept raw_json_schema or json_example isntead of schema_id so user can get flexibility while passing input.

### DIFF
--- a/services/extract/api/v1/jobs.py
+++ b/services/extract/api/v1/jobs.py
@@ -48,7 +48,6 @@ from extract.utils.job import (
     check_job_admission,
     process_batch_job,
     process_file,
-    resolve_schema,
     validate_and_resolve_file,
     validate_file_content,
     delete_all_job_files,
@@ -152,7 +151,6 @@ async def extract_sync(request: Request, body: ExtractionRequest) -> JSONRespons
         msg = "text field is empty"
         logger.error(msg)
         raise ExtractException(400, "INVALID_REQUEST", msg)
-
 
     # ------------------------------------------------------------------
     # 2. Semaphore check (non-blocking — reject immediately if saturated)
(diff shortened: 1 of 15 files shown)


services/extract/api/v1/jobs.py
dharaneeshvrd, line 292: In case resolve returns _EphemeralSchemaRow, how it would work here?

services/extract/api/v1/jobs.py
```diff
@@ -438,15 +409,60 @@ async def create_extract_job(
...
+            )
+
+    json_example_dict: Optional[dict] = None
+    if json_example is not None:
+        try:
+            json_example_dict = json.loads(json_example)
+        except json.JSONDecodeError as exc:
+            raise ExtractException(
+                400, "INVALID_REQUEST", f"json_example is not valid JSON: {exc}"
+            )
```
dharaneeshvrd, line 453: Seems this validation also happens in resolve_schema_input.
You can remove it here

dharaneeshvrd (commented):
overall lgtm

services/extract/api/v1/jobs.py
manalilatkar: In that case, schema_id will be null.

services/extract/api/v1/jobs.py
manalilatkar: You mean` if json_example is not None:`  on line no 447?
I have to do it if I want to execute line no 449 here so I can pass the JSON-loaded example dict to the resolve_schema_input method.

dharaneeshvrd: yes, keep the validation in the resolve_schema_input method.

services/extract/api/v1/jobs.py
dharaneeshvrd: But this would expect the id right?
how it would use the ephemeral schema?
I dont see a reference of the schema_row anywhere else.

dharaneeshvrd: I did nt look properly, this is valid.

services/extract/api/v1/jobs.py
dharaneeshvrd: this can also be there.

dharaneeshvrd (approved):

PR #1439, opened 2026-09-09
use worker join flow in catalog for local

yussufsh:
Make use of register API for local worker join in catalog configure.

Replace the LocalWorkerToken gateway bypass with a proper POST /api/v1/workers flow: catalog configure logs in once, pre-registers the local worker via the catalog API to get a real bootstrap token, and passes the authenticated client to JoinAsLocalWorker — no sentinel token, no second login.

Added case to not allow deleting Local worker.

### DIFF
--- a/ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
+++ b/ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
@@ -96,14 +96,14 @@ func handlePostDeployment(ctx context.Context, tp templates.Template, runtime *r
 		return fmt.Errorf("admin password verification failed: %w", err)
 	}
 
-	// Step 8 (was Step 7): Join as local worker
+	// Step 8: Join as local worker
 	if !opts.SkipLocalWorker {
 		if err := JoinAsLocalWorker(ctx, runtime, catalogClient); err != nil {
 			return fmt.Errorf("local worker join failed: %w", err)
 		}
 	}
 
-	// Step 9 (was Step 8): Print next steps with route URLs
+	// Step 9: Print next steps with route URLs
 	if err := helpers.PrintNextSteps(ctx, tp, runtime, catalogconstants.CatalogAppName, catalogconstants.CatalogAppTemplate); err != nil {
 		logger.Infof("failed to display next steps: %v\n", err)
 
--- a/ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
+++ b/ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/project-ai-services/ai-services/assets"
 	"github.com/project-ai-services/ai-services/internal/pkg/catalog/cli/configure"
+	configureutils "github.com/project-ai-services/ai-services/internal/pkg/catalog/cli/configure/utils"
 	catalogconstants "github.com/project-ai-services/ai-services/internal/pkg/catalog/constants"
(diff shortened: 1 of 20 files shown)


ai-services/internal/pkg/catalog/cli/configure/worker.go
```diff
@@ -0,0 +1,46 @@
...
+
+	catalogclient "github.com/project-ai-services/ai-services/internal/pkg/catalog/client"
+	"github.com/project-ai-services/ai-services/internal/pkg/logger"
+	workerconstants "github.com/project-ai-services/ai-services/internal/pkg/worker/constants"
+)
+
+const (
+	// CatalogAdminUser is the default admin username created during catalog configure.
+	CatalogAdminUser = "admin"
+)
```
mayuka-c: Can we define constant in common place and use it in other places in apiserver.go

ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
```diff
@@ -67,14 +77,33 @@ func DeployCatalog(ctx context.Context, opts catalogutils.OpenShiftConfigureOpti
...
+	if err != nil {
+		return fmt.Errorf("failed to resolve catalog API URL: %w", err)
+	}
+
+	catalogClient, err := configure.LoginToCatalog(ctx, catalogAPIURL, adminPassword)
+	if err != nil {
+		return fmt.Errorf("admin password verification failed: %w", err)
+	}
+
+	// Step 8 (was Step 7): Join as local worker
```
mayuka-c: Maybe `was Step 7` and also for next step `was Step 8` can be removed

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -119,8 +126,16 @@ func handlePostDeployment(ctx context.Context, caddyCtx *caddy.Context, deployCt
 		return fmt.Errorf("route registration failed: %w", err)
 	}
 
+	// Login to the catalog API — this both verifies the admin password and gives
+	// us a client to reuse for local worker registration without a second login.
+	catalogAPIURL := routeURLs["CATALOG_API_ROUTE"]
```
mayuka-c: Can we define constant for `CATALOG_API_ROUTE`

mayuka-c (commented):
Minor comments, Overall Approach LGTM

mayuka-c (approved):
LGTM

ai-services/internal/pkg/catalog/cli/configure/openshift/configure.go
```diff
@@ -67,14 +77,33 @@ func DeployCatalog(ctx context.Context, opts catalogutils.OpenShiftConfigureOpti
...
+	// us a client to reuse for local worker registration without a second login.
+	catalogAPIURL, err := getCatalogAPIURL(ctx, runtime)
+	if err != nil {
+		return fmt.Errorf("failed to resolve catalog API URL: %w", err)
+	}
+
+	catalogClient, err := configure.LoginToCatalog(ctx, catalogAPIURL, adminPassword)
+	if err != nil {
+		return fmt.Errorf("admin password verification failed: %w", err)
+	}
```
Shubhamag12, line 97: do we need to login to catalog even if localworker is skipped?
I feel it should be inside below if function if not used outside local worker

yussufsh: By default we will prompt for password and login. This will ensure configure validates the password using the actual login after catalog is deployed.

PR #1440, opened 2026-09-09
feat(connector) : Connect datasource to application 

Niranjan-J-Joshi:
# Connect Data Source to Application

## Summary

Adds a **Connect data source** modal to the **Deployment Details → Data Sources** table.

Users can now select one or more available data source connectors and connect them to the application without leaving the page.

### New Files

- `ConnectDatasourceModal.tsx`
  - Added a modal for selecting data source connectors.
  - Supports selecting multiple connectors.
  - Shows a loading skeleton while connectors are being loaded.
  - Shows an error message if loading fails.
  - Shows a message when there are no available connectors.
  - Shows selected data sources as tags below the dropdown.

- `ConnectDatasourceModal.module.scss`
  - Added styles for the new modal.
  - Uses Carbon spacing and typography tokens.

### Updated Files

| File | Changes |
|---|---|
| `ApplicationDatasourcesTable.tsx` | Connect button now opens the modal. The table refreshes after data sources are connected. |
| `applications.api.ts` | Added `connectApplicationDatasources()` API call and updated usages to use the shared data source endpoint. |
| `api-endpoints.constants.ts` | Replaced separate GET and PUT data source endpoints with a single `APPLICATION_DATASOURCES` endpoint. |


ui/catalog/src/api/applications.api.ts
```diff
@@ -425,3 +428,81 @@ export async function fetchAllDeployedServices(
...
+
+  return allData.map(transformDatasourceToRow);
+}
+
+// Connects one or more data source connectors to an application
+export async function connectApplicationDatasources(
+  applicationId: string,
+  datasourceIds: string[],
+): Promise<void> {
+  await api.put(APPLICATION_ENDPOINTS.APPLICATION_DATASOURCES(applicationId), {
```
ryanmarc: a failed connect is shown to the user as a success.

PUT /applications/:id/datasources does not fail the request when a datasource fails to connect. It collects the failures and returns them in the body with a 207 Multi-Status:
```
// datasource_service.go:497
if len(connErrors) == 0 { return nil, nil }
return &apimodels.ConnectDatasourcesResponse{Errors: connErrors}, nil
```
```
// datasource_handler.go:451
if resp == nil { c.Status(http.StatusNoContent); return }
c.JSON(http.StatusMultiStatus, resp)
```
There is no "everything failed" branch, so even selecting a single datasource that fails comes back 207. Axios's default validateStatus accepts anything in 2xx, so this await resolves.

So handleSubmit reaches onConnected(), the modal closes, the table refetches, and the rows that failed just are not there. No message, and the reasons the backend went to the trouble of sending (connectOneDatasource failures: connector load, decrypt, downstream POST) are dropped.

Suggestion: have this return the errors instead of void and in handleSubmit, only call onConnected() when that array is empty, otherwise keep the modal open and show them. You can map datasource_id back to the labels in selectedItems to get readable text.

ui/catalog/src/components/Table/components/CellRenderers.tsx
```diff
@@ -110,20 +138,39 @@ export const StatusCell = ({ value }: SharedCellRendererProps) => {
...
+   * Statuses that warrant the error icon instead of the in-progress icon.
+   * Defaults to ["Error", "Offline"] to preserve existing behaviour for the deployments table.
+   */
+  errorStatuses?: string[];
+}
+
+export const MessageCell = ({
+  value,
+  rowData,
+  hideStatuses = ["Running", "up to date"],
```
ryanmarc: This also silently reverts the fix made by #1344

ui/catalog/src/components/DeploymentDetails/components/ApplicationDatasourcesTable/ApplicationDatasourcesTable.tsx
```diff
@@ -0,0 +1,386 @@
...
+                  </>
+                )}
+              </DataTable>
+            )}
+
+            {/* Connect datasource modal */}
+            <ConnectDatasourceModal
+              open={isConnectModalOpen}
+              applicationId={applicationId}
+              connectedIds={new Set(state.rowsData.map((r) => r.id))}
```
ryanmarc: The dropdown builds its "already connected" set from state.rowsData, which holds only the current page of the datasources table (20 rows by default), while allConnectors is fetched exhaustively — so on any app with more than 20 connected sources, everything past page 1 is wrongly offered as available, and picking one PUTs a datasource that's already attached. Which entries appear also depends on which page the user is viewing, making bug reports hard to reproduce. The fix is to call fetchAllApplicationDatasources(applicationId) in the same effect that loads allConnectors and build the excluded set from that, costing one extra request but removing the dependency on table pagination.

ryanmarc (changes requested):

ui/catalog/src/api/applications.api.ts
Niranjan-J-Joshi: Updated the API to handle 207 responses and return datasource errors. For partial failures, successful connections are refreshed and removed from the dropdown, while failed connections stay visible with their specific error messages.

ui/catalog/src/components/Table/components/CellRenderers.tsx
Niranjan-J-Joshi: fixed

ui/catalog/src/components/DeploymentDetails/components/ApplicationDatasourcesTable/ApplicationDatasourcesTable.tsx
Niranjan-J-Joshi: Removed the dependency on state.rowsData. ConnectDatasourceModal now fetches all application datasources along with the connectors in Promise.all when the modal opens, ensuring all connected sources across pages are fetched and excluded correctly.

ryanmarc (dismissed):

pranithraoibm (approved):

PR #1442, opened 2026-09-10
[BYOS] Minor Fixes for custom component

mayuka-c:
- Call GetDeployOptions for application templates to use print the components along with provider types. Currently reads from embed asset
- Add name collision check so that if 2 components with same provider or 2 services have same name with diff ids, then it becomes easy to distinguish on CLient (UI/CLI)

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/templates/parameters.go
+++ b/ai-services/cmd/ai-services/cmd/application/templates/parameters.go
@@ -96,6 +96,7 @@ func displayArchitectureParameters(ctx context.Context, source catalogClient.Cat
 		deployOpts, err := source.GetServiceDeployOptions(ctx, svcRef.ID, runtimeType())
 		if err != nil {
 			logger.Warningf("skipping parameters for service '%s': %v", svcRef.ID, err)
+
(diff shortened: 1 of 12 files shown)


ai-services/cmd/ai-services/cmd/application/templates/parameters.go
```diff
@@ -65,90 +70,69 @@ func NewParametersCmd() *cobra.Command {
 }
 
 // displayServiceParameters displays all parameters for a specific service.
-func displayServiceParameters(ctx context.Context, source catalogClient.CatalogSource, serviceID string, dependencies []catalogTypes.DependencyReference) error {
+func displayServiceParameters(ctx context.Context, source catalogClient.CatalogSource, serviceID string, _ []catalogTypes.DependencyReference) error {
```
yussufsh: Get rid of the param.

mayuka-c: Done

yussufsh (dismissed):

yussufsh (approved):

PR #1443, opened 2026-09-10
feat(worker): propagate ctx cancellation to remote worker via COMMAND_TYPE_CANCEL

shwetha-s-poojary:
Problem

Ctrl+C or a mid-deployment delete had no effect on the remote worker - the in-flight Helm install, model download, or pod creation kept running until completion regardless.

What changed

- Added COMMAND_TYPE_CANCEL to the worker gRPC protocol. The control plane sends this when a deployment's context is cancelled; the worker cancels the per-command context so the in-flight operation (Helm, model download, pod creation) aborts immediately.
- Dispatcher now tracks in-flight commands by ID with individual cancellable contexts, allowing targeted cancellation without touching the stream.
- Sender.Send fires a best-effort COMMAND_TYPE_CANCEL when its context is cancelled before a result arrives.
- pollApplicationStatus (CLI application create) now handles ctx.Done() and calls DeleteApplication so the server-side cancellation chain fires on Ctrl+C.
- Readiness poll loops in helper.go replaced time.Sleep with select on ctx.Done() to unblock immediately on cancellation.

Testing

- Manually verified on Podman remote worker (with both Podman and Openshift as gateway) - DELETE mid model-download correctly sends COMMAND_TYPE_CANCEL, stops the download container, and completes deletion cleanly.

### DIFF
--- a/ai-services/cmd/ai-services/cmd/application/create.go
+++ b/ai-services/cmd/ai-services/cmd/application/create.go
@@ -433,6 +433,9 @@ func buildCatalogPayload(ctx context.Context, appClient *catalogClient.Applicati
 }
 
 // pollApplicationStatus polls the application status until it's ready or fails.
+// If ctx is cancelled (e.g. Ctrl+C) the in-flight deployment is deleted so the
+// server-side cancellation chain fires (DeploymentRegistry.Cancel →
+// Sender.sendCancel → COMMAND_TYPE_CANCEL on the worker).
 func pollApplicationStatus(ctx context.Context, appClient *catalogClient.ApplicationClient, appName, id string) error {
 	logger.Infof("Waiting for application '%s' to be ready...\n", appName)
 
@@ -443,6 +446,15 @@ func pollApplicationStatus(ctx context.Context, appClient *catalogClient.Applica
 
 	for {
 		select {
+		case <-ctx.Done():
+			logger.Infof("Interrupted — deleting application '%s' to stop the in-flight deployment...\n", appName)
+			deleteCtx := context.Background()
+			if err := appClient.DeleteApplication(deleteCtx, id, &catalogClient.DeleteApplicationParams{}); err != nil {
+				logger.Warningf("Failed to delete application '%s' after interruption: %v\n", appName, err)
+			}
+
+			return ctx.Err()
+
 		case <-timeout:
 			return fmt.Errorf("timeout waiting for application '%s' to be ready", appName)
 
--- a/ai-services/cmd/ai-services/cmd/application/create.go
+++ b/ai-services/cmd/ai-services/cmd/application/create.go
@@ -433,9 +433,6 @@ func buildCatalogPayload(ctx context.Context, appClient *catalogClient.Applicati
 }
 
 // pollApplicationStatus polls the application status until it's ready or fails.
-// If ctx is cancelled (e.g. Ctrl+C) the in-flight deployment is deleted so the
-// server-side cancellation chain fires (DeploymentRegistry.Cancel →
-// Sender.sendCancel → COMMAND_TYPE_CANCEL on the worker).
 func pollApplicationStatus(ctx context.Context, appClient *catalogClient.ApplicationClient, appName, id string) error {
 	logger.Infof("Waiting for application '%s' to be ready...\n", appName)
 
@@ -446,15 +443,6 @@ func pollApplicationStatus(ctx context.Context, appClient *catalogClient.Applica
 
 	for {
 		select {
-		case <-ctx.Done():
-			logger.Infof("Interrupted — deleting application '%s' to stop the in-flight deployment...\n", appName)
-			deleteCtx := context.Background()
-			if err := appClient.DeleteApplication(deleteCtx, id, &catalogClient.DeleteApplicationParams{}); err != nil {
-				logger.Warningf("Failed to delete application '%s' after interruption: %v\n", appName, err)
-			}
-
-			return ctx.Err()
-
 		case <-timeout:
 			return fmt.Errorf("timeout waiting for application '%s' to be ready", appName)
 
--- a/ai-services/internal/pkg/worker/dispatch/dispatch.go
+++ b/ai-services/internal/pkg/worker/dispatch/dispatch.go
@@ -1,9 +1,15 @@
 // Package dispatch implements the worker-side command dispatcher.
 //
 // When the gRPC CommandStream delivers a Command from the control plane, the
-// worker calls Dispatch: the payload is decoded, the appropriate local
-// runtime.Runtime method is called, and a CommandResult is returned to be
-// sent back on the stream.
+// worker calls Dispatcher.Dispatch: the payload is decoded, the appropriate
+// local runtime.Runtime method is called, and a CommandResult is returned to
+// be sent back on the stream.
+//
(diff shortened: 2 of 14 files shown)


ai-services/cmd/ai-services/cmd/application/create.go
```diff
@@ -443,6 +485,15 @@ func pollApplicationStatus(ctx context.Context, appClient *catalogClient.Applica
...
 		select {
+		case <-ctx.Done():
+			logger.Infof("Interrupted — deleting application '%s' to stop the in-flight deployment...\n", appName)
+			deleteCtx := context.Background()
+			if err := appClient.DeleteApplication(deleteCtx, id, &catalogClient.DeleteApplicationParams{}); err != nil {
+				logger.Warningf("Failed to delete application '%s' after interruption: %v\n", appName, err)
+			}
+
+			return ctx.Err()
+
```
mayuka-c: Lets remove Ctrl+C behavior in CLI. Also further changes done as part of this

ai-services/tests/e2e/e2e_suite_test.go
```diff
@@ -81,6 +84,9 @@ func init() {
 	flag.BoolVar(&runFailureTests, "run-failure-tests", false,
 		"Opt in to running failure test suites (bootstrap, catalog, similarity). "+
 			"Failure tests are skipped by default to prevent accidental execution during a normal suite run.")
+	flag.StringVar(&workerName, "worker-name", "",
+		"Name of the connected remote worker to target for mid-deployment delete tests. "+
+			"When empty, mid-deploy-delete tests are skipped.")
 }
```
mayuka-c: Any reason of adding this in e2e?. This is nothing to do with this PR. Pls remove

ai-services/internal/pkg/worker/dispatch/dispatch.go
```diff
@@ -30,12 +37,80 @@ import (
...
+// release the context node from the parent tree (avoids a context leak).
+func (d *Dispatcher) deregister(commandID string) {
+	d.mu.Lock()
+	cancel, ok := d.inflight[commandID]
+	delete(d.inflight, commandID)
+	d.mu.Unlock()
+
+	if ok {
+		cancel()
+	}
```
mayuka-c, line 73: cancel() is called upon deferred return of Dispatch(). Any asynchronous background goroutines spawned by a runtime handler that inherit ctx will immediately have their context cancelled once the command handler returns. Ensure all runtime command implementations have finished all work before returning.

ai-services/internal/pkg/worker/stream/sender.go
```diff
@@ -84,7 +105,53 @@ func (s *Sender) Send(ctx context.Context, cmdType workerpb.CommandType, payload
...
 	}
 }
+
+// sendCancel fires a COMMAND_TYPE_CANCEL for commandID to the worker.
+// It blocks for up to cancelEnqueueTimeout waiting for a slot in the worker's
+// command channel — this ensures the cancel is not silently dropped when the
+// channel is momentarily full (e.g. during a large architecture deployment
+// with many concurrent Helm installs). Errors are silently ignored — the
+// caller (Send) already has the error it needs to return.
+func (s *Sender) sendCancel(commandID string) {
```
mayuka-c, line 129: sendCancel enqueues the COMMAND_TYPE_CANCEL onto the worker's cmdCh. If the command was queued on cmdCh but has not started executing on the worker yet, the CANCEL could be processed before the command is registered in d.inflight. If this occurs, d.cancelCommand(req.CommandID) will be a no-op, and the subsequent command will run to completion.

ai-services/internal/pkg/worker/stream/sender.go
```diff
@@ -84,7 +105,53 @@ func (s *Sender) Send(ctx context.Context, cmdType workerpb.CommandType, payload
...
+	// Use a short timeout context so we never block indefinitely if the worker
+	// is overwhelmed or disconnected between WorkerCommandChannel and here.
+	ctx, cancel := context.WithTimeout(context.Background(), cancelEnqueueTimeout)
+	defer cancel()
+
+	select {
+	case cmdCh <- cancelCmd:
+	case <-ctx.Done():
+		// Channel still full after cancelEnqueueTimeout — worker is either
+		// overwhelmed or has disconnected. Best-effort: give up rather than block.
```
mayuka-c, line 155: If cmdCh is full and cannot accept the cancel command within 5 seconds (cancelEnqueueTimeout), the cancel command is silently dropped.

mayuka-c (commented):
Few of the other issues which Bob found. PTAL

ai-services/tests/e2e/e2e_suite_test.go
shwetha-s-poojary: No there was a test suite earlier that BOB created ,which I removed - this stale code ended up here.
Thanks for flagging!! Removed.

ai-services/internal/pkg/worker/stream/sender.go
shwetha-s-poojary: @mayuka-c How should we handle this?Should. we just log the miss or do you have anything else in mind?

ai-services/internal/pkg/worker/stream/sender.go
shwetha-s-poojary: Right, that's a race.
Bob suggested having  another entry in Dispatcher just to track that, something like this:
```
type Dispatcher struct {
    mu        sync.Mutex
    inflight  map[string]context.CancelFunc
    cancelled map[string]struct{}          // tombstone: cancelled before registered
}
```
cancelCommand — if the ID isn't in inflight yet, write a tombstone instead of silently returning;
register — check for the tombstone before creating a live context.

Should we go ahead with this approach?Also can it be taken as a follow-up given this PR needs to go in today?

ai-services/internal/pkg/worker/dispatch/dispatch.go
shwetha-s-poojary: Is the concern that a future handler might do` go func() { work(ctx) }(); return` and have its goroutine killed immediately?
I don't think any current handler does that AFAIK.
Should we add a  warning comment above handle(), so we don't use this pattern in future or is it something else?

mayuka-c: Yes lets mention it

ai-services/internal/pkg/worker/stream/sender.go
mayuka-c: Lets have a task created to keep track of this

ai-services/internal/pkg/worker/stream/sender.go
mayuka-c: Not sure, if I know of a best way to handle it as I couldnt spend much time analyze this. Lets have a task for this and see to handle it better next week or something?

ai-services/internal/pkg/worker/stream/sender.go
shwetha-s-poojary: Ok will create JIRA to track.

ai-services/internal/pkg/worker/stream/sender.go
shwetha-s-poojary: will add this in the JIRA to track.

yussufsh (approved):

PR #1444, opened 2026-09-10
fix: call uninstall worker as part of catalog uninstall

Shubhamag12:
Catalog uninstall was uninstalling worker pod since we were filtering via label: `ai-services.io/application: ai-services`. Ideally it should be handled when worker uninstall is called.

1. remove worker pod from the list of pods deleted as part of catalog uninstall.
2. invoking worker uninstall to remove worker pods now.
3. Added force flag in volume deletion. Reason: when catalog uninstall in invoked, it fails to delete podman auth secret as it is used by worker pod. Now force deleting it as there is no use of standalone worker pod now that catalog is uninstalled.

### DIFF
--- a/ai-services/Makefile
+++ b/ai-services/Makefile
@@ -1,6 +1,6 @@
 REGISTRY?=icr.io/ai-services-private
 IMAGE=ai-services
-TAG?=v0.0.349
+TAG?=v0.0.350
 CONTAINER_BUILDER?=podman
 CREDS_ARG := $(if $(and $(REGISTRY_USER),$(REGISTRY_PASSWORD)),--creds="$(REGISTRY_USER):$(REGISTRY_PASSWORD)")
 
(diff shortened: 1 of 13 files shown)


prachikurade1:
> Added force flag in volume deletion. Reason: when catalog uninstall in invoked, it fails to delete podman auth secret as it is used by worker pod. Now force deleting it as there is no use of standalone worker pod now that catalog is uninstalled.

Podman auth secret always being deleted was a defect, with my changes from this PR - https://github.com/IBM/project-ai-services/pull/1445 it will be deleted only if skip-cleanup flag is not passed.

Shubhamag12:
> > Added force flag in volume deletion. Reason: when catalog uninstall in invoked, it fails to delete podman auth secret as it is used by worker pod. Now force deleting it as there is no use of standalone worker pod now that catalog is uninstalled.
> 
> Podman auth secret always being deleted was a defect, with my changes from this PR - #1445 it will be deleted only if skip-cleanup flag is not passed.

This is different. I did not pass skip-cleanup flag.
And catalog was not able to delete podman-auth secret because it is consumed worker pod as well.

Shubhamag12:
closing this, not required anymore

PR #1446, opened 2026-09-10
Cover Connector and Sync logs schema as part of Import/Export APIs

manju956:
- Contains code changes for including connector and connector sync logs schema as part of Export and Import APIs
- Added only the method definitions for connector schema metadata import and export. 
- Existing definitions of import-export APIs are not modified to call the new methods on connectors as don't support backup and restore functionality for connectors yet. these methods will be called when we have cover it in the later release.

### DIFF
--- a/services/digitize/models.py
+++ b/services/digitize/models.py
@@ -199,8 +199,6 @@ class ImportExportData(BaseModel):
     """Shared payload structure for import/export APIs."""
     jobs: List["ExportJobRecord"] = Field(default_factory=list)
     documents: List["ExportDocumentRecord"] = Field(default_factory=list)
-    connectors: List["ExportConnectorRecord"] = Field(default_factory=list)
-    sync_logs: List["ExportSyncLogRecord"] = Field(default_factory=list)
 
 
 class ExportJobRecord(BaseModel):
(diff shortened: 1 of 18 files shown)


services/digitize/models.py
```diff
@@ -220,22 +223,65 @@ class ExportDocumentRecord(BaseModel):
...
     submitted_at: str
     completed_at: Optional[str] = None
     error: Optional[str] = None
     metadata: Dict[str, Any] = Field(default_factory=dict)
 
 
+class ExportConnectorRecord(BaseModel):
+    """Serializable connector record for export/import APIs.
+
+    connection_details is intentionally excluded — encrypted credentials
```
dharaneeshvrd, line 235: We need to document this behaviour properly.
Agree it's safe to avoid the security keys part of the export.
So with this connector syncing won't happen by default after importing it into a new deployment, they have to fix the connection details. But this would be really complicated for catalog backend I guess. Do we have plan to take care of this? @yussufsh @mayuka-c  @prachikurade1

dharaneeshvrd (commented):
@manju956 I still see traces of connector in existing import and export with empty list. Imo we can remove that too, let's keep the existing things untouched.

manju956:
> @manju956 I still see traces of connector in existing import and export with empty list. Imo we can remove that too, let's keep the existing things untouched.

Yeah, retained connectors in the export response with empty list. we could remove it as well

manju956:
> traces of connector in existing import and export with empty list. Imo we can remove that too, let's keep the existing things untouched.

removed the traces of connectors in existing import/export methods.

dharaneeshvrd:
@manju956 please update the description for this updated change

manju956:
> @manju956 please update the description for this updated change

done

dharaneeshvrd (approved):

PR #1452, opened 2026-09-11
use registered worker name case instead of lowering

yussufsh:
1. Always use user provided worker name.
2. Validation check for 3-64 chars and trim spaces.

### DIFF
--- a/ai-services/internal/pkg/catalog/apiserver/handlers/worker_handler.go
+++ b/ai-services/internal/pkg/catalog/apiserver/handlers/worker_handler.go
@@ -34,7 +34,7 @@ func NewWorkerHandler(reg *registry.Registry, repo repository.WorkerRepository, 
 
 // createWorkerReq is the request body for registering a new worker.
 type createWorkerReq struct {
-	WorkerName string `json:"worker_name" binding:"required,min=1,max=100"`
+	WorkerName string `json:"worker_name" binding:"required,max=64"`
 }
 
 // createWorkerResp is the response body for a newly registered worker.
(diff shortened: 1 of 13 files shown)


mayuka-c (dismissed):
LGTM

ai-services/internal/pkg/catalog/apiserver/handlers/worker_handler.go
```diff
@@ -66,19 +66,24 @@ func (h *WorkerHandler) CreateWorker(c *gin.Context) {
...
 		return
 	}
 
 	// "Local" is reserved for the catalog-machine worker registered by the
 	// configure flow. It is only allowed when LOCAL_WORKER=true, meaning this
 	// catalog instance is configured to host a co-located worker.
-	// Preserve the canonical casing so the DB row matches LocalWorkerName exactly.
+	// Normalise to the canonical casing so the DB row and token store always
+	// use "Local" regardless of how the operator typed it.
 	if strings.EqualFold(req.WorkerName, workerconstants.LocalWorkerName) {
```
mayuka-c, line 92: I hope for workerName local even if we use "local" it still works. Can we ensure we do case insensitive equality for it?

yussufsh: It will never be that case since it is using `workerconstants.LocalWorkerName` at both places: auto-join and here.

mayuka-c: I hope when we pass Local as worker name (client side) through API or CLI even there its taking care

mayuka-c (approved):

PR #1454, opened 2026-09-11
fix: WaitInferenceService method should only be called if a component is deploying isvc resource

mayuka-c:
- Currently if a component doesnt deploy isvc it simply checks and fails if overall application deployment doesnt depend on isvc.

### DIFF
--- a/ai-services/internal/pkg/catalog/utils/deployment_utils.go
+++ b/ai-services/internal/pkg/catalog/utils/deployment_utils.go
@@ -1,8 +1,10 @@
 package utils
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io/fs"
 	"sync"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


Shubhamag12 (approved):
lgtm

ai-services/internal/pkg/catalog/utils/deployment_utils.go
```diff
@@ -77,3 +79,31 @@ func RunConcurrently(
...
+
+// ChartHasInferenceService reports whether any template file under catalogPath
+// in fsys declares a KServe InferenceService (i.e. contains "kind: InferenceService").
+// It is used to decide whether WaitForInferenceServiceReady must be called after
+// a Helm install — Helm marks an InferenceService "Current" as soon as the CRD is
+// accepted, but KServe only sets Ready=True once the predictor pod is fully up.
+func ChartHasInferenceService(fsys fs.FS, catalogPath string) bool {
+	found := false
+
+	_ = fs.WalkDir(fsys, catalogPath, func(p string, d fs.DirEntry, err error) error {
```
yussufsh, line 91: What about BYOS here?

PR #1459, opened 2026-09-11
AIS-1808 - fix local worker join with skip-cleanup

prachikurade1:
Issue1: Worker rejoin after catalog uninstall with skip-cleanup
If user runs catalog uninstall with skip-cleanup and there is a local worker running, then subsequent runs of catalog configure, the local worker registration should be skipped (since it already exists in the db). 

Issue2: Worker configure re-run with mismatched skip-local flag
This is to ensure that the user does not attempt to change the skip-local-worker flag during re-runs of catalog configure

### DIFF
--- a/ai-services/internal/pkg/catalog/cli/configure/worker.go
+++ b/ai-services/internal/pkg/catalog/cli/configure/worker.go
@@ -3,6 +3,7 @@ package configure
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	catalogclient "github.com/project-ai-services/ai-services/internal/pkg/catalog/client"
 	catalogconstants "github.com/project-ai-services/ai-services/internal/pkg/catalog/constants"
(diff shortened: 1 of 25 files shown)


ai-services/internal/pkg/catalog/cli/configure/worker.go
```diff
@@ -39,4 +40,91 @@ func RegisterLocalWorker(ctx context.Context, c *catalogclient.Client) (string,
 	return resp.Token, nil
 }
 
+// CheckLocalWorkerSkip determines whether local-worker registration can be
+// skipped because valid on-disk TLS credentials already exist.
+//
+// It returns skip=true when BOTH conditions hold:
+//  1. The worker-mtls-encryption-secret is present — the TLS encryption key is
+//     on disk, so hasValidTLSCredentials will succeed inside the worker container.
```
mayuka-c, line 48: Didnt quite understand, if the encryption secret is present, how do we know the certs are present on volume?

mayuka-c: Both are completely diff, you could have secret present but certs might not be even present inside volume

ai-services/internal/pkg/catalog/cli/configure/worker.go
```diff
@@ -39,4 +40,91 @@ func RegisterLocalWorker(ctx context.Context, c *catalogclient.Client) (string,
...
+
+	return false, nil
+}
+
+// ResolveLocalWorkerToken decides whether to issue a new bootstrap token or
+// reuse existing credentials. It returns an empty token when skip conditions
+// are met (valid on-disk credentials + a completed catalog registration),
+// signalling to the caller that the worker will reconnect without re-registering.
+//
+// secretExists is rt.SecretExists from either PodmanClient or OpenshiftClient;
```
mayuka-c: WHat is rt.SecretExists?. Which secret is being tested whether it exists?

prachikurade1: rt.SecretExists is the runtime specific secret checking function. For this scenario, it is checking if the worker mtls secret exists or not.

ai-services/internal/pkg/catalog/cli/configure/worker.go
prachikurade1: This is just the worker-mtls-encryption secret. Do you want to add a check for worker-tls volume as well ?

ai-services/internal/pkg/catalog/cli/configure/worker.go
mayuka-c: Can we pls update the comments instead of it again mentioning rt.SecretExists?

prachikurade1: done

ai-services/internal/pkg/catalog/cli/configure/worker.go
mayuka-c: Yeah but lets say if volume is present but certs are not there, will it again recreate certs?
If not, then it will keep failing right?

prachikurade1: yes, both need to be present.

mayuka-c: I feel we might need to revisit later, I feel there is a gap if above occurs, then we need to ask user to uninstall completely and reinstall back for now I believe

mayuka-c (approved):
LGTM,

I hope you have validated all corner cases

Pls get approval from @yussufsh before merging

ai-services/internal/pkg/catalog/cli/configure/worker.go
```diff
@@ -39,4 +40,89 @@ func RegisterLocalWorker(ctx context.Context, c *catalogclient.Client) (string,
...
+	if !isReinstall {
+		return nil
+	}
+
+	workers, err := catalogclient.NewWorkerClientFromClient(c).ListWorkers(ctx)
+	if err != nil {
+		return fmt.Errorf("skip-local-worker validation: list workers: %w", err)
+	}
+
+	workersExist := len(workers) > 0
```
yussufsh: There may be more than 0 workers in the db which were previously connected.

ai-services/internal/pkg/catalog/cli/configure/worker.go
```diff
@@ -39,4 +40,89 @@ func RegisterLocalWorker(ctx context.Context, c *catalogclient.Client) (string,
...
+
+	// workersExist == true  → original flag was false (local worker was joined)
+	// workersExist == false → original flag was true  (local worker was skipped)
+	if skipLocalWorker == workersExist {
```
yussufsh: We might also need to check if first time skip was given and second time it was not passed (no local entry in the db).

prachikurade1: Image

ai-services/internal/pkg/catalog/cli/configure/worker.go
prachikurade1: done

yussufsh (approved):
lgtm

PR #1461, opened 2026-09-11
close grpc channel when worker is deregistered

yussufsh:
Close the grpc channel when the remote worker is deregistered.


mayuka-c (approved):
LGTM

PR #1463, opened 2026-09-12
Bump cryptography and pillow

yussufsh:


adarshagrawal38:
Similar changes will be applicable for `llmlite` and `tools` image

adarshagrawal38:
CVE scan is falling, due to image build is falling, we need to fix this flaking error.

yussufsh:
> Similar changes will be applicable for `llmlite` and `tools` image

Then I will fix the base-image for now and then we can fix all the services at once.

adarshagrawal38 (approved):

PR #1464, opened 2026-09-12
Make use of the latest base image

yussufsh:


ai-services/internal/pkg/vars/var.go
```diff
@@ -15,7 +15,7 @@ var (
 var (
 	// SpyreCardAnnotationRegex -> ai-services.io/<containerName>--spyre-cards.
 	SpyreCardAnnotationRegex = regexp.MustCompile(`^ai-services\.io\/([A-Za-z0-9][-A-Za-z0-9_.]*)--spyre-cards$`)
-	ToolImage                = "icr.io/ai-services/tools:0.11"
+	ToolImage                = "icr.io/ai-services/tools:0.12"
```
adarshagrawal38:
```suggestion
	ToolImage                = "icr.io/ai-services-cicd/tools:0.12"
```

image is not yet promoted to public namepsace.

adarshagrawal38:
separate PR for litellm?

yussufsh:
> separate PR for litellm?

I don't see any changes needed for LiteLLM, as it uses its own pip and venv. What issue are we looking at?

adarshagrawal38 (approved):

PR #1465, opened 2026-09-12
fix: digitize Containerfile pip not found

yussufsh:
Fix for pip not found error due to recent changes.
Also, updating only the digitize version since other Containerfile are just comment changes.


yussufsh:
@adarshagrawal38 the build tests are not running if I change any service code. Let's fix that.

adarshagrawal38 (approved):
LGTM

PR #1466, opened 2026-09-15
Minor worker related changes

pranithraoibm:
- Remove PowerVS option displayed in the StepOne of the deploy flows
- Add inline notification to show server errors if worker register fails in the deploy flow's worker register flow
- Pass Local as well in the resources API call


ui/catalog/src/components/DeployFlow/Shared/steps/SharedStepOne.tsx
```diff
@@ -194,8 +199,12 @@ export const SharedStepOne = ({
           gatewayAddress: result.gateway_address,
         },
       });
-    } catch {
-      dispatchRegister({ type: "SET_PHASE", payload: "error" });
+    } catch (err) {
+      dispatchRegister({
+        type: "ERROR",
+        payload:
+          err instanceof Error ? err.message : "Failed to register worker",
```
ryanmarc: Good change to make, but `err.message` on an axios rejection is just
`"Request failed with status code 400"`. The message the backend actually sent is in
`err.response.data.error`, and it's dropped here.

I checked what the user ends up seeing by serving the real handler's response body
(`400 {"error":"worker name must be between 3 and 64 characters"}`) and running this expression
against it with the repo's own axios:

```
subtitle shown:  "Request failed with status code 400"
backend sent:    "worker name must be between 3 and 64 characters"
```

That matters here more than usual, because nearly all of this endpoint's failures are the
actionable kind (`worker_handler.go:70-117`): name length, `worker name "local" is reserved`,
`invalid payload`. Those are exactly the cases where the notification will show the status code
instead of the reason, which leaves the user stuck at the thing the change was meant to unstick.

The repo already has a pattern for this in five or so places
(`DataSourceConnectorsTable.tsx:199`, `deployError.ts:14`, `DeploymentDetails.tsx:361`, others):

```
import { isAxiosError } from "axios";
const message =
  (isAxiosError(err) && typeof err.response?.data?.error === "string"
    ? err.response.data.error
    : err instanceof Error ? err.message : null) ?? "Failed to register worker";
```

One thing you may want to check while you're in here: the reserved-name check is case-insensitive
server-side (`strings.EqualFold(req.WorkerName, "Local")`), so `local` and `LOCAL` are rejected
too. Someone typing `local` in this modal and getting "Request failed with status code 400" has no
way to work out why.

ryanmarc (changes requested):

ui/catalog/src/components/DeployFlow/Shared/steps/SharedStepOne.tsx
pranithraoibm: Fixed it using the `deployError` util we had

image

ryanmarc (approved):

PR #1468, opened 2026-09-15
Fix [AISERVICES-1748]: add support for services without providers.

Niranjan-J-Joshi:
## Summary

This change adds support for deploying services that do not have provider components, as well as services that use service-level schemas. The deploy flow now handles all supported service configurations, including services with providers, services without providers, and services that expose schema parameters directly.

The update introduces service-level schema fetching, validation of required schema fields before deployment, support for deployment payloads without components, improved error handling for schema loading failures, and automatic population of schema default values. It also updates the deployment payload structure to align with the latest backend API contract and improves the UI by clearly highlighting missing required configuration fields.
### API & Store
- Added `fetchServiceSchemaParams()` to fetch service-level schemas.
- Added `serviceSchemas` state for storing schema data, loading, and errors.
- Clears schema state on reset.

### Services Deploy Flow
- Fetches service-level schemas when available.
- Supports services with no providers.
- Validates required schema fields before enabling Next/Deploy.
- Shows schema fetch errors in the UI.
- Supports deployment payloads with `components: []` and top-level params.

### Digital Assistant Deploy Flow
- Added support for services with no providers and/or service-level schemas.
- Auto-populates default schema values when available.
- Validates required schema fields before enabling Deploy.
- Updated deployment payload format to match the latest API contract.

### Shared Updates
- `ServiceConfigCard` now highlights missing required fields.
- Added `schema` support in deploy option types.


ui/catalog/src/components/DeployFlow/DigitalAssistant/utils/digitalAssistantDeploymentTransform.ts
```diff
@@ -180,9 +208,7 @@ export function transformToDeploymentPayload(
     };
 
     if (Object.keys(serviceBackendParams).length > 0) {
-      deploymentService.params = {
-        backend: serviceBackendParams,
-      };
+      deploymentService.params = serviceBackendParams;
```
ryanmarc: This changes the wire payload for digital assistant deploys that already work today. On main this is params = { backend: serviceBackendParams }; here the backend level is gone.

Everywhere else, I see `backend` as the wrapper. Was this intentional, Is there a corresponding backend change that supports this?

ui/catalog/src/components/DeployFlow/Services/hooks/useServiceDeployOptions.ts
```diff
@@ -231,6 +267,23 @@ export const useServiceDeployOptions = (
     // --- Path B: deploy options cached — retry only errored models on reopen ---
     if (!deployOptions) return;
 
+    // Retry schema fetch if it previously errored — mirrors the DA hook's hasError pattern.
+    const schemaError =
+      storeState.serviceSchemasError[`${serviceId}:${runtime}`];
+    if (schemaError && deployOptions.schema) {
```
ryanmarc: I think this leaves the flow permanently stuck for a fairly ordinary case, and silently.

The schema fetch only happens inside Path A (the not-cached branch). Path B retries it only when `serviceSchemasError` has an entry. But `partialize` persists `serviceDeployOptions` and `serviceSchemas` and not `serviceSchemasError` (`serviceDeploy.store.ts:489-496`), so after a reload you can be holding cached deploy options with a `schema` URL, no schema, and no error. Path A is skipped because `deployOptions` is truthy, Path B is skipped because there's no error, and nothing fetches.

`ServicesDeployFlow.tsx:224` then bails on every render (`if (deployOptions.schema && !serviceSchema && !serviceSchemaError) return;`), so the form never initializes. And it's invisible: `shellIsLoading` is false (nothing in flight) and `shellError` is null (no error recorded), so you get a blank step with a greyed Next and no indication why.

I seeded a fake `localStorage` with exactly what `partialize` writes, imported the real store, let it rehydrate, and evaluated the gates verbatim:

```
deployOptions.schema = "/api/v1/catalog/services/chat/schema"
serviceSchema = null, serviceSchemasError[key] = null
Path A taken? false | Path B retry taken? false | init guard bails? true
shellIsLoading false | shellError null
```

The bit that worries me most is that it doesn't need anything to fail first. `fetchServiceDeployOptions` returns `response.data` as-is and the store persists that object verbatim, so anyone who has used the Services flow before this lands already has `schema` sitting in their cached entry with no matching `serviceSchemas` entry. That's the dead state on first open after deploy, no error required.

Simplest fix I can see: make Path B fetch the schema whenever `deployOptions.schema` is set and there's no schema cached, rather than only when an error is recorded. Something like `if (deployOptions.schema && !getServiceSchema(serviceId, runtime))`, which covers the errored case too and makes the persisted-error question moot. Bumping the persist `version` to drop stale entries would also work for the upgrade case but wouldn't fix the errored-then-reloaded one.

ryanmarc (changes requested):

ui/catalog/src/components/DeployFlow/DigitalAssistant/utils/digitalAssistantDeploymentTransform.ts
Niranjan-J-Joshi: fixed it by adding a generic nestParamsBySchema() function. Instead of hardcoding the backend wrapper, it reads the service schema and wraps parameters under the correct key.
Some services require a wrapper (like backend) and some don't, so this approach supports both cases.

ui/catalog/src/components/DeployFlow/Services/hooks/useServiceDeployOptions.ts
Niranjan-J-Joshi: Updated Path B to fetch whenever deployOptions.schema is set and !getServiceSchema(serviceId, runtime).

ui/catalog/src/components/DeployFlow/DigitalAssistant/DigitalAssistantDeployFlow.tsx
```diff
@@ -287,12 +287,34 @@ export const DeployFlow = ({
...
+  const areAllRequiredFieldsFilled = useMemo(() => {
+    if (!deployOptions) return true;
+    const runtimePrefix = `${runtime}:`;
+    return deployOptions.services.every((service) => {
+      const serviceConfig = state.formData.services[service.id];
+      if (!serviceConfig?.enabled) return true;
+      const schema = serviceParams[`${runtimePrefix}${service.id}`]?.data;
+      if (!schema?.required) return true;
+      return schema.required.every((fieldKey: string) => {
+        const value = serviceConfig.params?.[fieldKey];
```
ryanmarc, line 321: I think this gate cannot pass on a wrapped service schema. schema.required here is the top-level array, which for the wrapped shape is ["backend"], but serviceConfig.params is keyed by leaf name (systemPrompt, maxTokens), which is the whole reason nestParamsBySchema had to be added in this PR. So the lookup is for a key that is never present.

The opposite case is worth noting too: a schema with no top-level required short-circuits at line 300, so leaf fields marked required inside the wrapper are never checked.

ui/catalog/src/components/DeployFlow/DigitalAssistant/utils/digitalAssistantDeploymentTransform.ts
```diff
@@ -16,6 +16,50 @@ import type {
...
 
+/**
+ * Re-nests a flat params object to match the structure of the given JSON Schema.
+ */
+function nestParamsBySchema(
+  flatParams: Record<string, unknown>,
+  schema: ProviderSchema | null | undefined,
+): Record<string, unknown> {
+  if (!schema?.properties || Object.keys(flatParams).length === 0) {
+    return flatParams;
```
ryanmarc, line 27: This early return is the one case where the new nesting can still send the old wrong shape. When schema is missing, nestParamsBySchema returns the params flat, but main wrapped them under backend unconditionally, so the server gets systemPrompt instead of backend.systemPrompt.

I do not think this needs a race to hit. serviceSchemas is built from the serviceParams cache at deploy time keyed by runtime:serviceId, and the Deploy button does not wait for those fetches: isProviderParamsLoading only counts global-component pairs, which the comment at useDeployOptions.ts:247-251 says explicitly. Fill the system prompt on podman, switch the tile to openshift (the params survive, handleUpdateFormData shallow-merges and the init effect is guarded by hasInitialized), then deploy before the openshift schema lands. A schema fetch that just fails gets there with no timing involved.

ryanmarc (changes requested):

ui/catalog/src/components/DeployFlow/DigitalAssistant/DigitalAssistantDeployFlow.tsx
Niranjan-J-Joshi: Fixed. Instead of using `schema.required` directly, it now uses `getRequiredFieldKeys(schema)` to get the actual required parameter names. This handles both wrapped and flat schemas.

ui/catalog/src/components/DeployFlow/DigitalAssistant/utils/digitalAssistantDeploymentTransform.ts
Niranjan-J-Joshi: Fixed. Deploy button will be disabled while a service schema is loading. If a service has a schema URL but the schema hasn't been fetched yet, the Deploy button stays disabled. If the schema request fails, the existing error handling displays the error. This prevents params from being sent in the wrong format.

ryanmarc (dismissed):

mkumatag (approved):

PR #1469, opened 2026-09-15
fix: honour livenessProbe timing fields in container readiness check

mkumatag:
Two bugs caused the readiness poller to fail healthy containers:

1. WaitForContainerReadiness enforced the deadline while Podman's health status was still 'starting' (inside initialDelaySeconds). No probe has run during that window so timing out is incorrect. Fix: skip the deadline check when healthStatus == constants.Starting.

2. doContainerReadinessCheck computed the timeout as startPeriod + 2 min (hardcoded), ignoring periodSeconds and failureThreshold declared in the pod spec. For vllm-cpu this gave a 4-minute budget against a spec-mandated 8-minute window (failureThreshold 12 × periodSeconds 30 s). Fix: derive both the timeout and the poll interval from the LivenessProbe fields on the parsed PodSpec using: timeout = initialDelaySeconds + (failureThreshold × periodSeconds) + 30s buffer pollInterval = periodSeconds

Additional cleanup:
- Remove now-unused FetchContainerStartPeriod helper.
- Rename inspectPollInterval → defaultInspectPollInterval (used as fallback when probe.PeriodSeconds is zero or the probe is absent).

Observed failure: container llm-5c32c9e93e-llm was healthy (GET /health returning 200 OK) but deployment reported 'operation timed out waiting for container readiness' because the 4-minute deadline fired while Podman was still in the 'starting' state, 18 minutes before the first unhealthy probe would have appeared.


ai-services/internal/pkg/cli/helpers/helper.go
```diff
@@ -36,20 +46,24 @@ func WaitForContainerReadiness(ctx context.Context, runtime runtime.Runtime, con
...
 			return nil
 		}
 
-		// if deadline exceeds, stop the container readiness check
-		if time.Now().After(deadline) {
+		// While Podman is inside the initialDelaySeconds window it reports "starting":
+		// no probe has executed yet so there is nothing meaningful to time out on.
+		// Only enforce the deadline once Podman has transitioned out of "starting"
+		// (i.e. the container is "unhealthy" or in an unknown state).
+		if healthStatus != string(constants.Starting) && time.Now().After(deadline) {
```
mayuka-c: Can we check for deadline irrespective of `healthStatus != string(constants.Starting)`. Since the deadline already accounts for initialDelaySeconds, I feel we need not have to check for health status not to be in starting.

If there is a problem with health status and if it stays with "Starting", then this might hang forever

ai-services/internal/pkg/cli/podman/deploy.go
```diff
@@ -78,39 +85,81 @@ func doContainersCreationCheck(ctx context.Context, rt runtime.Runtime, podSpec
...
+// or nil if no probe is defined or the container is not found (e.g. the infra container).
+func fetchLivenessProbe(podSpec *models.PodSpec, containerName string) *v1.Probe {
+	if podSpec == nil {
+		return nil
+	}
+	for _, c := range podSpec.Spec.Containers {
+		if c.Name == containerName {
+			return c.LivenessProbe
+		}
+	}
```
mayuka-c, line 138: I see the containerName is computed by inspecting the running container (which will have podname prefixed with it along with InstanceSlug). But the podSpec would give the containername present within the template which will not match with the running container name right?

I feel we might  need to get the podname also with running pod and concat with `c.Name` and do equality check

ai-services/internal/pkg/cli/helpers/helper.go
mkumatag: Good catch. Addressed in the latest commit.

Rather than skipping the deadline for `"starting"`, the logic is now simpler and cleaner:

- `"unhealthy"` → return error **immediately** (a conclusive probe failure, no point waiting further).
- `"starting"` → keep polling against the **single hard deadline** (= the full `timeout` computed from `initialDelaySeconds + failureThreshold × periodSeconds + buffer`). Since the timeout already accounts for the entire `initialDelaySeconds` window, the deadline will correctly fire if the container is permanently stuck in `"starting"` — no infinite loop possible.

ai-services/internal/pkg/cli/podman/deploy.go
mkumatag: You are correct — confirmed from the logs: the actual running container name is `llm-5c32c9e93e-llm`, while the pod spec has `name: llm`. Podman formats it as `<podname>-<specname>`.

Fixed in the latest commit: `fetchLivenessProbe` now strips the pod name prefix (`podSpec.Name + "-"`) from the running container name before matching against the spec containers. Tests updated to use the correct Podman-format names (`llm-test1234ab-llm` instead of `llm`).

mkumatag:
@mayuka-c fixed the review comments and tested the code as well, it is working fine.. please take a look at the recent changes.

Will bump the version in the end.

mayuka-c:
> @mayuka-c fixed the review comments and tested the code as well, it is working fine.. please take a look at the recent changes.
> 
> Will bump the version in the end.

Thanks @mkumatag , LGTM.

mayuka-c (dismissed):
LGTM

mayuka-c (approved):

PR #1474, opened 2026-09-16
Handle connector secrets via AllowList

sats-23:
-We are removing the unused password field for SSH Scanner
-Updating example to reflect the same
-Inverting to an AllowList based approach where only fields mentioned are allowed to be returned back in plaintext
-Everything else is treated as a secret
-Strict forbid policy (fail request) for unsupported fields
-Adding strict validation on payload (400 Bad request)


services/digitize/api/v1/connectors.py
```diff
@@ -37,10 +38,11 @@
     SyncLogStatus,
     SyncTriggerResponse,
 )
+from digitize.connectors.scanners.config import S3ConnectorConfig, SSHConnectorConfig
```
dharaneeshvrd: why not move it to models ?

yussufsh (dismissed):
lgtm
Thanks @sats-23 @dharaneeshvrd

dharaneeshvrd (dismissed):

services/digitize/api/v1/connectors.py
```diff
@@ -195,10 +197,30 @@ async def update_connector(connector_id: str, body: ConnectorUpdateRequest):
...
+                    )
+                elif existing.type == "object_storage":
+                    S3ConnectorConfig.model_validate(
+                        {**decrypted_existing, **body.connection_details}
+                    )
+            except ValidationError as exc:
+                APIError.raise_error(
+                    ErrorCode.INVALID_REQUEST,
+                    f"Invalid connection_details for connector type {existing.type!r}: {exc}",
+                )
```
mkumatag, line 229: **🔴 Critical — `ValidationError` error message may leak decrypted secret values in the 400 response body**

`ValidationError.__str__()` includes the **input values that failed validation** by default. Because `decrypt_secrets` runs before `model_validate` and its output is merged into the dict being validated, a validation failure on a field like `private_key` or `secret_access_key` will include the **decrypted plaintext secret** in the HTTP 400 response returned to the caller.

Fix by suppressing input values in the error serialization:

```python
except ValidationError as exc:
    APIError.raise_error(
        ErrorCode.INVALID_REQUEST,
        f"Invalid connection_details for connector type {existing.type!r}: "
        + str(exc.errors(include_input=False)),
    )
```

services/digitize/connectors/encryption.py
```diff
@@ -21,11 +33,14 @@
...
-# and stripped before API responses.
-_SECRET_FIELDS: dict[str, set[str]] = {
-    "file_system": {"private_key"},
-    "object_storage": {"secret_access_key"},
+# Fields that are safe to store in plaintext and safe to return in API
+# responses for each connector type.  Every field NOT listed here is treated
+# as a secret: encrypted at rest and stripped from responses.
+_SAFE_FIELDS: dict[str, frozenset[str]] = {
+    "file_system": frozenset({"host", "port", "username", "remote_path"}),
+    "object_storage": frozenset({"bucket_name", "access_key_id", "endpoint_url",
```
mkumatag, line 41: **🟡 Bug — `allowed_extensions` is missing from `_SAFE_FIELDS["file_system"]`**

`SSHConnectorConfig` declares an `allowed_extensions` field (config.py), and `from_connection_details` injects it into the payload via `payload["allowed_extensions"] = allowed_extensions`. If that field is present in the `file_system` JSONB column it will be **AES-encrypted at write time and silently stripped from all GET responses** — causing confusing missing-data behaviour with no error raised.

The `object_storage` entry already includes `"allowed_extensions"` on line 43. Unless there is a confirmed architectural decision that `file_system` connectors never store `allowed_extensions` in JSONB, add it here too:

```python
"file_system": frozenset({"host", "port", "username", "remote_path", "allowed_extensions"}),
```

services/digitize/api/v1/connectors.py
```diff
@@ -195,10 +197,30 @@ async def update_connector(connector_id: str, body: ConnectorUpdateRequest):
...
+                if existing.type == "file_system":
+                    SSHConnectorConfig.model_validate(
+                        {**decrypted_existing, **body.connection_details}
+                    )
+                elif existing.type == "object_storage":
+                    S3ConnectorConfig.model_validate(
+                        {**decrypted_existing, **body.connection_details}
+                    )
+            except ValidationError as exc:
```
mkumatag, line 224: **🟡 Bug — Unknown connector type silently skips validation**

If `existing.type` is neither `"file_system"` nor `"object_storage"` the `if/elif` falls through with no validation and no error. This directly contradicts the PR's closed-by-default principle and means a future connector type would be silently unvalidated on updates.

Add an `else` branch:

```python
else:
    APIError.raise_error(
        ErrorCode.INVALID_REQUEST,
        f"Schema validation not supported for connector type {existing.type!r}",
    )
```

The same gap also exists in `validate_connection_details_schema` in `models.py` (though that one is guarded earlier by `validate_type`).

services/digitize/api/v1/connectors.py
```diff
@@ -195,10 +197,30 @@ async def update_connector(connector_id: str, body: ConnectorUpdateRequest):
...
-        # encrypted details so untouched keys stay encrypted and intact.
+        # If connection_details is being updated, validate the partial payload
+        # against the typed config for this connector type before storing anything.
+        # Decrypt the existing stored details first — the DB holds ciphertext for
+        # secret fields, which would fail PEM / format validators if merged raw.
         merged_details: Optional[dict] = None
         if body.connection_details is not None:
+            try:
+                from digitize.connectors.encryption import decrypt_secrets
+                decrypted_existing = decrypt_secrets(
```
mkumatag, line 208: **🟢 Style — Deferred inline import; inconsistent with the rest of the file**

`encrypt_secrets`, `merge_and_encrypt_partial`, and `safe_connection_details` are all imported at the module level (lines 42–46). This single deferred `from digitize.connectors.encryption import decrypt_secrets` inside the function body is inconsistent and may suggest to future maintainers that there's a circular-import reason to defer it (there isn't).

Move it to the top-level import block:

```python
from digitize.connectors.encryption import (
    decrypt_secrets,
    encrypt_secrets,
    merge_and_encrypt_partial,
    safe_connection_details,
)
```

services/digitize/connectors/models.py
```diff
@@ -127,13 +127,33 @@ class ConnectorCreateRequest(BaseModel):
...
 
+    @model_validator(mode="after")
+    def validate_connection_details_schema(self) -> "ConnectorCreateRequest":
+        """Validate connection_details against the typed config for the connector type.
+
+        Ensures that invalid fields (e.g. 'password' on a file_system connector)
+        are rejected at the API boundary with a clear 422, before anything is
+        stored or encrypted.
+        """
+        from digitize.connectors.scanners.config import S3ConnectorConfig, SSHConnectorConfig
```
mkumatag, line 145: **🟡 Bug — `except Exception` is too broad; masks real programming errors as user-input failures**

This catch swallows `ImportError`, `AttributeError`, and any other unexpected exception from the config validators, re-raising them all as `ValueError("Invalid connection_details...")`. A real bug inside a validator would appear to users as a confusing 422 with a misleading message instead of a 500.

Since Pydantic config validators only raise `ValidationError`, narrow the catch:

```python
from pydantic import ValidationError
...
except ValidationError as exc:
    raise ValueError(
        f"Invalid connection_details for connector type {self.type!r}: {exc}"
    ) from exc
```

This is consistent with the explicit `from pydantic import ValidationError` already present at the top of `connectors.py`.

services/digitize/connectors/models.py
```diff
@@ -127,13 +127,33 @@ class ConnectorCreateRequest(BaseModel):
...
         }
     }
 
+    @model_validator(mode="after")
+    def validate_connection_details_schema(self) -> "ConnectorCreateRequest":
+        """Validate connection_details against the typed config for the connector type.
+
+        Ensures that invalid fields (e.g. 'password' on a file_system connector)
+        are rejected at the API boundary with a clear 422, before anything is
+        stored or encrypted.
```
mkumatag, line 143: **🟢 Style — No `else` branch for unknown connector type in the model validator**

In practice `validate_type` (a `field_validator`) runs before this `model_validator` and already rejects unknown types, so the `elif` fall-through here is unreachable. However, Pydantic validator ordering is an implementation detail. A short comment or a defensive `else` makes the invariant explicit and prevents a future refactor from silently opening a gap:

```python
else:
    # validate_type rejects unknown types before this point — unreachable,
    # but kept explicit to guard against future refactoring.
    raise ValueError(
        f"No schema validator registered for connector type {self.type!r}"
    )
```

mkumatag (changes requested):
## Review: Handle connector secrets via AllowList

The allowlist inversion is the correct security model — the old denylist silently leaked any credential field whose name wasn't on the list. The new design is fail-closed by default, `extra="forbid"` on both config models makes the schema the authoritative contract, and the test coverage is excellent (especially `test_unknown_connector_type_returns_empty_dict` which directly validates the closed-by-default guarantee).

**Requesting changes** for one critical security issue and a few medium/low items, all detailed as inline comments:

| # | File | Severity | Issue |
|---|------|----------|-------|
| 1 | `connectors.py` line ~222 | 🔴 Critical | `ValidationError.__str__()` includes input values — decrypted secrets leak into the 400 response body |
| 2 | `encryption.py` line ~40 | 🟡 Medium | `allowed_extensions` missing from `_SAFE_FIELDS["file_system"]` — would be encrypted and stripped silently |
| 3 | `connectors.py` line ~211 | 🟡 Medium | No `else` branch for unknown connector type in `update_connector` — validation silently skipped |
| 4 | `models.py` line ~151 | 🟡 Medium | `except Exception` too broad in model validator — masks real programming errors as user-input failures |
| 5 | `connectors.py` line ~207 | 🟢 Low | Deferred inline import of `decrypt_secrets` — inconsistent with other top-level imports |
| 6 | `models.py` line ~147 | 🟢 Low | No `else` branch for unknown type in model validator — add comment or defensive raise |

**What's working well:**
- `frozenset` for `_SAFE_FIELDS` — immutable and O(1) membership checks ✅
- `decrypt_secrets` before `model_validate` in `update_connector` is correct — avoids false-positive PEM validation failures on ciphertext ✅
- Removing `password` from the SSH example in `models.py` closes a documentation-driven misconfiguration path ✅
- Test suite correctly inverts the old `test_returns_all_fields_for_unknown_connector_type` into a security-asserting empty-dict test ✅

services/digitize/api/v1/connectors.py
sats-23: Fixed

services/digitize/connectors/encryption.py
sats-23: Fixed

services/digitize/api/v1/connectors.py
sats-23: Fixed

services/digitize/api/v1/connectors.py
sats-23: Done

services/digitize/connectors/models.py
sats-23: Fixed

services/digitize/connectors/models.py
sats-23: Fixed

mkumatag (approved):

mkumatag (approved):

PR #1475, opened 2026-09-16
AISERVICES-1816: fix: wait for Caddy TLS ready before catalog login

shwetha-s-poojary:
RegisterCatalogRoutes triggers a Caddy config reload that briefly interrupts TLS on port 443, causing tls: internal error if login is attempted immediately after. The Caddy liveness probe only checks HTTP port 8080, so the container appears healthy while TLS is still reloading.

Added WaitForHTTPSReady to the caddy package, which polls the catalog HTTPS endpoint until the TLS handshake succeeds before LoginToCatalog is called.


ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -126,9 +126,16 @@ func handlePostDeployment(ctx context.Context, caddyCtx *caddy.Context, deployCt
 		return fmt.Errorf("route registration failed: %w", err)
 	}
 
+	// Wait for Caddy's TLS layer to be ready before attempting login.
+	// RegisterCatalogRoutes triggers a Caddy config reload that briefly interrupts
+	// TLS, causing "tls: internal error" if login is attempted immediately.
+	catalogAPIURL := routeURLs[catalogconstants.CatalogAPIRouteKey]
+	if err := caddy.WaitForHTTPSReady(ctx, catalogAPIURL); err != nil {
+		return fmt.Errorf("caddy TLS readiness check failed: %w", err)
```
mayuka-c: @shwetha-s-poojary This can be made generic. As soon as we register a route, let us ensure we make curl call to it (just the raw endpoint). If that works, rest should work

mayuka-c: Atleast in CLI flow I feel we can make it generic during register routes and within a pod it might not be possible to test if custom domain is used (we would need host alias and we might not be able to test each application routes)

ai-services/internal/pkg/catalog/cli/common/podman/caddy/routes.go
```diff
@@ -2,18 +2,80 @@ package caddy
...
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec // self-signed certs are expected
+		},
+	}
+
+	deadline := time.Now().Add(httpsReadyTimeout)
+	ticker := time.NewTicker(httpsReadyPollInterval)
+	defer ticker.Stop()
+
+	for {
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, httpsURL, nil)
```
yussufsh: Let's use `tls.DialWithDialer` instead of GET service call. And make it a util func which can be use from anywhere in future as soon as the route is registered for quick access.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
shwetha-s-poojary: As discussed with @yussufsh , not going ahead with generic approach, as making it generic would require us to add all route entries to /etc/hosts

shwetha-s-poojary:
Tested:
```
Worker node setup complete.
worker joined successfully.
Next Steps:
-------
- Access the Catalog UI at https://catalog-ui.example.com

- Access the Catalog Backend at https://catalog-api.example.com

```

ai-services/internal/pkg/catalog/cli/common/podman/caddy/routes.go
```diff
@@ -2,18 +2,80 @@ package caddy
...
 )
 
+const (
+	httpsReadyPollInterval   = 2 * time.Second
+	httpsReadyTimeout        = 60 * time.Second
+	httpsReadyRequestTimeout = 5 * time.Second
+)
+
+// WaitForTLSReady polls rawURL via TLS handshake until Caddy finishes reloading TLS for the domain.
+func WaitForTLSReady(ctx context.Context, rawURL string) error {
```
yussufsh: Please create a new file named tls.go under the same package and move all the new code into it.

yussufsh (approved):

PR #1477, opened 2026-09-16
Minor CLI output helper fixes

mayuka-c:
- Fix next steps output in application create as `info` command shouldnt take runtime
- Also fixes `templates parameters` command CLI output to include appropriate runtime along with ai-services prefix.


yussufsh (approved):
lgtm
Approving since it is a small log message change for CLI o/p.

Shubhamag12 (approved):
lgtm

PR #1480, opened 2026-09-16
fix: Add minReadySeconds to Catalog Backend Pod

mayuka-c:
- Adding small readiness delay for catalog backend pod to allow ingress controller to pick it up.

Searched online and got below response if it is good practice to do it.
```
Yes, adding minReadySeconds is a well-established, industry-accepted best practice for handling the asynchronous routing lag in OpenShift and regular Kubernetes clusters using external Ingress controllers.
```

Fixes: https://jsw.ibm.com/browse/AISERVICES-1810


mayuka-c:
Tried multiple times redeploying and working fine

ai-services/assets/catalog/openshift/templates/catalog-backend.yaml
```diff
@@ -135,9 +136,9 @@ spec:
             httpGet:
               path: /health
               port: 8080
-            initialDelaySeconds: 10
-            periodSeconds: 30
-            timeoutSeconds: 5
+            initialDelaySeconds: 3
+            periodSeconds: 5
+            timeoutSeconds: 2
```
Shubhamag12, line 141: reason of reducing these values is adding `minReadySeconds` or something else?

mayuka-c: Yes once the pod is Ready, it will wait for 10s and the next readiness probe need not have to wait for 30s, can be checked immediately, so I have reduced the timings here and its considered to be good practice, so I have kept the same.

Shubhamag12 (approved):
lgtm

yussufsh (approved):
lgtm

PR #1485, opened 2026-09-17
Make Images List API unsupported for Openshift

mayuka-c:
- Currently the Images API are being consumed by `application images` CLI command which is being used for air-gapped env. We already are making `application images` CLI unsupported for Openshift.
- So currently ensuring the API responds with proper error with appropriate Status codes in case of Openshift.


yussufsh (dismissed):
lgtm

Shubhamag12 (dismissed):

Shubhamag12 (approved):

PR #1486, opened 2026-09-17
Adding ja_core_news_sm to approved license pkg list

adarshagrawal38:
# Summary
The license scan was failing due to `ja_core_news_sm` being licensed under `CC-BY-SA-4.0`, which is denied by default.

Since we have received approval to use this package, it has been added to the approved package list to allow the license scan to pass.

# Changes
- Added `ja_core_news_sm` to approved_pkg.json


yussufsh (approved):
lgtm

PR #1488, opened 2026-09-17
AIS-1802 - fix issue with ssl certs during re-configure

prachikurade1:


ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -156,6 +191,45 @@ func handlePostDeployment(ctx context.Context, caddyCtx *caddy.Context, deployCt
...
+// domain from the previous install but ComputeDomainSuffix had no cert paths
+// to work with and would have fallen back to the host IP.
+// Errors are non-fatal — the caller logs and continues with the host-IP domain.
+func recoverDomainFromCaddy(ctx context.Context, caddyCtx *caddy.Context) error {
+	proxyManager, err := caddyCtx.CreateProxyManager(ctx)
+	if err != nil {
+		return err
+	}
+
+	route, err := proxyManager.GetRouteByID(ctx, "catalog-api")
```
shwetha-s-poojary: Do we already a have constant for "catalog-api"? Use that if we do, we can define one if not.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -156,6 +191,45 @@ func handlePostDeployment(ctx context.Context, caddyCtx *caddy.Context, deployCt
...
+// When useExistingCert is true (the cert secret was preserved by --skip-cleanup
+// and no new paths were supplied), a non-empty sentinel is returned so the caddy
+// template renders the volume mount; the existing secret provides the actual bytes.
+func resolveCertPaths(certPath, keyPath string, useExistingCert bool) (string, string) {
+	if useExistingCert {
+		return existingCertSentinel, existingCertSentinel
+	}
+
+	return certPath, keyPath
+}
```
shwetha-s-poojary, line 237: We are passing and returning the same certPath and keyPath if ! useExistingCert 
Can we simplify this? IMO we can add a small if block instead of this function

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
prachikurade1: Done

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
prachikurade1: This was extracted out to fix lint function complexity issue. So only the part which is additional with minimal changes to the main function was extracted to a new function.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -156,6 +197,45 @@ func handlePostDeployment(ctx context.Context, caddyCtx *caddy.Context, deployCt
...
+	if err != nil {
+		return err
+	}
+
+	// route.Domain is the full host (e.g. "catalog-api.powervm-spyre-pok.cis.ibm.net").
+	// Strip the "catalog-api." prefix to get the domain suffix.
+	const prefix = catalogAPIRouteID + "."
+	if strings.HasPrefix(route.Domain, prefix) {
+		caddyCtx.SetDomainSuffix(strings.TrimPrefix(route.Domain, prefix))
+	}
```
mayuka-c, line 220: If we only need to get the domain, cant we not get it using regex?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -19,6 +20,17 @@ import (
 	"github.com/project-ai-services/ai-services/internal/pkg/utils"
 )
 
+// existingCertSentinel is a non-empty placeholder passed into sslCertContent/sslKeyContent
+// when the cert secret already exists from a previous run. It triggers the template
+// volume-mount guard without embedding real cert bytes (the secret is already stored).
+const (
+	existingCertSentinel = "_existing_"
```
mayuka-c, line 27: Do we really need this?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
```diff
@@ -156,6 +197,45 @@ func handlePostDeployment(ctx context.Context, caddyCtx *caddy.Context, deployCt
...
+}
+
+// resolveCertPaths returns the cert and key paths to use for template rendering.
+// When useExistingCert is true (the cert secret was preserved by --skip-cleanup
+// and no new paths were supplied), a non-empty sentinel is returned so the caddy
+// template renders the volume mount; the existing secret provides the actual bytes.
+func resolveCertPaths(certPath, keyPath string, useExistingCert bool) (string, string) {
+	if useExistingCert {
+		return existingCertSentinel, existingCertSentinel
+	}
```
mayuka-c: Do we really need to return this string and again do compare in caller with same string?

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
prachikurade1: Since it is a fixed prefix, the recommendation is to use it directly instead of regex to avoid compiling the regex.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
prachikurade1: The sentinel is needed to be non-empty so the {{- if …}} guards in both templates render their SSL blocks, without carrying real cert bytes that would overwrite the preserved secret.

ai-services/internal/pkg/catalog/cli/configure/podman/configure.go
prachikurade1: fixed

yussufsh (approved):

yussufsh (approved):
lgtm

PR #1489, opened 2026-09-17
Bump up service base image

adarshagrawal38:
Making use of latest `service-base` image for all the `services`


yussufsh:
Can we try docling upgrade tomorrow after most of the packages will be available in devpi?

yussufsh:
I think if we can merge this today for the upcoming rc build.
New packages will be available tomorrow evening which will be late.

yussufsh (approved):
lgtm

PR #1490, opened 2026-09-17
Bump up docling to v2.127.0

adarshagrawal38:
Bump docling to v2.127.0 and update transitive dependencies

Upgrades docling to version 2.127.0 and syncs all associated transitive dependency versions in images/service-base/requirements.txt.


adarshagrawal38:
Successfully parsed pdf document with docling v2.127.0
image

yussufsh:
@adarshagrawal38 the build is failing because the scipy version is not available on devpi. If not possible to downgrade can we use docling `v2.122.0`? where scipy-1.18.0 is used.

adarshagrawal38:
> @adarshagrawal38 the build is failing because the scipy version is not available on devpi. If not possible to downgrade can we use docling `v2.122.0`? where scipy-1.18.0 is used.

Scipy is available in devpi look here -> https://wheels.developerfirst.ibm.com/ppc64le/linux/scipy/stable

yussufsh:
> > @adarshagrawal38 the build is failing because the scipy version is not available on devpi. If not possible to downgrade can we use docling `v2.122.0`? where scipy-1.18.0 is used.
> 
> Scipy is available in devpi look here -> https://wheels.developerfirst.ibm.com/ppc64le/linux/scipy/stable

Oh, I didn't find it on the home page at https://wheels.developerfirst.ibm.com/ppc64le/linux. But why is the build failing?

yussufsh:
@adarshagrawal38 could you also set ubi9.8 for the prebuilder?

yussufsh:
@adarshagrawal38 can you please also bump the docling to the recent version v2.128.0?

yussufsh (approved):

PR #1496, opened 2026-09-18
Fix for Disconnect Datasource Issue

prachikurade1:
Issue:
While making changes to the "Disconnect Datasource" UI conditions, it was observed that intermittently, if a datasource was deleted during the syncing phase, the datasource would be returned in the application datasources listing API with the "delete pending" status (from downstream service) 
This was because we fetch connectors from downstream service first and then populate with our data (done so to ensure pagination works as per the downstream service).
So even if the datasource has been deleted from our side (because we received acknowledgement of the delete the request from the downstream service), it was not deleted from the service. So fix was added to filter the connectors received from downstream based on data in our db.


ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
```diff
@@ -1224,6 +1224,13 @@ func (s *DatasourceService) ListApplicationDatasources(ctx context.Context, req
 	// params. The service response is authoritative for both the page content and total count.
 	baseURL, servicePage := s.fetchServiceConnectors(ctx, proxy, appID, allConnectorIDs, req.PageSize, offset)
 
+	// The service pod may still return a connector with status="delete pending" after
+	// DisconnectDatasourcesFromApplication has removed its service_dependencies row but
+	// before the pod finishes its async teardown. Drop any such stale entries so they
+	// do not reappear in the list response. allConnectorIDs is the authoritative set of
+	// connectors that are still linked to this application in the DB.
+	filterServiceDeletionPendingDatasources(servicePage.ByID, allConnectorIDs)
```
yussufsh: After filter, won't the pagination count change?

mayuka-c:
`So even if the datasource has been deleted from our side (because we received acknowledgement of the delete the request from the downstream service), it was not deleted from the service. So fix was added to filter the connectors received from downstream based on data in our db.`

Why the datasource was not deleted from the service? Is it async? Didnt quite get it.

ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/datasource_service.go
prachikurade1: done

prachikurade1:
> `So even if the datasource has been deleted from our side (because we received acknowledgement of the delete the request from the downstream service), it was not deleted from the service. So fix was added to filter the connectors received from downstream based on data in our db.`
> 
> Why the datasource was not deleted from the service? Is it async? Didnt quite get it.

The datasource was in syncing state, so digitize acknowledges the request and marks the datasource for deletion. However they have not yet deleted it. For list application datasources we fetch their list first and then merge with the details required from our end, so a deleted datasource would appear, hence we are filtering it.

mayuka-c:
LGTM,

Once delete is initiated, the digitize service is taking care of deleting in background and ensuring their database is cleaned up with the entry even if anything fails in their async delete flow (as per @dharaneeshvrd ). So going ahead and merging this.

Still feel there is scope of handling it better, maybe take it up in next release to handle delete better.

mayuka-c (approved):

PR #1498, opened 2026-09-18
fix(BYOS): Do not allow certified_by to be set with IBM

mayuka-c:
- This is to not allow users to use certified_by set to "IBM" while bringing up their own service


yussufsh (approved):
lgtm

PR #1501, opened 2026-09-18
Worker UI bug fixes

pranithraoibm:
This PR fixes all the bugs mentioned in https://jsw.ibm.com/browse/AISERVICES-1824

_Dig Assist/Service - drilldown panel - details_
Data source tab isn’t visible until you click on it - **Code wasn't merged then, visible now**
Can this tag be moved into the header by status and service type? OR move it to the ‘Services’ tab - **Moved to header**
Align content left with graph component, align graph component with label - **Done**

image (2)

_List of registered workers_
Search field should match the table color; in this case it should be white - **Only on hover of that its grey or else its white itself, default carbon behaviour nothing changed**
Missing error icon (but do not add icons for success or normal messages) - **Added the error icon**

image (3)  

_Register a worker_
The field changes to the same color as the background ... it needs to be light grey - **Done**
image (4)

_Deploy experience - Select services panel_
Needs 16px margins between cards, the radio button icon looks huge - it should be 16px - **Done**
image (5)

We never use an all-page loading state – all components should load immediately, and those components that are looking for data should be in a skeleton-loading state - **Added skeleton loading states everywhere applicable**
image (6)
image (7)
image

is this normal behavior for a menu to display a scroll bar when there is no scrolling needed? - **Yes, its carbon default behaviour**

When RedHat is selected, the field should not be displaying “local” - **Yes, its fixed in the latest code**

https://github.com/user-attachments/assets/b642dab4-f601-4cf0-8869-172e42af280f

this popup should have a white background with grey fields ... this is opposite, the field changes colors after the token is generated, the run command content does some weird shifting - **All are fixed now**

https://github.com/user-attachments/assets/86336256-4998-4a47-a792-819b4cbc496e

Perhaps I was looking at an old version, but I assume the user could just copy the command, paste it and hit enter, then everything would be done.   But it looks like the are extra steps that the user has to do. Can we simplify this? - **User has to just run the command, nothing else**

disk storage should not be on a line by itself! All 4 should be together on the one line - **It's just the responsive behavior of the cards**

https://github.com/user-attachments/assets/6290e62a-408c-44bc-aa21-a95b4e6ce3c6

add space here ... should be 48px above the Inference credentials header
image (9)

All these are questions for the backend team: 
After the registration (generate key) popup closes, the table status shows "pending" ... should this be an informational tag or error, because the worker is not usable and will not appear in applicable menus
Pending workers need the message: ” Registration is not complete. Run the bootstrap command on the worker to connect it and finish setup - **Added**
image

Vector name and Embedding model need to move to the Configure services tab
Vector store and Embedding model fields should be enabled on this tab now – the should be removed from the Service Details tab, and be enabled here - **Will add this change only if the backend team makes the above change to move it from step 1 to step 2 as there will be some logic changes as well, or else will do it in the next release**

3 new changes:
- Updated 'Inference backend' label to 'LLM inference backend'
- Updated 'What is this model good at?' to 'What is this LLM good at?'

image
image

- Resource requirements cards height adjustment as requested by the design team
image


yussufsh:
> _Register a worker_
> The field changes to the same color as the background ... it needs to be light grey - **Done**
> image (4)

Just a suggestion to move `--runtime` parameter to the right, since that is the only one that needs to be provided.

pranithraoibm:
@yussufsh Updated the command as requested.

image

ui/catalog/vite.config.ts
```diff
@@ -15,7 +15,7 @@ export default defineConfig({
   server: {
     proxy: {
       "/api": {
-        target: "http://localhost:8080",
+        target: "https://catalog-api.10.20.185.203.nip.io",
```
ryanmarc: change this back.

ryanmarc (approved):

ryanmarc (approved):

PR #1503, opened 2026-09-18
[AISERVICES-1822] fix: improve datasource connector UI, validation, and error handling

Niranjan-J-Joshi:
### Summary

Improves the Data Source Connector experience by updating UI text, fixing layout issues, improving validation and error handling.

---

#### 1. Update button text and placeholder

**File:** `ConnectDatasourceModal.tsx`

- Changed button text from **Add** / **Adding...** to **Connect** / **Connecting...**.
- Updated the dropdown placeholder from **Select data sources** to **Choose an option**.

---

#### 2. Improve data source selection layout

**Files:** `SharedDatasourceStep.tsx`, `SharedDatasourceStep.module.scss`

- Moved the success checkmark icon to the top-right corner.
- Clear selected data sources when the upload option is turned off to prevent stale data from being submitted.
- Fixed alignment issues by removing unnecessary left padding.
- Added a white background to the dropdown menu.

---

#### 3. Improve delete confirmation dialog

**File:** `DeleteConfirmNameModal.tsx`

- Updated confirmation text from:
  - `Type [name] to confirm`
  - to `Type name to confirm`
- Added validation to ensure the item name is not empty before enabling confirmation.
- Enabled the danger button state only when the entered name matches.

---

#### 4. Select checkbox options by default

**File:** `schemaUtils.ts`

- Checkbox groups now start with all available options selected by default instead of starting empty.

---

#### 5. Fix tooltip positioning

**File:** `ConnectorFieldLabel.tsx`

- Updated tooltip positioning to prevent it from being clipped inside the side panel.
- Added automatic alignment support for better placement.

---

#### 6. Improve error handling and spacing

**Files:** `AddDataSourceModal.tsx`, `AddDataSourceModal.module.scss`, `types.ts`

##### Error Handling

- Added support for showing errors in the most relevant location:
  - Authentication errors are shown in authentication fields and section banners.
  - Access errors are shown in location-related fields and section banners.
  - Network and unexpected errors are shown in the top-level error banner.

##### Form State Updates

- Added section-level error tracking.
- Clear field and section errors when users update values.
- Reset all validation and error states when switching connector types.

##### Layout Improvements

- Updated spacing and margins to better match Carbon Design guidelines.

---

https://github.com/user-attachments/assets/156c9c73-d589-48f6-94f2-b84769c6bae3

image
image
image
image


ui/catalog/src/components/AddDataSourceModal/AddDataSourceModal.tsx
```diff
@@ -143,15 +152,114 @@ const AddDataSourceModal = ({
...
+        if (m[1] === field.key) {
+          matched = true;
+          break;
+        }
+      }
+
+      // B: word scan — only within the prefix section to avoid cross-section hits.
+      if (
+        !matched &&
+        (!prefixSection || field.sectionTitle === prefixSection)
```
ryanmarc: The !prefixSection || here undoes the restriction the comment above describes. When the server message has no [AUTH]/[ACCESS] prefix, every field in the form gets word-scanned, and since the name-conflict message contains the name the user typed, a data source called prod-host or s3-prefix matches a field key (\b counts - as a boundary). I ran the two matchers from this function over the real messages to check: datasource with name "host" already exists flags the host field, and datasource with name "prefix" already exists flags prefix.

The part that makes it more than cosmetic is that parseServerError then returns non-null, so handleSubmit takes the SET_FIELD_ERRORS branch and returns, and the actual name-conflict message never reaches the top-level banner. The user sees "Provide a valid hostname or IP address." under a hostname that is fine, and nothing at all about the name.

Could this just be `if (!matched && prefixSection && field.sectionTitle === prefixSection)`? That is what the comment already says it does, and the at '/key': pass above stays unrestricted, so schema validation errors still route correctly.

ui/catalog/src/components/AddDataSourceModal/AddDataSourceModal.tsx
```diff
@@ -143,15 +152,114 @@ const AddDataSourceModal = ({
...
+      structuredRe.lastIndex = 0;
+      let m: RegExpExecArray | null;
+      while ((m = structuredRe.exec(message)) !== null) {
+        if (m[1] === field.key) {
+          matched = true;
+          break;
+        }
+      }
+
+      // B: word scan — only within the prefix section to avoid cross-section hits.
```
ryanmarc: file_system.go says "verify the username and private key are correct", with "private key" spelled as two words, so `\bprivate_key\b `misses it and only username gets flagged.

The other four messages happen to come out right, but only because the wording happens to use the key spelling, so this will drift again the next time someone rewords a Go string.

Smallest fix that holds: build the pattern from the key with separators made flexible, e.g. `field.key.split("_").join("[\\s_-]*")`, so private_key also matches "private key". That flags both Username and Private key on the SSH message, which is the honest answer since the handshake failure does not tell them apart.

A much better fix would be a backend update so [ConnectionCheckError](https://github.com/IBM/project-ai-services/blob/main/ai-services/internal/pkg/catalog/apiserver/repository/datasource_service/interface.go#L24) could carry the offending property names and the UI could stop guessing.

ryanmarc (changes requested):

PR #1510, opened 2026-09-22
Addressing security on exceptions for Use-Cases

sats-23:
-Mask credentials related vars from being emitted 
-Aggregate exceptions into a single exception class for each service 
-Standardize Summarize & Extract exceptions


services/chatbot/tests/conftest.py
```diff
@@ -47,9 +38,9 @@ def mock_diagnostic_crash_handler():
     during imports of application modules, which conflicts with pytest capture
     and can trigger "OSError: [Errno 29] Illegal seek" on macOS.
     """
-    # The patching is already done at module level above
-    # This fixture just ensures it stays in place
-    yield
+    with patch.dict(os.environ, {"DISABLE_CRASH_HANDLER": "1"}), \
```
dharaneeshvrd: Can you please explain why this change required?

PR #1513, opened 2026-09-22
Resource requirement updates

prachikurade1:
Updates for the resource requirements for catalog and services (summarize, extract and translate, digitize load test was already done last release, hence skipped now)


yussufsh:
@prachikurade1 can you justify the new value, especially doubling or even quadrupling it in some places?
Also, Digitize service now has connectors and lots of ingestions will be done there. I feel we need to re-evaluate it.

prachikurade1:
> @prachikurade1 can you justify the new value, especially doubling or even quadrupling it in some places? Also, Digitize service now has connectors and lots of ingestions will be done there. I feel we need to re-evaluate it.

I have shared the stats on slack along with the recommendations. 
For Digitize, we will need to run large ingestion, or a large connector datasource. @sats-23 , can you help with this ?

PR #1514, opened 2026-09-22
jenkins: Adding missing images for code signature pipeline

adarshagrawal38:
Add pipeline stages for `translate-service`, `extract-service`, and `ai-services-mcp` images so that signatures are generated for these images as part of the signing pipeline.


yussufsh (approved):
lgtm

PR #1515, opened 2026-09-22
Docs stuck in Accepted state

sats-23:
-Fix docs stuck in accepted state
-Fix fire sync after restart


ai-services/assets/applications/rag-cpu/podman/values.yaml
```diff
@@ -24,7 +24,7 @@ digitize:
   # @description Database name for the DIGITIZE service. Default is "digitize_metadata".
   database: "digitize_metadata"
   # @description AES-256 encryption key for the DIGITIZE database (32 raw bytes, base64-encoded).
-  dbEncryptionKey: "B374A26A71490437AA024E4FADD5B497FDFF1A8EA6FF12F6FB65AF2720B59CCF"
```
dharaneeshvrd: Why are we changing this now?

services/digitize/connectors/scheduler.py
```diff
@@ -77,21 +77,24 @@ async def register_connector_job(
     from digitize.api.v1.connectors import dispatch_sync
 
     now = datetime.now(timezone.utc)
-    # APScheduler v3: IntervalTrigger always schedules the first run at
-    # start_date + N*interval.
-    # Passing next_run_time=now overrides the trigger's first computed fire time
-    # so the job executes immediately, then repeats on the normal interval.
+    # APScheduler v3: passing next_run_time=None puts the job in a paused state
```
dharaneeshvrd: When & how did this happen?

services/digitize/pipeline/digitize.py
```diff
@@ -131,8 +131,15 @@ def _fail(error: str) -> None:
             deadline, timeout_s, "start",
         )
 
-        # Mark IN_PROGRESS as soon as the dispatcher starts running.
-        if task is not None and task.status == ConversionTaskStatus.RUNNING and doc_id:
+        # Mark IN_PROGRESS once the task has been picked up — covers both the
```
dharaneeshvrd: can you please explain why this change is required?
IMO if its completing within the polling interval, that's completely fine, we don't need to show the in progress status

dharaneeshvrd (commented):
Can you please open a JIRA bug for this?
Btw is this caught by QA team?

ai-services/assets/applications/rag-cpu/podman/values.yaml
sats-23: Length is 64 bytes and this is preventing connector from being created with "DB_ENCRYPTION_KEY must be 32 bytes, got 64 bytes"

services/digitize/connectors/scheduler.py
sats-23: v3 deals it differently 
If we pass fire_immediately as false to our function, it does not run it on schedule. It waits for first manual trigger to resume normal behavior.
So fixed it with this change, now the fire happens 5 mins after restart (usual 300s window)

services/digitize/pipeline/digitize.py
sats-23: Right, not necessary to have interim state. 
I will drop it

ai-services/assets/applications/rag-cpu/podman/values.yaml
sats-23: Doesn't affect catalog way of deployment. 
Only affects legacy deployment

services/digitize/connectors/scheduler.py
dharaneeshvrd: OK please update this info on description as well as a comment here

dharaneeshvrd (commented):
Changes lgtm, please bump the version

sats-23:
> Can you please open a JIRA bug for this? Btw is this caught by QA team?

Yes - https://jsw.ibm.com/browse/AISERVICES-1841

dharaneeshvrd (dismissed):

dharaneeshvrd (approved):

PR #1516, opened 2026-09-22
Allow deletion while syncing

Niranjan-J-Joshi:
## Summary

Allows users to delete a Data Source Connector even when its status is `syncing`.

Previously, the Delete action was disabled while a connector was syncing.
### `ApplicationDatasourcesTable/CellRenderers.tsx`

- Removed the logic that disabled the Delete action when a connector was in the `syncing` state.
- The Delete action is now available for all visible connectors.
- Removed the additional check for `delete pending` status, as connectors in this state are already filtered out by the backend API and are not returned to the UI. https://github.com/IBM/project-ai-services/pull/1496

### `Table/components/CellRenderers.tsx` (shared)

- Kept the `delete pending` entry in `STATUS_CONFIG` as a fallback.

https://github.com/user-attachments/assets/db154462-d950-4133-9d78-3675b2e55558


ryanmarc (changes requested):
The content in this delete popup looks entirely incorrect. It talks about vector store when the type of the connector being removed is object storage.

Niranjan-J-Joshi:
I checked with design team about the remove datasource description ,As there are some questions around this,  they have asked us to keep the current behavior as is.

ryanmarc (approved):

